PR #8, opened 2018-02-07
Basic Travis CI support

aesedepece:


aesedepece:
If all is OK, there should be a message right on top of this one stating:

**All checks have passed**
1 successful check

PR #17, opened 2018-09-25
config: update cli, travis and read config file

Tommytrg:
Read default values from the config file and update cli to read peer address.

### DIFF
--- a/crypto/tests/lib.rs
+++ b/crypto/tests/lib.rs
@@ -4,5 +4,5 @@ use crypto::greetings;
 
 #[test]
 fn crypto_greeeting() {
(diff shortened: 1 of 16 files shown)


crypto/tests/lib.rs
```diff
@@ -4,5 +4,5 @@ use crypto::greetings;
 
 #[test]
 fn crypto_greeeting() {
-    assert_eq!(greetings(), String::from("Hello form crypto!"));
+    assert_eq!(greetings(), String::from("Hello from crypto!"));
```
aesedepece: :boom:

PR #21, opened 2018-09-27
Implement basic Storage module with different backends

aesedepece:
Hey! This is a WIP pull request for 3 reasons:
- It's not based upon `upstream/master` but `csmoe/2018`.
- It's missing tests
- It's missing an entry in the official documentation

The `Storage` trait is widely generic and can be implemented for different storage backend solutions.

This PR adds two specific backends:
- In-memory
- RocksDB

Dependencies of the different backends are optional, and they must me enabled in `Cargo.toml` behind a feature flag.

The `util` package will contain useful structs, traits, types, etc. that can be easily used across all the project.

For the time being, I have added some convenience structs, types and implementations for nicely handling our own `Result`s and `Error`s.


util/src/lib.rs
```diff
@@ -0,0 +1,10 @@
+//! The `util` package contains useful structs, traits, types, etc. that can be easily used across
+//! all the Witnet-rust project.
+
```
csmoe: add `#![deny(rust_2018_idioms)]` to force rust_2018, and let `cargo fix` help you to translate the patch.

PR #24, opened 2018-10-01
Document and include flatbuffers for network protocol

mariocao:
This PR includes the following:
 + Documentation for defining a basic network protocol (still under development)
 + Flatbuffers for the defined messages in the aforementioned documentation
 + Generated code for the flatbuffer schema
 + Test to provide a sample how to use the generated code


data_structures/src/lib.rs
```diff
@@ -1,6 +1,7 @@
 //! data structures
 
-#![deny(rust_2018_idioms)]
+// Removed due to flatbuffers generated code
```
aesedepece: `FIXME`

mariocao:
I will check how to exclude the flatbuffers generated code from `clippy`. :disappointed:

csmoe:
@mariocao add `#[allow(clippy)]` into the mod declaration

data_structures/src/lib.rs
csmoe: Instead of removing, it'll be better to turn off this by applying `#[allow(rust_2018_idioms)]` to generated mod as there may be other hand-writed code in the future.

mariocao: We tried to apply `#[allow(rust_2018_idioms)]` after declaring the `#![deny(rust_2018_idioms)]` but it seems that outer attributes cannot override inner attributes. Any ideas?

data_structures/src/lib.rs
```diff
@@ -12,3 +13,7 @@ pub fn greetings() -> String {
     println!("Hello from data structures!");
     String::from("Hello from data structures!")
 }
+
+/// Generated Message module from Flatbuffers compiler
+#[allow(missing_docs)]
```
csmoe: @mariocao add it at here, does it helps?

mariocao:
After discussing it with @aesedepece and @mmartinbar , we decided to leave the `lib.rs` without attributes, so that each mod can declare their own attributes. The reason behind is if we allow `#![allow(elided_lifetimes_in_paths)]` in `lib.rs`, then all submodules will have to override this attribute.

data_structures/tests/flatbuffers/mod.rs
```diff
@@ -0,0 +1,66 @@
...
+        },
+    );
+
+    // Serialize the root of the object, without providing a file identifier.
+    builder.finish(message, None);
+
+    // Check flatbuffer data of type `&[u8]`
+    let buf = builder.finished_data();
+
+    let expected_buf: [u8; 48] = [
```
aesedepece: :cool: :sunglasses:

docs/protocol/network.md
```diff
@@ -0,0 +1,206 @@
+# Witnet Network Protocol
+
+Witnet network protocol is inspired in other blockchain network protocols such as Bitcoin, Ethereum and Rchain. Additionally, other blockchains such as Grin and Exonum have been taken into consideration as their reference implementations are coded in Rust.
+
+## Initial considerations
+
+In order to start the network discovery process, at least one existing node on the network is required. However, knowning several existing nodes, extracted from different sources, is always recommended in order to avoid unnecessary risks or potential attack vectors.
```
aesedepece: `However, in order to avoid unnecessary risks or potential attack vectors, it is always recommended to know several existing nodes extracted from different sources.`

docs/protocol/network.md
```diff
@@ -0,0 +1,206 @@
...
+
+### Peer bootstrapping methods
+
+The strategy for selecting the first node to connect should be random, thus it should not depends on geographical position or any other network parameter.
```
aesedepece: `depends` → `depend`

docs/protocol/network.md
```diff
@@ -0,0 +1,206 @@
...
+
+Subsequently, the handshake initiator will expect a `version` message from the remote peer. The local peer will also acknowledge by replying with a `verack` message.
+
+Connection cannot be considered as establish until both `verack` messages have been received by both peers.
+
+After sending a `version` message the peers will define a timeout to wait for a response. If no response is received during this timeout (usually set to 10 seconds), the peer will be discarded from the known peers list.
+
+### Peer discovery (GET_PEERS, PEERS)
+
+In Witnet, a node must always establish establish different paths into de network by connecting to different peers, ideally randomly selected. As connections may disappear, network peers should assist others when they bootstrap.
```
aesedepece: - `establish establish` → `establish` 
- `de network` → `the network`

docs/protocol/network.md
```diff
@@ -0,0 +1,206 @@
...
+
+The `peers` message consists of the message header with the command `PEERS` and as payload it contains a list of known peers as:
+
+| Field  | Type | Description |
+|--------|:----:|-------------|
+| peers | addr[] | list of addresses of active known peers |
+
+### Heartbeat (PING, PONG)
+
+As in Bitcoin, the heartbeat protocol is defined with `ping` and `pong` messagesabnd it is defined as follows:
```
aesedepece: `messagesabnd` → `messages and`

PR #27, opened 2018-10-03
refactor(err): introduce failure 

csmoe:
r? @aesedepece 

edit: wip for some unoptimized snippets.


storage/src/backends/in_memory.rs
```diff
@@ -12,28 +12,28 @@ use std::collections::HashMap;
...
+    pub memory: HashMap<&'a [u8], Vec<u8>>,
 }
 
 /// Implement the Storage generic trait for the InMemoryStorage storage data structure.
-impl <'a> Storage<(), &'a [u8], Vec<u8>> for InMemoryStorage<'a> {
-
-    fn new(_: ()) -> Result<Box<Self>> {
-        Ok(Box::new(InMemoryStorage { memory: HashMap::new() }))
+impl<'a> Storage<&'a [u8], Vec<u8>> for InMemoryStorage<'a> {
+    fn new(_: String) -> StorageResult<Box<Self>> {
```
aesedepece: Why use `String` here as input type instead of `()`? We don't even want to pass any parameters to `InMemoryStorage::new()`. Am I missing some 2018 novelty?

storage/src/backends/rocks.rs
```diff
@@ -1,69 +1,69 @@
...
 use rocksdb::DB;
+
 use std::str;
 
+use witnet_util::error::Error;
+
 /// Data structure for the RocksDB storage whose only member is a rocksdb::DB object.
 pub struct RocksStorage {
-    db: DB
+    db: DB,
```
aesedepece: I like this :+1:

storage/src/error.rs
```diff
@@ -2,23 +2,52 @@
...
     /// Error while trying to open a connection to the storage backend.
-    ConnectionError,
+    #[fail(
+        display = "connection error: at \"{}\", rocksdb_msg {}",
+        path,
+        msg
+    )]
+    Connection {
+        /// Connection address
+        path: String,
```
aesedepece: Ok, I get it. You used a `String` in `InMemoryStorage::new()` because of this.

But the thing is that `Storage` and therefore `StorageError` must remain generic as we don't want to limit different implementors to only accept `String` as connection info. I mean, this argument could sometimes be a path, some times an URL and some times a data structure with user, password, etc. 

What about this?
```rust
pub enum StorageError<T: Debug> {
    #[fail(
        display = "connection error: with data \"{:?}\" and msg \"{}\"",
        path,
        msg
    )]
    Connection {
        data: T,
        msg: String
    }
}
```
For the same reason, we shouldn't mention here `rocksdb_msg` but simply `msg`.

Generally speaking, we want to decouple our architecture from specific dependencies, so that we can easily swap them in the future (or even let the users use whatever storage backend they want to use).

storage/src/storage.rs
```diff
@@ -1,25 +1,23 @@
...
 /// This trait can be easily implemented for any specific storage backend solution (databases,
 /// volatile memory, flat files, etc.)
-pub trait Storage<ConnData, Key, Value> {
-
+pub trait Storage<Key, Value> {
     /// Storage object constructor.
     /// `connection_data` can be used for passing credentials, urls, paths, etc. down to the storage
     /// backend.
-    fn new(connection_data: ConnData) -> Result<Box<Self>>;
+    fn new(connection_data: String) -> StorageResult<Box<Self>>;
```
aesedepece: As said before, we want `Storage` to be generic over different storage backends that may require different connection data types. Forcing a `String` here is doing the opposite.

util/src/error.rs
```diff
@@ -1,79 +1,57 @@
...
+    /// storage error
+    #[fail(display = "StorageError: {}", 0)]
+    Storage(K),
 }
 
-/// Implement the `::new()` constructor for the generic Error<K> type.
-impl <K: Debug> Error<K> {
+impl<K: Fail> WitnetError<K> {
+    /// create storage error
+    pub fn storage_err(err: K) -> Self {
```
aesedepece: The `util` crate should not know about the existence of the `storage` crate. Everything in this file is supposed to be generic.

storage/src/backends/rocks.rs
```diff
@@ -1,69 +1,71 @@
...
 }
 
 /// Implement the Storage generic trait for the RocksStorage storage data structure.
-impl <'a> Storage<&'a str, &'a [u8], Vec<u8>> for RocksStorage {
-
-    fn new(path: &str) -> Result<Box<Self>> {
+impl<'a> Storage<&'a [u8], Vec<u8>> for RocksStorage {
+    fn new(path: impl Debug) -> StorageResult<Box<Self>> {
```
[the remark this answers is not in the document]
csmoe: personally, I like this as I said in discord. But witnet team is more senior than me, if you reach a decision about this, ping me.

Issue #32, opened 2018-10-15
Serializers implementation


Tommytrg:
#### Description
Witnet-rust will use [Flatbuffers](https://google.github.io/flatbuffers/index.html) as serialization library, but, in spite of this, it is not discarded support other similar libraries in future. Because of that, the first approach will create a "serialization"/"codecs" folder where to implement methods as ```fromX``` and ```toX``` for a concrete library and manage which one should be used at compilation time.

Focusing on the point of view of a peer: ***as a peer, I want to be able to create a structure\* where add a message object that could be read from other peers.***

\* *A structure which uses flatbuffers in a first approach*

#### Actionables
- [ ] Define an API to wrap flatbuffers' native methods.
- [ ] Implement serializer methods of flatbuffers.
- [X] Define Flatbuffers schema.
- [ ] Define schema's message types mirroring flatbuffers ones.
- [X] Implement flatbuffers schema.
- [ ] Use flatbuffers ```verify``` from C library.

mariocao:
I created #69 in order to track the optional task/issue described in the actionables of this issue. :blush:

PR #62, opened 2018-10-23
feat(ftb-codec): add ftb encode and decode implementations

Tommytrg:
This PR includes implementation to encode and decode flatbuffers' traits and create a witnet
message from flatbuffers one and vice versa.

Current `protocol_generated.rs` file is build from flatc compiler in commit 49fed8c4f61f296d1d173a1a8df872a094020278 from master branch


data_structures/src/serializers.rs
```diff
@@ -0,0 +1,461 @@
...
+            }
+            FlatBuffersComand::GetPeers => create_get_peers_message(magic),
+            FlatBuffersComand::Peers => {
+                create_peers_message(magic, message.command_as_peers().unwrap())
+            }
+            FlatBuffersComand::Verack => create_verack_message(magic),
+            FlatBuffersComand::Version => {
+                let command = message.command_as_version().unwrap();
+                let ftb_sender_address = command.sender_address().unwrap();
+                let ftb_receiver_address = command.receiver_address().unwrap();
```
csmoe: to be honest, I'm a bit scared with those `.unwrap()`.

data_structures/src/serializers.rs
```diff
@@ -0,0 +1,461 @@
...
+                hextets.ip3(),
+                address.port(),
+            )
+        }
+        FlatBuffersIpAddress::NONE => panic!(),
+    }
+}
+
+// Create a get peers flatbuffer to encode a witnet's get peers message
+fn create_get_peers_flatbuffer(mut builder: FlatBufferBuilder, magic: u16) -> Vec<u8> {
```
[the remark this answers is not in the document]
Tommytrg: That's true, variable builder doesn't need to be mutable itself, just its reference.

data_structures/src/serializers.rs
Tommytrg: I have refactored almost all `.unwrap()`. I am working on refactor others.

data_structures/src/serializers.rs
```diff
@@ -0,0 +1,465 @@
+extern crate flatbuffers;
+
+use crate::flatbuffers::protocol_generated::protocol::{
+    get_root_as_message, Address as FlatBufferAddress, AddressArgs, Command as FlatBuffersComand,
```
mmartinbar: Typo:

```suggestion
    get_root_as_message, Address as FlatBufferAddress, AddressArgs, Command as FlatBuffersCommand,
```

mmartinbar:
`cargo build` gives one warning and `just clippy` also shows some warnings/errors

data_structures/src/serializers.rs
```diff
@@ -0,0 +1,572 @@
+extern crate flatbuffers;
+
+use crate::flatbuffers::protocol_generated::protocol::{
+    get_root_as_message, Address as FlatBufferAddress, AddressArgs, Command as FlatBuffersCommand,
+    GetPeers as FlatBuffersGetPeers, GetPeersArgs, IpAddress as FlatBuffersIpAddress,
+    Ipv4 as FlatBuffersIpv4, Ipv4Args as FlatBuffersIpv4Args, Ipv6 as FlatBuffersIpv6,
+    Ipv6Args as FlatBuffersIpv6Args, Message as FlatBuffersMessage, MessageArgs,
+    Peers as FlatBuffersPeers, Peers, PeersArgs, Ping as FlatBuffersPing, PingArgs,
```
mmartinbar: I think that `Peers` may be imported with two names here, as `Peers`and `FlatBuffersPeers` which may be confusing

data_structures/src/serializers.rs
```diff
@@ -0,0 +1,572 @@
...
+#[derive(Debug, Clone, Copy)]
+struct PeersFlatbufferArgs<'a> {
+    magic: u16,
+    peers: &'a [Address],
+}
+
+#[derive(Debug, Clone, Copy)]
+struct PeersWitnetArgs<'a> {
+    magic: u16,
+    peers: Peers<'a>,
```
mmartinbar:
```suggestion
    peers: FlatBuffersPeers<'a>,
```

data_structures/src/serializers.rs
```diff
@@ -0,0 +1,572 @@
...
+    receiver_address: Address,
+    sender_address: Address,
+    timestamp: u64,
+    user_agent: String,
+    version: u32,
+}
+
+pub trait FromBytes<T> {
+    fn from_bytes(bytes: Vec<u8>) -> Option<T>;
+}
```
mmartinbar: I think (though I'm not 100% sure) that we could perhaps be using here the already existing [TryFrom](https://doc.rust-lang.org/std/convert/trait.TryFrom.html) trait and not define our own trait

data_structures/src/serializers.rs
```diff
@@ -0,0 +1,572 @@
...
+
+pub trait ToBytes {
+    fn to_bytes(&self) -> Vec<u8>;
+}
```
mmartinbar: I think (though I'm not 100% sure) that we could perhaps be implementing the already existing [TryInto](https://doc.rust-lang.org/std/convert/trait.TryInto.html) or [Into](https://doc.rust-lang.org/std/convert/trait.Into.html) trait and not defining our own

data_structures/src/serializers.rs
```diff
@@ -0,0 +1,572 @@
...
+
+// #[cfg(feature = "flatbuffers")]
+impl FromBytes<Message> for Message {
+    fn from_bytes(bytes: Vec<u8>) -> Option<Self> {
```
mmartinbar: I would suggest using a slice as parameter (unless we want the vector to be moved in here, which might be our case). This way both vectors and array slices could be used as parameters to this function.

```suggestion
    fn from_bytes(bytes: &[u8]) -> Option<Self> {
```

data_structures/src/types.rs
```diff
@@ -0,0 +1,50 @@
+#[derive(Debug, Eq, PartialEq)]
+pub enum Command {
```
mmartinbar: :heart:

data_structures/src/serializers.rs
```diff
@@ -0,0 +1,572 @@
...
+            command_type: FlatBuffersCommand::Ping,
+            command: Some(ping_command.as_union_value()),
+        },
+    );
+
+    build_flatbuffer(builder, message)
+}
+
+// Create a witnet's ping message to decode a flatbuffers' ping message
+fn create_ping_message(ping_args: &PingWitnetArgs) -> Message {
```
mmartinbar: I would make this function take the ownership of its parameter. Similar functions like `create_get_peers_message` and `create_verack_message` already do it like that. To me, it makes more sense as this function transforms its input parameters into another different thing and returns it.

```suggestion
fn create_ping_message(ping_args: PingWitnetArgs) -> Message {
```

data_structures/src/serializers.rs
```diff
@@ -0,0 +1,572 @@
...
+            command_type: FlatBuffersCommand::Pong,
+            command: Some(pong_command.as_union_value()),
+        },
+    );
+
+    build_flatbuffer(builder, message)
+}
+
+// Create a witnet's pong message to decode a flatbuffers' pong message
+fn create_pong_message(pong_args: &PongWitnetArgs) -> Message {
```
mmartinbar: Same thing as with the `create_ping_message`:
```suggestion
fn create_pong_message(pong_args: PongWitnetArgs) -> Message {
```

data_structures/src/serializers.rs
```diff
@@ -0,0 +1,572 @@
...
+                let receiver_address = command.receiver_address().and_then(create_address);
+                // Check if sender address and receiver address exist
+                if sender_address.and(receiver_address).is_some() {
+                    Some(create_version_message(VersionWitnetArgs {
+                        version: command.version(),
+                        timestamp: command.timestamp(),
+                        capabilities: command.capabilities(),
+                        sender_address: sender_address.unwrap(),
+                        receiver_address: receiver_address.unwrap(),
+                        // user_agent field should be required as specified in ftb schema
```
mmartinbar: Please, mark this as TODO so that we can easily locate this in the future

data_structures/src/serializers.rs
```diff
@@ -0,0 +1,572 @@
...
+
+pub trait FromBytes<T> {
+    fn from_bytes(bytes: Vec<u8>) -> Option<T>;
+}
+
+pub trait ToBytes {
+    fn to_bytes(&self) -> Vec<u8>;
+}
+
+// #[cfg(feature = "flatbuffers")]
```
mmartinbar: Not sure what these are for, but I would remove them or mark them with some TODO comment / issue for the future.

data_structures/src/serializers.rs
```diff
@@ -0,0 +1,572 @@
...
+                } else {
+                    None
+                }
+            }),
+            FlatBuffersCommand::NONE => None,
+        }
+    }
+}
+
+// #[cfg(feature = "flatbuffers")]
```
mmartinbar: Same thing here.

data_structures/src/serializers.rs
```diff
@@ -0,0 +1,572 @@
...
+            nonce: ping_args.nonce,
+        },
+        magic: ping_args.magic,
+    }
+}
+
+// Create a pong flatbuffer to encode a witnet's pong message
+fn create_pong_flatbuffer(
+    builder: &mut FlatBufferBuilder,
+    pong_args: &PongFlatbufferArgs,
```
mmartinbar: I would make this function take the ownership of its PongFlatbuffersArgs parameter. It would be more consistent with the rest of functions.
```suggestion
    pong_args: PongFlatbufferArgs,
```

data_structures/src/serializers.rs
```diff
@@ -0,0 +1,572 @@
...
+) -> Vec<u8> {
+    let pong_command = FlatBuffersPong::create(
+        builder,
+        &PongArgs {
+            nonce: pong_args.nonce.to_owned(),
```
mmartinbar: Unnecessary call to `to_owned()` here

data_structures/src/serializers.rs
```diff
@@ -0,0 +1,572 @@
...
+            FlatBuffersCommand::Peers => message
+                .command_as_peers()
+                .and_then(|peers| create_peers_message(PeersWitnetArgs { magic, peers })),
+            FlatBuffersCommand::Verack => Some(create_verack_message(VerackWitnetArgs { magic })),
+            FlatBuffersCommand::Version => message.command_as_version().and_then(|command| {
+                // Get ftb addresses and create witnet addresses
+                let sender_address = command.sender_address().and_then(create_address);
+                let receiver_address = command.receiver_address().and_then(create_address);
+                // Check if sender address and receiver address exist
+                if sender_address.and(receiver_address).is_some() {
```
mmartinbar: witty :clap:

data_structures/src/serializers.rs
```diff
@@ -0,0 +1,572 @@
...
+    Message {
+        kind: Command::Verack,
+        magic: verack_args.magic,
+    }
+}
+
+// Create a version flatbuffer to encode a witnet's version message
+fn create_version_flatbuffer(
+    builder: &mut FlatBufferBuilder,
+    version_args: &VersionFlatbufferArgs,
```
mmartinbar: I would make this function take the ownership of its `VersionFlatbuffersArgs` parameter. It would be more consistent with the rest of functions.
```suggestion
    version_args: VersionFlatbufferArgs,
```

data_structures/src/serializers.rs
```diff
@@ -0,0 +1,572 @@
...
+            Some(hextets) => Some(create_ipv6_address(
+                hextets.ip0(),
+                hextets.ip1(),
+                hextets.ip2(),
+                hextets.ip3(),
+                address.port(),
+            )),
+            None => None,
+        },
+        FlatBuffersIpAddress::NONE => panic!(),
```
mmartinbar: I would mark this with a TODO and create an issue to rethink about the flatbuffers schema and the fields that have to be required

mmartinbar: I'm now realizing that the function returns an `Option`, sorry. Perhaps change it to return a `None` in that case for the time being? (as well as marking it with the TODO).

data_structures/src/serializers.rs
```diff
@@ -0,0 +1,572 @@
...
+}
+
+// Create a version flatbuffer to encode a witnet's version message
+fn create_version_flatbuffer(
+    builder: &mut FlatBufferBuilder,
+    version_args: &VersionFlatbufferArgs,
+) -> Vec<u8> {
+    let sender_address_command = match version_args.sender_address.ip {
+        WitnetIpv4 { ip } => {
+            let ip_command = FlatBuffersIpv4::create(builder, &FlatBuffersIpv4Args { ip });
```
mmartinbar: This code is repeated later on in this function and also it was written in the `create_peers_flatbuffer` function. I would suggest a small refactor into a function to avoid repeating code.

data_structures/src/serializers.rs
```diff
@@ -0,0 +1,572 @@
...
+                &AddressArgs {
+                    ip_type: FlatBuffersIpAddress::Ipv4,
+                    ip: Some(ip_command.as_union_value()),
+                    port: version_args.sender_address.port,
+                },
+            )
+        }
+        WitnetIpv6 { ip0, ip1, ip2, ip3 } => {
+            let ip_command =
+                FlatBuffersIpv6::create(builder, &FlatBuffersIpv6Args { ip0, ip1, ip2, ip3 });
```
mmartinbar: Same thing here but with `FlatBuffersIpv6`. This code is repeated later on in this function and also it was written in the `create_peers_flatbuffer` function. I would suggest a small refactor into a function to avoid repeating code.

data_structures/src/serializers.rs
mmartinbar: But if you want to borrow the parameter it's ok by me, as long all similar functions are consistent in their taking the ownership/borrowing policy

data_structures/src/serializers.rs
mmartinbar: Or make the other functions borrow their parameter to keep the consistency :)

data_structures/src/serializers.rs
```diff
@@ -0,0 +1,572 @@
...
+                },
+            )
+        }
+    };
+
+    let user_agent = Some(builder.create_string(&version_args.user_agent));
+    let version_command = FlatBuffersVersion::create(
+        builder,
+        &VersionArgs {
+            version: version_args.version.to_owned(),
```
mmartinbar: Remove all `to_owned` here, they are unnecessary as they are applied on `Copy` types

data_structures/src/serializers.rs
csmoe: personally, I'd like the workflow of `FIXME/TODO` in `rust-lang/rust`:
 1. use ***temporary*** `TODO` marker in commits;
2. the contributor and reviewer work together to clean up the `TODO`s as more as possible;
3. and turn unfixed `TODO`s into ***permanent*** `FIXME`;
4. ensure no TODO in the PR before final merging(`rust-lang/rust` ensures this by rejecting `TODO` in CI).

And the format of `FIXME`:
```
// FIXME(issue-number ): must have comments 

//snippet from rustc codebase

// FIXME: this should be `cannot borrow` (issue #18330)
let node_id = item.as_local_node_id();
```

the comments and issue-number is ***must-have***, and the unfixed case should be exposed with test case if possible.

mariocao:
This PR closes #32

PR #63, opened 2018-10-26
feat(core): add sessions and connections manager

mariocao:
This PR is still a WIP as minor tasks are still missing:

- [x] Query `Config Manager` to retrieve configuration parameters (#57 #59)
- [x] Potential minor refactor of p2p sessions library
- [x] Tests for the p2p sessions library
- [x] Sessions Manager documentation

This PR includes the following:

- Connections Manager actor
- Connections Manager documentation
- Introduction to Managers documentation (in order to avoid redundant information across all managers)
- Sessions Manager actor
- P2P sessions library

PR #64, opened 2018-10-29
Config manager actor and library code

anler:
## Actor

Config manager (`ConfigManager`, actor which is in charge of loading the configuration from a file and storing it in its state, then, any other actor can ask for a copy of that configuration.

Messages supported by the actor:
- `LoadConfig { filename: String }`: Load the configuration from a file.
- `GetConfig`: Obtain a copy of the configuration

The actor automatically derives `std::default::Default`, meaning you can create an instance of the actor and it will be initialized with a default configuration.

Take a look at `core/tests/config_manager_test.rs`, there live the integration tests for that actor.

More specific information is present in tha api docs (`cargo doc -p witnet_core --open`)

## Library code

The library code contains the actual definition of the `Config` (`witnet_config`) struct, along with the different loaders (`witnet_config::loaders`), a loader is just a module that knows how to create an instance of `Config` parsing a raw string, a file, or any other possible source of configuration written in a given format. So far the only loader is `witnet_config::loaders::toml` which supports configuration written in Toml and is able to load it from a raw string (`witnet_config::loaders::toml::from_str`) and from a file (`witnet_config::loaders::toml::from_file`)

## Current limitations and future improvements

- When parsing a configuration string written in a given format, the parser fails at the first error it encounters, so if there are many errors, the user won't see all the errors reported in one step. Ideally we could have the parser and coercion into data structures in separate steps so once the source is parsed we can validate all the parameters and tell the user all the errors she made in one shot.
- We rely on the `Display` implementation of the errors to show a meaningful message of what's wrong when parsing the configuration from a source, some times the generated error is good, but other times it's a bit obscure
- The configuration object implements `Clone` in order for the actor to be able to return a copy of it whenever someone asks for it with the message `GetConfig`. Currently the configuration object is not too big, but in the future if it grows we should think about other ways of making this copy less expensive


mmartinbar:
Had some problems when building upstream/master rebased with this branch. Left those errors in this [issue](https://github.com/witnet/witnet-rust/issues/67)

config/src/lib.rs
```diff
@@ -1,75 +1,113 @@
...
+            known_peers: Self::default_known_peers(),
         }
-        Err(error) => {
-            panic!("failed to parse TOML: {}", error);
+    }
+}
+
+impl StorageConfig {
+    fn default_db_path() -> String {
+        "witnet_db".to_string()
```
mmartinbar: This should be agreed with the team. In this other [PR](https://github.com/witnet/witnet-rust/pull/63), the default db path (added to the .gitignore) is `.wit`. One of them should prevail :)

config/src/lib.rs
```diff
@@ -1,75 +1,113 @@
...
-pub fn read_config() -> Option<Config> {
-    let name: String = String::from(CONFIG_FILE);
-    let mut input = String::new();
+impl ConnectionsConfig {
+    fn default_server_addr() -> SocketAddr {
+        "127.0.0.1:1234".parse().unwrap()
+    }
+
+    fn default_inbound_limit() -> u16 {
+        256
```
mmartinbar: I think that we have usually talked about a smaller default limit (around 128). Please talk to the team to check the concrete number.

config/src/lib.rs
csmoe: I have seen several `String`s while dealing with path, are there any reasons rejecting `std::path`?

config/src/loaders/toml.rs
```diff
@@ -0,0 +1,106 @@
...
+use std::result;
+use toml;
+
+/// Error type denoting the different errors this module can fail with.
+/// Parsing the configuration from Toml might fail with a
+/// `toml::de::Error`, but loading that configuration from a file
+/// might also fail with a `std::io::Error`.
+#[derive(Debug)]
+pub enum Error {
+    /// Indicates there was an error when trying to load a file from a file.
```
mmartinbar:
```suggestion
    /// Indicates there was an error when trying to load configuration from a file.
```

config/src/lib.rs
anler: @csmoe I considered it but found out it is an unsized type and that I had to use it behind a `Box` and so I thought that maybe it was a bit too much. I'm going to give it a try again now that everything is more stable and see how it looks like :wink:

config/src/loaders/toml.rs
```diff
@@ -0,0 +1,106 @@
...
+        match self {
+            Error::IOError(e) => e.fmt(f),
+            Error::ParseError(e) => e.fmt(f),
+        }
+    }
+}
+
+/// Just like `std::result::Result` but withe error param fixed to
+/// `Error` type in this module.
+pub type Result<T> = result::Result<T, Error>;
```
mmartinbar: Really like this error definition! However, I wonder if the error handling in [util](https://github.com/witnet/witnet-rust/blob/master/util/src/error.rs) should be used instead. I don't know if it is supposed that all the subcrates results should be `WitnetResult` like in the [storage library](https://github.com/witnet/witnet-rust/blob/master/storage/src/error.rs). Any thoughts @witnet/devs ?

config/src/lib.rs
anler: @csmoe I tried again and now I remember, `Path::new` returns a reference type, that forces me to handle the lifetime and that makes my life more difficult so I decided to just go straight to the `String` type. Nevertheless if you want to give it a try if you think that with Path is going to be better.

config/src/loaders/toml.rs
```diff
@@ -0,0 +1,106 @@
...
+    from_str(&contents)
+}
+
+#[cfg(not(test))]
+fn read_file_contents(filename: &str, contents: &mut String) -> io::Result<usize> {
+    let mut file = File::open(filename)?;
+    file.read_to_string(contents)
+}
+
+#[cfg(test)]
```
mmartinbar: :heart:

config/src/loaders/toml.rs
anler: I don't see how I can use that currently, because in my case I'm not just defining a custom error, but wrapping two already existing ones: `std::io::Error` and `toml::de::Error`

config/src/lib.rs
mariocao: AFAIK, for now we are being inspired by Bitcoin parameters. They have 117 inbound peers, so we could set the limit to something similar (e.g. 128 if we want to have a power of two number :smile: ).

mariocao: The outbound limit is correct! :+1:

core/src/actors/config_manager.rs
```diff
@@ -0,0 +1,82 @@
...
+use witnet_config::loaders::toml;
+use witnet_config::Config;
+
+/// Config manager actor: manages the application configuration
+///
+/// This actor is in charge of reading the configuration for the
+/// application from a given source and using a given format, and
+/// supports messages for giving access to the configuration it holds.
+#[derive(Default, Debug)]
+pub struct ConfigManager {
```
mmartinbar: Please start this actor together with the rest of them [here](https://github.com/witnet/witnet-rust/blob/master/core/src/actors/node.rs)

core/tests/config_manager_test.rs
```diff
@@ -0,0 +1,47 @@
...
+
+        System::current().stop();
+        future::result(Ok(()))
+    }));
+
+    sys.run();
+}
+
+#[test]
+fn test_config_manager_load_config() {
```
mmartinbar: :1st_place_medal:

src/main.rs
```diff
@@ -24,58 +23,28 @@ fn main() {
...
-    let default_db_root = configuration
-        .server
-        .db_root
-        .unwrap_or_else(|| DB_ROOT.to_string());
+    let default_db_root = configuration.storage.db_path;
 
     // TODO
     let matches = app_from_crate!()
-        .subcommand(cli::node::get_arg(default_address))
+        .subcommand(cli::node::get_arg("127.0.0.1:1234"))
```
mmartinbar: I would use if possible the `default_address.to_string()`

core/src/actors/config_manager.rs
mmartinbar: Currently, this actor requires that a `LoadConfig` message is sent to it to load a configuration. Do you think it's reasonable for this actor to have a `new(filename: String)` method so that the configuration can be loaded before the actor is even started? This way, there would be no need to send an initial message to the actor for it to load its configuration.

mariocao: As discussed, we know that in the near future we will have to rethink how to start the actors so that, if a failure occurs, they could recover gracefully. However, I would suggest to move (having the Config Manager using a `witnet.toml` file) in order to boost the integration of all the current Managers. :tada: What do you think guys?

PR #71, opened 2018-10-30
feat(core): add encoder and decoder traits for p2p codec

Tommytrg:
Encoder and decoder trait implementation for P2PCodec

PR #73, opened 2018-10-31
feat(core): add encoder and decoder traits for p2p codec v2

tmpolaczyk:
Followup to https://github.com/witnet/witnet-rust/pull/71


core/src/actors/codec.rs
```diff
@@ -25,25 +25,34 @@ pub enum Response {
...
         let mut ftb: Option<Self::Item> = None;
-        let msg_len = src.len() as u16;
+        let msg_len = src.len();
         if msg_len > 2 {
-            let mut header_vec = Cursor::new([src.to_vec()[0], src.to_vec()[1]]);
-            let msg_size = header_vec.read_u16::<BigEndian>().unwrap();
-            if msg_len >= (msg_size + HEADER_SIZE) {
-                src.split_to(HEADER_SIZE as usize);
-                ftb = Some(Request::Message(src.split_to(msg_size as usize)));
+            let mut header_vec = Cursor::new(&src[0..2]);
```
Tommytrg: :ok_hand:

mariocao: :heart_eyes:

core/src/actors/codec.rs
```diff
@@ -29,21 +45,17 @@ impl Decoder for P2PCodec {
...
+        let msg_len = src.len();
+        if msg_len > 2 {
+            let mut header_vec = Cursor::new(&src[0..2]);
+            let msg_size = header_vec.read_u16::<BigEndian>().unwrap() as usize;
+            if msg_len >= msg_size + HEADER_SIZE {
+                src.split_to(HEADER_SIZE);
+                ftb = Some(Request::Message(src.split_to(msg_size)));
+            }
         }
+        Ok(ftb)
```
mariocao: I liked it! Perhaps, I would include some comments of what is happening in case the *if* condition is not fulfilled. This would help to understand the code if future changes are needed. :relaxed:

core/src/actors/codec.rs
```diff
@@ -53,11 +65,26 @@ impl Encoder for P2PCodec {
...
+        let Response::Message(bytes) = msg;
+
+        let mut encoded_msg = vec![];
 
+        if bytes.len() > u16::max_value() as usize {
+            return Err(io::Error::new(
+                io::ErrorKind::InvalidInput,
+                format!("Message size {} bytes too big for u16", bytes.len()),
+            ));
+        }
```
mariocao: :1st_place_medal:

core/src/actors/codec.rs
```diff
@@ -29,21 +45,17 @@ impl Decoder for P2PCodec {
...
-
-            // Remove the last two bytes of the string (corresponding to \r\n)
-            res.truncate(res.len() - 2);
-
-            Ok(Some(Request::Message(res)))
-        } else {
-            Ok(None)
+        let mut ftb: Option<Self::Item> = None;
+        let msg_len = src.len();
+        if msg_len > 2 {
```
aesedepece: `2` her should be `HEADER_SIZE`

core/src/actors/codec.rs
aesedepece: :heart_decoration:

core/src/actors/codec.rs
mariocao: True, I missed that! :sweat_smile:

core/src/actors/codec.rs
mariocao: Shouldn't we use `HEADER_SIZE` here too?

```suggestion
            let mut header_vec = Cursor::new(&src[0..HEADER_SIZE]);
```

PR #75, opened 2018-10-31
docs(dev): improve doc for developers with mkdocs

lrubiorod:
This PR add instructions to install rust and mkdocs, and clang requirement.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -5,13 +5,25 @@ Witnet-rust is build using [The Rust Programming language](https://rust-lang-nur
 
 1. Install dependencies
     - Rust 1.31 (currently on the `nightly` release channel)
+    ```
+    $ curl https://sh.rustup.rs -sSf | sh
+    $ rustup default nightly
+    $ rustc --version
+    ```
+    - `clang` Clang compiler
     - `flatc` FlatBuffers compiler (optional, only if recompiling schemas) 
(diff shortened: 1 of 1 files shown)


docs/development.md
```diff
@@ -5,13 +5,25 @@ Witnet-rust is build using [The Rust Programming language](https://rust-lang-nur
 
 1. Install dependencies
     - Rust 1.31 (currently on the `nightly` release channel)
+    ```
+    $ curl https://sh.rustup.rs -sSf | sh
+    $ rustup default nightly
+    $ rustc --version
+    ```
+    - `clang` Clang compiler
```
mariocao: :1st_place_medal:

mariocao:
I would also update some instructions of the Development guide that are outdated regarding the `server` component. I think that right now we are calling it `node`.

For that reason, step 5 should be:
```mkdocs
$ RUST_LOG=witnet=trace cargo run node
```

Additionally, we should also update the `Justfile` and its documentation (also in `development.md`) accordingly,.

lrubiorod:
> I would also update some instructions of the Development guide that are outdated regarding the `server` component. I think that right now we are calling it `node`.
> 
> For that reason, step 5 should be:
> 
> ```
> $ RUST_LOG=witnet=trace cargo run node
> ```
> 
> Additionally, we should also update the `Justfile` and its documentation (also in `development.md`) accordingly,.

It is fixed in PR #80

mariocao:
This PR is superseded by #80

PR #76, opened 2018-11-02
refactor(core): add function in ConfigManager to send GetConfig message

mmartinbar:
This PR refactors the way actors communicate with the `ConfigManager`.

Some repetitive code was being used when an actor needed to communicate with the `ConfigManager` in order to get the configuration and extract some parameter from it.

With this PR, the `ConfigManager` provides a public method `send_get_config_request` that other actors can use to send a `GetConfig` message to the `ConfigManager`. Those other actors need to provide a function to process the configuration when received.

This PR implements this new way of communicating with `ConfigManager` for the actors: `PeersManager`, `SessionsManager` and `ConnectionsManager`.

### DIFF
--- a/core/src/actors/sessions_manager.rs
+++ b/core/src/actors/sessions_manager.rs
@@ -9,15 +9,12 @@ use actix::{
 
 use log::debug;
 
-use crate::actors::config_manager::{process_get_config_response, ConfigManager, GetConfig};
+use crate::actors::config_manager::send_get_config_request;
 use crate::actors::connections_manager::{ConnectionsManager, OutboundTcpConnect};
 use crate::actors::session::Session;
 
 use crate::actors::peers_manager::{GetPeer, PeersManager, PeersSocketAddrResult};
-use witnet_p2p::sessions::{
-    error::SessionsResult,
-    {SessionStatus, SessionType, Sessions},
-};
+use witnet_p2p::sessions::{error::SessionsResult, SessionStatus, SessionType, Sessions};
 
 /// Period of the bootstrap peers task (in seconds)
 const BOOTSTRAP_PEERS_PERIOD: u64 = 5;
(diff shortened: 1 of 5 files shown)


core/src/actors/sessions_manager.rs
```diff
@@ -116,6 +114,17 @@ impl SessionsManager {
             .map(actix::fut::ok)
             .unwrap_or_else(|| actix::fut::err(()))
     }
+
+    /// Method to process the configuration received from the config manager
+    fn process_config(config: &Config, act: &mut Self, _ctx: &mut <Self as Actor>::Context) {
```
csmoe: the signature is a bit strange, why not
```suggestion
    fn process_config(&mut self, config: &Config, _ctx: &mut <Self as Actor>::Context) {
```
and `process_config` impl'd for several managers, but the parameters are optional sometimes, it's not a trait method, why keep the signature unchanged as unneeded `_ctx` in this case?

mmartinbar: I agree that the signature of the `process_config` methods looks strange, I will change it to your suggestion, thanks!

The signature always has those three parameters (even if unneeded) because those `process_config` functions are passed as parameters to the `send_get_config_request` function:
```
/// Method to send a GetConfig message to the ConfigManager
pub fn send_get_config_request<T, U: 'static>(act: &mut T, ctx: &mut T::Context, process_config: U)
where
    T: Actor,
    T::Context: AsyncContext<T>,
    U: FnOnce(&Config, &mut T, &mut T::Context)
```
Some `process_config` functions do need the context to perform their functionality (for instance, the one in the `ConnectionsManager`), so this was the best I could think of so that all `process_config` functions got what they needed.

I'm open to suggestions/improvements if this looks weird!

core/src/actors/sessions_manager.rs
```diff
@@ -127,34 +136,8 @@ impl Actor for SessionsManager {
...
-                    .set_server_address(config.connections.server_addr);
-                act.sessions.set_limits(
-                    config.connections.inbound_limit,
-                    config.connections.outbound_limit,
-                );
-                actix::fut::ok(())
-            })
-            .wait(ctx);
+        // Send message to config manager and process its response
+        send_get_config_request(self, ctx, SessionsManager::process_config);
```
tmpolaczyk: You can use a closure instead of a function here, to avoid creating a new function:

```rust
// Send message to config manager and process its response
send_get_config_request(self, ctx, |s, _ctx, config| {
    s.sessions
        .set_server_address(config.connections.server_addr);
    s.sessions.set_limits(
        config.connections.inbound_limit,
        config.connections.outbound_limit,
    )
});
```

PR #80, opened 2018-11-02
feat(cli): read config filename

lrubiorod:
Remove unused arguments on cli/node.rs
Add "config" argument to read a config filename
Modify storage manager to read the data base path from config manager

PR #83, opened 2018-11-02
refactor(cli): rewrite cli with structopt

csmoe:
Instead of parsing args manually, generate a cli command with an exact struct/enum.

~~block until https://github.com/witnet/witnet-rust/pull/80 merged.~~


csmoe:
wit

tmpolaczyk:
Hi, thanks for the PR, I will review it.

In the future (in another PR), we would like to be able to change the configuration with command-line parameters, for example:

    cargo run -- node --inbound-limit 16

This would have higher precedence, but it won't modify the config file.

Would it be possible to integrate the `Config` struct from config/src/lib.rs with structopt?
We could create a new argument for each possible field from `Config`, but that would duplicate the code. Since structopt can already parse args into a struct, maybe there is some method to "update" a struct with new args? We could simply pass the additional arguments to `ConfigManager`:

```rust
// core/src/actors/node.rs
let config_manager_addr = ConfigManager::new_with_extra_args(config_filename, args).start();
```
And then
* ConfigManager reads config from filename
* ConfigManager updates config according to extra args

core/src/actors/config_manager.rs
```diff
@@ -15,40 +16,40 @@ pub const CONFIG_DEFAULT_FILENAME: &str = "witnet.toml";
...
     config: Config,
-    filename: Option<String>,
+    config_file: PathBuf,
+}
+
+impl Default for ConfigManager {
+    fn default() -> Self {
+        Self {
+            config: Config::default(),
+            config_file: PathBuf::from(CONFIG_DEFAULT_FILENAME),
```
tmpolaczyk: :heart:

data_structures/src/lib.rs
```diff
@@ -2,7 +2,6 @@
 // To enable `#[allow(clippy::all)]`
 //#![feature(tool_lints)]
 
-
```
tmpolaczyk: Remove data_structures/ from the commit

csmoe:
@tmpolaczyk yep, that's kind of easy based on my previous experience with cli stuff.

src/cli.rs
```diff
@@ -1,3 +1,60 @@
...
+
+        /// Config file path
+        #[structopt(name = "config", short = "c")]
+        #[structopt(parse(from_os_str))]
+        config: Option<PathBuf>,
+
+        /// Run the server in the background
+        #[structopt(name = "background", short = "b")]
+        background: bool,
+    },
```
tmpolaczyk: Currently only the `config` parameter is used, the other ones are not implemented so they can be removed to avoid confusion.

src/main.rs
```diff
@@ -5,55 +5,31 @@
...
-            })
-            .expect("Server error");
-        }
-
-        // unknown run mode
-        _ => {
-            println!("Unrecognized command. Run with '--help' to learn more.");
-            exit(1);
+    if let Err(e) = run() {
+        eprintln!("Error: {}", e);
```
tmpolaczyk: Use `error!` (`log::error`) for printing errors.

data_structures/src/lib.rs
csmoe: this diff is generated by rustfmt, the code should be formatted as listed in `Justfile` taskes.

tmpolaczyk: True, but you should put it in a separated commit: `refactor(data_structures): fix format`

tmpolaczyk:
Sorry but we can't merge this PR because some of the commits are not signed :sweat_smile:
For your convenience, the following command will sign the last 5 commits:

    git rebase --exec 'git commit --amend --no-edit -n -S' -i HEAD~5

PR #84, opened 2018-11-02
fix(travis): Switch to beta Rust version (1.31)

tmpolaczyk:
The minimum Rust version required for this project is 1.31 because of the use of the 2018 edition.

This PR changes the version from nightly to beta, and removes the use of the unstable `try_from` feature (only available in nightly), which is replaced with a `TryFrom` in `data_structures/src/serializers.rs`, at least until `try_from` is stabilized:

```rust
pub trait TryFrom<T>: Sized {
    type Error;

    fn try_from(value: T) -> Result<Self, Self::Error>;
}
```

To switch between the std `TryFrom` and "our" `TryFrom`, just change the import:

```rust
//use std::convert::TryFrom;
use witnet_data_structures::serializers::TryFrom;
```

### DIFF
--- a/data_structures/src/serializers.rs
+++ b/data_structures/src/serializers.rs
@@ -1,4 +1,4 @@
-use std::convert::{Into, TryFrom};
+use std::convert::Into;
 extern crate flatbuffers;
 
 use crate::flatbuffers::protocol_generated::protocol::{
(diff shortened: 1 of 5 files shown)


data_structures/src/serializers.rs
```diff
@@ -104,7 +104,13 @@ struct VersionWitnetArgs {
     version: u32,
 }
 
-impl TryFrom<Vec<u8>> for Message {
+pub trait MyTryFrom<T>: Sized {
```
csmoe:
```suggestion
pub trait TryFrom<T>: Sized {
```
`TryFrom` from `std` can be overlapped locally.

tmpolaczyk: True :+1:

aesedepece:
Merged as 4bfcf9b25c462bc08e6963dd71fd99c28cfd495f.

PR #94, opened 2018-11-06
refactor(p2p): add separate consolidated/unconsolidated collections for outbound sessions

mmartinbar:
This PR adds separate consolidated / unconsolidated collections for outbound sessions in `p2p sessions library`. The p2p sessions library is the one that contains the state and most of the logic of the session manager. Inbound sessions are left in the same collection, no matter what their status is.

Tests for the `p2p sessions library` have been adapted to this new logic. Also, the actors that used that refactored logic in some way have been adapted.

Additionally, some extra logic that was left in the `SessionsManager` actor has been extracted from it and written in the library instead. New tests have been implemented for this new logic.

### DIFF
--- a/p2p/src/sessions/mod.rs
+++ b/p2p/src/sessions/mod.rs
@@ -31,14 +31,20 @@ pub enum SessionStatus {
 
 /// Sessions struct contains:
 /// - server address used to listen to incoming connections
-/// - lists of of inbound and outbound sessions parametrized with their reference (T)
+/// - list of inbound sessions parametrized with their reference (T)
+/// - list of consolidated outbound sessions parametrized with their reference(T)
+/// - list of unconsolidated outbound sessions parametrized with their reference(T)
 pub struct Sessions<T> {
     /// Server address listening to incoming connections
     pub server_address: Option<SocketAddr>,
     /// Inbound sessions: __untrusted__ peers that connect to the server
-    pub inbound_sessions: BoundedSessions<T>,
-    /// Outbound sessions: __known__ peers that the node is connected
-    pub outbound_sessions: BoundedSessions<T>,
+    pub inbound: BoundedSessions<T>,
+    /// Outbound consolidated sessions: __known__ peer sessions that the node is connected to (in
+    /// consolidated status)
+    pub outbound_consolidated: BoundedSessions<T>,
+    /// Outbound unconsolidated sessions: __known__ peer sessions that the node is connected to
+    /// (in unconsolidated status)
+    pub outbound_unconsolidated: BoundedSessions<T>,
 }
 
 /// Default trait implementation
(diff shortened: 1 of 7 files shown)


aesedepece:
In the absence of an issue for this PR, we should give it an estimate so the effort is reflected in our sprint.

p2p/src/sessions/mod.rs
```diff
@@ -47,41 +53,69 @@ impl<T> Default for Sessions<T> {
...
+            .collection
+            .contains_key(&candidate_addr);
+        let is_outbound_unconsolidated = self
+            .outbound_unconsolidated_sessions
             .collection
             .contains_key(&candidate_addr);
+
         // Check if address is the server address
         let is_server =
             self.server_address.is_some() && self.server_address.unwrap() == candidate_addr;
```
aesedepece: Maybe for the sake of uniformity and avoiding wild unwraps:
```suggestion
            self.server_address.map(|address| address == candidate_addr).unwrap_or(false);
```

p2p/src/sessions/mod.rs
```diff
@@ -31,14 +31,20 @@ pub enum SessionStatus {
...
 /// - server address used to listen to incoming connections
-/// - lists of of inbound and outbound sessions parametrized with their reference (T)
+/// - list of inbound sessions parametrized with their reference (T)
+/// - list of consolidated outbound sessions parametrized with their reference(T)
+/// - list of unconsolidated outbound sessions parametrized with their reference(T)
 pub struct Sessions<T> {
     /// Server address listening to incoming connections
     pub server_address: Option<SocketAddr>,
     /// Inbound sessions: __untrusted__ peers that connect to the server
     pub inbound_sessions: BoundedSessions<T>,
```
csmoe: the name of `xxxx_session` fields is kind of long, could we just call them `xxxx` since they are wrapped in `Session` "already"?

anler:
Not sure if these improvements belong here, but regarding the P2P library I'm missing:

1. More documentation at the library level telling me what is the purpose of the library and what one should expect from it
image

as an example this is what the config library looks like:
image

2. I'm missing more detail when describing the purpose of the modules, for example, in the sessions module says:
> Library for managing the node sessions, including outbound and inbound sessions
I would like that to also explain what is a session, and what does it mean to be outbound and inbound. Same for the rest of the modules in the crate.

mmartinbar:
@anler, love your suggestion about the documentation, 100% agree. As this PR is blocking other issues, if it's ok for you I'll merge it as it is, and then create an issue to improve the p2p library documentation so that we do not forget about it.

mmartinbar:
Created an issue to implement requested changes in documentation: [#106](https://github.com/witnet/witnet-rust/issues/106)

aesedepece:
So happy to see this merged 🤩

PR #104, opened 2018-11-06
feat(storage): create Storable trait

tmpolaczyk:
Define a `Storable` trait which defines a conversion from a given type to raw bytes.
The default implementation uses MessagePack, but since the encoding is specific to each type, users can create their own encoders/decoders.

### DIFF
--- a/docs/architecture/storage.md
+++ b/docs/architecture/storage.md
@@ -153,7 +153,90 @@ These are the specific types for this implementor:
 
 The full source code of the `Storage` implementor for `RocksStorage` can be found at [`rocks.rs`][rocks].
 
+## `Storable` trait
+
+The `Storable` trait defines a conversion from any type to bytes.
+It is useful because the storage backends expect keys and values to be raw bytes,
+so the data needs to be serialized and deserialized.
+
+The simplest way to implement this trait is to add
+`#[derive(Serialize, Deserialize)]` from `serde` to the type definition:
+
+```rust
+#[derive(Serialize, Deserialize)]
+struct Foo {
+    data: Vec<String>
+}
+```
+
+The default implementation uses [MessagePack][msgpack], but the implementor is free to choose
+a different encoding for their custom types.
+
+The preferred way to work with this trait is using the `StorageHelper`,
+described below:
+
+## `StorageHelper` trait
+
+To enable better ergonomics when working with the storage, users can
+import the `StorageHelper` trait which adds two additional methods to
+the `Storage`:
+
+```rust
+pub trait StorageHelper {
+    /// Insert an element into the storage
+    fn put_t<T: Storable>(&mut self, key: &[u8], value: T) -> StorageResult<()>;
+    /// Get an element from the storage
+    fn get_t<T: Storable>(&mut self, key: &[u8]) -> StorageResult<Option<T>>;
+}
+```
+
+This trait is implemented by default for all the storage backends which
+work with raw bytes.
+
+It allows for inserting and removing typed values, as long as the types
+implement the `Storable` trait.
+
+### Example
+
+```rust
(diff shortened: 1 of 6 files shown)


tmpolaczyk:
Unresolved questions:

* Are `get_t` and `put_t` good names?

aesedepece:
Some suggestions for naming:
- `put_object` / `get_object`
- `put_storable` / `get_storable`
- `serialize_and_put` / `get_and_deserialize`
- `write` / `read`
- `store` / `recover`
- `insert` / `find`

aesedepece:
@witnet/devs feedback on naming :pray:

tmpolaczyk:
Currently the only way to interact with the storage is by sending messages to the `StorageManager`, and since actix enforces some limitations on the messages, we can't implement a generic `Get` message.

A generic `Put` can be implemented, but it does not need to be, since we can convert any type to bytes before sending the message.

Therefore, `get_t` and `put_t` won't be as useful as expected, so maybe we can keep the temporary names?

docs/architecture/storage.md
```diff
@@ -153,7 +153,90 @@ These are the specific types for this implementor:
...
+pub trait StorageHelper {
+    /// Insert an element into the storage
+    fn put_t<T: Storable>(&mut self, key: &[u8], value: T) -> StorageResult<()>;
+    /// Get an element from the storage
+    fn get_t<T: Storable>(&mut self, key: &[u8]) -> StorageResult<Option<T>>;
+}
+```
+
+This trait is implemented by default for all the storage backends which
+work on raw bytes.
```
jrmoreau: "on" or "with" raw bytes?

storage/src/storage.rs
```diff
@@ -22,3 +24,108 @@ pub trait Storage<ConnData: Debug, Key, Value> {
     /// Delete an entry from the storage, identified by its key.
     fn delete(&mut self, key: Key) -> StorageResult<()>;
 }
+
+/// Trait which marks a type as storable.
+/// The simplest way to implement this trait
+/// is to add `#[derive(Serialize, Deserialize)]` to the type definition.
+/// The storage works on raw bytes, so we need to serialize and deserialize the data.
+/// The default implementation uses MessagePack, but the implementor is free to chose
```
jrmoreau: "choose" rather than "chose"

storage/src/storage.rs
```diff
@@ -22,3 +24,108 @@ pub trait Storage<ConnData: Debug, Key, Value> {
...
+/// a different encoding for their custom types.
+pub trait Storable: Sized {
+    /// Convert `Self` into `Vec<u8>`
+    fn to_bytes(&self) -> StorageResult<Vec<u8>>;
+    /// Convert `Vec<u8>` into `Self
+    fn from_bytes(x: &[u8]) -> StorageResult<Self>;
+}
+
+// By default, mark as `Storable` all the types which can be serialized and deserialized
```
jrmoreau: // By default, mark all the types which can be serialized and deserialized
// using serde as `Storable`.

docs/architecture/storage.md
```diff
@@ -153,7 +153,90 @@ These are the specific types for this implementor:
...
+`#[derive(Serialize, Deserialize)]` from `serde` to the type definition:
+
+```rust
+#[derive(Serialize, Deserialize)]
+struct Foo {
+    data: Vec<String>
+}
+```
+
+The default implementation uses [MessagePack][msgpack], but the implementor is free to chose
```
jrmoreau: "choose" instead of "chose"

tmpolaczyk:
Fixed, thank you!

aesedepece:
Merged as 8a5ddc082bf9e77eea72feddf796b576267ce560.

Issue #92, opened 2018-11-06
Add "GetPeers" handler to Session actor


aesedepece:
#### Context
- Each `Session` actor should be capable of handling `GetPeers` messages (normally coming from the `Anycast` handler of `SessionManager`).
- These `GetPeers` messages are actor messages, not protocol messages (`SessionsManager` should never construct protocol messages, and `Session` should construct them but never handle them from the outside).
- Inside this `GetPeers` handler, a `GetPeers` protocol message should be constructed, serialized and piped to the connection. Serialization can be centralized using a special handler that takes any protocol message, encodes it into bytes and write it into the connection.

#### Actionables
- Create `ProtocolMessage` protocol message handler in `Session`. This should encode the input `ProtocolMessage`
- Create `GetPeers` actor message handler in `Session`. This should construct a `GetPeers` protocol message and forward it to the `Session` itself so that it is encoded and piped into the connection.

#### Remarks
- The `ProtocolMessage` name used in this issue description is currently called `Message` in the source code. This can be quite confusing because it may lead to confusion between protocol messages and actor messages. For the sake of clarity, we should take this opportunity to rename it to `ProtocolMessage`.

PR #109, opened 2018-11-07
feat(core): add anycast handler in SessionsManager

mmartinbar:
This PR adds a message handler in `SessionsManager` for the so-named anycast messages.

Anycast messages are messages that are sent to the `SessionsManager` containing some Actix Session message. The handler for the anycast messages just selects a random consolidated outbound session and sends that actix message to the random session actor.

Upon reception of that message, the session actor will perform whichever task is needed.

The function to select a random consolidated outbound session is in the `p2p sessions library`. Tests have been added for it.

Also, a very basic `GetPeers` handler has been implemented in the `Session` actor so that the Anycast handler could be tested. It should be completed in issue #92  .

Additionally, some functionality has been included in the `data_structures` library to define builder functions for each protocol message. Tests have been added for this as well.


data_structures/src/builders.rs
```diff
@@ -0,0 +1,133 @@
+extern crate rand;
```
csmoe: unneeded in edition-2018.

p2p/tests/sessions.rs
```diff
@@ -217,46 +209,103 @@ fn p2p_sessions_is_outbound_bootstrap_needed() {
...
+        .is_ok());
+    assert!(sessions
+        .consolidate_session(SessionType::Outbound, outbound_address)
+        .is_ok());
+
+    // Get random session for a "big" number
+    let mut diff: i16 = 0;
+    for _ in 0..100000 {
+        // Get a random anycast sessions (there are only 2)
+        match sessions.get_random_anycast_session() {
```
csmoe: does `&sessions.get_random_anycast_session()` help here instead of clone later?

data_structures/src/builders.rs
```diff
@@ -0,0 +1,133 @@
...
+pub const CAPABILITIES: u64 = 0x0000_0000_0000_0001;
+
+/// User agent
+pub const USER_AGENT: &str = "full-node-desktop-edition";
+
+/// Genesis block
+pub const GENESIS: u64 = 0x0123_4567_89AB_CDEF;
+
+////////////////////////////////////////////////////////////////////////////////////////
+// BUILDER PUBLIC FUNCTIONS
```
csmoe: I see...these builder generate a `Message` with `Command`, so why not `impl Message {}` instead of setting them free.

data_structures/src/builders.rs
```diff
@@ -0,0 +1,133 @@
...
+
+/// Function to get a random nonce
+fn random_nonce() -> u64 {
+    thread_rng().gen()
+}
+
+/// Function to get current timestamp (ms since Unix epoch)
+fn current_timestamp() -> u64 {
+    let now = SystemTime::now();
+    let now_duration = now.duration_since(UNIX_EPOCH).unwrap();
```
csmoe: I'm personally afraid of such kind of "brutely" unwrap everytime :smile: 
thus, how about ask help from `precise_time_ns -> u64` in the offical [rust-lang/time](https://docs.rs/time/0.1.40/time/fn.precise_time_ns.html) crate?

p2p/tests/sessions.rs
mmartinbar: Sure, thanks!

data_structures/src/builders.rs
mmartinbar: I think that what we need in order to fill the timestamp field of the `Version` messages is a Unix timestamp (seconds or milliseconds since the Unix epoch). I see at the [`precise_time_ns`](https://docs.rs/time/0.1.40/time/fn.precise_time_ns.html) documentation that it uses an unspecified epoch so I'm not sure if we can use it.

I agree that the unwrap is dangerous.  I think that in the future this function may belong to the `EpochManager` which will detect if the system time is before the Unix epoch. If that happens, I think that it is ok the Witnet node fails. But anyway, I agree that we should control that error better.

Would it be ok if I returned a `Result` wrapping the timestamp?

csmoe: oh, sorry, my bad. timestamp should be a raw `u64`, `Result` mighty be unnecessarily complicated , so  just `.expect("err message")`  here?

mmartinbar: Perfect, I'll change it, thanks!

mmartinbar: The team told me that there is already a function in [`util`](https://github.com/witnet/witnet-rust/blob/master/util/src/timestamp.rs#L6) doing this, my bad :woman_facepalming: . @csmoe, would it be ok for you if I use the one that's already defined? It already wraps the error in a `WitnetResult` to be handled by the caller.

csmoe: @mmartinbar that's ok :)

mmartinbar: The team also told me to refactor the [`get_timestamp`](https://github.com/witnet/witnet-rust/blob/master/util/src/timestamp.rs#L6) function to use [**chrono**](https://docs.rs/chrono/0.4.0/chrono/struct.DateTime.html#method.timestamp).

Changes have been made across the code to change the timestamp type from u64 to i64 (flatbuffer schema changed, our own types as well...).

aesedepece:
This is looking super good @mmartinbar!

What do you think of starting to split our actor files into folders like this?:
- `/actors`
  - `/session`
    - `/mod.rs` (`struct Session`)
    - `/actor.ts` (`impl Actor for Session` and other actor `impl`s for functions)
    - `/messages.rs` (Actor messages)
    - `/handlers.rs` (Handlers `impl`s. Normally for actor messages, but also protocol messages in the case of `Session`)

Any counterproposal @witnet/devs ?

mmartinbar:
Love your proposal @aesedepece ! As this issue is blocking some other issues and that is quite a huge refactor, would you be ok with leaving it for a separate issue?

aesedepece:
@mmartinbar Moving my proposal to a separate issue.

PR #113, opened 2018-11-08
feat(managers): periodically persist peers into storage 

### LINKED ISSUE #97: Persist peers into local storage periodically
#### Context
- Peer addresses obtained from our peers through the peer discovery protocol need to be persisted from volatile memory into local storage to enhance trustlessness of future cold-bootstrapping when restarting the node.
- This process doesn't necessarily need to be performed every time the peer addresses list is updated. Instead, we can just back it up periodically.
- The peers backup period should be configurable.

#### Actionables
- Add a variable for the peers backup period in the configuration file.
- Make `ConfigManager` read the peers backup period variable.
- Make `PeersManager` query `ConfigManager` for the value of the peers backup period variable.
- Make the peer addresses list implement the `Storable` trait.
- Make `PeersManager` schedule the serialization and storage of the peer addresses list by sending it to the `StorageManager` through `Put`.
- Document the process in the project documentation (mkdocs)

lrubiorod:
Define a new config parameter named storage_peers_period.
Implement a function on Peers Manager to dump periodically peers into storage through sending a Put message to Storage Manager
Add documentation to peers manager doc 

Close #97

### DIFF
--- a/core/src/actors/peers_manager.rs
+++ b/core/src/actors/peers_manager.rs
@@ -1,9 +1,15 @@
 use std::net::SocketAddr;
+use std::time::Duration;
 
-use actix::{Actor, Context, Handler, Message, Supervised, SystemService};
+use actix::{
+    Actor, ActorFuture, AsyncContext, Context, ContextFutureSpawner, Handler, Message, Supervised,
+    System, SystemService, WrapFuture,
+};
 use log::{debug, error};
 
 use crate::actors::config_manager::send_get_config_request;
+use crate::actors::storage_keys::PEERS_KEY;
+use crate::actors::storage_manager::{Put, StorageManager};
 
 use witnet_p2p::peers::{error::PeersResult, Peers};
 
@@ -20,6 +26,37 @@ pub struct PeersManager {
     peers: Peers,
 }
 
+impl PeersManager {
+    /// Method to periodically persist peers into storage
+    fn persist_peers(&self, ctx: &mut Context<Self>, storage_peers_period: Duration) {
+        // Schedule the discovery_peers with a given period
+        ctx.run_later(storage_peers_period, move |act, ctx| {
+            // Get storage manager address
+            let storage_manager_addr = System::current().registry().get::<StorageManager>();
+
+            // Persist peers into storage. `AsyncContext::wait` registers
+            // future within context, but context waits until this future resolves
+            // before processing any other events.
+            storage_manager_addr
+                .send(Put::from_value(PEERS_KEY, &act.peers).unwrap())
+                .into_actor(act)
+                .then(|res, _act, _ctx| {
+                    match res {
+                        Ok(Ok(_)) => debug!("PeersManager successfully persist peers to storage"),
+                        _ => {
+                            debug!("Peers manager persist peers to storage failed");
+                            // FIXME(#72): handle errors
+                        }
+                    }
+                    actix::fut::ok(())
+                })
+                .wait(ctx);
+
+            act.persist_peers(ctx, storage_peers_period);
+        });
+    }
+}
+
 /// Make actor from `PeersManager`
 impl Actor for PeersManager {
     /// Every actor has to provide execution `Context` in which it can run.
@@ -29,15 +66,21 @@ impl Actor for PeersManager {
         debug!("Peers Manager actor has been started!");
 
         // Send message to config manager and process response
-        send_get_config_request(self, ctx, |s, _ctx, config| {
+        send_get_config_request(self, ctx, |act, ctx, config| {
             // Get known peers
-            let known_peers = config.connections.known_peers.iter().cloned().collect();
+            let known_peers: Vec<_> = config.connections.known_peers.iter().cloned().collect();
+
+            // Get storage peers period
+            let storage_peers_period = config.connections.storage_peers_period;
 
             // Add all peers
-            match s.peers.add(known_peers) {
+            match act.peers.add(known_peers.clone()) {
                 Ok(peers) => debug!("Added the following peer addresses: {:?}", peers),
                 Err(e) => error!("Error when adding peer addresses: {}", e),
             }
+
+            // Start the storage peers process on sessions manager start
+            act.persist_peers(ctx, storage_peers_period);
         });
     }
 }
--- a/docs/architecture/managers/peers-manager.md
+++ b/docs/architecture/managers/peers-manager.md
@@ -92,11 +92,10 @@ The way other actors will communicate with the storage manager is:
 
 These are the messages sent by the peers manager:
 
-| Message   | Destination   | Input type | Output type                 | Description               |
-| --------- | ------------- | ---------- | --------------------------- | ------------------------- |
-| GetConfig | ConfigManager | `()`       | `Result<Config, io::Error>` | Request the configuration |
-
-It is foreseen that in the future, the peers manager will call the `Storage Manager` to store a list of known peers.
+| Message     | Destination       | Input type                                | Output type                 | Description                               |
+| ----------- | ----------------- | ----------------------------------------- | --------------------------- | ----------------------------------------- |
+| `GetConfig` | `ConfigManager`   | `()`                                      | `Result<Config, io::Error>` | Request the configuration                 |
+| `Put`       | `StorageManager`  | `&'static [u8]`, `Vec<u8>`                | `StorageResult<()>`         | Wrapper to RocksStorage `put()` method    |
 
 #### GetConfig
 
@@ -104,9 +103,18 @@ This message is sent to the [`ConfigManager`][config_manager] actor when the pee
 
 The return value is used to initialize the list of known peers. For further information, see  [`ConfigManager`][config_manager].
 
+#### Put
+
+This message is sent to the [`StorageManager`][storage_manager] actor periodically using a period 
+obtained from [`ConfigManager`][config_manager]
+
+The return value is used to check if the storage process has been successful.
+
 ## Further information
 
 The full source code of the `PeersManager` can be found at [`peers_manager.rs`][peers_manager].
 
 [peers]: https://github.com/witnet/witnet-rust/blob/master/p2p/src/peers.rs
 [peers_manager]: https://github.com/witnet/witnet-rust/blob/master/core/src/actors/peers_manager.rs
+[config_manager]: https://github.com/witnet/witnet-rust/blob/master/core/src/actors/config_manager.rs
+[storage_manager]: https://github.com/witnet/witnet-rust/blob/master/core/src/actors/storage_manager.rs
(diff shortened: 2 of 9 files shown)


config/src/lib.rs
```diff
@@ -59,6 +59,10 @@ pub struct ConnectionsConfig {
     /// used as a bootstrap mechanism to gain access to the P2P
     /// network
     pub known_peers: Vec<SocketAddr>,
+
+    #[serde(default = "ConnectionsConfig::default_storage_peers_period")]
+    /// Period to dump peers into storage
+    pub storage_peers_period: u64,
```
tmpolaczyk: It is more idiomatic to store a time period using the [`Duration` type](https://doc.rust-lang.org/std/time/struct.Duration.html):
```suggestion
    pub storage_peers_period: Duration,
```

And then use it as:
```rust
let period = Duration::from_secs(30);
let seconds = period.as_secs();
```

core/src/actors/peers_manager.rs
```diff
@@ -20,6 +25,42 @@ pub struct PeersManager {
...
+                    .send(Put::from_value(PEERS_KEY, &act.peers).unwrap())
+                    .into_actor(act)
+                    .then(|res, _act, ctx| {
+                        match res {
+                            Ok(Ok(_)) => debug!("Peers manager successfully dump peers to storage"),
+                            _ => {
+                                debug!("Peers manager dump peers to storage failed");
+                                // FIXME(#72): a full stop of the session is not correct (unregister should
+                                // be skipped)
+                                ctx.stop()
```
tmpolaczyk: The `PeersManager` should not stop itself because of an error in the `StorageManager` 
```suggestion
```

config/src/lib.rs
lrubiorod: done!!

core/src/actors/peers_manager.rs
lrubiorod: done!!

core/src/actors/peers_manager.rs
```diff
@@ -20,6 +25,37 @@ pub struct PeersManager {
     peers: Peers,
 }
 
+impl PeersManager {
+    /// Method to periodically dump peers into storage
+    fn storage_peers(&self, ctx: &mut Context<Self>, storage_peers_period: Duration) {
+        // Schedule the discovery_peers with a given period
+        ctx.run_later(storage_peers_period, move |act, ctx| {
```
aesedepece: If `act` stands for actor, I'd rather use the whole word.
I'm OK with `ctx` because that's a generally accepted and well-known apocope.

core/src/actors/peers_manager.rs
```diff
@@ -20,6 +25,37 @@ pub struct PeersManager {
...
+            // Get storage manager address
+            let storage_manager_addr = System::current().registry().get::<StorageManager>();
```
aesedepece: Can't see why would it be necessary to get this address every time. Why not querying it and storing in the actor state upon start up?

core/src/actors/peers_manager.rs
```diff
@@ -20,6 +25,37 @@ pub struct PeersManager {
...
+
+            // Dump peers into storage. `AsyncContext::wait` register
```
aesedepece:
```suggestion
            // Dump peers into storage. `AsyncContext::wait` registers
```

core/src/actors/peers_manager.rs
```diff
@@ -20,6 +25,37 @@ pub struct PeersManager {
...
+            // future within context, but context waits until this future resolves
+            // before processing any other events.
+            storage_manager_addr
+                .send(Put::from_value(PEERS_KEY, &act.peers).unwrap())
+                .into_actor(act)
+                .then(|res, _act, _ctx| {
+                    match res {
+                        Ok(Ok(_)) => debug!("Peers manager successfully dump peers to storage"),
```
aesedepece:
```suggestion
                        Ok(Ok(_)) => debug!("PeersManager successfully persisted peers into storage."),
```

core/src/actors/peers_manager.rs
```diff
@@ -20,6 +25,37 @@ pub struct PeersManager {
...
+            // Get storage manager address
+            let storage_manager_addr = System::current().registry().get::<StorageManager>();
+
+            // Dump peers into storage. `AsyncContext::wait` register
+            // future within context, but context waits until this future resolves
+            // before processing any other events.
+            storage_manager_addr
+                .send(Put::from_value(PEERS_KEY, &act.peers).unwrap())
+                .into_actor(act)
+                .then(|res, _act, _ctx| {
```
aesedepece:
```suggestion
                .then(|res, _actor, _ctx| {
```

core/src/actors/peers_manager.rs
```diff
@@ -29,15 +65,21 @@ impl Actor for PeersManager {
         debug!("Peers Manager actor has been started!");
 
         // Send message to config manager and process response
-        send_get_config_request(self, ctx, |s, _ctx, config| {
+        send_get_config_request(self, ctx, |act, ctx, config| {
```
aesedepece:
```suggestion
        send_get_config_request(self, ctx, |actor, ctx, config| {
```

core/src/actors/storage_manager.rs
```diff
@@ -13,6 +13,9 @@ use witnet_util::error::WitnetError;
 type ValueStorageResult<T> = StorageResult<Option<T>>;
 type UnitStorageResult = StorageResult<()>;
 
+/// Constant to specify the peers key for the storage
+pub static PEERS_KEY: &'static [u8] = b"peers";
```
aesedepece: I don't think this constant belongs here. We may need a separate file somewhere for static storage keys.

docs/architecture/managers/peers-manager.md
```diff
@@ -92,21 +92,28 @@ The way other actors will communicate with the storage manager is:
...
 These are the messages sent by the peers manager:
 
-| Message   | Destination   | Input type | Output type                 | Description               |
-| --------- | ------------- | ---------- | --------------------------- | ------------------------- |
-| GetConfig | ConfigManager | `()`       | `Result<Config, io::Error>` | Request the configuration |
-
-It is foreseen that in the future, the peers manager will call the `Storage Manager` to store a list of known peers.
+| Message   | Destination     | Input type                                | Output type                 | Description                               |
+| --------- | --------------- | ----------------------------------------- | --------------------------- | ----------------------------------------- |
+| GetConfig | ConfigManager   | `()`                                      | `Result<Config, io::Error>` | Request the configuration                 |
```
aesedepece:
```suggestion
| `GetConfig` | `ConfigManager`   | `()`                                      | `Result<Config, io::Error>` | Request the configuration                 |
```

docs/architecture/managers/peers-manager.md
```diff
@@ -92,21 +92,28 @@ The way other actors will communicate with the storage manager is:
...
+| Put       | StorageManager  | `&'static [u8]`, `Vec<u8>`                | `StorageResult<()>`         | Wrapper to RocksStorage `put()` method    |
```
aesedepece:
```suggestion
| `Put`       | `StorageManager`  | `&'static [u8]`, `Vec<u8>`                | `StorageResult<()>`         | Wrapper to RocksStorage `put()` method    |
```

aesedepece: I don't feel comfortable with the mention to `RocksStorage` here. This actor shouldn't give a ʞɔnɟ for the concrete storage backend we use, and neither should its documentation.

docs/architecture/managers/peers-manager.md
```diff
@@ -92,21 +92,28 @@ The way other actors will communicate with the storage manager is:
...
 
 #### GetConfig
 
 This message is sent to the [`ConfigManager`][config_manager] actor when the peers manager actor is started.
 
 The return value is used to initialize the list of known peers. For further information, see  [`ConfigManager`][config_manager].
 
+#### Put
+
+This message is sent to the [`StorageManager`][storage_manager] actor periodically using a period obtained from config file.
```
aesedepece: The period is not necessarily taken from the config file as it can be overridden by CLI arguments. All this actor knows about is the `ConfigManager`.

config/src/lib.rs
```diff
@@ -110,6 +115,10 @@ impl ConnectionsConfig {
     pub fn default_known_peers() -> Vec<SocketAddr> {
         Vec::default()
     }
+
+    fn default_storage_peers_period() -> Duration {
+        Duration::from_secs(30)
```
aesedepece: This is a magic number, please:
- Move it to a more visible spot
- Make it static
- Add a comment for it, potentially reasoning about considerations when deciding the value

core/src/actors/peers_manager.rs
tmpolaczyk: I disagree. We are already using `|act, ctx|` in many places in the code so there is no point in changing it now. In the actix documentation they use `act` so I think it's better to use their conventions.

aesedepece: image

core/src/actors/peers_manager.rs
tmpolaczyk: Are we sure that the address will not change when the storage actor restarts? I think that the added code complexity does not justify the potential performance improvement because:

* In actix, the address is stored in a hashmap

* This function is called once every 30 seconds

And if we decide to store the address in the actor state, should all the actors store addresses to their most used actors?

core/src/actors/peers_manager.rs
```diff
@@ -20,6 +25,37 @@ pub struct PeersManager {
     peers: Peers,
 }
 
+impl PeersManager {
+    /// Method to periodically dump peers into storage
+    fn storage_peers(&self, ctx: &mut Context<Self>, storage_peers_period: Duration) {
```
tmpolaczyk: `storage_peers` should be renamed to a more descriptive name: `dump_peers`, `store_peers`, ...

core/src/actors/peers_manager.rs
aesedepece: > And if we decide to store the address in the actor state, should all the actors store addresses to their most used actors?

That depends on whether we assume manager actors' lifespan to be everlasting or that some of them can die while others are still alive and thus need fresh addresses.

None of those assumptions are explicitly stated anywhere. If we all agree that we are better off taking the second assumption, I withdraw my former comment.

core/src/actors/storage_manager.rs
lrubiorod: Where do you think is the best option to storage it?

aesedepece:
We are in time for changing `dump` to `persist` for the sake of adopting a more formal register in our comments and documentation.

core/src/actors/peers_manager.rs
lrubiorod: If you don't mind, I think we could discuss it tomorrow. As @tmpolaczyk wrote, if we save the addresses in the state, we should also modify other actors to do the same. What do you think we create an issue for that and we finish this PR?

core/src/actors/storage_manager.rs
lrubiorod: Created a storage keys file to solve this

PR #116, opened 2018-11-08
refactor(config): add bootstrap period variable

### LINKED ISSUE #88: Add bootstrap period variable to configuration file
We need a new entry in the configuration file for specifying the bootstrapping period, that is, how many seconds will pass between attempts to fulfill the targeted amount of peers.

#### Actionables
- [ ] Decide the name of the variable
- [ ] Make config manager load the variable from the configuration file
- [ ] Update documentation to reflect existence, purpose and usage of this variable

lrubiorod:
Add and doc a bootstrap period variable in config lib
Use this variable in session manager instead of the old constant

Close #88

### DIFF
--- a/config/src/lib.rs
+++ b/config/src/lib.rs
@@ -21,6 +21,7 @@ extern crate failure;
 use std::default::Default;
 use std::net::SocketAddr;
 use std::path::PathBuf;
+use std::time::Duration;
 
 pub mod loaders;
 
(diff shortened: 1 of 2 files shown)


config/src/lib.rs
```diff
@@ -110,6 +115,11 @@ impl ConnectionsConfig {
     pub fn default_known_peers() -> Vec<SocketAddr> {
         Vec::default()
     }
+
+    /// Default `outbound_limit` value (`5`)
```
aesedepece: Wrong comment?

core/src/actors/sessions_manager.rs
```diff
@@ -124,18 +121,19 @@ impl Actor for SessionsManager {
         debug!("Sessions Manager actor has been started!");
 
         // Send message to config manager and process its response
-        send_get_config_request(self, ctx, |s, _ctx, config| {
+        send_get_config_request(self, ctx, |act, ctx, config| {
```
aesedepece: What does `act` stand for? Actix? If so, I'd rather use `actix`. Let's make our code as eloquent as possible!

PR #118, opened 2018-11-11
feat(config env): Add config environment defaults

### LINKED ISSUE #86: Move protocol constants into its own file and pivot them over a environment variable
Protocol related and consensus critical constants should be hard coded into the source code in their own file.

Different values should exist for different environments (testnet-1, mainnet, etc.).

At start up, the configuration manager should read the environment from any of the available settings sources and select the values of the protocol constants according to that.

The settings sources are the `witnet.toml` configuration file and the `argv` of the CLI (take into account that the latter should always override the former).

The values of the constants should be kept in memory by the configuration manager. Having them in a separate structure or merging them into the general configuration are both OK. The former sounds cleaner if paired with its own querying methods. However, the latter may be an acceptable compromise and takes way less effort.

anler:
Closes #86 

This commit refactors the witnet_config codebase in order to support different default constants and parameters depending on the environment in which the node is going to run (tesnet, mainnet, etc).

It does so by separating the the configuration used in the application:

`config::Config`

from the one that is used as temporary container of config parameters loaded from configuration files or command line arguments:

`config::partial::Config`

All the loaders defined in `loaders` will return a `config::partial::Config` object, then that object will be consumed by:

`config::Config::from_partial`

in order to merge it with the declared environment defaults (`defaults` module).

### Example:

As a complete example suppose you want to build a config object containing all the defaults for mainnet, you'll do it like this:

```
use witnet_config::config::{partial, Config, Environment};

fn get_mainnet_config() -> Config {
    Config::from_partial(partial::Config::default_mainnet())
}
```

### Improvements:

Currently there's a lot of duplication between `Config` and `partial::Config`, the later is just the same as the former but with all the fields marked as optional `Option`, this is a good candidate for being generated by a macro. I found a crate that could help as a starting point:

[OptionalStruct](https://github.com/pLesur/OptionalStruct)

the reason why I did not used it was because it does not play nice with other derive attributes (issue #8) and because is gives no chance of hiding the auto-generated `apply_options` method.


docs/configuration/environment.md
```diff
@@ -0,0 +1,25 @@
+# Environment defaults
+
+You may specify the environment in `witnet.toml` in which the Witnet protocol will run. The environment can be: `testnet-1` or `mainnet` at the moment. Each environment comes with a set of default values which you may override in the configuration file.
```
aesedepece:
```suggestion
Environments create contention between different instances of the Witnet network (e.g.: testnet vs. mainnet).

Each environment comes with a set of default values which you can later override in the configuration file. You can specify which environment to use in the `witnet.toml` configuration file.

At the moment, the available environments are: `testnet-1` and `mainnet`.
```

PR #125, opened 2018-11-13
fix(config): read duration as seconds from file

tmpolaczyk:
Allow to set

```toml
[connections]
bootstrap_peers_period_seconds = 10
```


tmpolaczyk:
Renamed the variable name to `{}_seconds` and added a test. Ready to merge now.

Issue #72, opened 2018-10-30
Define / improve error handling strategy


mmartinbar:
The error handling strategy is yet to be properly defined across the different components in the witnet-rust node. Several tasks need to be studied and accomplished regarding this error handling / graceful stop:

In main thread (`core/src/actors/node.rs`):
- [ ] Actix system needs to be stopped properly
- [ ] Decisions shall be made about what to do when different type of signals are received by the application

In session actor (`core/src/actors/session.rs`):
- [ ] Decisions shall be made about what to do with the actor when it could not be registered. Currently it is being stopped but a full stop is not the right thing to do as a full stop will try to unregister the session.

In connections manager actor (`core/src/actors/connections_manager.rs`):
- [ ] Decisions shall be made about what to do with the actor when the server could not be started. For the time being, if that happens, the application panics.

anler:
I'm going to comment here my findings so far.

I'll divide my comments in:

1. Different scenarios for error handling
1. A proposal for our implementation
1. Info about Actix components related to error handling

## 1. Different scenarios for error handling

Here I'll expose the different error-handling scenarios that I can think of:

1. Panic inside actors
1. Error inside an actor
1. Error inside an actor that can be restarted
1. Error inside an actor that needs to notify its manager
1. Error inside an actor that needs to stop the system

### Panic inside actors

This scenario is the most destructive one since panics are irrecoverable errors, we should try to avoid these situations altogether, but if they cannot be dodged then we have two options:

- The panic should take the system down: This is the default behaviour for actors scheduled in the system arbiter.

``` rust
extern crate actix;

use actix::prelude::*;
use std::process::exit;

fn main() {
    exit(System::run(|| {
        let addr = MyActor.start();
        // Panic in actor will shut down the system
        addr.do_send(Panic);
    }));
}

struct MyActor;

impl Actor for MyActor {
    type Context = Context<Self>;
}

#[derive(Message)]
struct Panic;

impl Handler<Panic> for MyActor {
    type Result = ();

    fn handle(&mut self, _: Panic, _: &mut Self::Context) {
        panic!("MyActor panic");
    }
}
```

- The panic should NOT take the system down: In this case before running the actor we need to create a new arbiter with the `stop_system_on_panic` set to `false` and execute the actor that might panic in that arbiter.

``` rust
extern crate actix;

use actix::prelude::*;
use std::process::exit;

fn main() {
    exit(System::run(|| {
        // Panic in actor won't shut down the system
[...]

PR #133, opened 2018-11-13
feat(core): add getPeers and peerDiscovery handlers to session actor

### LINKED ISSUE #92: Add "GetPeers" handler to Session actor

### LINKED ISSUE #96: Implement peer discovery protocol in Session actor
#### Context
- While #90 implements the `SessionsManager` side of the protocol, this implements the `Session` side.
- The peer discovery protocol allows new nodes in the network to get a more global vision of the network so that they can freely choose their peers and they don't have to trust the same nodes they used for bootstrapping.
- This strategy is very positive for descentralization, trustlessness and privacy as it makes nodes constantly change their position in the network graph.
- New peers are discovered by asking our peers for the addresses of their own peers.
- `Session` actors never initiate the peer discovery protocol by themselves. Instead, they are requested to do so by the `SessionManager`, which periodically sends `GetPeers` to a random session through `Anycast` (see #87, #89 and #90).
- Part of this functionality (adding the `GetPeers` handler to `Session`) is already tackled by issue #92.

#### Actionables
- Make `Session` react to `GetPeer` messages coming any kind of connections (both inbound and outbound) and respond immediately a `Peers` protocol message with the list of known peers (as returned by `PeersManager`'s `GetPeers` handler).

Tommytrg:
This PR includes:
- Implement get peers handler to session actor(#92)
- Implement peer discovery protocol in session actor(#96)

Tests for #92 and #96 will be implemented in #134 and #135.

Close #92 
Close #96

### DIFF
--- a/core/src/actors/session.rs
+++ b/core/src/actors/session.rs
@@ -1,3 +1,4 @@
+use log::warn;
 use std::io::Error;
 use std::net::SocketAddr;
 use std::time::Duration;
@@ -11,9 +12,14 @@ use log::debug;
 use tokio::io::WriteHalf;
 use tokio::net::TcpStream;
 
-use crate::actors::codec::{P2PCodec, Request};
+use crate::actors::codec::{P2PCodec, Request, Response};
+use crate::actors::peers_manager;
 use crate::actors::sessions_manager::{Register, SessionsManager, Unregister};
 
+use witnet_data_structures::{
+    serializers::TryFrom,
+    types::{Command, Message as WitnetMessage},
+};
 use witnet_p2p::sessions::{SessionStatus, SessionType};
 
 ////////////////////////////////////////////////////////////////////////////////////////
@@ -35,7 +41,7 @@ pub struct Session {
     status: SessionStatus,
 
     /// Framed wrapper to send messages through the TCP connection
-    _framed: FramedWrite<WriteHalf<TcpStream>, P2PCodec>,
+    framed: FramedWrite<WriteHalf<TcpStream>, P2PCodec>,
 
     /// Handshake timeout
     _handshake_timeout: Duration,
@@ -48,7 +54,7 @@ impl Session {
         _server_addr: SocketAddr,
         remote_addr: SocketAddr,
         session_type: SessionType,
-        _framed: FramedWrite<WriteHalf<TcpStream>, P2PCodec>,
+        framed: FramedWrite<WriteHalf<TcpStream>, P2PCodec>,
         _handshake_timeout: Duration,
     ) -> Session {
         Session {
@@ -56,7 +62,7 @@ impl Session {
             remote_addr,
             session_type,
             status: SessionStatus::Unconsolidated,
-            _framed,
+            framed,
             _handshake_timeout,
         }
     }
@@ -135,7 +141,7 @@ impl WriteHandler<Error> for Session {}
 /// Implement `StreamHandler` trait in order to use `Framed` with an actor
 impl StreamHandler<Request, Error> for Session {
     /// This is main event loop for client requests
-    fn handle(&mut self, msg: Request, _ctx: &mut Self::Context) {
+    fn handle(&mut self, msg: Request, ctx: &mut Self::Context) {
         // Handle different types of requests
         match msg {
             Request::Message(message) => {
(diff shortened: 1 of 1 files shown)


core/src/actors/session.rs
```diff
@@ -125,11 +130,53 @@ impl WriteHandler<Error> for Session {}
...
+                                                // FIXME(#72): a full stop of the session is not correct (unregister should
+                                                // be skipped)
+                                                ctx.stop();
+                                            }
+                                        }
+                                        actix::fut::ok(())
+                                    })
+                                    .wait(ctx);
+                            }
+                            _ => panic!(""),
```
mmartinbar: Don't know if panicking is the best option here. I would discuss this with the team or link this as well to issue #72 , to be discussed in the near future.

core/src/actors/session.rs
```diff
@@ -125,11 +131,21 @@ impl WriteHandler<Error> for Session {}
...
             Request::Message(message) => {
                 debug!("Session {} received message: {:?}", self.address, message);
+                let decoded_msg = WitnetMessage::try_from(message.to_vec());
+                match decoded_msg {
+                    Err(err) => debug!("Error decoding message: {:?}", err),
+                    Ok(msg) => match msg.kind {
+                        Command::GetPeers => {
+                            handle_get_peers_message(self, ctx);
+                        }
+                        _ => warn!("{:?} message not implemented", msg.kind),
```
aesedepece:
```suggestion
                        _ => warn!("Received a message of kind \"{:?}\", which is not implemented yet", msg.kind),
```

core/src/actors/session.rs
```diff
@@ -154,5 +173,43 @@ impl Handler<GetPeers> for Session {
...
+    let peers_manager_addr = System::current()
+        .registry()
+        .get::<peers_manager::PeersManager>();
+
+    // Start chain of actions
+    peers_manager_addr
+        // Send GetPeer message to PeersManager actor
+        // This returns a Request Future, representing an asynchronous message sending process
+        .send(peers_manager::GetPeers)
+        // Convert a normal future into an ActorFuture
```
jrmoreau: Is `normal future` the same as `Request Future`?

core/src/actors/session.rs
```diff
@@ -154,5 +173,43 @@ impl Handler<GetPeers> for Session {
 
     fn handle(&mut self, _msg: GetPeers, _: &mut Context<Self>) {
         debug!("GetPeers message should be sent through the network");
+        // Create get peers message
+        let get_peers_msg: Vec<u8> = WitnetMessage::build_get_peers().into();
+        // Write get peers message in session
```
jrmoreau:
```suggestion
        // Write GetPeers message in session
```

core/src/actors/session.rs
```diff
@@ -154,5 +173,43 @@ impl Handler<GetPeers> for Session {
 
     fn handle(&mut self, _msg: GetPeers, _: &mut Context<Self>) {
         debug!("GetPeers message should be sent through the network");
+        // Create get peers message
```
jrmoreau:
```suggestion
        // Create GetPeers message
```

core/src/actors/session.rs
```diff
@@ -135,14 +141,27 @@ impl WriteHandler<Error> for Session {}
 /// Implement `StreamHandler` trait in order to use `Framed` with an actor
 impl StreamHandler<Request, Error> for Session {
     /// This is main event loop for client requests
-    fn handle(&mut self, msg: Request, _ctx: &mut Self::Context) {
+    fn handle(&mut self, msg: Request, ctx: &mut Self::Context) {
         // Handle different types of requests
```
jrmoreau:
```suggestion
        // Handles different types of requests
```

core/src/actors/session.rs
```diff
@@ -135,14 +141,27 @@ impl WriteHandler<Error> for Session {}
 /// Implement `StreamHandler` trait in order to use `Framed` with an actor
 impl StreamHandler<Request, Error> for Session {
     /// This is main event loop for client requests
```
jrmoreau:
```suggestion
    /// This is the main event loop for client requests
```

core/src/actors/session.rs
```diff
@@ -135,14 +141,27 @@ impl WriteHandler<Error> for Session {}
 /// Implement `StreamHandler` trait in order to use `Framed` with an actor
```
jrmoreau:
```suggestion
/// This implements the `StreamHandler` trait in order to use `Framed` with an actor
```

Issue #123, opened 2018-11-12
Chain Manager


aesedepece:
#### Description
The ~~blocks~~ chain manager will be in charge of:
- Initializing the chain info upon running the node for the first time and persisting it into storage.
- Recovering the chain info from storage and keeping it in its state.
- Validating block candidates as they come from `Session`.
- Consolidating multiple block candidates for the same checkpoint into a single valid block.
- Putting valid blocks into storage by sending them to `StorageManager`.
- Having a method for letting other components to get blocks by hash or checkpoint.
- Having a method for letting other components to get the epoch of the current tip of the blockchain (e.g. last epoch field required for the handshake in the Witnet network protocol)
- __WIP__: _this is a living list. More features will be added as they are needed._

#### Data structures
These are first-approach proposals for some of the required data structures:

```rust
struct ChainInfo {
  environment: Environment,
  protocol_constants: ProtocolConstants,
  highest_block_checkpoint: Checkpoint
}
```
```rust
enum Environment {
  Mainnet,
  Testnet1
}
```
```rust
struct ConsensusConstants {
  genesis_timestamp: u64,
  genesis_hash: Array<u8>,
  checkpoints_period: u16,
  reputation_demurrage: float,
  reputation_punishment: float,
}
```
```rust
struct Checkpoint {
  number: u32,
  hash: Arrray<u8>
}
```

PR #154, opened 2018-11-19
feat(data) implement ChainInfo data structure

### LINKED ISSUE #150: Implement ChainInfo data structure
As specified in #123

lrubiorod:
Create chain.rs file with ChainInfo data structure as required in #123 

Close #150

### DIFF
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -1,0 +1,43 @@
+#[derive(Debug)]
+pub struct ChainInfo {
+    environment: Environment,
+    consensus_constants: ConsensusConstants,
+    highest_block_checkpoint: Checkpoint
+}
+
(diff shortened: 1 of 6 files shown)


data_structures/src/chain.rs
```diff
@@ -0,0 +1,27 @@
...
+pub struct ChainInfo {
+    environment: Environment,
+    consensus_constants: ConsensusConstants,
+    highest_block_checkpoint: Checkpoint
+}
+
+#[derive(Debug)]
+pub enum Environment {
+    Mainnet,
+    Testnet1
```
aesedepece: This is not guaranteed to match the `enum Environment` in `ConfigManager`, but I don't think we can do much to avoid that.

Issue #152, opened 2018-11-19
Receive inventory items through JSON-RPC and broadcast them through SessionManager


aesedepece:
## Context
For us to be able to introduce new inventory items into the network before the node itself is capable of producing them organically, we need a TCP socket server exposing a JSON-RPC interface with a method for manually introducing and announcing inventory items.

This method will in the future be used for introducing transactions and data requests, but for the time being and for the scope of Sprint 3, we will exceptionally allow introducing only blocks.

## Actionables
- [x] Add `jsonrpc.server_address` parameter in configuration for setting TCP binding address and port.
- [x] Document `jsonrpc.server_address`.
- [x] Make `ConfigManager` read `jsonrpc.server_address`.
- [x] Repeat same three tasks for `jsonrpc.enable` (default value is `true`)
- [x] Create a `JsonRpcServer` actor for the JSON-RPC server. This kind of looks like a manager, but it's not a manager.
- [x] Start `JsonRpcServer` from main node entry point only if `jsonrpc.enable == true`.
- [x] Make `JsonRpcServer` query `ConfigManager` for `jsonrpc.server_address` 
- [x] Make `JsonRpcServer` start a TCP server on the specified address and port. Stream delimiter can be `\n`.
- [x] Make `JsonRpcServer` parse the received string as JSON. [Return `-32700` on failure][jsonrpc-errors].
- [x] Make `JsonRpcServer` check conformity of the decoded data structure as JSON-RPC. [Return `-32600` on failure][jsonrpc-errors].
- [x] Make `JsonRpcServer` match the method in the JSON-RPC. [Return `-32601` if method is not supported][jsonrpc-errors].
- [ ] Add support in `JsonRpcServer` for the `inventory` method, which should take any inventory item as payload (only blocks right now). See #100 for the block data structure. [Return `-32602` if the payload is wrong][jsonrpc-errors].
- [ ] Make the `inventory` method send `Broadcast(AnnounceItem(block))` to `SessionsManager`.

[jsonrpc-errors]: https://www.jsonrpc.org/specification#error_object

mmartinbar:
Regarding the `jsonrpc_address` param in the configuration file, in the end a new section `jsonrpc` was added and a `server_address` param under it:
```rust
[jsonrpc]
server_address = "127.0.0.1:4321"
```

mmartinbar:
I'm going to PR the first three actionables as @tmpolaczyk is going to inherit this issue.

aesedepece:
[...]

PR #166, opened 2018-11-20
feat(config): add jsonrpc section and server_address under it

mmartinbar:
This PR implements the first three actionables of #152 :

- Add a new `[jsonrpc]` section in configuration and a `server_address` param under it.
- Add tests for it
- Add documentation for this new section and parameter

`ConfigManager` already reads the whole configuration, no change was required for that.

This PR does not close the issue #152 , as more work is required.

### DIFF
--- a/config/src/defaults.rs
+++ b/config/src/defaults.rs
@@ -63,6 +63,9 @@ pub trait Defaults {
     fn consensus_constants_checkpoints_period(&self) -> u16 {
         90
     }
+
+    /// Default JSON-RPC server addr
+    fn jsonrpc_server_address(&self) -> SocketAddr;
 }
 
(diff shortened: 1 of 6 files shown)


config/src/defaults.rs
```diff
@@ -76,9 +79,14 @@ impl Defaults for Mainnet {
         SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 11337)
     }
 
+    fn jsonrpc_server_address(&self) -> SocketAddr {
+        SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 11338)
```
aesedepece: `11338` is perfect.

PR #171, opened 2018-11-21
feat(core): Create ChainInfo data structure and make BlocksManager load it from Storage

### LINKED ISSUE #130: Make BlocksManager recover chain info from storage and keep it in its state
#### Context
- Epic for `BlocksManager` is #123.
- `ChainInfo` is a data structure holding critical information about the chain state and protocol constants. It is crucial for making sure that the node is running on the same environment as the one used to produce the chain that we read from storage. Otherwise, consensus would be broken.

#### Actionables
- [ ] Make `BlocksManager` query `ConfigManager` for environment name and protocol constants and keep all that info into state.
- [ ] Implement `ChainInfo` data structure.
- [ ] Make `BlocksManager` query `StorageManager` for the stored `ChainInfo` upon startup.
  - [ ] If `StorageResult<Option<ChainInfo>>` matches `Ok(Some(chain_info))`, verify if `chain_info` matches the environment info (compare environment name and protocol constants).
  - [ ] If otherwise it matches `Ok(None)`, create new `ChainInfo`, put into `BlockManager` state and persist into storage.
- [ ] Update documentation to reflect `BlocksManager` actor state.
- [ ] Update documentation to reflect all the new behavior and outcoming actor messages.
- [ ] Create PR and connect to this issue.

#### Data structures
A suggestion for `ChainInfo` data structure can be found in Epic #123.

### LINKED ISSUE #146: Fix typo in config parameter `checkpoints_period`
According to the documentation, the following should work:

```toml
[consensus_constants]
checkpoints_period_seconds = 10
```

However, in core/src/config/partial.rs the field name is `checkpoint_period`, missing an `s`, so the following works instead:

```toml
[consensus_constants]
checkpoint_period_seconds = 10
```

### LINKED ISSUE #150: Implement ChainInfo data structure

lrubiorod:
This PR implements the requirements for the issues #130 #146 #150 

- Fix typo in config parameter `checkpoints_period`.
- Implement ChainInfo data structure.
- Make BlocksManager query ConfigManager for environment name and protocol constants and keep all that info into state.
- Make BlocksManager query StorageManager for the stored ChainInfo upon startup.
- Update documentation to reflect BlocksManager actor state.
- Update documentation to reflect all the new behavior and outcoming actor messages.

Close #130 
Close #146 
Close #150

### DIFF
--- a/config/src/config/partial.rs
+++ b/config/src/config/partial.rs
@@ -8,12 +8,12 @@
 //! later, the `config` module will use this partial config object and
 //! the environment-specific defaults (see the `environment` module)
 //! to produce a __total__ (no `Option` fields) configuration object.
-use crate::config::Environment;
 use std::collections::HashSet;
 use std::default::Default;
 use std::net::SocketAddr;
 use std::path::PathBuf;
 use std::time::Duration;
+use witnet_data_structures::chain::Environment;
 
 /// The partial configuration object that contains all other, more
 /// specific, configuration objects (connections, storage, etc).
@@ -102,10 +102,23 @@ pub struct ConsensusConstants {
     /// Timestamp at checkpoint 0 (the start of epoch 0)
     #[serde(default)]
     pub checkpoint_zero_timestamp: Option<i64>,
+
     /// Seconds between the start of an epoch and the start of the next one
     #[serde(default)]
-    #[serde(rename = "checkpoint_period_seconds")]
-    pub checkpoint_period: Option<u16>,
+    #[serde(rename = "checkpoints_period_seconds")]
+    pub checkpoints_period: Option<u16>,
+
+    /// Genesis block hash value
+    #[serde(default)]
+    pub genesis_hash: Option<Vec<u8>>,
+
+    /// Decay value for reputation demurrage function
+    #[serde(default)]
+    pub reputation_demurrage: Option<f64>,
+
+    /// Punishment value for dishonestly use
+    #[serde(default)]
+    pub reputation_punishment: Option<f64>,
 }
 
 /// JSON-RPC API configuration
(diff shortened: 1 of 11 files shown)


config/src/config/partial.rs
```diff
@@ -102,10 +102,23 @@ pub struct ConsensusConstants {
...
+    #[serde(rename = "checkpoints_period_seconds")]
+    pub checkpoints_period: Option<u16>,
+
+    /// Genesis block hash value
+    #[serde(default)]
+    pub genesis_hash: Option<Vec<u8>>,
+
+    /// Decay value for reputation demurrage function
+    #[serde(default)]
+    pub reputation_demurrage: Option<f64>,
```
aesedepece: @witnet/devs Floating-point values are not suitable for consensus-critical constants because they are not deterministic. They can arbitrarily trade off accuracy for precision, and they can be rounded up or down differently on distinct architectures, processors, operating systems and specific devices. Many [FPU]s even assimilate [subnormal values][denormal] to zero.

To put this into perspective, this is the very reason why the Ethereum Virtual Machine does not support float types either.

Even though Rust complies with IEEE 754, it still behaves inconsistently when dealing with floating-point values across different platforms. This has been [discussed at length by the community][broken-floats], but no final RFC has been adopted.

We will be better off using [fixed-point arithmetics][fixed-point], as we can compromise range and even precision for the sake of accuracy and determinism. A bunch of good-looking fixed-point arithmetic libraries are just [one Cargo search away][cargo].

[FPU]: https://en.wikipedia.org/wiki/Floating-point_unit
[denormal]: https://en.wikipedia.org/wiki/Denormal_number
[broken-floats]: https://internals.rust-lang.org/t/pre-rfc-dealing-with-broken-floating-point/2673
[fixed-point]: https://en.wikipedia.org/wiki/Fixed-point_arithmetic
[cargo]: https://crates.io/search?q=fixed%20point

lrubiorod: I have been checking fixed-point arithmetic libraries and currently they do not implement deserialize traits. Due to demurrage algorithm it is not necessary to Testnet, I would add an issue instead to solve now. What do you think?

aesedepece: I agree, we can defer the decision on this for the time being. Issue #172 has been opened to track progress on assessing, presenting, discussing, deciding and documenting our strategy regarding consensus-critical real numbers.

Issue #164, opened 2018-11-20
 Implement BLOCK protocol message and FlatBuffers builder


aesedepece:
### Inventory exchange protocol

The inventory exchange protocol describes the interaction between nodes in order to synchronize themselves with the network and reestablish the full blockchain (in terms of blocks and transactions).

Inventory exchanges envision 2 main scenarios:

 1. **Block Download**, for nodes in need of synchronizing their blockchains by requesting and downloading blocks.

 2. **Inventory Broadcasting**, for nodes willing to broadcast transactional information, such as blocks and transactions.

#### Block Download

**Block Download** is performed by nodes that need to catch up with other nodes by downloading missing blocks in order to reconstruct their blockchains. For example, this is the case of nodes connecting to the blockchain for the first time or after a significant amount of downtime.

Starting from the block #0 (the hardcoded genesis block), the nodes need to validate all blocks up to the current tip of the blockchain. Therefore, in case of missing blocks, a local node will initiate the following process of synchronization with its outbound peers:

 1. The local node will have already exchanged `version` messages with its remote outbound peers. Those `version` messages contain the last epoch known to others peers, i.e. the local peer can already compare how many blocks they each have and identify how many are missing.

 2. The local node will send a `get_blocks` message to all its outbound nodes (after successful handshake protocol). These messages contain the hash of the top block of the local blockchain.

 3. Remote peers will reply by sending another `get_blocks` message containing the hash of their top block of their respective blockchains.

 4. The peer with the longest blockchain will identify which blocks are required by the other peer in order to allow it to synchronize to its blockchain. The peer will select up to the first consecutive 500 blocks and it will transmit their hashes using an `inv` (inventory) message.

 5. After identifying which blocks are missing (it may already have some of them), the node may request them by using a `get_data` message, containing the hashes of the needed blocks. This message will help the node to catch up with the current full blockchain.

 6. After receiving the `get_data` message, the peer sends the requested blocks individually by using `block` messages.
The following diagram depicts the previously described process under the assumption that the peer with the longest blockchain is `NodeB` (step 4).

```ascii
         NodeA                            NodeB
           +                                +
           |           GET_BLOCKS           |
           +------------------------------->+
           |           GET_BLOCKS           |
           +<-------------------------------+
           |              INV               |
           +<-------------------------------+
           |                                |
           |                                |
           |            GET_DATA            |
           +------------------------------->+
           |             BLOCK              |
           +<-------------------------------+
           |             BLOCK              |
[...]

PR #175, opened 2018-11-22
feat(core): Create GetHighestBlockCheckpointBeacon msg and its handler

### LINKED ISSUE #142: Implement GetHighestBlockCheckpoint method in BlocksManager
## Context
`BlocksManager` should provide a method for other components to learn the checkpoint of the highest known block, that is, the block in the tip of the chain.

This _highest block checkpoint_ doesn't necessarily need to be the same as the last passed checkpoint (the current epoch) for a number of reasons:
- Multiple blocks with the same checkpoint may need to be consolidated into only one before considered final.
- The known chain for any peer can lag behind the consensus agreed tip of the chain at any time, specially after network disruption or downtimes of any nature.
- New nodes need time before they are fully synchronized up to the tip of the chain.

The best example of how the `GetHighestBlockCheckpoint` method will be used is during the handshake protocol. The first protocol message sent through an outbound peer connection is `Version`, which includes a field for exchanging the checkpoint of the highest block. This allows each node to learn if their peers are lagging behind and proactively send them inventory lists starting from the checkpoint they specified.

## Actionables
- Define `GetHighestBlockCheckpoint` actor message in `core/actors/blocks_manager/messages.rs`
- `impl Handler<GetHighestBlockCheckpoint> for BlocksManager` at `core/actors/blocks_manager/handlers.rs`.
- __WIP__
- Document the `GetHighestBlockCheckpoint` method at `docs/architecture/managers/config-manager.md`
- Create PR and link it to this issue.

lrubiorod:
Renamed Checkpoint to CheckpointBeacon in chain.rs to avoid misunderstandings with Epoch checkpoint.
Added Epoch struct in chain.rs 
Create GetHighestBlockCheckpointBeacon msg and its handler
Update doc in BlocksManager

Close #142


core/src/actors/blocks_manager/messages.rs
```diff
@@ -0,0 +1,16 @@
...
+use std::io;
+
+use witnet_data_structures::chain::CheckpointBeacon;
+
+/// Message to obtain the highest block checkpoint managed by the `BlocksManager`
+/// actor.
+pub struct GetHighestBlockCheckpointBeacon;
+
+/// Result of the GetHighestBlockCheckpoint message handling
+pub type CheckpointBeaconResult = Result<CheckpointBeacon, io::Error>;
```
aesedepece: We can define our `Result` and `Error`s for `BlocksManager` in the `witnet_chain` crate, where the `BlocksManager` library will be located.

For reference, look at how `witnet_storage` defines its own `Result` and `Error`s:
https://github.com/witnet/witnet-rust/blob/master/storage/src/error.rs

data_structures/src/chain.rs
```diff
@@ -50,17 +50,24 @@ pub struct ConsensusConstants {
...
-impl Default for Checkpoint {
-    fn default() -> Checkpoint {
-        Checkpoint {
-            number: 0,
-            hash: Vec::new(),
+impl Default for CheckpointBeacon {
+    fn default() -> CheckpointBeacon {
+        CheckpointBeacon {
+            checkpoint: Epoch(0),
+            hash_prev_block: Vec::new(),
```
aesedepece: I like this more, but anyway we will need to change it in the future to account generalization over different hash functions. @mariocao is working on it as part of #164 .
```suggestion
            hash_prev_block: vec![0; 32],
```

core/src/actors/blocks_manager/handlers.rs
```diff
@@ -33,3 +37,24 @@ impl Handler<EpochNotification<EveryEpochPayload>> for BlocksManager {
...
+        _msg: GetHighestBlockCheckpointBeacon,
+        _ctx: &mut Context<Self>,
+    ) -> Self::Result {
+        if let Some(chain_info) = &self.chain_info {
+            Ok(chain_info.highest_block_checkpoint.clone())
+        } else {
+            error!("Not Chain Info loaded in BlocksManager");
+            Err(Error::new(
+                ErrorKind::Other,
+                "Not Chain Info loaded in BlocksManager",
```
tmpolaczyk:
```suggestion
                "No ChainInfo loaded in BlocksManager",
```

core/src/actors/blocks_manager/messages.rs
lrubiorod: Done!!

data_structures/src/chain.rs
```diff
@@ -50,17 +50,24 @@ pub struct ConsensusConstants {
...
-pub struct Checkpoint {
-    pub number: u32,
-    pub hash: Vec<u8>,
+/// Checkpoint beacon structure
+#[derive(Debug, Eq, PartialEq, Clone, Serialize, Deserialize)]
+pub struct CheckpointBeacon {
+    /// The serial number for an epoch
+    pub checkpoint: Epoch,
+    /// The 256-bit hash of the previous block header
+    pub hash_prev_block: Vec<u8>,
```
anler: Just a suggestion, but I would love to see more elaborated documentation, currently the auto-generated docs for this module looks like this:
image
I would love to see described what the ChainInfo is, why it is needed and where it is used, the "serial number for an epoch" is a bit vague IMO, the documentation already creates a link for that type so I would add something like why the checkpoint is of type Epoch or why it is necessary. Take for example how the documentation for the config looks like:
image

anler:
Just a suggestion,  I would love to see more elaborated auto-generated documentation, currently the docs for most of the modules looks like this:
image
I would love to see described what the ChainInfo is, why it is needed and where it is used, and the same for all the modules and types. Take for example how the documentation for the config looks like:
image

PR #185, opened 2018-11-26
Feat(serializers): implement GetBlocks serializers

### LINKED ISSUE #160: Implement GET_BLOCKS protocol message and FlatBuffers builder
### Inventory protocol
Starting from the block #0 (the hardcoded genesis block), the nodes need to validate all blocks up to the current tip of the blockchain. Therefore, in case of missing blocks, a local node will initiate the following process of synchronization with its outbound peers:

 1. The local node will have already exchanged `version` messages with its remote outbound peers. Those `version` messages contain the checkpoint of the last block known to others peers, i.e. the local peer can already compare how many blocks they each have and identify how many are missing.

 2. The local node will send a `get_blocks` message to all its outbound nodes (after successful handshake protocol). These messages contain the hash of the top block of the local blockchain.

 3. Remote peers will reply by sending another `get_blocks` message containing the hash of their top block of their respective blockchains.

_(the protocol continues, but it is not relevant for the scope of this issue)_

```ascii
         NodeA                            NodeB
           +                                +
           |           GET_BLOCKS           |
           +------------------------------->+
           |           GET_BLOCKS           |
           +<-------------------------------+
           |                                |
           +                                +
```

### `get_blocks` message

The `get_blocks` messages are used in order to notify the hash of the highest known block by the peer. After exchanging `get_blocks` messages between peers, the one with the longest blockchain in terms of blocks will send an `inv` message to the other peer. This message will include the list of block hashes starting right after the last known block hash provided by the other peer.

The `get_blocks` message consists of a message header with the `GET_BLOCKS` command and a payload containing the tip of the chain (the hash of the latest block) as known to the local peer:

| Field        | Type       | Description                                    |
| ------------ | :--------: | ---------------------------------------------- |
| `last_block` | `[u32; 8]` | Hash of the last known block to the local node |

## Actionables
- [x] Implement `get_blocks` as a `Command` in the FlatBuffers schema.
- [x] Implement `get_blocks` as a native data structure
- [ ] Implement `get_blocks` and its builder as a `WitnetMessage` in `witnet_data_structures::types::Message`.

Tommytrg:
Fixes #160


data_structures/src/serializers.rs
```diff
@@ -182,35 +213,42 @@ impl Into<Vec<u8>> for Message {
         let mut builder = flatbuffers::FlatBufferBuilder::new_with_capacity(FTB_SIZE);
         // Create flatbuffer to encode a witnet message
```
jrmoreau:
```suggestion
        // Create Flatbuffer to encode a Witnet message
```

data_structures/src/serializers.rs
```diff
@@ -268,6 +306,47 @@ fn create_address(address: FlatBufferAddress) -> Option<Address> {
     }
 }
 
+// Create a ping flatbuffer to encode a witnet's ping message
```
jrmoreau:
```suggestion
// Create a ping Flatbuffer to encode a Witnet ping message
```

data_structures/src/serializers.rs
```diff
@@ -268,6 +306,47 @@ fn create_address(address: FlatBufferAddress) -> Option<Address> {
...
+        &MessageArgs {
+            magic: get_blocks_args.magic,
+            command_type: FlatBuffersCommand::GetBlocks,
+            command: Some(get_blocks_command.as_union_value()),
+        },
+    );
+    build_flatbuffer(builder, message)
+}
+
+// Create a witnet's ping message to decode a flatbuffers' ping message
```
jrmoreau:
```suggestion
// Create a Witnet ping message to decode Flatbuffers' ping message
```

data_structures/src/serializers.rs
```diff
@@ -268,6 +306,47 @@ fn create_address(address: FlatBufferAddress) -> Option<Address> {
...
+fn create_get_blocks_message(get_blocks_args: GetBlocksWitnetArgs) -> Message {
+    Message {
+        kind: Command::GetBlocks(GetBlocks {
+            last_block: get_blocks_args.last_block,
+        }),
+        magic: get_blocks_args.magic,
+    }
+}
+
 // Create a get peers flatbuffer to encode a witnet's get peers message
```
jrmoreau:
```suggestion
// Create a get peers Flatbuffer to encode Witnet's get peers message
```

data_structures/src/serializers.rs
```diff
@@ -565,7 +644,7 @@ fn create_version_flatbuffer(
 // Create a witnet's version message to decode a flatbuffers' version message
```
jrmoreau:
```suggestion
// Create a Witnet version message to decode a Flatbuffers' version message
```

data_structures/src/serializers.rs
```diff
@@ -472,7 +551,7 @@ fn create_verack_flatbuffer(
 // Create a witnet's verack message to decode a flatbuffers' verack message
```
jrmoreau:
```suggestion
// Create a Witnet verack message to decode a Flatbuffers' verack message
```

data_structures/src/serializers.rs
```diff
@@ -444,9 +523,9 @@ fn create_pong_flatbuffer(
 // Create a witnet's pong message to decode a flatbuffers' pong message
```
jrmoreau:
```suggestion
// Create a Witnet pong message to decode a Flatbuffers' pong message
```

data_structures/src/serializers.rs
```diff
@@ -411,9 +490,9 @@ fn create_ping_flatbuffer(
 // Create a witnet's ping message to decode a flatbuffers' ping message
```
jrmoreau:
```suggestion
// Create a Witnet ping message to decode a Flatbuffers' ping message
```

data_structures/src/serializers.rs
jrmoreau:
```suggestion
// Create a Witnet ping message to decode a Flatbuffers' ping message
```

data_structures/src/serializers.rs
```diff
@@ -289,7 +368,7 @@ fn create_get_peers_flatbuffer(
 // Create a witnet's get peers message to decode a flatbuffers' get peers message
```
jrmoreau:
```suggestion
// Create Witnet's get peers message to decode a Flatbuffers' get peers message
```

schemas/protocol.fbs
```diff
@@ -4,7 +4,7 @@
 namespace Protocol;
 
 // List of available commands
-union Command (required) { Version, Verack, GetPeers, Peers, Ping, Pong, Block }
+union Command (required) { Version, Verack, GetBlocks, GetPeers, Peers, Ping, Pong, Block }
```
mariocao: It is really important to keep the order in Flatbuffers. That's why we could add new commands into the schema, re-generate the code and everything should work and be compatible.

If the order is changed, then the serialized FlatBuffers will have different internal field IDs.
As a consequence, we will have to update all the tests that use bytes to represent Flatbuffers (what you had to do with the `serializers.rs` tests).

mariocao:
```suggestion
union Command (required) { Version, Verack, GetPeers, Peers, Ping, Pong, Block, GetBlocks }
```

data_structures/src/serializers.rs
```diff
@@ -32,6 +34,18 @@ use flatbuffers::FlatBufferBuilder;
 
 const FTB_SIZE: usize = 1024;
 
+#[derive(Debug, Clone, Copy)]
+struct GetBlocksFlatbufferArgs {
```
mariocao: I think that we don't need 2 structs that are the same (i.e. `GetBlocksFlatbufferArgs` and `GetBlocksWitnetArgs`).
Unfortunately, the same happened with other commands (e.g. `PingFlatbufferArgs`, ...). :disappointed:

data_structures/src/serializers.rs
```diff
@@ -158,6 +172,36 @@ impl TryFrom<Vec<u8>> for Message {
                     })
                 })
                 .ok_or(""),
+            FlatBuffersCommand::GetBlocks => message
+                .command_as_get_blocks()
+                .map(|get_blocks| {
+                    let hash_prev_block = match get_blocks
+                        .highest_block_checkpoint()
+                        .unwrap()
```
mariocao: This `unwrap()` can be avoided if the related part of the schema is required (see comment below).

schemas/protocol.fbs
```diff
@@ -86,6 +86,10 @@ table CheckpointBeacon {
     hash_prev_block: Hash (required);
 }
 
+table GetBlocks {
+    highest_block_checkpoint: CheckpointBeacon;
```
mariocao: To avoid having unnecessary `Options` around our code, i would advise to use the `required` attribute. 
```suggestion
    highest_block_checkpoint: CheckpointBeacon (required);
```

core/src/actors/blocks_manager/actor.rs
```diff
@@ -137,7 +137,7 @@ impl Actor for BlocksManager {
                         initialize and store a new chain."
                         );
                         // Create a new ChainInfo
```
jrmoreau: What is a `ChainInfo`? Is it a file? Is it a message? We may want to be more descriptive.

Tommytrg: You are right, it should be defined [here](https://github.com/witnet/witnet-rust/blob/master/data_structures/src/chain.rs). I will include it in this PR.

jrmoreau: Great!

PR #188, opened 2018-11-26
fix(core): add outbound/inbound checks in Session's StreamHandler

mmartinbar:
The handling of some protocol messages is only allowed for a certain type of sessions (outbound/inbound).

This PR introduces the check of the session type in the `StreamHandler`of the `Session`. It also fixes this check for the `Peers` message (`Peers` message should only be allowed for outbound sessions).

### DIFF
--- a/core/src/actors/session/handlers.rs
+++ b/core/src/actors/session/handlers.rs
@@ -19,7 +19,7 @@ use witnet_data_structures::{
     serializers::TryFrom,
     types::{Address, Command, Message as WitnetMessage, Peers, Version},
 };
-use witnet_p2p::sessions::SessionStatus;
+use witnet_p2p::sessions::{SessionStatus, SessionType};
 
 use super::{
     messages::{GetPeers, SessionUnitResult},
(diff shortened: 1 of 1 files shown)


core/src/actors/session/handlers.rs
```diff
@@ -55,29 +57,34 @@ impl StreamHandler<BytesMut, Error> for Session {
...
                     ////////////////////
-                    (SessionStatus::Consolidated, Command::GetPeers(_)) => {
+                    // Handle GetPeers message
+                    (_, SessionStatus::Consolidated, Command::GetPeers(_)) => {
                         peer_discovery_get_peers(self, ctx);
                     }
-                    (SessionStatus::Consolidated, Command::Peers(Peers { peers })) => {
+                    // Handle Peers message
+                    (
+                        SessionType::Outbound,
```
mariocao: :heart_eyes:

core/src/actors/session/handlers.rs
```diff
@@ -55,29 +57,34 @@ impl StreamHandler<BytesMut, Error> for Session {
...
-                    (SessionStatus::Consolidated, _) => {
-                        warn!("Not implemented message command received!");
-                    }
-                    (_, kind) => {
+                    (session_type, session_status, msg_type) => {
                         warn!(
-                            "Received a message of kind \"{:?}\", which is not implemented yet",
-                            kind
+                            "Received a message of type \"{:?}\" for a session with {:?} type and\
+                             in {:?} status, which is not implemented yet",
```
mariocao: Just to be in line with the previous "Not supported" comment. :)

```suggestion
                             in {:?} status, which is not supported",
```

I am not 100% sure, but perhaps I would try to shorten the log (to avoid having really long log traces).

```
"Message of type \"{:?}\" for session (type: {:?}, status: {:?}) is not supported"
```
What do you think? :blush:

mmartinbar: 100% agree, I'll change it now, thanks!

PR #189, opened 2018-11-26
feat(data-structures): add get_data protocol message

### LINKED ISSUE #163: Implement GET_DATA protocol message and FlatBuffers builder
### Inventory protocol
The inventory exchange protocol describes the interaction between nodes in order to synchronize themselves with the network and reestablish the full blockchain (in terms of blocks and transactions).

Inventory exchanges envision 2 main scenarios:

 1. **Block Download**, for nodes in need of synchronizing their blockchains by requesting and downloading blocks.

 2. **Inventory Broadcasting**, for nodes willing to broadcast transactional information, such as blocks and transactions.

#### Block Download

**Block Download** is performed by nodes that need to catch up with other nodes by downloading missing blocks in order to reconstruct their blockchains. For example, this is the case of nodes connecting to the blockchain for the first time or after a significant amount of downtime.

Starting from the block #0 (the hardcoded genesis block), the nodes need to validate all blocks up to the current tip of the blockchain. Therefore, in case of missing blocks, a local node will initiate the following process of synchronization with its outbound peers:

 1. The local node will have already exchanged `version` messages with its remote outbound peers. Those `version` messages contain the last epoch known to others peers, i.e. the local peer can already compare how many blocks they each have and identify how many are missing.

 2. The local node will send a `get_blocks` message to all its outbound nodes (after successful handshake protocol). These messages contain the hash of the top block of the local blockchain.

 3. Remote peers will reply by sending another `get_blocks` message containing the hash of their top block of their respective blockchains.

 4. The peer with the longest blockchain will identify which blocks are required by the other peer in order to allow it to synchronize to its blockchain. The peer will select up to the first consecutive 500 blocks and it will transmit their hashes using an `inv` (inventory) message.

 5. After identifying which blocks are missing (it may already have some of them), the node may request them by using a `get_data` message, containing the hashes of the needed blocks. This message will help the node to catch up with the current full blockchain.

The following diagram depicts the previously described process under the assumption that the peer with the longest blockchain is `NodeB` (step 4).

_(the protocol continues, but it is not relevant for the scope of this issue)_

```ascii
         NodeA                            NodeB
           +                                +
           |           GET_BLOCKS           |
           +------------------------------->+
           |           GET_BLOCKS           |
           +<-------------------------------+
           |              INV               |
           +<-------------------------------+
           |                                |
           |                                |
           |            GET_DATA            |
           +------------------------------->+
           |                                |
           +                                +
```

 #### Inventory Broadcasting

Similarly to the previously described process of synchronization, any node may contribute to the synchronization of their outbound peers by advertising inventory objects such as blocks and transactions. Inventory broadcasting is also used in case a node creates transactions or mine blocks.

The inventory broadcasting can be described as the following sequence of steps:

 1. A remote node broadcasts its inventory by sending an `inv` message, containing all the hashes of the advertised inventory objects.

 2. After receiving an `inv` message and filtering the inventory objects that may be missing in the local blockchain, the local node sends a `get_data` message with the hashes of the needed objects.

_(the protocol continues, but it is not relevant for the scope of this issue)_

The following diagram depicts the previous step under the assumption that the local node (`NodeA`) sends a `get_data` message requesting 3 blocks and 2 transactions.

```ascii
         NodeA                            NodeB
           +                                +
           |              INV               |
           +<-------------------------------+
           |                                |
           |           GET_DATA             |
           +------------------------------->+
           |                                |
           +                                +
```

## Get data messages

The `get_data` messages are used in order to request specific objects from other nodes. Usually, `get_data` messages are sent after receiving a `inv` (inventory) message and filtering the already known objects. The response to a `get_data` message is often a message carrying the requested object (e.g.: `block`, `tx`, etc.).

The `get_data` message consists of a message header with the `GET_DATA` command and a payload following this format:

| Field       | Type         | Description                 |
| ----------- | :----------: | --------------------------- |
| `count`     | `u16`        | Number of inventory entries |
| `inventory` | `inv_vect[]` | Inventory vectors           |

## Actionables
- [ ] Implement `get_data` as a `Command` in the FlatBuffers schema.
- [ ] Implement `get_data` as a native data structure
- [ ] Implement `get_data` and its builder as a `WitnetMessage` in `witnet_data_structures::types::Message`.

lrubiorod:
Implement get_data as a Command in the FlatBuffers schema.
Implement get_data as a native data structure.
Implement get_data and its builder as a WitnetMessage in witnet_data_structures::types::Message
Add tests for the serializer and builder functions.

Close #163

### DIFF
--- a/schemas/protocol.fbs
+++ b/schemas/protocol.fbs
@@ -6,7 +6,7 @@ namespace Protocol;
 // MAIN TYPES
 /////////////////////////////////////////////////////////
 // List of available commands
-union Command (required) { Version, Verack, GetPeers, Peers, Ping, Pong, Block, Inv }
+union Command (required) { Version, Verack, GetPeers, Peers, Ping, Pong, Block, Inv, GetData }
 
 // Message format with header
 table Message {
@@ -127,4 +127,7 @@ table Inv {
     inventory: [InvVector] (required);
 }
 
+table GetData {
+    inventory: [InvElem] (required);
+}
 root_type Message;
--- a/schemas/protocol.fbs
+++ b/schemas/protocol.fbs
@@ -128,6 +128,6 @@ table Inv {
 }
 
 table GetData {
-    inventory: [InvElem] (required);
+    inventory: [InvVector] (required);
 }
 root_type Message;
--- a/data_structures/src/builders.rs
+++ b/data_structures/src/builders.rs
@@ -6,7 +6,8 @@ use rand::{thread_rng, Rng};
 use crate::chain::{Block, BlockHeaderWithProof, InvVector, Transaction};
 
 use crate::types::{
-    Address, Command, GetPeers, Inv, IpAddress, Message, Peers, Ping, Pong, Verack, Version,
+    Address, Command, GetData, GetPeers, Inv, IpAddress, Message, Peers, Ping, Pong, Verack,
+    Version,
 };
 
 use witnet_util::timestamp::get_timestamp;
@@ -100,6 +101,13 @@ impl Message {
     pub fn build_inv(inv_vectors: Vec<InvVector>) -> Message {
         Message::build_message(Command::Inv(Inv {
             inventory: inv_vectors,
+        }))
+    }
+
+    /// Function to build GetData messages
+    pub fn build_get_data(inv_elems: Vec<InvElem>) -> Message {
+        Message::build_message(Command::GetData(GetData {
+            inventory: inv_elems,
         }))
     }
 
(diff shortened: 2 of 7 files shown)


schemas/protocol.fbs
```diff
@@ -127,4 +127,7 @@ table Inv {
     inventory: [InvVector] (required);
 }
 
+table GetData {
+    inventory: [InvElem] (required);
```
mmartinbar: This should be `InvVector` I think, I don't even know how this compiles as `InvElem` is not on your schema. Please fix this and then recompile to get a new `protocol_generated.rs`
```suggestion
    inventory: [InvVector] (required);
```

mmartinbar:
Looks good to me! :100:

PR #184, opened 2018-11-26
feat(data-structures): add inv protocol message

mmartinbar:
This PR implements the inv message as part of the Witnet's inventory exchange protocol:
- Add inv message in flatbuffers' schema (including auto-generated code)
- Add inv message as a Witnet data structure
- Add functions to convert inv message to/from flatbuffers' types from/to Witnet's types
- Add function to build an inv message
- Add tests for the serializer and builder functions

This PR is to be merged after #183

PR #191, opened 2018-11-27
feat(core): add inv handler for the session

mmartinbar:
This PR adds a basic handler for inv messages at the `Session`.

It is to be merged after #184, #188 and #189.

When #189 is merged, a small change will need to be performed to really build `GetData` messages (for the time being, `Inv` messages are built instead).

### DIFF
--- a/core/src/actors/session/handlers.rs
+++ b/core/src/actors/session/handlers.rs
@@ -27,7 +27,7 @@ use witnet_data_structures::{
     builders::from_address,
     chain::{Block, Hash, InvVector},
     serializers::TryFrom,
-    types::{Address, Command, GetData, Message as WitnetMessage, Peers, Version},
+    types::{Address, Command, GetData, Inv, Message as WitnetMessage, Peers, Version},
 };
 use witnet_p2p::sessions::{SessionStatus, SessionType};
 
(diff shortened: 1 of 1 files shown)


core/src/actors/session/handlers.rs
```diff
@@ -77,6 +77,13 @@ impl StreamHandler<BytesMut, Error> for Session {
                     ) => {
                         peer_discovery_peers(&peers);
                     }
+                    ////////////////////
+                    // INVENTORY      //
+                    ////////////////////
+                    // Handle Inv message
+                    (SessionType::Outbound, SessionStatus::Consolidated, Command::Inv(inv)) => {
```
aesedepece: If we are sending `inv`s only to outbounds, we strictly need to accept `inv`s from inbounds. Otherwise, inventory items diffusion will just not work. 
```suggestion
                    (_, SessionStatus::Consolidated, Command::Inv(inv)) => {
```

core/src/actors/session/handlers.rs
```diff
@@ -207,6 +214,24 @@ fn peer_discovery_peers(peers: &[Address]) {
     });
 }
 
+/// Function to process an Inv message
+fn inventory_process_inv(session: &mut Session, _ctx: &mut Context<Session>, inv: &Inv) {
+    // Check how many of the received inventory vectors need to be requested
+    let inv_vectors = &inv.inventory;
+
+    // TODO missing check of how many of these items really need to be requested
```
aesedepece: Agree. Items already known to us can be just filtered out from the list. TODO.

mmartinbar: I'm working on it right now but I'm not sure if it will be on time for today's review :(. Would you be ok with approving it as it is? If I finish this functionality this morning, I'll create another PR.

aesedepece: That's why I explicitly said TODO :stuck_out_tongue_winking_eye:

PR #193, opened 2018-11-27
chore(config): Add link to config-params docs

anler:
#### Context
- We try to use very expressive names for configuration variables.
- Each variable is encumbered with a comment.
- Even though, we should add a header to the `witnet.toml` configuration file, including a link to the documentation page for the same file.
- This lets us keep the comments in the configuration file succinct while providing a way for the users to learn more about each variable and discover example values for them.

#### Actionables
- [X] Find the URL for the documentation of the configuration file in https://docs.witnet.io
- [X] Modify `witnet.toml` to add a friendly and expressive header pointing to the URL.
- [X] Create PR and link it to this issue.

### DIFF
--- a/witnet.toml
+++ b/witnet.toml
@@ -1,3 +1,5 @@
+# You can read more about all the configuration params in
+# https://docs.witnet.io/configuration/toml-file/
 [connections]
 server_addr = "0.0.0.0:21337"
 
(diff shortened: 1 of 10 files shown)


witnet.toml
```diff
@@ -1,3 +1,5 @@
+# You can read more about all the documentation params in
```
mmartinbar:
```suggestion
# You can read more about all the configuration params in
```

PR #173, opened 2018-11-21
JSON-RPC

tmpolaczyk:
JSON-RPC server.

Missing: improve docs

To be merged after #201

PR #194, opened 2018-11-27
feat(core): AnnounceItems actor message

tmpolaczyk:
Add `AnnounceItems` message to `Session` actor. Allows to announce a new inventory item to the outbound peers.

### DIFF
--- a/core/src/actors/session/handlers.rs
+++ b/core/src/actors/session/handlers.rs
@@ -22,7 +22,7 @@ use witnet_data_structures::{
 use witnet_p2p::sessions::{SessionStatus, SessionType};
 
 use super::{
-    messages::{GetPeers, SessionUnitResult},
+    messages::{AnnounceItems, GetPeers, SessionUnitResult},
     Session,
 };
 
@@ -103,6 +103,22 @@ impl Handler<GetPeers> for Session {
         let get_peers_msg = WitnetMessage::build_get_peers();
         // Write get peers message in session
         self.send_message(get_peers_msg);
+    }
+}
+
+/// Handler for AnnounceItems message (sent by other actors)
+impl Handler<AnnounceItems> for Session {
+    type Result = SessionUnitResult;
+
+    fn handle(&mut self, msg: AnnounceItems, _: &mut Context<Self>) {
+        debug!(
+            "Sending AnnounceItems message to peer at {:?}",
+            self.remote_addr
+        );
+        // Create AnnounceItems message
+        let announce_items_msg = WitnetMessage::build_inv(msg.items);
+        // Write message in session
+        self.send_message(announce_items_msg);
     }
 }
 
--- a/core/src/actors/session/messages.rs
+++ b/core/src/actors/session/messages.rs
@@ -1,4 +1,5 @@
 use actix::Message;
+use witnet_data_structures::chain::InvVector;
 
 /// Message result of unit
 pub type SessionUnitResult = ();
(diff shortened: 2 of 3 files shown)


core/src/actors/session/messages.rs
```diff
@@ -9,3 +10,10 @@ pub struct GetPeers;
 impl Message for GetPeers {
     type Result = SessionUnitResult;
 }
+
+/// Message to announce new inventory items through the network
+#[derive(Clone, Message)]
```
mmartinbar: why is the clone needed? :wink:

tmpolaczyk: The `Clone` is needed for the `Broadcast` method, because this message will be used by #173 to announce a new block to all the outbound peers.

This means that each session will construct an `Inv` message from the `Vec<InvVector>`, which is actually unnecessary because all those messages are exactly the same.

Therefore, the `Broadcast` method could be optimized by constructing the raw message before sending a message to each session actor, although this would require some refactoring.

docs/architecture/session.md
```diff
@@ -42,15 +42,16 @@ Session::create(move |ctx| {
 
 These are the messages supported by the session handlers:
 
-| Message          | Input type            | Output type                       | Description            |
-| ---------------- | --------------------- | --------------------------------- | ---------------------- |
-| `GetPeers`       | `()`                  | `()`                              | Empty                  |
+| Message          | Input type            | Output type                       | Description                            |
+| ---------------- | --------------------- | --------------------------------- | -------------------------------------- |
+| `GetPeers`       | `()`                  | `()`                              | Request peers from an outbound session |
+| `AnnounceItems`  | `Vec<InvElem>`        | `()`                              | Announce new inventory items           |
```
mmartinbar:
```suggestion
| `AnnounceItems`  | `Vec<InvVector>`        | `()`                              | Announce new inventory items           |
```

core/src/actors/session/messages.rs
mmartinbar: Agreed, would you be ok with creating an issue for that optimisation? I think it would be nice to have it in the future for network messages that are exactly the same for all sessions.

docs/architecture/session.md
```diff
@@ -42,15 +42,16 @@ Session::create(move |ctx| {
...
 
 #### GetPeers
-The handler of `GetPeers` message is currently empty.
-
-// TODO Update documentation when `GetPeers` gets any actual functionality.
+Request peers from an outbound session.
```
mmartinbar: At the moment, this is not exactly like that. Upon reception of a `GetPeers` message, the `Session` creates a `GetPeers` message and sends it through the network connection... but does not check if the session is outbound or inbound.

It's the caller (the `SessionsManager` through the `Anycast` method) the one who selects an outbound `Session` to send the `GetPeers` message to it.

However, I think that not checking the session type at the handler of the `GetPeers` in the `Session` might be a mistake. I'll create an issue for that.

core/src/actors/session/messages.rs
aesedepece: Discussed in #198.

core/src/actors/session/handlers.rs
```diff
@@ -106,6 +106,19 @@ impl Handler<GetPeers> for Session {
     }
 }
 
+/// Handler for AnnounceItems message (sent by other actors)
+impl Handler<AnnounceItems> for Session {
+    type Result = SessionUnitResult;
+
+    fn handle(&mut self, msg: AnnounceItems, _: &mut Context<Self>) {
+        debug!("AnnounceItems message should be sent through the network");
```
aesedepece:
```suggestion
        debug!("Sending AnnounceItems message to peer at {?:}" self.remote_addr);
```

core/src/actors/session/messages.rs
```diff
@@ -9,3 +10,10 @@ pub struct GetPeers;
...
+pub struct AnnounceItems {
+    /// Inventory items
+    pub inventory: Vec<InvVector>,
```
aesedepece:
```suggestion
    pub items: Vec<InvVector>,
```

PR #201, opened 2018-11-28
feat(core): add AddNewBlock message in BlocksManager

tmpolaczyk:
Blocked by #194


core/Cargo.toml
```diff
@@ -10,6 +10,7 @@ edition = "2018"
 actix = "0.7.4"
 byteorder = "1.2.6"
 bytes = "0.4"
+rust-crypto = "0.2"
```
mmartinbar: This dependency might not be needed here I think

core/src/actors/blocks_manager/handlers.rs
```diff
@@ -12,7 +12,11 @@ use witnet_util::error::WitnetError;
 
 use log::{debug, error};
 
-use super::messages::GetHighestBlockCheckpoint;
+use super::messages::{AddNewBlock, GetHighestBlockCheckpoint};
+use crate::actors::session::messages::AnnounceItems;
+use crate::actors::sessions_manager::messages::Broadcast;
+use crate::actors::sessions_manager::SessionsManager;
```
mmartinbar: perhaps these two `use` could be grouped into one? not sure what our code guidelines say about this

core/src/actors/blocks_manager/handlers.rs
```diff
@@ -63,3 +67,24 @@ impl Handler<GetHighestBlockCheckpoint> for BlocksManager {
...
+impl Handler<AddNewBlock> for BlocksManager {
+    type Result = ();
+
+    fn handle(&mut self, msg: AddNewBlock, _ctx: &mut Context<Self>) {
+        match self.process_new_block(msg.block) {
+            Ok(hash) => {
+                // Get sessions manager address
+                let sessions_manager_addr = System::current().registry().get::<SessionsManager>();
+
+                let items = vec![witnet_data_structures::chain::InvVector::Block(hash)];
```
mmartinbar: I would add here a comment to describe what this code does and perhaps add a `use` statement to avoid the full path to the `InvVector`:
```suggestion
                // up in the file header
                use witnet_data_structures::chain::InvVector;
                
                // Send a broadcast message to the SessionsManager to announce the new block
                let items = vec![InvVector::Block(hash)];
```

core/src/actors/blocks_manager/handlers.rs
```diff
@@ -63,3 +67,24 @@ impl Handler<GetHighestBlockCheckpoint> for BlocksManager {
...
+                sessions_manager_addr.do_send(Broadcast {
+                    command: AnnounceItems { items },
+                });
+            }
+            Err(_) => {
+                debug!("Block already exists");
```
mmartinbar: Hmmm, perhaps for the time being an `Err` returned by `process_new_block` means exactly that. But more errors could appear (`InvalidBlock`, `AlreadyPresentBlock`...). Perhaps the `process_new_block` method should return a `WitnetResult` with the `Err` type being defined like we do with the [`PeersError`](https://github.com/witnet/witnet-rust/blob/master/p2p/src/peers/error.rs)

tmpolaczyk: I created a simple `BlocksManagerError`, but it will be replaced in the future by a `InventoryError` when the `InventoryManager` is created.

mmartinbar:
Also, please add docs for the new message in the [`BlocksManager` docs](https://github.com/witnet/witnet-rust/blob/master/docs/architecture/managers/blocks-manager.md)

core/src/actors/blocks_manager/mod.rs
```diff
@@ -84,4 +103,161 @@ impl BlocksManager {
             })
             .wait(ctx);
     }
+
+    fn process_new_block(&mut self, block: Block) -> Result<Hash, BlocksManagerError> {
+        // Calculate the hash of the block
+        let hash = calculate_sha256(&block.to_bytes().unwrap());
```
aesedepece: This is not safu.

tmpolaczyk: Right, fixed!

PR #202, opened 2018-11-28
refactor(core): Modify static to dynamic key value in storage messages

lrubiorod:
Modified StorageManager generic messages to accept dynamic key values


tmpolaczyk:
It is possible to make the `Get` and `Put` message accept both `'static` and owned values, see:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&gist=f94a0a12505bd130eff06a43b5722e58

This would remove the need for all the `PEERS_KEY.to_vec()` lines, but it may require some changes in the message handlers.

PR #212, opened 2018-11-29
feat(core): Handle GetBlocks message in Session actor

anler:
This is a fast-and-loose solution. It is intended mainly for demo purpose and we should discuss further the implementation that is missing.

### DIFF
--- a/core/src/actors/session/handlers.rs
+++ b/core/src/actors/session/handlers.rs
@@ -10,7 +10,7 @@ use log::{debug, error, info, warn};
 
 use crate::actors::{
     blocks_manager::{
-        messages::{AddNewBlock, GetHighestCheckpointBeacon},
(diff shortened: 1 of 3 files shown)


core/src/actors/session/handlers.rs
```diff
@@ -365,3 +389,74 @@ fn send_block_msg(session: &mut Session, ctx: &mut Context<Session>, hash: &Hash
...
+    session: &Session,
+    ctx: &mut Context<Session>,
+    CheckpointBeacon {
+        checkpoint: received_checkpoint,
+        ..
+    }: CheckpointBeacon,
+) {
+    // Get BlocksManager address from registry
+    let blocks_manager_addr = System::current().registry().get::<BlocksManager>();
+    // Send get highest checkpoint beacon message to BlocksManager
```
aesedepece:
```suggestion
    // Send GetHighestCheckpointBeacon message to BlocksManager
```

PR #214, opened 2018-11-29
feat(core): discard existing inventory vectors in inv handler

mmartinbar:
This PR adds a message in the `BlocksManager` to discard existing inventory vectors (in memory) from a list of inventory vectors. The handler of this message just calls a library function (tests are also added for it).

When an `Inv` message is received at the `Session`, a message is sent to the `BlocksManager` to discard the inventory vectors that already exist and to send a `GetData` message only with those. This is also added in this PR.

### DIFF
--- a/core/src/actors/blocks_manager/mod.rs
+++ b/core/src/actors/blocks_manager/mod.rs
@@ -189,27 +189,11 @@ mod tests {
     fn add_block() {
         let mut bm = BlocksManager::default();
 
-        use witnet_data_structures::chain::*;
+        // Build hardcoded block
         let checkpoint = 2;
-        let block_a = Block {
-            header: BlockHeaderWithProof {
-                block_header: BlockHeader {
-                    version: 1,
-                    beacon: CheckpointBeacon {
-                        checkpoint,
-                        hash_prev_block: Hash::SHA256([4; 32]),
-                    },
-                    hash_merkle_root: Hash::SHA256([3; 32]),
-                },
-                proof: LeadershipProof {
-                    block_sig: None,
-                    influence: 99999,
-                },
-            },
-            txn_count: 1,
-            txns: vec![Transaction],
-        };
+        let block_a = build_hardcoded_block(checkpoint, 99999);
 
+        // Add block to BlocksManager
         let hash_a = bm.process_new_block(block_a.clone()).unwrap();
 
         // Check the block is added into the blocks map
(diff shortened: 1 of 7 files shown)


core/src/actors/blocks_manager/mod.rs
```diff
@@ -154,6 +155,30 @@ impl BlocksManager {
...
+                InvVector::Error(hash)
+                | InvVector::Block(hash)
+                | InvVector::Tx(hash)
+                | InvVector::DataRequest(hash)
+                | InvVector::DataResult(hash) => hash,
+            };
+
+            // Add the inventory vector to the missing vectors if it is not found
+            if self.blocks.get(&hash).is_none() {
+                missing_inv_vectors.push(inv_vector.clone());
```
tmpolaczyk: This function would be more clear if written in functional style:

```rust
let missing_inv_vectors = inv_vectors
    .iter()
    .filter(|inv_vector| {
        let hash = /**/
        self.blocks.get(&hash).is_none()
    })
    .collect()
```

core/src/actors/session/handlers.rs
```diff
@@ -327,21 +329,55 @@ fn inventory_process_block(_session: &mut Session, _ctx: &mut Context<Session>,
...
+                },
+            }
+        })
+        // Process the received filtered inv elems
+        // This returns a FutureResult containing a success
+        .and_then(|missing_inv_vectors, act, _ctx| {
+            // Check if there are any vectors to be requested
+            if !missing_inv_vectors.is_empty() {
+                // Create GetData message with requested inventory vectors
+                let get_data_msg = WitnetMessage::build_get_data(missing_inv_vectors.to_vec());
```
tmpolaczyk: Question: what would happen if we built a GetData message with an empty vector? Will it crash?

core/src/actors/blocks_manager/mod.rs
tmpolaczyk: Also, it may make sense to take ownership of the `inv_vectors`, which would allow to use `into_iter`:
```rust
fn discard_existing_inv_vectors(&mut self, inv_vectors: Vec<InvVector>) -> InvVectorsResult {
```

core/src/actors/session/handlers.rs
mmartinbar: It shouldn't crash (we're testing it though :wink: ). Anyway, I think it may be unnecessary to send a `GetData` protocol message through the network if there's nothing to request

tmpolaczyk: An empty `GetData` is sent through the network:
```
-----> Session (V4(127.0.0.1:60208)) sending message: Message { kind: GetData(GetData { inventory: [] }), magic: 43981 }
```
So should we somehow move the check inside the message builder, so that it is impossible to send an empty `GetData` message. The same logic may also apply to `Inv` messages.

PR #243, opened 2018-12-11
feat(crypto): create merkle_tree_root function

tmpolaczyk:


aesedepece:
This is a quick and cool implementation of ["Fast Merkle Trees"](https://gist.github.com/maaku/41b0054de0731321d23e9da90ba4ee0a).

aesedepece:
This kind of trees were described in depth by the Stampery Whitepaper ([v5], [v6]).

[v5]: https://s3.amazonaws.com/stampery-cdn/docs/Stampery-BTA-v5-whitepaper.pdf
[v6]: https://s3.amazonaws.com/stampery-cdn/docs/Stampery-BTA-v6-whitepaper.pdf

tmpolaczyk:
I decided not to use any external crate because none of them meets the requirements: most crates expect the input to be data, not the hash of the data, and only a few allow custom definition of the "hash_concat" method. The most promising crates were:

https://www.crates.io/crates/merkle-generator

But the example in the README does not compile, and

https://www.crates.io/crates/mrkl

Which is relatively young, but supports the necessary customization to implement our variation of the merkle tree.

PR #248, opened 2018-12-13
feat(core): Add epoch and PoE validations to the AddNewBlock Handler of ChainManager

lrubiorod:
Added current epoch validation
Added Proof of Eligibility validation requesting to Reputation Manager
Created validation library in witnet_core to be implemented after Transaction struct has been defined
Updated docs of ReputationManager and ChainManager
Modified ValidatePoE message

### DIFF
--- a/core/src/actors/chain_manager/handlers.rs
+++ b/core/src/actors/chain_manager/handlers.rs
@@ -1,23 +1,25 @@
-use actix::{Context, Handler, System};
+use actix::{ActorFuture, Context, ContextFutureSpawner, Handler, System, WrapFuture};
 
-use crate::actors::chain_manager::{ChainManager, ChainManagerError};
+use crate::actors::chain_manager::{messages::SessionUnitResult, ChainManager, ChainManagerError};
 use crate::actors::epoch_manager::messages::EpochNotification;
 
+use crate::actors::reputation_manager::{messages::ValidatePoE, ReputationManager};
+
 use witnet_data_structures::{
-    chain::{Block, CheckpointBeacon, Hash, InventoryEntry},
+    chain::{Block, CheckpointBeacon, InventoryEntry},
     error::{ChainInfoError, ChainInfoErrorKind, ChainInfoResult},
 };
 
+use crate::validations::{validate_coinbase, validate_merkle_tree};
+
 use witnet_util::error::WitnetError;
 
-use log::{debug, error};
+use log::{debug, error, warn};
 
 use super::messages::{
     AddNewBlock, DiscardExistingInventoryEntries, GetBlock, GetBlocksEpochRange,
     GetHighestCheckpointBeacon, InventoryEntriesResult,
 };
-use crate::actors::session::messages::AnnounceItems;
-use crate::actors::sessions_manager::{messages::Broadcast, SessionsManager};
 
 ////////////////////////////////////////////////////////////////////////////////////////
 // ACTOR MESSAGE HANDLERS
@@ -45,6 +47,7 @@ impl Handler<EpochNotification<EveryEpochPayload>> for ChainManager {
 
     fn handle(&mut self, msg: EpochNotification<EveryEpochPayload>, _ctx: &mut Context<Self>) {
         debug!("Periodic epoch notification received {:?}", msg.checkpoint);
+        self.current_epoch = Some(msg.checkpoint);
     }
 }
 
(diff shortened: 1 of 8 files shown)


core/src/actors/chain_manager/mod.rs
```diff
@@ -83,6 +83,8 @@ pub struct ChainManager {
     epoch_to_block_hash: HashMap<Epoch, HashSet<Hash>>,
     /// Map that stores blocks by their hash
     blocks: HashMap<Hash, Block>,
+    /// Current Epoch
+    current_epoch: Epoch,
```
tmpolaczyk: This should probably be an `Option<Epoch>` (to handle the edge case when we don't know the current epoch), and it should be updated by the `Handler<EpochNotification<EveryEpochPayload>>`, like so:

```rust
fn handle... {
    self.current_epoch = Some(msg.checkpoint);
```

core/src/actors/chain_manager/handlers.rs
```diff
@@ -71,37 +73,62 @@ impl Handler<GetHighestCheckpointBeacon> for ChainManager {
...
+            .into_actor(self)
+            .then(|res, act, _ctx| {
+                match res {
+                    Err(e) => {
+                        // Error when sending message
+                        error!("Unsuccessful communication with reputation manager: {}", e);
+                    }
+                    Ok(valid_poe) => {
+                        let block_epoch = msg.block.header.block_header.beacon.checkpoint;
+                        if block_epoch == act.current_epoch {
```
aesedepece: This verification should be made before querying `ReputationManager` at all in order to save unnecessary asynchronous IO.

core/src/actors/chain_manager/handlers.rs
```diff
@@ -71,37 +73,62 @@ impl Handler<GetHighestCheckpointBeacon> for ChainManager {
...
+                block_header: msg.block.header.clone(),
+            })
+            .into_actor(self)
+            .then(|res, act, _ctx| {
+                match res {
+                    Err(e) => {
+                        // Error when sending message
+                        error!("Unsuccessful communication with reputation manager: {}", e);
+                    }
+                    Ok(valid_poe) => {
```
aesedepece: Huge branches like this qualify for becoming a separate callback function.

core/src/actors/chain_manager/handlers.rs
```diff
@@ -71,37 +72,59 @@ impl Handler<GetHighestCheckpointBeacon> for ChainManager {
...
+                })
+                .into_actor(self)
+                .then(|res, act, _ctx| {
+                    match res {
+                        Err(e) => {
+                            // Error when sending message
+                            error!("Unsuccessful communication with reputation manager: {}", e);
+                        }
+                        Ok(valid_poe) => {
+                            if !valid_poe {
```
aesedepece: Even if checking for errors first looks like a good idea from a _fail-fast_ mindset, I think that checking for positive assertions should be always preferred for the sake of readability:
```suggestion
                            if valid_poe {
```
Otherwise, the triggering logic for the `else` clause could become not evident, because it's a double negation.

Also, because of optimization, it's not a bad idea to expect the first branch in an assertion to be the most common one. In addition, we save the `!` operator, which comes at a mostly negligible yet different-from-zero cost.

Only case when `if !expression` is acceptable IMO is when we have no `else` clause.

What do you think? @witnet/devs

core/src/actors/chain_manager/mod.rs
```diff
@@ -192,6 +208,14 @@ impl ChainManager {
 
         Ok(missing_inv_entries)
     }
+
+    fn validate_coinbase(&mut self) -> bool {
+        true
```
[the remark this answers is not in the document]
aesedepece: The reason for the wrapping here is giving future chance for abstracting this actor away from specific library errors, and for allowing multiple parts of the actor to use these methods with a single implementation of the call to the actual validators, which guarantees uniformity.

core/src/actors/chain_manager/handlers.rs
tmpolaczyk: When checking for multiple failure conditions, it makes sense to leave the success branch as the last one.

I like `if !valid { return err; }`, as it eliminates the need for an else branch, but in this case it can be nicely replaced with:

```rust
                    match res {
                        Err(e) => {
                            // Error when sending message
                            error!("Unsuccessful communication with reputation manager: {}", e);
                        }
                        Ok(false) => {
                            warn!("Block PoE not valid"); 
                        }
                        Ok(true) => {
                            // Valid
                        }
```

core/src/actors/chain_manager/mod.rs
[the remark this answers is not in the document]
lrubiorod: I have created a library named validations.rs in witnet_core which has these two functions (Now they only return true)

core/src/actors/chain_manager/handlers.rs
lrubiorod: Nice!!

PR #251, opened 2018-12-17
feat(core): add BuildBlock message to ChainManager

tmpolaczyk:
This PR adds a method to create new blocks with a given proof of eligibility. It gets the maximum number of transactions (currently unimplemented). The block is then validated in the `AddNewBlock` handler, following the same path as external blocks (blocks from the network).

TODOs:

* Add coinbase transaction
* Calculate the total fee of the transactions
* Should there be any helper methods to build Block/BlockHeader/LeadershipProof?

### DIFF
--- a/core/src/actors/chain_manager/handlers.rs
+++ b/core/src/actors/chain_manager/handlers.rs
@@ -24,6 +24,7 @@ use super::messages::{
 };
 
 use witnet_crypto::{hash::calculate_sha256, merkle::merkle_tree_root};
+use witnet_data_structures::chain::Transaction;
 
 ////////////////////////////////////////////////////////////////////////////////////////
 // ACTOR MESSAGE HANDLERS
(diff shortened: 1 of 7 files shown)


core/src/actors/chain_manager/handlers.rs
```diff
@@ -93,7 +97,7 @@ impl Handler<AddNewBlock> for ChainManager {
             // Request proof of eligibility validation to ReputationManager
             reputation_manager_addr
                 .send(ValidatePoE {
-                    beacon: msg.block.block_header.beacon.clone(),
+                    beacon: msg.block.block_header.beacon,
```
tmpolaczyk: This is unrelated to the PR, but clippy fails on master

tmpolaczyk:
Right now the only missing parts are related to the definition of the `Transaction` struct, so I propose to merge this PR now.

core/src/actors/chain_manager/mod.rs
```diff
@@ -186,6 +190,40 @@ impl ChainManager {
...
+        let mut transaction_fees = 0;
+        let transactions: Vec<Transaction> = self
+            .transactions_pool
+            .iter()
+            .map(|t| {
+                // TODO: t.fee()
+                transaction_fees += 1;
+                *t
+            })
+            .collect();
```
anler: _Reminder_
This is pending a refactor: take only as many transactions as they can fit in a block.

PR #255, opened 2018-12-18
feat(core): Add block_candidate state and chain_info update

lrubiorod:
Added block_candidate to the state of ChainManager
Updated chain_info beacon in EpochNotification handler
Added comparing bloc received  to block_candidate validation in AddNewBlock handler
Updated block_candidate in AddNewBlock handler
Added an empty method to send a valid block to InventoryManager (Another Issue)

### DIFF
--- a/core/src/actors/chain_manager/handlers.rs
+++ b/core/src/actors/chain_manager/handlers.rs
@@ -8,7 +8,7 @@ use crate::actors::epoch_manager::messages::EpochNotification;
 use crate::actors::reputation_manager::{messages::ValidatePoE, ReputationManager};
 
 use witnet_data_structures::{
-    chain::{Block, CheckpointBeacon, InventoryEntry},
+    chain::{Block, CheckpointBeacon, Hashable, InventoryEntry},
     error::{ChainInfoError, ChainInfoErrorKind, ChainInfoResult},
 };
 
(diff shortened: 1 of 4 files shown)


core/src/actors/chain_manager/handlers.rs
```diff
@@ -45,9 +45,33 @@ impl Handler<EpochNotification<EpochPayload>> for ChainManager {
 impl Handler<EpochNotification<EveryEpochPayload>> for ChainManager {
     type Result = ();
 
-    fn handle(&mut self, msg: EpochNotification<EveryEpochPayload>, _ctx: &mut Context<Self>) {
+    fn handle(&mut self, msg: EpochNotification<EveryEpochPayload>, ctx: &mut Context<Self>) {
         debug!("Periodic epoch notification received {:?}", msg.checkpoint);
         self.current_epoch = Some(msg.checkpoint);
+
+        if let Some(candidate) = self.block_candidate.clone() {
```
tmpolaczyk: You can use the `take()` method to return the `Block` and replace the value of an `Option` with `None`.

```suggestion
        if let Some(candidate) = self.block_candidate.take() {
```

core/src/actors/chain_manager/handlers.rs
```diff
@@ -45,9 +45,33 @@ impl Handler<EpochNotification<EpochPayload>> for ChainManager {
...
+                    chain_info.highest_block_checkpoint = beacon;
+                    self.chain_info = Some(chain_info);
+                }
+                None => {
+                    error!("No ChainInfo loaded in ChainManager");
+                }
+            }
+
+            // Delete current block candidate
+            self.block_candidate = None;
```
tmpolaczyk: Not needed if you use `take()` above.
```suggestion
```

core/src/actors/chain_manager/handlers.rs
```diff
@@ -45,9 +45,33 @@ impl Handler<EpochNotification<EpochPayload>> for ChainManager {
...
+    fn handle(&mut self, msg: EpochNotification<EveryEpochPayload>, ctx: &mut Context<Self>) {
         debug!("Periodic epoch notification received {:?}", msg.checkpoint);
         self.current_epoch = Some(msg.checkpoint);
+
+        if let Some(candidate) = self.block_candidate.clone() {
+            // Send block to Inventory Manager
+            self.persist_block(ctx, &candidate);
+
+            // Update chain_info
+            match self.chain_info.clone() {
```
tmpolaczyk: To mutate the value inside of an `Option`, you can use the `as_mut()` method:

```rust
            // Update chain_info
            match self.chain_info.as_mut() {
                Some(chain_info) => {
                    let beacon = CheckpointBeacon {
                        checkpoint: msg.checkpoint,
                        hash_prev_block: candidate.hash(),
                    };

                    chain_info.highest_block_checkpoint = beacon;
                }
                None => {
                    error!("No ChainInfo loaded in ChainManager");
                }
            }
```

PR #256, opened 2018-12-18
feat(data structures): add TransactionsPool struct

anler:
Add TransactionsPool struct which is a container of Transaction objects providing constant time lookup by transaction-hash and iteration over the transactions sorted by K, where:

```
    transaction fee
K = ------------------
    transaction weight
```

This pool is used internally by the ChainManager as its mempool.

See the generated docs for the data structure for more examples.

### DIFF
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -1,4 +1,5 @@
 use super::serializers::{build_block_flatbuffer, BlockArgs};
+use std::collections::{BTreeSet, HashMap};
 use witnet_crypto::hash::{calculate_sha256, Sha256};
 
 pub trait Hashable {
@@ -143,7 +144,7 @@ pub struct Secp256k1Signature {
 }
 
 /// Hash
-#[derive(Debug, Eq, PartialEq, Copy, Clone, Serialize, Deserialize, Hash)]
+#[derive(Debug, Eq, PartialEq, PartialOrd, Ord, Copy, Clone, Serialize, Deserialize, Hash)]
 pub enum Hash {
     /// SHA-256 Hash
     SHA256(SHA256),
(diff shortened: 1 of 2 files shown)


tmpolaczyk:
Would it make sense to cache transaction-fee and transaction-weight somewhere? It would take 2 `u64` and it saves us manually calculating these values every time. However they should be accessible externally, we may want to iterate until a certain total size, and keep track of the total transaction fee for the coinbase.

What about:

```rust
pub struct TransactionsPool {
    transactions: HashMap<Hash, (Transaction, Fee, Weight)>,
    sorted_index: BTreeSet<WeightedHash>
}
```

And the iterator:
```rust
pub fn iter(&self) -> impl Iterator<Item = &(Transaction, Fee, Weight)> {
    self.sorted_index.iter().rev().filter_map(move |WeightedHash(_, h)| self.transactions.get(h))
}
```

Or maybe even a dedicated `TransactionWithCache` struct.

tmpolaczyk:
There are some missing methods: `get`, `remove`, `clear`.

Since one common operation is to remove many elements at once, a `retain` method would also be nice, along the lines of:

```rust
pub fn retain<F>(&mut self, mut f: F)
where
    F: FnMut(&K, &mut V) -> bool 
{
    self.transactions.retain(|k, v| {
        let retain_item = f(k, v);
        if !retain_item {
            self.sorted_index.remove(k);
        }

        retain_item
    });
}
```

anler:
@tmpolaczyk Regarding caching `fee` and `weight` of the transaction, currently I don't see the need, because is only calculated once when inserting a transaction into the pool, and after that is not used anymore, at least by the pool, so I wouldn't cache it or I'd do it in another place outside the pool.

tmpolaczyk:
@anler yes, I was thinking outside of the pool, we at least need to calculate the total value of the transactions when building a new block, but this can be added later if we see that it really is a bottleneck.

aesedepece:
@tmpolaczyk it sounds to me as if that should be an accumulator in the iterator that takes transactions from the pool into the block until full.

data_structures/src/chain.rs
```diff
@@ -164,6 +165,166 @@ pub type SHA256 = [u8; 32];
...
+    /// let mut pool = TransactionsPool::new();
+    ///
+    /// pool.insert(Hash::SHA256([0 as u8; 32]), Transaction);
+    /// pool.insert(Hash::SHA256([0 as u8; 32]), Transaction);
+    ///
+    /// let mut iter = pool.iter();
+    /// let tx1 = iter.next();
+    /// let tx2 = iter.next();
+    ///
+    /// // TODO: assert!(tx1.weight() >= tx2.weight());
```
aesedepece: That's paranoid :stuck_out_tongue_closed_eyes:

data_structures/src/chain.rs
```diff
@@ -164,6 +165,202 @@ pub type SHA256 = [u8; 32];
...
+    /// let mut pool = TransactionsPool::new();
+    /// pool.insert(Hash::SHA256([0 as u8; 32]), Transaction);
+    /// pool.insert(Hash::SHA256([1 as u8; 32]), Transaction);
+    /// assert_eq!(pool.len(), 2);
+    /// pool.retain(|h, _| match h { Hash::SHA256(n) => n[0]== 0 });
+    /// assert_eq!(pool.len(), 1);
+    /// ```
+    pub fn retain<F>(&mut self, mut f: F)
+    where
+        F: FnMut(&Hash, &mut Transaction) -> bool,
```
tmpolaczyk: The retain method is more useful if we allow to mutate the values, but that conflicts with our cache idea (because we would need to check for every item that did change and recalculate the cache). Also, the transactions are inherently immutable so I think a good compromise would be:

```suggestion
        F: FnMut(&Hash, &Transaction) -> bool,
```

tmpolaczyk: It should be `FnMut`, not `Fn`, like

https://doc.rust-lang.org/std/vec/struct.Vec.html#method.retain

PR #260, opened 2018-12-19
docs(network): redesign transactions to include inputs

mariocao:
This PR includes:

- add signatures to transaction claims, as it is necessary to avoid that transactions could be malleable
- redesign transaction data structure to include additional input fields and removes some unnecessary claim fields

### DIFF
--- a/docs/protocol/network/data-structures/transaction.md
+++ b/docs/protocol/network/data-structures/transaction.md
@@ -110,32 +110,34 @@ In Witnet, different output types implicitly have different spending conditions 
 
 Value transfer claims prove ownership of a certain private key by providing a signature of the identifier of the transaction produced with such key and the serialization of the matching public key.
 
-| Field        |  Type  | Description                                     |
-| ------------ | :----: | ----------------------------------------------- |
-| `signature`  | `[u8]` | Signed hash of the transaction                  |
-| `public_key` | `[u8]` | Public Key of the P2PKH outpoint to be consumed |
+| Field        |  Type  | Description                                                |
+| ------------ | :----: | ---------------------------------------------------------- |
+| `signature`  | `[u8]` | Signature of the transaction digest, i.e. `transaction_id` |
+| `public_key` | `[u8]` | Public Key of the P2PKH outpoint to be consumed            |
 
 ### Data request claim (**commit stage**)
 
 A **data request** output has to be consumed/used by a number of **witnesses**, i.e. there will be as many claims as **witnesses** have been defined. For a committer to be able to pledge a share of the reward from the data request, they must provide a _Proof of Eligibility_ (PoE) that proves their eligibility as witnesses for such data request in the current epoch. In addition, for everyone in the network to be able to verify such proof, they must include the public key that matches the private key that produced the PoE.
 
-| Field        |  Type  | Description                           |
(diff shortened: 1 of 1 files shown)


aesedepece:
General comments:
- The name _claims_ no longer make sense as this field no longer accommodates any kind of response to challenges but only signatures. I can't see why wouldn't this be called _signatures_ now*. 
- Inputs are no longer outpoints. They now couple (1) an outpoint to (2) the _response to the spending conditions sported by the output referred by the outpoint_. 
- Be careful when saying `spending conditions`. That's the term for the explicit or implicit script found in the outputs, not the elements included in inputs for fulfilling the unlock of the funds in those inputs. Therefore _spending conditions_ is synonym to _spending requirements_ or _spending challenge_. This is Bitcoin's _scriptPubKey_.
- The second element in each input, which tries to satisfy the spending conditions, doesn't have a name now, but I think we can reuse _claims_ or come up with a new name. This is equivalent to Bitcoin's _scriptSig_.
- I don't think we should speak about different input types. They are all the same type, but they just try to satisfy the requirements of different output types. Therefore, just like before, it's actually the _claims_ (the second element inside each input) which have different types, each matching an output type.  

<sub>* I know, I know, because of P2SH. But that's simply not a concern for us right now.</sub>

docs/protocol/network/data-structures/transaction.md
```diff
@@ -2,24 +2,58 @@
...
-| `version` |   `u32`    | The transaction data format version number  |
-| `inputs`  | `[input]`  | A list of transaction inputs                |
-| `outputs` | `[output]` | A list of 1 or more transaction outputs     |
-| `claims`  | `[claim]`  | A list of claims (i.e. spending conditions) |
+| Field     |    Type    | Description                                |
+| --------- | :--------: | ------------------------------------------ |
+| `version` |   `u32`    | The transaction data format version number |
+| `inputs`  | `[input]`  | A list of transaction inputs               |
+| `outputs` | `[output]` | A list of 1 or more transaction outputs    |
+| `claims`  | `[claim]`  | A list of claims                           |
```
girazoki: Might we want to change claims for signatures here? I find the word ambiguous

docs/protocol/network/data-structures/transaction.md
```diff
@@ -85,59 +120,45 @@ commit_value = (data_request_value / witnesses) - commit_fee
...
-| `poe`        | `[u8]` | Proof of eligibility                  |
-| `public_key` | `[u8]` | Public Key used for computing the PoE |
-
-### Commit claim (**reveal stage**)
+```math
+reveal_value = data_request_value - committers * commit_fee - revealers * (reveal_fee + tally_fee + reward)
+```
 
-A commit claim is used to prove that the witness has a valid `reveal`, i.e. the signed data request result and nonce match the previous commitment.
+## Claims
```
girazoki: Same as before, I would just call them signatures. These they can serve as claims (in the cases of Value Transfer Outputs). I feel the name is confusing

aesedepece:
Merged as 5139a0cf86e41f5b79250a4a08e31f84684ea5d8 and 3313136c458b4e01c3c1e46e567b037ee8157fe9.

PR #261, opened 2018-12-19
feat(core): Add validation merkle tree function

lrubiorod:
Added create_merkle_tree function
Added validate_merkle_tree function

### DIFF
--- a/core/src/validations.rs
+++ b/core/src/validations.rs
@@ -1,4 +1,6 @@
-use witnet_data_structures::chain::Block;
+use witnet_crypto::hash::Sha256;
+use witnet_crypto::merkle::merkle_tree_root as crypto_merkle_tree_root;
+use witnet_data_structures::chain::{Block, Hash, Hashable, Transaction};
 
 /// Function to validate block's coinbase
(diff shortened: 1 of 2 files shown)


core/src/validations.rs
```diff
@@ -1,13 +1,29 @@
...
+use witnet_data_structures::chain::{Block, Hash, Hashable, Transaction};
 
 /// Function to validate block's coinbase
 pub fn validate_coinbase(_block: &Block) -> bool {
     // TODO Implement validate coinbase algorithm
     true
 }
 
+/// Function to calculate a merkle tree from a transaction vector
+pub fn create_merkle_tree(transactions: &[Transaction]) -> Hash {
```
anler: _Suggestion_
I don't like the name `create_merkle_tree` too much since is basically the same function as `merkle_tree_root` but lifted to work with Transaction and Hash instead of Sha256. I would just call it `merkle_tree_root` to be consistent with the one in `witnet_crypto::merkle`.

aesedepece:
Hmmm good point, as we are always initializing, filling and closing the
tree at once.

PR #264, opened 2018-12-20
Adapt transaction format to latest discussions

aesedepece:
This supersedes #260, but it's still WIP as it's missing _coinbase_ documentation.


mariocao:
Great! :100:

aesedepece:
This documentation page now accounts for the specific validation rules applicable to different output types, and introduces the concept of _mints_, a specific validation rule for _value transfer_ outputs:
 
> - A single VTO spending from no inputs is considered to be a _mint_ transaction, which is only acceptable if located first in the list of transactions of a block.

PR #266, opened 2018-12-20
Fix bugs found while trying mining demo

tmpolaczyk:
The following works correctly:

* The block with the lowest hash will be chosen when there are 2 candidates
* A new node will ask for the all the blocks that it is missing
* The blocks are correctly mined and sent
* The blocks are successfully stored and retrieved from the `InventoryManager`

Aside from a few minor bugs, the following changes should be discussed:

* replace MessagePack with Json as a temporary workarround for a MessagePack bug
* accept blocks from a past epoch
* add randomness to mining manager to help testing the block candidate logic

### DIFF
--- a/core/src/actors/chain_manager/handlers.rs
+++ b/core/src/actors/chain_manager/handlers.rs
@@ -71,6 +71,9 @@ impl Handler<EpochNotification<EveryEpochPayload>> for ChainManager {
 
             // Send block to Inventory Manager
             self.persist_item(ctx, InventoryItem::Block(candidate));
+
+            // Persist chain_info into storage
+            self.persist_chain_info(ctx);
         }
     }
 }
--- a/core/src/actors/chain_manager/handlers.rs
+++ b/core/src/actors/chain_manager/handlers.rs
@@ -16,12 +16,14 @@ use crate::validations::{validate_coinbase, validate_merkle_tree};
 
 use witnet_util::error::WitnetError;
 
-use log::{debug, error, warn};
+use log::{debug, error, info, warn};
 
 use super::messages::{
     AddNewBlock, BuildBlock, DiscardExistingInventoryEntries, GetBlock, GetBlocksEpochRange,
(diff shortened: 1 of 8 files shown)


core/src/actors/chain_manager/mod.rs
```diff
@@ -483,20 +480,174 @@ mod tests {
...
+            influence,
+        };
+
         Block {
             block_header: BlockHeader {
                 version: 1,
                 beacon: CheckpointBeacon {
                     checkpoint,
-                    hash_prev_block: Hash::SHA256([4; 32]),
+                    hash_prev_block: Hash::SHA256([111; 32]),
```
lrubiorod: Very important change

core/src/actors/chain_manager/mod.rs
```diff
@@ -483,20 +480,174 @@ mod tests {
...
+            s: signed_beacon_hash,
+            v: 0,
+        }));
+        let msp = a.to_bytes().unwrap();
+
+        assert_eq!(Option::<Signature>::from_bytes(&msp).unwrap(), a);
+    }
+
+    #[test]
+    fn som_de_uno() {
```
lrubiorod:
```suggestion
    fn som_som_de_uno() {
```

core/src/actors/chain_manager/handlers.rs
```diff
@@ -110,14 +113,65 @@ impl Handler<AddNewBlock> for ChainManager {
...
+            );
+        } else if Some(block_epoch) < self.current_epoch {
+            warn!(
+                "Block epoch mismatch: current: {}, block: {}",
+                self.current_epoch.unwrap(),
+                block_epoch
+            );
+            // Add old block
+            // FIXME(#235): check proof of eligibility from the past
+            // ReputationManager should have a method to validate PoE from a past epoch
```
aesedepece: This is in line with my suggestion on Discord. Is someone working on it?

PR #270, opened 2018-12-21
Update blocks specification to reflect latest thoughts

aesedepece:
This PR updates blocks specification to reflect latest thoughts and changes regarding signatures and malleability challenges.

As you can see, this forces miners to sign the blocks they produce. That's not necessary for Bitcoin, but it is for us because we don't take the merkle root into account when computing the proof of eligibility (to prevent people from using _PoW_ for gaming our algorithm). We need the signature in order to ensure the merkle tree is not malleable by third parties.  

I also took this chance to introduce the distinction between _keyed_ signatures (those accompanied by the public key) and plain signatures, for which we are supposed to be already in knowledge of their matching key.

E.g.: the block signature is a _keyed_ signature, while the _Proof of Eligibility_ is a plain signature, as the public key is already provided in the block signature.

Also note that I have dropped the _Proof of Leadership_ term in favor for the general _Proof of Eligibility_.


aesedepece:
This also removes the distinction between the block header data structure _with_ and _without_ proof, as it takes the signature out of the header itself :boom:

docs/protocol/network/data-structures/transaction.md
```diff
@@ -240,13 +240,11 @@ The only distinctive feature of _reveal_ inputs is that they do not require matc
 
 As aforementioned, transactions should include as many signatures as inputs. In every transaction, signatures complement the material required for satisfying the spending conditions that encumbered the past transaction outputs that the inputs in the transaction are trying to spend. Signatures and inputs are matched positionally, i.e. the first claim is checked against the first input and so forth.
 
-Signatures prove ownership of a certain private key by providing a signature of the identifier of the transaction produced with such key and the serialization of the matching public key.
+Signatures prove ownership of a certain private key by providing a signature of the identfier of the transaction produced with such key and the serialization of the matching public key.
```
mariocao:
```suggestion
Signatures prove ownership of a certain private key by providing a signature of the identifier of the transaction produced with such key and the serialization of the matching public key.
```

docs/protocol/network/data-structures/transaction.md
```diff
@@ -2,12 +2,12 @@
...
-| `version`    | `u32`         | The transaction data format version number |
-| `inputs`     | `[input]`     | A list of transaction inputs               |
-| `outputs`    | `[output]`    | A list of 1 or more transaction outputs    |
-| `signatures` | `[signature]` | A list of claims (as many as inputs)       |
+| Field        | Type                | Description                                |
+|--------------|---------------------|--------------------------------------------|
+| `version`    | `u32`               | The transaction data format version number |
+| `inputs`     | `[input]`           | A list of transaction inputs               |
+| `outputs`    | `[output]`          | A list of 1 or more transaction outputs    |
+| `signatures` | `[keyed_signature]` | A list of signatures (as many as inputs)   |
```
mariocao: Perhaps if not too verbose?
```suggestion
| `signatures` | `[keyed_signature]` | A list of keyed signatures (as many as inputs)   |
```

PR #272, opened 2018-12-21
feat(chain manager): add as many transactions as fit in block

ghost:
Store as many transactions as they can fit when building a block, using the weight parameter.

### DIFF
--- a/config/src/defaults.rs
+++ b/config/src/defaults.rs
@@ -96,6 +96,11 @@ pub trait Defaults {
     fn mining_enabled(&self) -> bool {
         true
     }
+
+    fn consensus_constants_max_block_weight(&self) -> u32 {
+        // TODO: Replace  with real max_block_weight value used in mainnet
+        10_000
+    }
 }
 
 /// Struct that will implement all the mainnet defaults
--- a/core/src/actors/chain_manager/mod.rs
+++ b/core/src/actors/chain_manager/mod.rs
@@ -41,7 +41,7 @@ use std::collections::HashMap;
 use std::collections::HashSet;
 use witnet_data_structures::chain::{
     Block, BlockHeader, ChainInfo, Epoch, Hash, Hashable, InventoryEntry, InventoryItem,
-    Transaction, TransactionsPool,
+    TransactionsPool,
 };
 
 use crate::actors::session::messages::AnnounceItems;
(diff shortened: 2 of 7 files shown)


core/src/actors/chain_manager/mod.rs
```diff
@@ -226,15 +228,28 @@ impl ChainManager {
...
-                // TODO: t.fee()
-                transaction_fees += 1;
-                *t
-            })
-            .collect();
+        let max_block_weight = self.max_block_weight;
+        let mut block_weight = 0;
+        let mut transactions = Vec::new();
+
+        for transaction in self.transactions_pool.iter() {
```
Tommytrg: Not functional enough

anler: Nah, the thing is that I added a small optimization: if the weight of the included transactions is equal to the maximum weight allowed for the block, then there's no need to keep processing more transactions so a `break` is used to stop the loop.

config/src/defaults.rs
```diff
@@ -91,6 +91,11 @@ pub trait Defaults {
 
     /// Default JSON-RPC server addr
     fn jsonrpc_server_address(&self) -> SocketAddr;
+
+    fn consensus_constants_max_block_weight(&self) -> u32 {
+        // TODO: Replace with real max_block_weight value used in mainnet
```
Tommytrg: Replace TODO by FIXME(#issue_number) to be able to tack this.

core/src/actors/chain_manager/mod.rs
```diff
@@ -225,17 +227,29 @@ impl ChainManager {
...
+
+        let max_block_weight = self.max_block_weight;
+        let mut block_weight = 0;
+        let mut transactions = Vec::new();
+
+        for transaction in self.transactions_pool.iter() {
+            // Currently, 1 weight unit is equivalent to 1 byte
+            let transaction_weight = transaction.size();
+            // TODO: Calculate transaction fee
+            let transaction_fee = 1;
```
aesedepece: Let's take this chance to create `transaction.fee()` returning a static `1`. Then incorporating the actual fee calculation logic will be plug-and-play.

data_structures/src/chain.rs
```diff
@@ -94,6 +105,13 @@ pub struct Block {
     pub txns: Vec<Transaction>,
 }
 
+/// Any reference to a Hashable type is also Hashable
+impl<'a, T: Hashable> Hashable for &'a T {
```
aesedepece: This is wonderful :heart_eyes:

PR #290, opened 2018-12-31
feat(chain_manager): implement AddTransaction handler

Tommytrg:
### [WIP] Need to refactor: refactor methods into library code

Scratched lines should be validated in other steps like block validation.

Validations made:

- Value transfer output
  - [x] VTOs take their value from the aggregate of all the inputs in the transactions.
  - [x] The number of VTOs in a single transaction is virtually unlimited as long as the VTOs are all contiguous and located at the end of the outputs list.
  - [ ] A single VTO spending from no inputs is considered to be a mint transaction, which is only acceptable if located first in the list of transactions of a block.
  - [x] The value brought into a transaction by an input pointing to a VTO can be freely assigned to any output of any type unless otherwise restricted by the specific validation rules for such output type.
- Data request output
  - [x] Multiple data request outputs can be included into a single transaction as long as the inputs are greater than outputs rule still hold true. The difference with VTOs is that the total output value for data request outputs also includes the commit fee, reveal fee and tally fee.
  - [x] The value brought into a transaction by an input pointing to a data request output can only be spent by commit outputs.
- Commit outputs
  - [x] Commit outputs can only take value from data request inputs whose index in the inputs list is the same as their own index in the outputs list.
  - [x] Multiple commit outputs can exist in a single transaction, but each of them needs to be coupled with a data request input occupying the same index in the inputs list as their own in the outputs list. Predictably, as a result of the previous rule, each of the multiple commit outputs only takes value from the data request input with the same index.
  - [x] The value brought into a transaction by an input pointing to a commit output can only be spent by reveal or tally outputs.
- Reveal output
  - [x] Reveal outputs can only take value from commit inputs whose index in the inputs list is the same as their own index in the outputs list.
  - [x] Multiple reveal outputs can exist in a single transaction, but each of them needs to be coupled with a commit input occupying the same index in the inputs list as their own in the outputs list. Predictably, as a result of the previous rule, each of the multiple reveal outputs only takes value from the commit input with the same index.
  - [x] The value brought into a transaction by an input pointing to a reveal output can only be spent by value transfer outputs.
  - [x] Any transaction including an input pointing to a reveal output must also include exactly only one tally output.
- Tally output
  - [x] Any transaction can contain at most one tally output.
  - [ ] Transactions containing tally outputs must not be broadcast through the inventory announcement protocol.
  - [ ]  As a result of the previous rule, transactions containing tally outputs can only be included into a block by the miner of the block.
  - [ ] The value brought into a transaction by an input pointing to a tally output can be freely assigned to any output of any type unless otherwise restricted by the specific validation rules for such output type.


core/src/actors/chain_manager/handlers.rs
```diff
@@ -102,20 +102,168 @@ impl Handler<AddNewBlock> for ChainManager {
...
+
+    outputs
+        .iter()
+        .rev()
+        .take_while(is_value_transfer_output)
+        .eq(outputs.iter().rev().filter(is_value_transfer_output))
+}
+
+fn count_consensus_outputs(outputs: &[Output]) -> usize {
+    outputs.iter().fold(0, |mut counter, output| {
```
tmpolaczyk: What about

```suggestion
    outputs.iter().filter(|x| is_consensus_output(x)).count()
```

PR #291, opened 2018-12-31
Filter out 0.0.0.0 "unspecified" address when adding new peer addresses

### LINKED ISSUE #282: Don't accept 0.0.0.0 as peer address
When processing peers addresses lists, any `0.0.0.0` address should be dropped, not stored. This would reduce the failed connections ratio and speed up the bootstrapping process.

aesedepece:
Fix #282

### DIFF
--- a/p2p/src/peers/mod.rs
+++ b/p2p/src/peers/mod.rs
@@ -36,6 +36,8 @@ impl Peers {
         // Note: if the peer address exists, the peer info will be overwritten
         Ok(addrs
             .into_iter()
+            // Filter out unspecified addresses (aka 0.0.0.0)
+            .filter(|address| !address.ip().is_unspecified())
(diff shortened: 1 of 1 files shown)


tmpolaczyk:
The standard library already provides some of these methods: `is_unspecified` and `is_loopback`, but the `is_global` method is not stable so we can implement it ourselves.

https://doc.rust-lang.org/std/net/enum.IpAddr.html

aesedepece:
Oh my LOL.

tmpolaczyk:
No, wait, https://doc.rust-lang.org/std/net/struct.Ipv4Addr.html provides the missing methods, you can safely remove most of the logic from this PR

aesedepece:
Past-testnet, switchable `loopback` and `global` filters may be added.

Issue #176, opened 2018-11-22
Put JsonRpcServer into its own Arbiter


aesedepece:

mariocao:
This issue is still relevant as it is an attack vector for those nodes with public Json RPC server.

PR #292, opened 2018-12-31
getBlockChain

### LINKED ISSUE #280: getBlockChain JSON-RPC method
The JSON-RPC interface should expose a `getBlockChain` method that will return a list containing tuples of epochs and block header digests:

```json
[
  [ 0, "digest0" ],
  [ 1, "digest1" ]
]
```

For this feature to be implemented, we have two options:
- (efficient) have such list cached in memory and persisted to storage 
- (naive) calculate the list on the fly by tracing the chain back from the tip to the genesis

tmpolaczyk:
Closes #280 

Implement the getBlockChain JSON-RPC method.

One major change is that now JSON-RPC requests are handled asynchronously, which looks like it solves #176.

Example request:

     {"jsonrpc": "2.0","method": "getBlockChain", "id": 1}

Example response:

    {"jsonrpc":"2.0","result":[[0,"ed28899af8c3148a4162736af942bc68c4466da93c5124dabfaa7c582af49e30"],[1,"9c9038cfb31a7050796920f91b17f4a68c7e9a795ee8962916b35d39fc1efefc"]],"id":1}

The hashes are obtained from the `ChainManager`, from the `epoch_to_block_hash` map, which should be stored in the storage as it is used to synchronize two nodes. Right now it is not stored, so the blockchain is lost when all the nodes shutdown. (Alternatively, the map could be rebuilt when the node starts, but it would need to naively read all the blocks from the storage one by one). Right now the blocks are stored, but the node is unable to send them to new nodes unless some node asks specifically for the block hash...

### DIFF
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -4,6 +4,7 @@ use super::serializers::encoders::{
 };
 use std::collections::{BTreeSet, HashMap};
 use std::convert::AsRef;
+use std::fmt;
 use witnet_crypto::hash::{calculate_sha256, Sha256};
 
 pub trait Hashable {
@@ -203,6 +204,19 @@ impl Default for Hash {
 impl From<Sha256> for Hash {
     fn from(x: Sha256) -> Self {
         Hash::SHA256(x.0)
+    }
+}
+
+impl fmt::Display for Hash {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            Hash::SHA256(h) => f.write_str(
+                &h.into_iter()
+                    .fold(String::new(), |acc, x| format!("{}{:02x}", acc, x)),
+            )?,
+        };
+
+        Ok(())
     }
 }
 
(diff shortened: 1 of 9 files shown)


data_structures/src/chain.rs
```diff
@@ -206,6 +207,19 @@ impl From<Sha256> for Hash {
     }
 }
 
+impl fmt::Display for Hash {
+    fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
```
aesedepece: Why not this?
```suggestion
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
```

tmpolaczyk:
When two nodes interchange the `LastBeacon` message, they use the `epoch_to_block_hash` map to compare their respective chains. I understand that this is not correct, and they should only consider consolidated blocks?

I think this should be a separated issue: it's related to the blocks mempool, currently we store the block candidates in the storage, and it shouldn't be that way. The `epoch_to_block_hash` should store only consolidated blocks, and a new structure would keep the block candidates (and discard candidates from old epochs?).

aesedepece:
Unless my current headache is leading me into ill-judging, I do whink we should only consider consolidated blocks for the sake of communicating `LastBeacon` to our peers.

I agree with you that we'd be better off moving those changes to a separate issue.

aesedepece:
I have created #295 to track the aforementioned changes.

Issue #295, opened 2018-12-31
Change purpose of `epoch_to_block_hash` map in ChainManager 


aesedepece:
As discussed in #292:

> currently we store the block candidates in the storage, and it shouldn't be that way. The `epoch_to_block_hash` should store only consolidated blocks, and a new structure would keep the block candidates (and discard candidates from old epochs?).

Bottom line:
- `epoch_to_block_hash` is supposed to map checkpoint numbers to the digest of the best block in the subsequent epoch
- `epoch_to_block_hash` is thus bound to be useful as a linear representation of the block chain
- `epoch_to_block_hash` is a persisted collection (see #293 and #294)
- Thus, block candidates shouldn't be present in `epoch_to_block_hash`
- If block candidates happen to need such mapping (because of slow validations), we can have a separate mapping.
- We can rename `epoch_to_block_hash` to just `block_chain`. Elegant, right?

aesedepece:
@mariocao Up to what point does #309 solve or invalidate this issue?

tmpolaczyk:
Fixed in #373

PR #309, opened 2019-01-04
feat(chain_manager): persist blocks index into storage

### LINKED ISSUE #293: Persist `epoch_to_block_hash` blocks index on write
We should persist the `epoch => block_digest` into local storage upon write so that we can load it again when we start the app (see #294).

mariocao:
Closes #293 

Remarks:

1. I already took the opportunity to rename `epoch_to_block_hash` to `block_chain` (mentioned in #295)
2. Maybe we could refactor the way in which we persist block chain status into storage, as we are storing `chain_info` and `block_chain` separately (i.e. using 2 `Put` messages).

### DIFF
--- a/core/src/actors/chain_manager/mod.rs
+++ b/core/src/actors/chain_manager/mod.rs
@@ -35,7 +35,7 @@ use ansi_term::Color::Purple;
 use crate::actors::{
     chain_manager::messages::InventoryEntriesResult,
     inventory_manager::{messages::AddItem, InventoryManager},
-    storage_keys::CHAIN_KEY,
+    storage_keys::{BLOCK_CHAIN_KEY, CHAIN_KEY},
     storage_manager::{messages::Put, StorageManager},
 };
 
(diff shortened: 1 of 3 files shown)


core/src/actors/chain_manager/mod.rs
```diff
@@ -147,6 +147,33 @@ impl ChainManager {
...
+        // future within context, but context waits until this future resolves
+        // before processing any other events.
+        let msg = Put::from_value(BLOCK_CHAIN_KEY, &self.block_chain).unwrap();
+        storage_manager_addr
+            .send(msg)
+            .into_actor(self)
+            .then(|res, _act, _ctx| {
+                match res {
+                    Ok(Ok(_)) => {
+                        debug!("ChainManager successfully persisted block_chain into storage")
```
aesedepece:
```suggestion
                        debug!("Successfully persisted block_chain into storage")
```

core/src/actors/chain_manager/mod.rs
```diff
@@ -147,6 +147,33 @@ impl ChainManager {
...
+                    }
+                    _ => {
+                        error!("ChainManager failed to persist block_chain into storage");
```
aesedepece:
```suggestion
                        error!("Failed to persist block_chain into storage");
```

PR #318, opened 2019-01-09
feat(data-structures): Add encode and decode to transaction structure

lrubiorod:
Co-authored-by: Tommytrg <tomastrg22@gmail.com>

Implemented encode and decode functions to Transaction Command
Add build_transaction test

### DIFF
--- a/data_structures/src/serializers/decoders.rs
+++ b/data_structures/src/serializers/decoders.rs
@@ -129,7 +129,17 @@ impl TryFrom<Vec<u8>> for Message {
                 .and_then(|peers| create_peers_message(PeersWitnetArgs { magic, peers }))
                 .ok_or(""),
 
-            protocol::Command::Transaction => panic!("Unimplemented"), // Unimplemented,
+            protocol::Command::Transaction => message
+                .command_as_transaction()
+                .and_then(|transaction| {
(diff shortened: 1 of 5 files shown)


data_structures/src/serializers/encoders.rs
```diff
@@ -464,6 +487,32 @@ fn build_block_message_flatbuffer<'a>(
     build_message_flatbuffer(builder, block_message_wipoffset)
 }
 
+// Build a Transaction flatbuffer to encode a Witnet's Block message
```
jrmoreau:
```suggestion
// Build a Transaction flatbuffer to encode a Witnet Block message
```

data_structures/src/serializers/decoders.rs
```diff
@@ -130,7 +130,17 @@ impl TryFrom<Vec<u8>> for Message {
...
+            protocol::Command::Transaction => message
+                .command_as_transaction()
+                .and_then(|transaction| {
+                    let transaction = create_transaction(transaction);
+                    Some(Message {
+                        kind: Command::Transaction(transaction),
+                        magic,
+                    })
+                })
+                .ok_or(""),
```
anler: Would be nice to use an appropriate error message and not just `""`

PR #323, opened 2019-01-10
feat(core): Update utxo set and validate blocks in order

### LINKED ISSUE #319: Ensure UTXO set is being updated properly
Ensure that functions to update UTXO set are called properly when a block is added to the chain

lrubiorod:
Added blocks_to_validate state in ChainManager to keep blocks that we can't validate because we don't have their previous hash.
Added a recursive call to process_block to validate blocks in blocks_to_validate hashmap 
Added a function to update utxo_set with block candidate and with older blocks
Added tests

Close #319

### DIFF
--- a/core/src/actors/chain_manager/mod.rs
+++ b/core/src/actors/chain_manager/mod.rs
@@ -98,6 +98,8 @@ pub struct ChainManager {
     block_chain: BTreeMap<Epoch, HashSet<Hash>>,
     /// Map that stores blocks by their hash
     blocks: HashMap<Hash, Block>,
+    /// Map that stores blocks without validation by their hash
+    blocks_to_validate: HashMap<Hash, Block>,
     /// Current Epoch
     current_epoch: Option<Epoch>,
     /// Transactions Pool (_mempool_)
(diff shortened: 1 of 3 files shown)


core/src/actors/chain_manager/mod.rs
```diff
@@ -267,6 +299,29 @@ impl ChainManager {
         }
     }
 
+    fn keep_block_without_validation(&mut self, block: Block) -> Result<Hash, ChainManagerError> {
+        // Calculate the hash of the block
+        let hash: Hash = block.hash();
+
+        // Check if we already have a block with that hash
+        if let Some(_block) = self.blocks_to_validate.get(&hash) {
+            Err(ChainManagerError::BlockAlreadyExists)
```
anler: I think is unnecessary to err if the block already exist, the method could silently ignore this fact IMHO

core/src/actors/chain_manager/mod.rs
```diff
@@ -364,11 +419,16 @@ impl ChainManager {
         Ok(missing_inv_entries)
     }
 
+    fn get_block_to_validate(&mut self, hash: &Hash) -> Option<Block> {
```
anler: I would call this function `remove_block_to_validate` instead, `get_` kind of implies it won't modify anything which is in contradiction with the `&mut self` demanded by the method

core/src/actors/chain_manager/mod.rs
```diff
@@ -146,6 +148,36 @@ impl ChainManager {
...
+                    transaction_id: txn_hash,
+                    output_index: index as u32,
+                };
+
+                utxo_set.insert(output_pointer, output.clone());
+            }
+        }
+
+        utxo_set
+    }
```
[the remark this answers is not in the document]
lrubiorod: The idea of this function is obtain a copy of the utxo_set with this blocks modifications. Because in a future we will handle different utxo_set and will only persist one of them.

PR #332, opened 2019-01-14
feat(data_structures): add Transaction::new factory method

anler:
The Transaction::new factory method takes inputs, outputs and a signing key and returns a Transaction object containing those inputs and outputs and with the appropriate signatures. Since keys and signatures needs to be implemented first in witnet_crypto crate, the method returns dummy signatures.

### DIFF
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -238,6 +238,31 @@ pub struct Transaction {
 }
 
 impl Transaction {
+    /// Creates a new transaction from inputs and outputs.
+    // TODO: Transaction::new is missing the signatures which depend
+    // on a sign key that needs to be implemented first in
+    // witnet_crypto.
+    pub fn new(version: u32, inputs: Vec<Input>, outputs: Vec<Output>, _sign_key: ()) -> Self {
+        let signatures = inputs
+            .iter()
+            .map(|_input| KeyedSignature {
+                signature: Signature::Secp256k1(Secp256k1Signature {
+                    r: [0; 32],
+                    s: [0; 32],
+                    v: 0,
+                }),
+                public_key: [0; 32],
+            })
+            .collect();
+
+        Transaction {
+            version,
+            inputs,
+            outputs,
+            signatures,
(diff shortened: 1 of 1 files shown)


data_structures/src/chain.rs
```diff
@@ -8,6 +8,9 @@ use std::convert::AsRef;
 use std::fmt;
 use witnet_crypto::hash::{calculate_sha256, Sha256};
 
+/// Current version used when creating [`Transaction`](Transaction)s
+const TRANSACTION_VERSION: u32 = 0;
```
aesedepece: I would add here:
```rust
// TODO: move TRANSACTION_VERSION into a consensus constant
```

data_structures/src/chain.rs
```diff
@@ -238,6 +241,32 @@ pub struct Transaction {
...
+        for _ in inputs.iter() {
+            signatures.push(KeyedSignature {
+                signature: Signature::Secp256k1(Secp256k1Signature {
+                    r: [0; 32],
+                    s: [0; 32],
+                    v: 0,
+                }),
+                public_key: [0; 32],
+            })
+        }
```
aesedepece: Why not functional? :stuck_out_tongue_closed_eyes: 

```rust
let signatures = inputs.iter()
    .map(|_input| KeyedSignature {
        signature: Signature::Secp256k1(Secp256k1Signature {
            r: [0; 32],
            s: [0; 32],
            v: 0,
        }),
        public_key: [0; 32],
    })
    .collect();
```

data_structures/src/chain.rs
anler: Then I'm removing the constant and passing it as a parameter to the factory

data_structures/src/chain.rs
anler: I was worried that `.collect()` didn't create a vector of the right capacity and thus were a bit less performant, but I was wrong, [this is how it is implemented](https://doc.rust-lang.org/src/alloc/vec.rs.html#1782)

PR #345, opened 2019-01-15
RADON engine base functionality

aesedepece:
This implements:
- `String::parse_json() -> Mixed`
- `String::identity() -> String`
- `Mixed::identity() -> Mixed`


rad/src/script.rs
```diff
@@ -0,0 +1,144 @@
...
+use crate::operators::{operate, RadonOpCodes};
+use crate::types::RadonTypes;
+
+use log::error;
+use num_traits::FromPrimitive;
+use rmpv::{self, Value};
+use std::{error::Error, io::Cursor};
+use witnet_util::error::WitnetError;
+
+pub type RadonCall = (RadonOpCodes, Option<Vec<Value>>);
```
anler: I find `Option<Vec<...>>` a bit redundant since `Vec<...>` alone is able to express the same as `Option<...>`. Is there a reason that prevents us from using just `pub type RadonCall = (RadonOpCodes, Vec<Value>);`

rad/src/operators/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+pub trait Operable {
+    fn operate(self, call: &RadonCall) -> RadResult<RadonTypes>;
+}
+
+pub fn operate(input: RadonTypes, call: &RadonCall) -> RadResult<RadonTypes> {
+    match input {
+        RadonTypes::String(radon_string) => radon_string.operate(call),
+        RadonTypes::Mixed(radon_mixed) => radon_mixed.operate(call),
+    }
+}
```
anler: If `RadonTypes` variants, all contain types that implement `Operable`, then `RadonTypes` may implement it to with this exact same implementation, if we do so then this function will just be:

```
pub fn operate(input: RadonTypes, call: &RadonCall) -> RadResult<RadonTypes> {
    input.operate(call)
}
```

rad/src/script.rs
aesedepece: The reason is that you can't tell an empty vector from a filled one without using a guard :rofl:

anler: But, isn't the same situation with `Option`? You need to test if it filled or not right?

aesedepece: But testing for length in a guard for every branch of the `match` is a pain in the ass.

aesedepece: I mean, the `match` is explicit about whether each operator accepts arguments or not.

PR #346, opened 2019-01-15
feat(data_structures): add method to validate Block's mint

anler:
Add several methods to Transaction and Block that ultimately add up to support Block's validate method that so far it only validates the mint transaction. This method probably should grow to validate everything related to the block that currently is spread out through the ChainManager actor handlers.

Things to highlight:
- Merged two different `impl` blocks that existed for `Transaction`
- Removed the `validate_coinbase` function inside `validations.rs` module
- Created a `validate` method inside `Block` that so far it only validates the mint transaction, but my recommendation is that everything related to block validation goes in there
- Implemented the `fee` method for `Transaction`
- Implemented an `inputs_sum` method for `Transaction`
- Implemented an `is_mint` method for `Transaction`
- Implemented an `output_value` method for `Transaction` that returns the value of an output given its index
- Renamed `sum_outputs` to `outputs_sum` to mirror `inputs_sum`
- Removed `calculate_outputs_sum` and `calculate_outputs_sum_of` methods in `Transaction`, they are very similar if not the same as `sum_outputs`
- Implemented `value` method for `Output` that returns the value of an output no matter what type of output it is
- Implemented `output_pointer` method for `Input` that returns the output pointer of an input no matter what type of input is
- Implemented `get` method for `TransactionsPool`

~WIP: I still need to add tests, but apart from that it can be reviewed if necessary~

### DIFF
--- a/core/src/actors/chain_manager/handlers.rs
+++ b/core/src/actors/chain_manager/handlers.rs
@@ -154,14 +154,12 @@ impl Handler<AddTransaction> for ChainManager {
                 // Check that all inputs point to unspent outputs
                 if self.find_unspent_outputs(&msg.transaction.inputs) {
                     // Validate transaction
-                    let inputs_sum = msg
-                        .transaction
-                        // DRO RULE 1. Multiple data request outputs can be included into a single transaction
-                        // as long as the inputs are greater than outputs rule still hold true. The difference
-                        // with VTOs is that the total output value for data request outputs also includes the
-                        // commit fee, reveal fee and tally fee.
-                        .calculate_outputs_sum_of(&outputs_from_inputs);
-                    let outputs_sum = msg.transaction.calculate_outputs_sum();
+                    // DRO RULE 1. Multiple data request outputs can be included into a single transaction
+                    // as long as the inputs are greater than outputs rule still hold true. The difference
+                    // with VTOs is that the total output value for data request outputs also includes the
+                    // commit fee, reveal fee and tally fee.
+                    let inputs_sum = outputs_from_inputs.iter().map(Output::value).sum();
+                    let outputs_sum = msg.transaction.outputs_sum();
 
                     if outputs_sum < inputs_sum {
                         let mut is_valid_input = true;
--- a/core/src/actors/chain_manager/handlers.rs
+++ b/core/src/actors/chain_manager/handlers.rs
@@ -154,8 +154,8 @@ impl Handler<AddTransaction> for ChainManager {
                 // Check that all inputs point to unspent outputs
                 if self.find_unspent_outputs(&msg.transaction.inputs) {
                     // Validate transaction
-                    // DRO RULE 1. Multiple data request outputs can be included into a single transaction
-                    // as long as the inputs are greater than outputs rule still hold true. The difference
+                    // DRO RULE 1. Multiple data request outputs can be included into a single transaction.
+                    // As long as the inputs are greater than the outputs, the rule still hold true. The difference
                     // with VTOs is that the total output value for data request outputs also includes the
                     // commit fee, reveal fee and tally fee.
                     let inputs_sum = outputs_from_inputs.iter().map(Output::value).sum();
--- a/core/src/actors/chain_manager/mod.rs
+++ b/core/src/actors/chain_manager/mod.rs
@@ -53,7 +53,7 @@ use witnet_data_structures::chain::{
     OutputPointer, TransactionsPool,
 };
 
-use crate::validations::{validate_merkle_tree, validate_mint, validate_transactions};
+use crate::validations::{validate_merkle_tree, validate_transactions};
 
 use self::data_request::DataRequestPool;
 use witnet_data_structures::chain::CheckpointBeacon;
@@ -426,8 +426,8 @@ impl ChainManager {
                             error!("Unexpected error");
                         }
                     }
-                } else if !validate_mint(&block) {
-                    warn!("Block mint not valid");
+                } else if block.validate(0, &self.transactions_pool).is_err() {
+                    warn!("Block's mint transaction is not valid");
                 } else {
                     if block_epoch < current_epoch {
                         // FIXME(#235): check proof of eligibility from the past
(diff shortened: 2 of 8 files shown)


core/src/actors/chain_manager/handlers.rs
```diff
@@ -154,14 +154,12 @@ impl Handler<AddTransaction> for ChainManager {
...
-                    let inputs_sum = msg
-                        .transaction
-                        // DRO RULE 1. Multiple data request outputs can be included into a single transaction
-                        // as long as the inputs are greater than outputs rule still hold true. The difference
-                        // with VTOs is that the total output value for data request outputs also includes the
-                        // commit fee, reveal fee and tally fee.
-                        .calculate_outputs_sum_of(&outputs_from_inputs);
-                    let outputs_sum = msg.transaction.calculate_outputs_sum();
+                    // DRO RULE 1. Multiple data request outputs can be included into a single transaction
+                    // as long as the inputs are greater than outputs rule still hold true. The difference
```
jrmoreau:
```suggestion
                    // As long as the inputs are greater than the outputs, the rule still hold true. The difference
```

core/src/actors/chain_manager/handlers.rs
```diff
@@ -154,14 +154,12 @@ impl Handler<AddTransaction> for ChainManager {
...
                     // Validate transaction
-                    let inputs_sum = msg
-                        .transaction
-                        // DRO RULE 1. Multiple data request outputs can be included into a single transaction
-                        // as long as the inputs are greater than outputs rule still hold true. The difference
-                        // with VTOs is that the total output value for data request outputs also includes the
-                        // commit fee, reveal fee and tally fee.
-                        .calculate_outputs_sum_of(&outputs_from_inputs);
-                    let outputs_sum = msg.transaction.calculate_outputs_sum();
+                    // DRO RULE 1. Multiple data request outputs can be included into a single transaction
```
jrmoreau:
```suggestion
                    // DRO RULE 1. Multiple data request outputs can be included into a single transaction.
```

data_structures/src/chain.rs
```diff
@@ -99,6 +99,44 @@ pub struct Block {
...
+    Transaction(TransactionError),
+    /// Indicates that the total value created by the mint transaction
+    /// of the block, and the output value of the rest of the
+    /// transactions, plus the block reward, don't add up
+    MismatchedMintValue,
+}
+
+impl Block {
+    /// Check if the block is valid.
+    pub fn validate(&self, block_reward: u64, pool: &TransactionsPool) -> Result<(), BlockError> {
```
aesedepece: Why `Result` and not `WitnetResult`?

data_structures/src/chain.rs
```diff
@@ -237,7 +275,57 @@ pub struct Transaction {
     pub signatures: Vec<KeyedSignature>,
 }
 
+/// The error type for operations on a [`Transaction`](Transaction)
+#[derive(Debug, PartialEq)]
+pub enum TransactionError {
+    /// Error indicating the transaction creates value
```
aesedepece: I'd rather use this style of error description:
```suggestion
    /// The transaction creates value
```

core/src/actors/chain_manager/handlers.rs
```diff
@@ -154,14 +154,12 @@ impl Handler<AddTransaction> for ChainManager {
...
-                        // as long as the inputs are greater than outputs rule still hold true. The difference
-                        // with VTOs is that the total output value for data request outputs also includes the
-                        // commit fee, reveal fee and tally fee.
-                        .calculate_outputs_sum_of(&outputs_from_inputs);
-                    let outputs_sum = msg.transaction.calculate_outputs_sum();
+                    // DRO RULE 1. Multiple data request outputs can be included into a single transaction.
+                    // As long as the inputs are greater than the outputs, the rule still hold true. The difference
+                    // with VTOs is that the total output value for data request outputs also includes the
+                    // commit fee, reveal fee and tally fee.
+                    let inputs_sum = outputs_from_inputs.iter().map(Output::value).sum();
```
aesedepece: This is beautiful :heart_eyes_cat:

core/src/actors/chain_manager/mod.rs
```diff
@@ -341,7 +341,7 @@ impl ChainManager {
         for transaction in self.transactions_pool.iter() {
             // Currently, 1 weight unit is equivalent to 1 byte
             let transaction_weight = transaction.size();
-            let transaction_fee = transaction.fee();
+            let transaction_fee = transaction.fee(&self.transactions_pool).unwrap();
```
aesedepece: I'm concerned about this `unwrap`.

anler: The issue is that I changed the `fee` function signature, before it returned `u64`, now it returns `Result<u64, ...>`. I'm using unwrap to avoid changing this function signature too, because that would lead to a bigger refactor and maybe is better to address it later.

data_structures/src/chain.rs
anler: Meh, I forgot about it, I'll change it to `WtinetResult` although I kind of find it more cumbersome to use :(

core/src/actors/chain_manager/mod.rs
aesedepece: Ok, then let's put a FIXME here :+1:

data_structures/src/chain.rs
anler: I changed the code in `chain.rs` to use [`failure`](https://rust-lang-nursery.github.io/failure/) directly, what do you think? IMO our `WitnetResult` is in the middle of the approach followed in vanilla Rust and the approach followed in `failure` crate, so we should use just one: the `failure` approach since it let us create functions that may fail with different errors without any additional work (see the example in `failure`'s documentation). If we do this then I could refactor the code we already have to remove the `WitnetResult`

core/src/actors/chain_manager/handlers.rs
anler: :blush:

PR #348, opened 2019-01-15
feat(core): add method for requesting a lost block from a random session

lrubiorod:
Created a `RequestBlock` msg and handler in `session` actor to send a inventory_request_message with the `InventoryEntry::Block(Hash)` of the lost block.
Created a function to send an `Anycast<RequestBlock>` to `SessionsManager` actor.
Added updating `chain_info` for older blocks.
Added docs

### DIFF
--- a/core/src/actors/chain_manager/mod.rs
+++ b/core/src/actors/chain_manager/mod.rs
@@ -38,8 +38,11 @@ use self::messages::InventoryEntriesResult;
 use crate::actors::{
     inventory_manager::{messages::AddItem, InventoryManager},
     reputation_manager::{messages::ValidatePoE, ReputationManager},
-    session::messages::{AnnounceItems, SendBlock},
-    sessions_manager::{messages::Broadcast, SessionsManager},
+    session::messages::{AnnounceItems, RequestBlock, SendBlock},
+    sessions_manager::{
+        messages::{Anycast, Broadcast},
+        SessionsManager,
+    },
     storage_keys::{BLOCK_CHAIN_KEY, CHAIN_KEY},
     storage_manager::{messages::Put, StorageManager},
 };
@@ -53,6 +56,7 @@ use witnet_data_structures::chain::{
 use crate::validations::{validate_merkle_tree, validate_mint, validate_transactions};
 
 use self::data_request::DataRequestPool;
+use witnet_data_structures::chain::CheckpointBeacon;
 use witnet_storage::error::StorageError;
 use witnet_util::error::WitnetError;
 
@@ -299,7 +303,7 @@ impl ChainManager {
     }
 
     fn broadcast_announce_items(&mut self, hash: Hash) {
-        // Get SessionsManager's address
+        // Get SessionsManager address
         let sessions_manager_addr = System::current().registry().get::<SessionsManager>();
 
         // Tell SessionsManager to announce the new block through every consolidated Session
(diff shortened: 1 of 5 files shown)


core/src/actors/chain_manager/mod.rs
```diff
@@ -315,6 +319,16 @@ impl ChainManager {
         });
     }
 
+    fn request_block(&mut self, block_hash: Hash) {
+        // Get SessionsManager's address
```
jrmoreau:
```suggestion
        // Get SessionsManager address
```

core/src/actors/session/handlers.rs
```diff
@@ -208,7 +208,20 @@ impl Handler<SendBlock> for Session {
             "Sending SendBlock message to peer at {:?}",
             self.remote_addr
         );
-        send_block_msg(self, &msg.block)
+        send_block_msg(self, msg.block)
+    }
+}
+
+/// Handler for SendBlock message (sent by other actors)
```
tmpolaczyk:
```suggestion
/// Handler for RequestBlock message (sent by other actors)
```

And add this message to the session documentation

core/src/actors/session/messages.rs
```diff
@@ -45,3 +45,16 @@ impl fmt::Display for SendBlock {
         write!(f, "SendBlock")
     }
 }
+
+/// Message to send blocks through the network
```
tmpolaczyk:
```suggestion
/// Message to request blocks through the network
```

core/src/actors/chain_manager/mod.rs
```diff
@@ -439,7 +453,8 @@ impl ChainManager {
                         );
                     }
                 } else if !self.blocks.contains_key(&hash_prev_block) {
-                    //TODO Request this lost block to our peers
+                    self.request_block(hash_prev_block);
```
aesedepece: I'd put a comment here explaining what's happening.

core/src/actors/session/handlers.rs
```diff
@@ -542,16 +555,32 @@ fn send_item_msg(session: &mut Session, ctx: &mut Context<Session>, hash: &Hash)
...
+    let proof = block.proof;
+    let txns = block.txns;
     // Build Block msg
     let block_msg = WitnetMessage::build_block(block_header, proof, txns);
     // Send Block msg
     session.send_message(block_msg);
 }
 
+fn request_block_msg(session: &mut Session, block_entry: InventoryEntry) {
+    let mut inv_entries: Vec<InventoryEntry> = Vec::new();
```
aesedepece: These two lines can be blended into a single one:
```suggestion
    // Initialize a new inventory entries vector with the given block entry as its sole member
    let mut inv_entries: Vec<InventoryEntry> = vec![block_entry];
```

tmpolaczyk:
I found a few bugs while trying a few nodes, but they were present before this PR:

* Since the blocks are now properly validated, we can't just start a node and expect it to work, because the previous block hash is 0000000..., which does not have a corresponding valid block. So we must somehow hard-code the genesis block as a valid block, at least for testing.

* Mining is broken, but it's a one line fix
<details>
  <summary>Show diff</summary>

```
diff --git a/core/src/actors/chain_manager/mining.rs b/core/src/actors/chain_manager/mining.rs
index 0de375b..33bd5ba 100644
--- a/core/src/actors/chain_manager/mining.rs
+++ b/core/src/actors/chain_manager/mining.rs
@@ -27,6 +27,11 @@ impl Handler<EpochNotification<MiningNotification>> for ChainManager {
     fn handle(&mut self, msg: EpochNotification<MiningNotification>, ctx: &mut Context<Self>) {
         debug!("Periodic epoch notification received {:?}", msg.checkpoint);
 
+        // Fix bug where this handler is executed before the other EpochNotification handler
+        // and it thinks the block it just mined is from the future
+        self.current_epoch = Some(msg.checkpoint);
+
+
         // Check eligibility
         // S(H(beacon))
         let mut beacon = match self.handle(GetHighestCheckpointBeacon, ctx) {
```
</details>

core/src/actors/chain_manager/mod.rs
```diff
@@ -476,6 +493,22 @@ impl ChainManager {
                                 // Update UTXO set with older block transactions
                                 self.unspent_outputs_pool = utxo_set;
                                 self.transactions_pool = txn_mempool;
+
+                                // Update chain_info
```
tmpolaczyk: Why is `chain_info` updated here? The block is still only a candidate, right?

PR #361, opened 2019-01-21
feat(chain_manager, sessions_manager): add network ready flag

mariocao:
This PR adds a `network_ready` flag in order to *pause* the `ChainManager` until the network is considered ready, i.e. there are enough connected peers. It should avoid potential forks if mining capability is enabled.


anler:
Merged with #368

PR #363, opened 2019-01-22
Fix mint and transactions

tmpolaczyk:
The mint transaction is now validated correctly.

Transaction validation incorrectly used the `TransactionPool` instead of the `UnspentOutputsPool`.

Block consolidation now correctly updates the `TransactionPool`.

In `AddTransaction`, the validations are disabled because they failed to recognize a valid commitment transaction. The only checks enabled are whether the input of the transaction is present in the unspent outputs pool.

### DIFF
--- a/core/src/actors/chain_manager/mining.rs
+++ b/core/src/actors/chain_manager/mining.rs
@@ -16,13 +16,13 @@ use crate::actors::{
 };
 
 use witnet_crypto::hash::calculate_sha256;
+use witnet_data_structures::chain::UnspentOutputsPool;
 use witnet_data_structures::chain::{
     Block, BlockHeader, CheckpointBeacon, Hash, Input, LeadershipProof, Output, PublicKeyHash,
     RevealInput, Secp256k1Signature, Signature, TallyOutput, Transaction, TransactionsPool,
     ValueTransferOutput,
 };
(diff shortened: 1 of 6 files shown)


data_structures/src/chain.rs
```diff
@@ -369,6 +370,16 @@ impl Transaction {
...
                 .get_output_value(index)
                 .ok_or_else(|| TransactionError::OutputNotFound(transaction_id, index))?;
+                */
+            let pointed_value = pool
+                .get(&input.output_pointer())
+                .ok_or_else(|| {
+                    TransactionError::OutputNotFound(
+                        input.output_pointer().transaction_id,
+                        input.output_pointer().output_index as usize,
+                    )
```
anler: Why don't we just define `TransactionError::OutputNotFound` as TransactionError::OutputNotFound(OutputPointer)`, then this snippet of code would look like:

```rust
TransactionError::OutputNotFound(input.output_pointer())
```

core/src/actors/chain_manager/mining.rs
```diff
@@ -187,15 +205,94 @@ impl ChainManager {
             }
         }
     }
+
+    // This funtion needs to be async
```
Tommytrg: Remove comment, this is async.

PR #367, opened 2019-01-22
fix(chain_manager): block_chain should only have consolidated blocks

### LINKED ISSUE #294: Load `epoch_to_block_hash` blocks index on start
We should load the `epoch => block_digest` map from local storage upon start. This depends on #293.

### LINKED ISSUE #295: Change purpose of `epoch_to_block_hash` map in ChainManager

tmpolaczyk:
This fixes the `getBlockChain` cli method, which should only list the consolidated blocks.

Fixes #295
Fixes #294

### DIFF
--- a/core/src/actors/chain_manager/handlers.rs
+++ b/core/src/actors/chain_manager/handlers.rs
@@ -63,18 +63,22 @@ impl Handler<EpochNotification<EveryEpochPayload>> for ChainManager {
             // Update chain_info
             match self.chain_state.chain_info.as_mut() {
                 Some(chain_info) => {
+                    let block_hash = candidate.block.hash();
+                    let block_epoch = candidate.block.block_header.beacon.checkpoint;
+
                     let beacon = CheckpointBeacon {
                         checkpoint: msg.checkpoint,
-                        hash_prev_block: candidate.block.hash(),
+                        hash_prev_block: block_hash,
                     };
 
                     chain_info.highest_block_checkpoint = beacon;
+                    self.chain_state.block_chain.insert(block_epoch, block_hash);
 
                     info!(
                         "{} Block {} consolidated for epoch #{}",
                         Purple.bold().paint("[Chain]"),
                         Purple.bold().paint(candidate.block.hash().to_string()),
-                        Purple.bold().paint(beacon.checkpoint.to_string()),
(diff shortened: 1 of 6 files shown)


core/src/actors/chain_manager/handlers.rs
```diff
@@ -63,12 +63,18 @@ impl Handler<EpochNotification<EveryEpochPayload>> for ChainManager {
...
+                    let block_hash = candidate.block.hash();
+
                     let beacon = CheckpointBeacon {
                         checkpoint: msg.checkpoint,
-                        hash_prev_block: candidate.block.hash(),
+                        hash_prev_block: block_hash,
                     };
 
                     chain_info.highest_block_checkpoint = beacon;
+                    // TODO: epoch or epoch - 1?
```
tmpolaczyk: @lrubiorod ?

core/src/actors/chain_manager/mod.rs
```diff
@@ -265,19 +236,11 @@ impl ChainManager {
...
                 let beacon = &block.block_header.beacon;
-                let hash_set = &mut self
-                    .block_chain
-                    .entry(beacon.checkpoint)
-                    .or_insert_with(HashSet::new);
-                hash_set.insert(hash);
-
                 info!(
-                    "{} Epoch #{} has {} block candidates now",
+                    "{} Epoch #{} has a new block candidate",
```
aesedepece: :sob:

mariocao:
Superseded by #373

PR #368, opened 2019-01-22
Fix: send inventory blocks in same order as requested

anler:
When requesting blocks with an `InventoryRequest` message, the node was sending the blocks in inverse order causing a bit of trouble. This fix makes the node send the blocks in the requested order.

This PR supersedes #361

### DIFF
--- a/core/src/actors/session/handlers.rs
+++ b/core/src/actors/session/handlers.rs
@@ -1,10 +1,11 @@
-use std::io::Error;
-
 use actix::io::WriteHandler;
 use actix::{
-    Actor, ActorContext, ActorFuture, AsyncContext, Context, ContextFutureSpawner, Handler,
-    StreamHandler, System, WrapFuture,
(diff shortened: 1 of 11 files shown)


core/src/actors/session/handlers.rs
```diff
@@ -97,17 +101,31 @@ impl StreamHandler<BytesMut, Error> for Session {
...
+                            .collect();
+
+                        future::join_all(item_requests)
+                            .into_actor(self)
+                            .map_err(|e, _, _| error!("Inventory request error: {}", e))
+                            .and_then(|item_responses, session, _| {
+                                for item_response in item_responses {
+                                    match item_response {
+                                        Ok(item) => send_inventory_item_msg(session, item),
+                                        Err(e) => error!("Inventory result is error: {}", e),
```
lrubiorod:
```suggestion
                                        Err(e) => warn!("Inventory result is error: {}", e),
```

PR #371, opened 2019-01-23
Do not announce candidate

anler:
When sending the inventory announcement, do not send block candidate of the current epoch. Right now we are sending it, problem is that it won't be persisted until the next epoch, so when the inventory is requested, that block won't be found in the storage.

This PR supersedes #368


tmpolaczyk:
This change is not needed after #367 is merged, because the `GetBlocksEpochRange` message will only return consolidated blocks.

PR #373, opened 2019-01-23
refactor(chain_manager): remove storing blocks in ChainManager state

mariocao:
This PR refactors the `ChainManager` state so that the blocks are not longer required to be on memory in order to validate newer blocks. 

This PR supersedes the PR #367

### DIFF
--- a/core/src/actors/chain_manager/actor.rs
+++ b/core/src/actors/chain_manager/actor.rs
@@ -12,9 +12,9 @@ use super::{
 };
 
 use crate::actors::{
-    chain_manager::{data_request::DataRequestPool, Blockchain},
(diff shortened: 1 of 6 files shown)


core/src/actors/chain_manager/actor.rs
```diff
@@ -164,42 +165,13 @@ impl ChainManager {
                             chain_info: Some(chain_info),
                             unspent_outputs_pool: UnspentOutputsPool::default(),
                             data_request_pool: ActiveDataRequestPool::default(),
+                            block_chain: BTreeMap::default(),
```
tmpolaczyk:
```suggestion
                            block_chain: BlockChain::default(),
```

PR #393, opened 2019-02-13
Replace Flatbuffers with Protocol Buffers

tmpolaczyk:


tmpolaczyk:
After this PR is merged, a new build dependency will be needed:

```
apt install protobuf-compiler
```

PR #421, opened 2019-02-21
feat(codec): update u16 types to u32 in tcp codec

Tommytrg:


tmpolaczyk:
Please update the documentation at `architecture/p2p/encoding`

PR #424, opened 2019-02-21
feat(core): disable communication between nodes with different consensus constants

### LINKED ISSUE #307: Nodes with different consensus constants should not peer
We need to add an additional validation to the handshaking process so that sessions for peers announcing different consensus constants than ours should be closed, and their addresses should be removed from `PeersManager` through the `RemovePeer` command.

This will also prevent our `testnet-1` nodes from peering with our local test nodes and future testnets or even mainnet.

lrubiorod:
Added a magic number (`u16`) for messages built with the 2 first `u8` of the Sha256 of the `ConsensusConstants`
Added protobuff conversion to `ConsensusConstants`
Added a validation of this number in each message received. Closing communication in case of mismatching magic numbers

Close #307

### DIFF
--- a/data_structures/tests/serializers.rs
+++ b/data_structures/tests/serializers.rs
@@ -270,15 +270,14 @@ fn message_version_to_bytes() {
             receiver_address: receiver_address,
             user_agent: "asdf".to_string(),
             last_epoch: 8,
-            genesis: 2,
             nonce: 1,
         }),
         magic: 1,
     };
     let expected_buf: Vec<u8> = [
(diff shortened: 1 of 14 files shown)


data_structures/tests/serializers.rs
```diff
@@ -305,11 +304,14 @@ fn message_version_from_bytes() {
...
             last_epoch: 8,
-            genesis: 2,
             nonce: 1,
         }),
         magic: 1,
     };
+
+    // TODO: Ensure protobuf decryption, because this buf
+    // still have the genesis field of the old `Version` message
+    // and this test actually pass
```
tmpolaczyk: This is expected behavior: protocol buffers are backwards compatible, so unknown extra fields are ignored.

lrubiorod: Thanks, it was that!!

PR #432, opened 2019-02-28
feat(wallet): add address derivation methods

Tommytrg:
This PR contains commits from #432.

WIP: 
 - ~~Replace secp256k1 lib with openssl~~
 - Add test

### DIFF
--- a/wallet/src/key.rs
+++ b/wallet/src/key.rs
@@ -12,18 +12,20 @@
 
 use failure::Fail;
 use hmac::{Hmac, Mac};
-use secp256k1;
+use secp256k1::{PublicKey, SecretKey};
 use sha2;
+
+const HARDENED_BIT: u32 = 1 << 31;
 
 /// Default HMAC key used when generating a Master Key with
 /// [generate_master](generate_master)
-pub static DEFAULT_HMAC_KEY: &str = "Witnet seed";
+pub static DEFAULT_HMAC_KEY: &[u8] = b"Bitcoin seed";
 
 /// secp256k1 Secret Key with Chain Code
-pub type ExtendedSK = ExtendedKey<secp256k1::SecretKey>;
+// pub type ExtendedSK = ExtendedKey<SecretKey>;
 
 /// secp256k1 Public Key with Chain Code
-pub type ExtendedPK = ExtendedKey<secp256k1::PublicKey>;
+// pub type ExtendedPK = ExtendedKey<secp256k1::PublicKey>;
 
 /// The error type for [generate_master](generate_master)
 #[derive(Debug, PartialEq, Fail)]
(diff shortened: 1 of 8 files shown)


wallet/src/key.rs
```diff
@@ -12,18 +12,20 @@
...
+const HARDENED_BIT: u32 = 1 << 31;
+
 /// Default HMAC key used when generating a Master Key with
 /// [generate_master](generate_master)
-pub static DEFAULT_HMAC_KEY: &str = "Witnet seed";
+pub static DEFAULT_HMAC_KEY: &[u8] = b"Bitcoin seed";
 
 /// secp256k1 Secret Key with Chain Code
-pub type ExtendedSK = ExtendedKey<secp256k1::SecretKey>;
+// pub type ExtendedSK = ExtendedKey<SecretKey>;
```
tmpolaczyk: Why is this commented out? If it's not used, just remove it.

wallet/src/key.rs
```diff
@@ -78,22 +80,130 @@ where
...
+pub enum KeyDerivationError {
+    /// Invalid hmac key length
+    #[fail(display = "The length of the hmac key is invalid")]
+    InvalidKeyLength,
+    /// Invalid seed length
+    #[fail(display = "The length of the seed is invalid, must be between 128/256 bits")]
+    InvalidSeedLength,
+
+    #[fail(display = "Error in secp256k1 crate")]
+    Secp256k1Error,
```
tmpolaczyk: This should be a wrapper of the secp256k1 error:
```suggestion
    Secp256k1Error(secp256k1::Error),
```

wallet/src/key.rs
```diff
@@ -78,22 +80,130 @@ where
...
     pub chain_code: [u8; 32],
 }
 
+/// A child number for a derived key
+#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+pub struct ChildNumber(u32);
+
+impl ChildNumber {
+    pub fn is_hardened(&self) -> bool {
+        self.0 & HARDENED_BIT == HARDENED_BIT
```
tmpolaczyk: Assuming that a ChildNumber cannot be normal and hardened at the same time:

```suggestion
        !self.is_normal()
```

wallet/src/key.rs
```diff
@@ -78,22 +80,130 @@ where
...
+            hmac512.input(&[0]);
+            hmac512.input(&self.secret_key[..]);
+        } else {
+            hmac512.input(
+                &PublicKey::from_secret_key(&secp256k1::Secp256k1::new(), &self.secret_key)
+                    .serialize(),
+            );
+        }
+        hmac512.input(&child.to_bytes());
+        let i = hmac512.result().code();
```
tmpolaczyk: This chunk of code is repeated in `derive` and `child`, so it should be extracted into a helper function.

tmpolaczyk:
These are the lints that should be added to `wallet/src/lib.rs`:

```rust
#![deny(rust_2018_idioms)]
#![deny(non_upper_case_globals)]
#![deny(non_camel_case_types)]
#![deny(non_snake_case)]
#![deny(unused_mut)]
#![deny(missing_docs)]
```

PR #434, opened 2019-03-01
refactor(config): Simplify Config Manager actor API

anler:
- Move Config Manager actor into `core/src/config_mngr.rs`
- Expose public functions: `get`, `load_from_file`, and `start`. They
  return futures for each task: getting the actual config, loading a
  new config from a file, and starting the actor.
- Refactor rest of actors to consume Config Manager's public API

Also, a fallback strategy for loading config from a file has been added.
The configuration file used by the node will behave as the following:

| os-config dir | cwd         | cli-config flag | configuration loaded | errors if not present |
|---------------|-------------|-----------------|----------------------|-----------------------|
| not present   | not present | not present     | default              | -                     |
| not present   | not present | C               | C                    | true                  |
| not present   | B           | not present     | B                    | false (uses default)             |
| A             | not present | not present     | A                    | false (uses default)                |
| A             | B           | C               | C                    | false (uses default)                |
| A             | B           | not present               | B                    | false (uses default)                |


core/src/actors/connections_manager/mod.rs
```diff
@@ -27,12 +27,54 @@ impl SystemService for ConnectionsManager {}
...
+                    "P2P server has been started at {:?}",
+                    &config.connections.server_addr
+                );
+
+                fut::ok(())
+            })
+            .map_err(|err, _, _| log::error!("P2P server failed to start: {}", err))
+            .spawn(ctx);
+
+        // ctx.spawn(server);
```
aesedepece: This comment can be removed altogether, right?

aesedepece:
@anler That behavior matrix makes sense to me :+1:

PR #435, opened 2019-03-02
Add RadonString::hash() operator

aesedepece:
- Adds a new `hash_functions` module to the RADON engine
- Adds support for `RadonString::hash`, which automatically serializes the digest as hexadecimal lowercase
- Every each of the functions that have been added has its own test
- The only hash function currently implemented is `SHA2_256` (RADON constant `0x0A`)
- In the future we may add support for `RadonArray::hash` or even `RadonBinary::hash` if we decide to support a native bytearray-alike type, which now has started to make more sense to me.

### DIFF
--- a/rad/Cargo.toml
+++ b/rad/Cargo.toml
@@ -8,6 +8,7 @@ description = "RAD component"
 
 [dependencies]
 failure = "0.1.5"
+hex = "0.3.2"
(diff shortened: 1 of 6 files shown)


aesedepece:
Merged along #481! :dancer:

PR #460, opened 2019-03-04
feat(core): Create State Machine in ChainManager

### LINKED ISSUE #437: Create ChainManager state machine
Refactor ChainManager state machine:

- [ ] Implement Match clause
- [ ] Move existing business logic to default

lrubiorod:
Close #437

### DIFF
--- a/node/src/actors/chain_manager/mod.rs
+++ b/node/src/actors/chain_manager/mod.rs
@@ -83,6 +83,23 @@ impl From<WitnetError<StorageError>> for ChainManagerError {
     }
 }
 
+/// State Machine
+#[derive(Debug)]
+pub enum StateMachine {
+    /// First state, ChainManager is waiting to consensus between its peers
+    WaitingConsensus,
+    /// Second state, ChainManager synchronization process
+    Synchronizing,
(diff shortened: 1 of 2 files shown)


node/src/actors/chain_manager/mod.rs
```diff
@@ -83,6 +83,25 @@ impl From<WitnetError<StorageError>> for ChainManagerError {
...
+    Synchronizing,
+    /// Third state, ChainManager is ready to mine and consolidated blocks
+    Synced,
+    /// Default case,
+    Default,
+}
+
+impl Default for StateMachine {
+    fn default() -> Self {
+        StateMachine::Default
```
tmpolaczyk: I think it would be easier to use `WaitingConsensus` as the default state, and leave the `match`es as

``` rust
        match self.sm_state {
            StateMachine::WaitingConsensus => {
            }
            StateMachine::Synchronizing => {
                unimplemented!();
            }
            StateMachine::Synced => {
                unimplemented!();
            }
        }
        // Actual code goes here:
```

Then when we add the actual functionality we can replace the `unimplemented` with `log::debug` and early return.

lrubiorod: Good idea!!

PR #463, opened 2019-03-05
WIP PeersBeacons

### LINKED ISSUE #443: Implement establish consensus from connected peers
Add a message `PeersBeacons` to `ChainManager`, which will be used by `SessionManager` to send the list of `(peer, beacon)` used to establish consensus when our node is bootstrapping the chain.

### LINKED ISSUE #447: ChainManager: add new field `targetBeacon`
Add field `targetBeacon` which will be used for synchronization: it will contain the epoch and hash of the target block. Once that target block is reached, the chain manager will go into state `Synchronized` or `WaitingConsensus`, depending on if the current epoch matches the target epoch.

tmpolaczyk:
Close #443 
Close #447

### DIFF
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -82,7 +82,9 @@ pub struct ConsensusConstants {
 }
 
 /// Checkpoint beacon structure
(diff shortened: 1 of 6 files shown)


tmpolaczyk:
This PR is being integrated with other changes in the cm-refactor branch

PR #465, opened 2019-03-06
refactor(core): change AddNewBlock by AddBlocks and AddCandidates in ChainManager

lrubiorod:


node/src/actors/json_rpc/json_rpc_methods.rs
```diff
@@ -59,7 +59,9 @@ pub fn inventory(inv_elem: InventoryItem) -> JsonRpcResult {
             let chain_manager_addr = System::current().registry().get::<ChainManager>();
             // If this function was called asynchronously, it could wait for the result
             // But it's not so we just assume success
-            chain_manager_addr.do_send(AddNewBlock { block });
+            chain_manager_addr.do_send(AddBlocks {
```
tmpolaczyk:
```suggestion
            chain_manager_addr.do_send(AddCandidates {
```

tmpolaczyk: Because we can only add new blocks, so candidates for the current epoch.

PR #466, opened 2019-03-07
feat(core): change AddNewBlock to AddCandidates and AddBlocks in Session

lrubiorod:
Removed `DiscardExistingInventoryEntries`
Added `EpochNotification` to Session
Changed AddNewBlock to AddCandidates and AddBlocks in Session

### DIFF
--- a/node/src/actors/session/handlers.rs
+++ b/node/src/actors/session/handlers.rs
@@ -11,7 +11,7 @@ use log::{debug, error, info, trace, warn};
 
 use witnet_data_structures::{
     builders::from_address,
-    chain::{Block, CheckpointBeacon, InventoryEntry, InventoryItem, Transaction},
+    chain::{Block, CheckpointBeacon, Hashable, InventoryEntry, InventoryItem, Transaction},
     proto::ProtobufConvert,
     types::{
         Address, Command, InventoryAnnouncement, InventoryRequest, LastBeacon,
@@ -196,12 +196,12 @@ impl StreamHandler<BytesMut, Error> for Session {
                         session_last_beacon_outbound(self, ctx, highest_block_checkpoint);
                     }
 
-                    ////////////////////
-                    // INVENTORY      //
-                    ////////////////////
+                    ////////////////////////////
+                    // INVENTORY ANNOUNCEMENT //
+                    ////////////////////////////
                     // Handle InventoryAnnouncement message
                     (_, SessionStatus::Consolidated, Command::InventoryAnnouncement(inv)) => {
-                        inventory_process_inv(self, ctx, &inv);
+                        inventory_process_inv(self, &inv);
                     }
                     /////////////////////
                     // NOT SUPPORTED   //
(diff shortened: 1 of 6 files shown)


node/src/actors/session/handlers.rs
```diff
@@ -386,14 +412,37 @@ fn peer_discovery_peers(peers: &[Address]) {
...
+        let mut blocks_vector = vec![];
+        for hash in session.requested_block_hashes.clone() {
+            if let Some(block) = session.requested_blocks.remove(&hash) {
+                blocks_vector.push(block);
+            } else {
+                break;
+            }
+        }
+
+        // Send a message to the ChainManager to try to add a new block
```
tmpolaczyk: If I remember correctly, the blocks must be sorted by epoch before sending them to the `ChainManager`

node/src/actors/session/handlers.rs
```diff
@@ -386,14 +412,37 @@ fn peer_discovery_peers(peers: &[Address]) {
...
+        chain_manager_addr.do_send(AddBlocks {
+            blocks: blocks_vector,
+        });
+    } else {
+        // Add block to requested_blocks
+        session.requested_blocks.insert(block_hash, block);
```
tmpolaczyk: Here we are storing all the blocks, regardless if we requested them or not. This could be a good way to cache blocks, but since this map is unique to each session, we probably won't need the unrequested blocks.

node/src/actors/session/handlers.rs
```diff
@@ -386,14 +412,37 @@ fn peer_discovery_peers(peers: &[Address]) {
...
+
+    if Some(block_epoch) == session.current_epoch {
+        // Send a message to the ChainManager to try to add a new candidate
+        chain_manager_addr.do_send(AddCandidates {
+            blocks: vec![block],
+        });
+    // TODO: Add timeout
+    } else if session.requested_blocks.len() == session.requested_block_hashes.len() {
+        let mut blocks_vector = vec![];
+        for hash in session.requested_block_hashes.clone() {
```
tmpolaczyk: Add comment:
Iterate over requested block hashes ordered by epoch

node/src/actors/session/handlers.rs
```diff
@@ -386,14 +412,37 @@ fn peer_discovery_peers(peers: &[Address]) {
...
+            if let Some(block) = session.requested_blocks.remove(&hash) {
+                blocks_vector.push(block);
+            } else {
+                break;
```
tmpolaczyk: Add comment:
As soon as there is a missing block, stop processing the other blocks and send a message to the ChainManager

node/src/actors/session/handlers.rs
```diff
@@ -386,14 +412,37 @@ fn peer_discovery_peers(peers: &[Address]) {
...
+    if Some(block_epoch) == session.current_epoch {
+        // Send a message to the ChainManager to try to add a new candidate
+        chain_manager_addr.do_send(AddCandidates {
+            blocks: vec![block],
+        });
+    // TODO: Add timeout
+    } else if session.requested_blocks.len() == session.requested_block_hashes.len() {
+        let mut blocks_vector = vec![];
+        for hash in session.requested_block_hashes.clone() {
+            if let Some(block) = session.requested_blocks.remove(&hash) {
```
tmpolaczyk: We should also remove the hash from `session.requested_block_hashes`?

node/src/actors/session/handlers.rs
tmpolaczyk: Oh nevermind, they are sorted in `requested_block_hashes`

tmpolaczyk: No, wait, we can't assume they are sorted.

node/src/actors/session/handlers.rs
lrubiorod: Now I clear the vec and the hash

PR #467, opened 2019-03-07
Store beacons in sessions manager, and send them to chain manager

tmpolaczyk:


node/src/actors/sessions_manager/handlers.rs
```diff
@@ -210,3 +212,39 @@ where
...
+    type Result = ();
+
+    fn handle(&mut self, _msg: EpochNotification<()>, ctx: &mut Context<Self>) {
+        let all_ready_before = self.beacons.iter().all(|(_k, v)| v.is_some());
+        if !all_ready_before {
+            // Some peers sent us beacons, but not all of them
+            self.send_peers_beacons(ctx);
+        }
+        // New epoch, new beacons
+        self.clear_beacons();
```
tmpolaczyk: There is a race condition here: we must receive the beacons after this handler has been executed. We could avoid this by only clearing beacons from past epochs, and accepting beacons for future epochs in this line:

             // We already got all the beacons for this epoch, do nothing

node/src/actors/session/handlers.rs
```diff
@@ -571,6 +581,7 @@ fn session_last_beacon_inbound(
         ..
     }: CheckpointBeacon,
 ) {
+    // TODO: LastBeacon on inbound peers?
```
lrubiorod: This is the function that handle the LastBeacon messages **from** inbound peers

lrubiorod:
```rust
} else if chain_beacon.checkpoint == received_checkpoint {
                        info!("Our chain is on par with our peer's",);
                        // Create a last_beacon message
                        let last_beacon_msg = WitnetMessage::build_last_beacon(act.magic_number, chain_beacon);
                        // Write a last_beacon message in session
                        act.send_message(last_beacon_msg);
}
```
In this part it is already not necessary to respond with another last beacon, because synchronizing process is done through consensus

node/src/actors/session/handlers.rs
```diff
@@ -636,51 +647,17 @@ fn session_last_beacon_inbound(
...
+        beacon,
+    })
+}
 
-                    actix::fut::err(())
-                }
-            }
-        })
-        .wait(ctx);
+fn send_last_beacon(session: &mut Session, beacon: CheckpointBeacon) {
```
lrubiorod: I will add a comment to indicate that it would be cool to choose send to inbounds or outbounds or both

node/src/actors/sessions_manager/handlers.rs
lrubiorod: Please write a TODO

node/src/actors/chain_manager/handlers.rs
```diff
@@ -486,7 +476,21 @@ impl Handler<PeersBeacons> for ChainManager {
...
+                        .chain_info
+                        .as_ref()
+                        .unwrap()
+                        .highest_block_checkpoint;
+
+                    // Maybe we are already synchronized?
+                    self.sm_state = if our_beacon == beacon {
+                        StateMachine::Synced
+                    } else {
+                        StateMachine::Synchronizing
```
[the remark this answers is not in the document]
lrubiorod: Done!!

PR #471, opened 2019-03-08
[WIP] Fix state transition from Synchronizing to Synced

tmpolaczyk:


tmpolaczyk:
A better alternative was found

PR #474, opened 2019-03-08
refactor(core): remove useless attributes, methods and messages

lrubiorod:


src/json_rpc_client.rs
```diff
@@ -51,24 +49,6 @@ pub(crate) fn run(last_config: Option<PathBuf>, cmd: CliCommand) -> Result<(), f
...
-        } => {
-            let config = config.or(last_config);
-            let mut stream = start_client(config)?;
-            let output_pointer = OutputPointer::from_str(&output_index)?;
-            let request_payload = serde_json::to_string(&output_pointer)?;
-            let request = format!(
-                r#"{{"jsonrpc": "2.0","method": "getOutput", "params": [{}], "id": "1"}}"#,
-                request_payload,
-            );
-            let response = send_request(&mut stream, &request)?;
```
tmpolaczyk: Instead of removing the getOutput command, we can just leave an "unimplemented" warning

             //let response = send_request(&mut stream, &request)?;
             let response = "Unimplemented";

PR #481, opened 2019-03-12
Make witnet-rad module use Failure properly

aesedepece:
This kind of biggie refactor tries to exemplify how we are supposed to use the Failure framework properly.

The benefits of this approach are manyfold:
- Error messages are not created in place but defined in the `RadonError` enum, which prevents inconsistent error reporting and keeps the code cleaner.
- Every usage of `RadonResult<T>` is now `Result<T, RadonError>`, where `Result` is Rust's primitive `Result`. 
- The API for each module can now be uniform and predictable.

### DIFF
--- a/rad/src/error.rs
+++ b/rad/src/error.rs
@@ -57,6 +57,8 @@ pub enum RadErrorKind {
     ScriptNotArray,
     /// The given operator code is unknown
     UnknownOperator,
+    /// The given hash function is not implemented
+    UnsupportedHashFunction,
     /// The given operator is not implemented for the input type
     UnsupportedOperator,
     /// The given reducer is not implemented for the type of the input Array
--- a/rad/src/error.rs
+++ b/rad/src/error.rs
@@ -8,16 +8,10 @@ use rmpv::{Integer, Value};
 pub enum RadError {
     /// Failed to decode a type from other
     #[fail(display = "Failed to decode {} from {}", to, from)]
-    Decode {
-        from: &'static str,
-        to: &'static str,
-    },
+    Decode { from: String, to: String },
     /// Failed to encode a type into other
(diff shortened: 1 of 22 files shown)


rad/src/error.rs
```diff
@@ -1,87 +1,100 @@
...
 
 /// RAD errors.
-#[derive(Debug, PartialEq)]
-pub enum RadErrorKind {
-    /// Failed to encode or decode a RADON type into / from bytes
-    EncodeDecode,
+#[derive(Debug, PartialEq, Fail)]
+pub enum RadError {
+    /// Failed to decode a type from other
+    #[fail(display = "Failed to decode {} from {}", from, to)]
```
tmpolaczyk:
```suggestion
    #[fail(display = "Failed to decode {} from {}", to, from)]
```

ParseInt: Decode from string to int

rad/src/types/mod.rs
```diff
@@ -37,24 +37,23 @@ pub enum RadonTypes {
...
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
-            RadonTypes::Float(float) => write!(f, "RadonFloat: {}", float.value()),
-            _ => write!(f, "RadonType: {:?}", self),
+            RadonTypes::Array(_array) => write!(f, "RadonArray"),
+            RadonTypes::Float(_float) => write!(f, "RadonFloat"),
+            RadonTypes::Map(_map) => write!(f, "RadonMap"),
+            RadonTypes::Mixed(_mixed) => write!(f, "RadonMixed"),
+            RadonTypes::String(_string) => write!(f, "RadonString"),
+            //_ => write!(f, "RadonTypes::{:?}", self),
```
tmpolaczyk: The display implementation should show some information about the inner value, not only the type.

PR #476, opened 2019-03-11
refactor(core): ChainManager refactor

lrubiorod:
- Created `SendLastBeacon` message and handler in `Session`
- Sent a `SendLastBeacon` message in state Synced in `EpochNotification` handler
- Deprecated `DiscardExistingEntries`
- Deprecated `GetOutput`
- Deprecated `PeerLastBeacon`
- Deprecated `SendNetworkReady`
- Created `AddBlocks` and `AddCandidates` messages and handlers
- Added requested_block_hashes, requested_blocks, blocks_timeout and blocks_timestamp to `Session`
- Added candidates hashmap in `ChainManager`
- Refactored `AddNewBlock` in `AddBlocks` and `AddCandidates`: 
  - In `Session`: 
          blocks in current_epoch -> AddCandidates
          blocks in old blocks -> Wait to obtain all requested blocks and later create AddBlocks
  - In `ChainManager`:
         AddBlocks: synchronize blocks in Synchronizing state
         AddCandidates: add blocks to `candidates` hashmap

- Renamed `Candidate` struct to `BlockInChain`
- Refactored updating candidate process in `ChainManager` `EpochNotification` handler
- Added blocks_timeout in config
- Read blocks_timeout from `ConfigManager` in `Session`
- Added `EpochNotification` in `Session` and SessionsManager`
- Used blocks_timeout in Synchronizing state of `AddBlocks` handler
- Created a `PeersBeacons` and `PeerBeacon` message and hanlder
- Implemented a establish consensus method in `ChainManager`
- Implemented a send_peers_beacon function to send a `PeersBeacons` message and remove peers out of consensus in `SessionsManager` and mark as Safu peers the rest of the peers in consensus
- Handled `LastBeacon` message in `Session` to send a `InventoryAnnouncement` to the other node or a `PeerBeacon` to `SessionsManager`
- Added `target_beacon` in `ChainManager`
- Removed unused fields in `ChainManager`
- Refactored validate_block  and validate_transactions methods to return a `BlockInChain` struct
- Sent a empty `AddBlocks` when timeout and also in `Session` stop

It turns out that calling unregister_session is not enough: the session actor needs to be explicitly stopped.
Added only_inbound flag in Broadcast messages: to allow broadcasting the LastBeacon only to inbound peers.

Close #440 
Close #441
Close #442 
Close #443 
Close #444 
Close #445 
Close #446 
Close #447 
Close #448 
Close #450 
Close #451 
Close #453 
Close #454 
Close #459

PR #482, opened 2019-03-12
Fix unregister of sessions in cm-refactor 

tmpolaczyk:
It turns out that calling `unregister_session` is not enough: the session actor needs to be explicitly stopped.

Added `only_inbound` flag in Broadcast messages: to allow broadcasting the `LastBeacon` only to inbound peers.


tmpolaczyk:
This PR was integrated into #476

PR #487, opened 2019-03-14
feat(data-structures): use failure Error in validations

lrubiorod:
Modified Result<_, ()> to Result<_, failure::Error> in validations.rs

### DIFF
--- a/data_structures/src/validations.rs
+++ b/data_structures/src/validations.rs
@@ -10,8 +10,14 @@ use super::{
     data_request::DataRequestPool,
 };
 
-use log::{debug, warn};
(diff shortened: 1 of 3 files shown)


tmpolaczyk:
Since we are removing log traces from here, we should add logs when calling these functions, because errors can be easily ignored when using `if let Ok(_) = ...`

tmpolaczyk:
In `Handler<AddCandidates>` there is an `is_ok()` which hides the error:

```rust
if !self.candidates.contains_key(&hash_block)
                && validate_candidate(&block, current_epoch).is_ok()
            {
```

Issue #510, opened 2019-03-18
Transaction hash should not depend on the signatures field


tmpolaczyk:
A.k.a implement segwit.

A naive solution would be to set the signatures field to an empty vector before hashing:

```rust
impl Hashable for Transaction {
    fn hash(&self) -> Hash {
        let th = Transaction { signatures: vec![], ..self.clone() };
        calculate_sha256(&th.to_pb_bytes().unwrap()).into()
    }
}
```

This is a breaking change.

anler:
Can we think of a way that avoids cloning the entire transaction? It is a bit inefficient imho

tmpolaczyk:
We could follow the same strategy as with blocks: move all the fields that need to be hashed to a `TransactionHeader` and hash that, but this would be a pretty big refactor.

anler:
I guess another approach is to use [interior mutability](https://doc.rust-lang.org/reference/interior-mutability.html) so we can clear the signatures when generating the hash and restore them afterwards, for example:

![image](https://user-images.githubusercontent.com/83847/54738982-b4258680-4bb6-11e9-9ee3-1419051e7a3e.png)

This could be implemented easily but not in our case because virtually all of our abstractions in `witnet_data_structures` crate leak. Probably we should open an issue to fix this and implement [abstract data types](https://en.wikipedia.org/wiki/Abstract_data_type?oldformat=true) instead?

tmpolaczyk:
We could also use interior mutability to cache the hash as a `cached_hash: Cell<Option<Hash>>`, because right now every call to `hash()` will have to recalculate the hash, even if the transaction has not changed. We could make all the transaction fields private and provide setters (`fn inputs_mut(&mut self) -> &mut Vec<u8>`), and set the cached hash to `None` every time a setter is called.

aesedepece:
This may save us one copy:
```rust
impl Hashable for Transaction {
  fn hash(&self) -> u8 {
    let pb = self.to_pb();
    pb.set_signatures(vec![]);
    calculate_sha256(pb.to_bytes())
  }
}
```
However, what @tmpolaczyk suggested sounds even better to me :+1:

PR #527, opened 2019-03-19
feat(signature_mngr): add signature manager

anler:
The Signature Manager is in charge of storing a secret key and sign data.

The API consists of three functions:

- `start`: Starts the manager with no key
- `set_key`: Sets the key the manager is going to use to sign data
- `sign`: Signs the given data

Example:

```rust
use witnet_node::signature_mngr;

signature_mngr::start();

let key = SK::from_slice(&[1 as u8; 32]).unwrap();

signature_mngr.set_key(key).and_then(|_| {
  let data = [0xab; 32];
  signature_mngr::sign(&data)
})
//...
```

### DIFF
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -180,10 +180,9 @@ pub struct BlockInChain {
     pub data_request_pool: DataRequestPool,
 }
 
-/// Any reference to a Hashable type is also Hashable
-impl<'a, T: Hashable> Hashable for &'a T {
+impl<T: AsRef<[u8]>> Hashable for T {
     fn hash(&self) -> Hash {
-        (*self).hash()
+        calculate_sha256(self.as_ref()).into()
     }
 }
 
--- a/node/src/lib.rs
+++ b/node/src/lib.rs
@@ -13,5 +13,7 @@ pub mod actors;
 /// Config Manager Actor API
 pub mod config_mngr;
 
+pub mod signature_mngr;
+
 /// Utilities for actor behaviour
 pub mod utils;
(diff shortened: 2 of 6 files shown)


data_structures/src/chain.rs
```diff
@@ -179,10 +179,9 @@ pub struct BlockInChain {
     pub data_request_pool: DataRequestPool,
 }
 
-/// Any reference to a Hashable type is also Hashable
-impl<'a, T: Hashable> Hashable for &'a T {
+impl<T: AsRef<[u8]>> Hashable for T {
```
aesedepece: This may break the former property "_any reference to a Hashable type is also Hashable_".

anler: Yes, I thought about it, but the case for "_any ref to a Hashable is also Hashable_" is needed only if we define a generic function like:
```rust
pub fn do_something<T: Hashable>(x: T) {
  // ...
}
```
and we want to allow `T` to be either a Hashable or a ref to a Hashable. After some thought I think this _feature_ won't be really needed and the more natural approach if we don't care about the ownership is to just ask for a ref to a Hashable:
```rust
pub fn do_something<T: Hashable>(x: &T) {
  // ...
}
```
in fact this is the case for the `sign` function in the signature manager:
```rust
pub fn sign<T>(data: &T) -> impl Future
where
    T: Hashable,
{
  // ...
}
```

Unfortunately I cannot leave both impls because they clash.

data_structures/src/chain.rs
```diff
@@ -179,10 +179,9 @@ pub struct BlockInChain {
...
     fn hash(&self) -> Hash {
-        (*self).hash()
+        calculate_sha256(self.as_ref()).into()
```
aesedepece: Could this hinder any attempt to implement #510? That's super important!

anler: hmm I don't think so :thinking: `Transaction` does not (nor can) implement `AsRef<[u8]>` so it will always use its own `Hashable` impl. That's what you mean?

PR #540, opened 2019-03-26
Segregate signatures from transaction bodies

### LINKED ISSUE #510: Transaction hash should not depend on the signatures field

aesedepece:
Also implements hash caching for `Transaction` / `TransactionBody` (hashes only get computed once).

Fix #510


node/src/actors/messages.rs
```diff
@@ -13,7 +13,7 @@ use std::{
 use witnet_data_structures::{
     chain::{
         Block, CheckpointBeacon, Epoch, Hash, InventoryEntry, InventoryItem, RADConsensus,
-        RADRequest, Transaction,
```
lrubiorod: You remove here and import in line 35

aesedepece: F*ck CLion.

node/src/actors/session/handlers.rs
```diff
@@ -11,7 +11,7 @@ use log::{debug, error, info, trace, warn};
 
 use witnet_data_structures::{
     builders::from_address,
-    chain::{Block, CheckpointBeacon, Hashable, InventoryEntry, InventoryItem, Transaction},
```
lrubiorod: Here again xD

validations/src/validations.rs
```diff
@@ -4,21 +4,22 @@ use std::collections::HashMap;
 use witnet_data_structures::{
     chain::{
         Block, BlockError, BlockInChain, CheckpointBeacon, DataRequestOutput, Epoch, Hash,
-        Hashable, Input, Output, OutputPointer, RADRequest, Transaction, TransactionError,
```
lrubiorod: One more

PR #541, opened 2019-03-26
[Storage] Transparent Data Encryption support

### LINKED ISSUE #499: Implement Encrypted RocksDB backend for witnet_storage

### LINKED ISSUE #547: ref(storage): use failure::Error instead of StorageError

anler:
This PR achieves these things:
- Add new `protected` crate
- Refactor `storage` crate
- Refactor storage manager actor, and continue with the style applied to the config manager and signature manager actors
- Implement a new storage backend that stores the data encrypted with AES in CBC mode

## New `protected` crate

This crate is based in [ethsign](https://crates.io/crates/ethsign) and defines only one struct `Protected` that represents a protected set of bytes. By protected we mean that the bytes the type contains will be automatically zeroed whenever it gest is dropped.

Example:
```rust
use witnet_protected::Protected;

let p: Protected = vec![].into();
std::mem::drop(p) // Vec's bytes zeroed
```

## Refactor `storage` crate

This refactor removes the dependencies to:
- `witnet_util::error`: we have talked about this and using that module is very painful
- `serde` and `serde_json`: this crate had a `Storable` trait that has been removed because its API is exactly the same as the API of a serde serializer. Also, depending on `serde_json` is a bit too much since the application could decide to use another serde serializer that do not uses json format

Also the interface for the `Storage` trait is reduced to work with bytes and not parameterized in order to force compatibility between all the backends, because for example, the key type for the rocksdb backend might be `K: AsRef<[u8]>` but `K: Hash + Eq` for hashmap backend.

## Refactor storage manager

This refactor reduces the public API of the storage manager actor to four functions:
1. `start`: starts the actor
2. `get<K: Serialize, V: DeserializeOwned>`: gets a value from the storage
3. `put<K: Serialize, V: Serialize>`: puts a value in the storage`
4. `delete<K: Serialize>`: deletes a value from the storage

Example:
```rust
storage_mngr::get::<_, ChainState>(&CHAIN_STATE_KEY).and_then(|opt_value| ...)

storage_mngr::put(&CHAIN_STATE_KEY, &chain_state)

storage_mngr::delete(&CHAIN_STATE_KEY)
```

The new storage manager will query the config manager on startup in order to decide which backend to use. This means that now we can specify the backend used by the application in the configuration file:
```
[storage]
backend = "rocksdb"
```

Additionally, if a password for the backend is set, the backend will store the data encrypted with that password:
```
[storage]
backend = "rocksdb"
password = "my strong password"
```

## Add new encrypted backend

This a proxy backend, that is, it is just in charge of encrypting/decrypting information and uses another backend to actually store the data.

Example:
```rust
let backend = backends::encrypted::Backend::new(password, backends::hashmap::Backend::new());
```

Close #499
Close #547


aesedepece:
Looking good thus far :man_shrugging:

PR #558, opened 2019-04-03
feat(signatures): sign blocks in mining manager

### LINKED ISSUE #563: Remove `influence` field from `LeadershipProof` data structure
Unless I'm missing something, the `influence` field in `LeadershipProof` is self-declarative (not attested by any cryptographic proof) and therefore can be removed as it provides a piece of data that we already now and that will never be trusted or used.
https://github.com/witnet/witnet-rust/blob/2fc73f57f2961c038e5d73ff9577ee13743ea8e8/data_structures/src/chain.rs#L232-L237
In other words, we don't punish miners for alleging a fake influence value, we just compute the influence ourselves and reject the block if the eligibility requirements are not met.

anler:
## Highlights

- Signature manager returns a `KeyedSignature` as the result of signing
- Added the conversion `From<secp256k1::Secp256k1> for Signature`
- Signing of: beacon, reveal, commit and tally, in minging manager

## Pending

Start signature manager and give it a key in order to be able to sign

## Issues

Closes #563 #554

### DIFF
--- a/node/src/actors/chain_manager/mining.rs
+++ b/node/src/actors/chain_manager/mining.rs
@@ -1,9 +1,9 @@
+use actix::prelude::*;
 use actix::{
     ActorFuture, AsyncContext, Context, ContextFutureSpawner, Handler, System, WrapFuture,
 };
 use ansi_term::Color::{White, Yellow};
 use log::{debug, error, info, warn};
-use serde_json;
 
 use futures::future::{join_all, Future};
 use std::{collections::HashMap, time::Duration};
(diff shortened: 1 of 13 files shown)


node/src/actors/chain_manager/mining.rs
```diff
@@ -109,8 +83,17 @@ impl ChainManager {
...
+                    .join(
+                        signature_mngr::sign(&beacon)
+                            .map_err(|e| error!("Couldn't sign beacon: {}", e)),
+                    )
                     .into_actor(act)
-                    .and_then(move |tally_transactions, act, ctx| {
+                    .and_then(move |(tally_transactions, keyed_signature), act, ctx| {
+                        let leadership_proof = LeadershipProof {
+                            block_sig: Some(keyed_signature.signature),
+                            influence: 0,
```
aesedepece: See #563

aesedepece:
This is a true milestone, team! :muscle:

PR #559, opened 2019-04-03
feat(node): generate secret key upon node initialization

### LINKED ISSUE #553: Generate secret key upon node initialization

lrubiorod:
## Highlights

- On node initialization, the Signature Manager will query the Storage Manager for the stored secret key and will use that one to sign blocks. If the key hasn't been stored yet, the Signature Manager will generate one and persist it using the Storage Manager
- `witnet_data_structures::chain` now contains two new types `SecretKey` and `ExtendedSecretKey` which are suitable for being persisted because they implement `Serializable`/`Deserializable`
- Fixed the key derivation method in `witnet_crypto::key` module and added tests

Close #553


node/src/signature_mngr.rs
```diff
@@ -48,11 +59,79 @@ struct SignatureManager {
...
+                                set_key(extended_sk.secret_key);
+                            }
+                            Err(e) => log::warn!("{}", e),
+                        }
+                    },
+                );
+
+                fut::ok(())
+            })
+            .spawn(ctx);
```
tmpolaczyk: This actor should not process any messages before it is initialized, therefore a `wait` is more correct here.

```suggestion
            .wait(ctx);
```

data_structures/src/chain.rs
```diff
@@ -733,6 +787,21 @@ pub struct PublicKey {
     pub bytes: [u8; 32],
 }
 
+/// Public Key data structure
+#[derive(Debug, Default, Eq, PartialEq, Clone, Serialize, Deserialize)]
+pub struct SecretKey {
```
tmpolaczyk: Why define a new data structure? Wasn't a `SecretKey` already in use before?

data_structures/src/chain.rs
```diff
@@ -331,6 +341,50 @@ impl TryInto<Secp256k1_PublicKey> for PublicKey {
...
+impl From<Secp256k1_SecretKey> for SecretKey {
+    fn from(secp256k1_sk: Secp256k1_SecretKey) -> Self {
+        let mut bytes: [u8; 32] = [0; 32];
+        bytes.copy_from_slice(&secp256k1_sk[..]);
+
+        SecretKey { bytes }
+    }
+}
+
+impl TryInto<Secp256k1_SecretKey> for SecretKey {
```
anler: **(1)** `TryInto` can be safely replace by just an `Into`, because both formats are equivalent and correct. If `secp256k1::SecretKey::from_slice` returns an error I think is perfectly ok to just panic since it is something that shouldn't have happened in the first place

data_structures/src/chain.rs
```diff
@@ -331,6 +341,50 @@ impl TryInto<Secp256k1_PublicKey> for PublicKey {
...
+impl From<ExtendedSK> for ExtendedSecretKey {
+    fn from(extended_sk: ExtendedSK) -> Self {
+        ExtendedSecretKey {
+            secret_key: SecretKey::from(extended_sk.secret_key),
+            chain_code: extended_sk.chain_code,
+        }
+    }
+}
+
+impl TryInto<ExtendedSK> for ExtendedSecretKey {
```
anler: same as **(1)**

data_structures/src/chain.rs
```diff
@@ -733,6 +787,21 @@ pub struct PublicKey {
     pub bytes: [u8; 32],
 }
 
+/// Public Key data structure
+#[derive(Debug, Default, Eq, PartialEq, Clone, Serialize, Deserialize)]
+pub struct SecretKey {
+    pub bytes: [u8; 32],
```
anler: We should use `protected::Protected` type here, an array is a `Copy` type which is kind of spooky because the key will be copied all over the stack implicitly

data_structures/src/chain.rs
```diff
@@ -733,6 +787,21 @@ pub struct PublicKey {
...
+}
+
+/// Public Key data structure
+#[derive(Debug, Default, Eq, PartialEq, Clone, Serialize, Deserialize)]
+pub struct ExtendedSecretKey {
+    /// Secret key
+    pub secret_key: SecretKey,
+    /// Chain code
+    pub chain_code: [u8; 32],
```
anler: And being paranoid, probably we should use `protected::Protected` here also

node/src/signature_mngr.rs
```diff
@@ -48,11 +59,79 @@ struct SignatureManager {
 struct SetKey(SK);
 struct Sign(Vec<u8>);
 
+/// Auxiliary methods for SignatureManager actor
+impl SignatureManager {
+    /// Method to persist the extended secret key into storage
+    fn persist_extended_sk(&self, ctx: &mut Context<Self>, extended_sk: ExtendedSK) {
```
anler: **(2)** I'm against having actor-related methods inside type methods. IMHO this method shouldn't exist and instead generate the appropriate key in `node.rs`, persist it, and tell the signature manager about it (just as we communicate the config to load to the storage manager)

data_structures/src/chain.rs
lrubiorod: Because we want a SecretKey storable, and the other is from Secp256k1

node/src/signature_mngr.rs
anler: If we do **(2)** all these changes aren't needed

data_structures/src/chain.rs
lrubiorod: Created issue #560

data_structures/src/chain.rs
lrubiorod: Created issue #560

crypto/src/key.rs
```diff
@@ -226,48 +217,72 @@ mod tests {
...
+            .derive(vec![
+                ChildNumber(0x8000002c), // purpose: BIP-44
+                ChildNumber(0x80000000), // coin_type: Bitcoin
+                ChildNumber(0x80000000), // account: hardened 0
+                ChildNumber(0),          // change: 0
+                ChildNumber(0),          // address: 0
+            ])
+            .unwrap();
+
+        let expected_account = [
```
aesedepece:
```suggestion
        let expected_secret_key = [
```

crypto/src/key.rs
```diff
@@ -226,48 +217,72 @@ mod tests {
...
             62, 6, 109, 125, 238, 45, 191, 143, 205, 63, 226, 64, 163, 151, 86, 88, 202, 17, 138,
             143, 111, 76, 168, 28, 249, 145, 4, 148, 70, 4, 176, 90, 80, 144, 167, 157, 153, 229,
             69, 112, 75, 145, 76, 160, 57, 127, 237, 184, 47, 208, 15, 214, 167, 32, 152, 112, 55,
             9, 200, 145, 160, 101, 238, 73,
         ];
 
-        assert_eq!(&expected_seed[..], seed.as_bytes());
+        let extended_sk = MasterKeyGen::new(&seed[..]).generate().unwrap();
+
+        let account = extended_sk
```
aesedepece:
```suggestion
        let secret_key = extended_sk
```

node/src/signature_mngr.rs
```diff
@@ -48,11 +56,56 @@ struct SignatureManager {
...
 
-    fn started(&mut self, _ctx: &mut Self::Context) {
+    fn started(&mut self, ctx: &mut Self::Context) {
         log::debug!("Signature Manager actor has been started!");
+
+        storage_mngr::get::<_, ExtendedSecretKey>(&EXTENDED_SK_KEY)
+            .and_then(move |extended_sk_from_storage| {
+                extended_sk_from_storage.map_or_else(
+                    || -> Box<dyn Future<Item = (), Error = failure::Error>> {
+                        log::warn!("No extended secret key in storage");
```
aesedepece: We may inform the user here that nothing bad actually happened, as we will just be generating and persisting a new secret key.
```rust
log::info("Generating and persisting a new master key for this node")
```

aesedepece:
```suggestion
                        log::warn!("No node master key found in storage");
```

node/src/signature_mngr.rs
```diff
@@ -48,11 +56,56 @@ struct SignatureManager {
...
+
+                        // Create a new Secret Key
```
aesedepece:
```suggestion
                        // Create a new master key
```

node/src/actors/storage_keys.rs
```diff
@@ -3,3 +3,6 @@ pub static PEERS_KEY: &'static [u8] = b"peers";
 
 /// Constant to specify the chain state key for the storage
 pub static CHAIN_STATE_KEY: &'static [u8] = b"chain";
+
+/// Constant to specify the secret key key for the storage
+pub static EXTENDED_SK_KEY: &'static [u8] = b"secret";
```
aesedepece: This is an extended secret key, but not just any extended secret key. Let's clearly identify the master key as such.
```suggestion
pub static MASTER_KEY: &'static [u8] = b"master_key";
```

node/src/signature_mngr.rs
```diff
@@ -48,11 +56,56 @@ struct SignatureManager {
...
 impl Actor for SignatureManager {
     type Context = Context<Self>;
 
-    fn started(&mut self, _ctx: &mut Self::Context) {
+    fn started(&mut self, ctx: &mut Self::Context) {
         log::debug!("Signature Manager actor has been started!");
+
+        storage_mngr::get::<_, ExtendedSecretKey>(&EXTENDED_SK_KEY)
+            .and_then(move |extended_sk_from_storage| {
+                extended_sk_from_storage.map_or_else(
```
aesedepece: Can we extract these two closures as testable functions?

PR #571, opened 2019-04-10
Make ChainManager verify transaction and block using the witnet_crypto crate

### LINKED ISSUE #555: Make ChainManager verify transaction and block using the witnet_crypto crate

lrubiorod:
Modified `Secp256k1Signature` (r,s to der)
Modified `LeadershipProof` (Signature to KeyedSignature)
Updated Protobuff
Updated broken tests
Created function to validate block signature
Added transaction and block signature validations 
Added tests to check block signature and Secp256k1Signature conversions
Modified pkh in mining with a random value to create different hashes

Close #555


data_structures/src/chain.rs
```diff
@@ -131,6 +131,9 @@ pub enum BlockError {
     /// The first transaction of the block is no mint.
     #[fail(display = "The block first transaction is not a mint transactions")]
     NoMint,
+    /// The block has not signature
+    #[fail(display = "The block has not signature")]
+    NoSignature,
```
tmpolaczyk: This error was useful when signature was an `Option`, now it can never happen: blocks must have a signature.

data_structures/src/chain.rs
```diff
@@ -199,7 +205,7 @@ impl<T: AsRef<[u8]>> Hashable for T {
 
 impl Hashable for Block {
     fn hash(&self) -> Hash {
-        calculate_sha256(&self.block_header.to_pb_bytes().unwrap()).into()
+        calculate_sha256(&self.proof.to_pb_bytes().unwrap()).into()
```
tmpolaczyk: Please revert this change. If you want unique block hashes for testing, change the mint pkh in mining.rs:370 to a random value:

```rust
let pkh = thread_rng().gen();
```

data_structures/src/chain.rs
```diff
@@ -181,6 +184,9 @@ pub enum BlockError {
         commits, rf
     )]
     MismatchingCommitsNumber { commits: u32, rf: u32 },
+    /// Block verification signature process failed.
+    #[fail(display = "Block {}. Verification signature process failed.", hash)]
```
aesedepece:
```suggestion
    #[fail(display = "Failed to verify the signature of block {}", hash)]
```

tmpolaczyk:
The signature of a transaction input is implemented as the signature of `output_pointer.transaction_id`, but that would imply that two inputs that point to the same transaction (but different output_index) will have the same signature. I think the `message` used in the signature should be the hash of the output pointer.

data_structures/src/chain.rs
```diff
@@ -1474,17 +1474,16 @@ mod tests {
...
-            hash_merkle_root: Hash::SHA256([0; 32]),
+            hash_merkle_root: Hash::default(),
         };
-        let signature = Signature::Secp256k1(Secp256k1Signature {
-            r: [0; 32],
-            s: [0; 32],
-            v: 0,
-        });
+        let signature = Signature::Secp256k1(Secp256k1Signature::default());
         let proof = LeadershipProof {
```
aesedepece: Just for the record, at some point `LeadershipProof` will change to:
```rust
type ProofOfEligibitlity {
  y: [u8; 32], // `SEEDBYTES` in Algorand's implementation
  ⍴: [u8; 80] // `PROOFBYTES` in Algorand's implementation
}
```

crypto/src/signature.rs
```diff
@@ -21,11 +21,11 @@ pub fn sign(secret_key: SecretKey, data: &[u8]) -> Signature {
     secp.sign(&msg, &secret_key)
 }
 /// Verify signature with a provided public key
-pub fn verify(public_key: PublicKey, data: &[u8], sig: Signature) -> Result<(), failure::Error> {
+pub fn verify(public_key: &PublicKey, data: &[u8], sig: &Signature) -> Result<(), failure::Error> {
```
aesedepece: Much better this way so as not to move the bytes :+1:

validations/src/validations.rs
```diff
@@ -398,42 +398,58 @@ pub fn validate_tally_transaction(
...
+            TransactionError::VerifySignatureFail {
+                hash: transaction.hash(),
+            }
+        })?;
+    } else {
+        Err(TransactionError::SignatureNotFound)?
+    }
+
+    for (input, keyed_signature) in inputs.iter().zip(signatures.iter().next()) {
+        validate_pkh_signature(input, keyed_signature)?
```
aesedepece: This also needs to be passed the transaction or at least the transaction's hash (in terms of performance is the same as the hash is memoized, but IMO it's always better to operate on the minimum set of data possible, so I'd prefer the latter)

node/src/actors/chain_manager/mining.rs
```diff
@@ -443,6 +448,68 @@ mod tests {
         assert_ne!(block.txns[0], transaction);
     }
 
+    #[test]
+    fn build_signed_empty_block() {
+        // Initialize transaction_pool with 1 transaction
+        let mut transaction_pool = TransactionsPool::default();
+        // In protocol buffers, when version is 0 and all the other fields are empty vectors, the
```
aesedepece: Does this comment belong here?

node/src/actors/chain_manager/mining.rs
```diff
@@ -369,7 +370,9 @@ fn build_block(
 
     // Include Mint Transaction by miner
     // TODO: Include Witnet's node PKH (keyed signature is not needed as there is no input)
```
aesedepece: Good point, keyed signatures are only needed as proofs when we try to satisfy the requirements of a P2PKH challenge, i.e. a VTT output.

PR #572, opened 2019-04-10
High level transaction factory

### LINKED ISSUE #570: High-level transaction factory
We need a way to create transactions without having to manually add the utxos and the signatures. The node should keep track of all its mint transactions from mined blocks, and use them as inputs to value transfer transactions and data requests.

The API should be similar to:

```rust
fn build_vtt(vto: ValueTransferOutput) -> Transaction {}
fn build_drt(dro: DataRequestOutput) -> Transaction {}
```

tmpolaczyk:
* Create `PublicKeyHash` struct, change protobuf schema.
* Add `GetPkh` message to signature manager, use that pkh for the mint transaction
* Keep track of the outputs that can be spent by this node:
  * When consolidating a block, update the `own_outputs` set, adding new outputs with pkh == own_pkh, and removing spent outputs
* High level transaction factory: automatically choose inputs from `own_outputs`:
  * Assumes that the node will only use one `pkh` for all the transactions (key derivation not implemented)
  * It has a limitation of 1 transaction per block: trying to create 2 transactions will likely result in a double spend (because `own_outputs` is only updated on block consolidation)
  * The algorithm used to select utxos is very naive
  * Returns a `TransactionBody`, can be signed using `sign_transaction` (requires running the `signature_mngr`
* Many new helper methods
* Tests for transaction factory and `own_outputs` logic.
* Validate `PublicKeyHash` in value transfer transactions
* Add `buildDataRequest` and `buildValueTransfer` JSON-RPC methods
* Add random poe to commits to avoid hash collisions when two commits have the same value

Close #570


node/src/actors/chain_manager/actor.rs
```diff
@@ -105,9 +105,7 @@ impl ChainManager {
                         };
                         act.chain_state = ChainState {
                             chain_info: Some(chain_info),
-                            unspent_outputs_pool: UnspentOutputsPool::default(),
-                            data_request_pool: DataRequestPool::default(),
-                            block_chain: Blockchain::default(),
+                            ..ChainState::default()
```
lrubiorod: Like!!

node/src/actors/chain_manager/mod.rs
```diff
@@ -341,6 +353,167 @@ fn update_transaction_pool(transactions_pool: &mut TransactionsPool, transaction
...
+    own_pkh: PublicKeyHash,
+    all_utxos: &UnspentOutputsPool,
+) -> Result<TransactionBody, u64> {
+    let dro = Output::DataRequest(dr);
+    let output_value: u64 = dro.value();
+    match take_enough_utxos(own_utxos, all_utxos, output_value + fee) {
+        Err(sum_of_own_utxos) => Err(sum_of_own_utxos),
+        Ok((output_pointers, input_value)) => {
+            let mut outputs = vec![];
+            let change_amount = input_value - output_value - fee;
```
lrubiorod: Maybe you can use a method to produce a VTO for the change instead of repeat code

node/src/actors/json_rpc/json_rpc_methods.rs
```diff
@@ -534,17 +534,17 @@ mod tests {
...
         });
         let consensus_output = Output::Tally(TallyOutput {
-            pkh: [0; 20],
+            pkh: PublicKeyHash::default(),
             result: [0; 32].to_vec(),
             value: 0,
         });
         let value_transfer_output = Output::ValueTransfer(ValueTransferOutput {
-            pkh: [0; 20],
+            pkh: PublicKeyHash::default(),
```
lrubiorod: In line 558 and 777 use transaction_example() to use a transaction with default fields

node/src/actors/json_rpc/json_rpc_methods.rs
```diff
@@ -860,7 +860,7 @@ mod tests {
             backup_witnesses: 0,
             commit_fee: 0,
             data_request,
-            pkh: [0; 20],
+            pkh: PublicKeyHash::default(),
```
lrubiorod: You could use:
```rust
DataRequestOutput { 
data_request,
..DataRequestOutput::default()
}
```

data_structures/src/data_request.rs
```diff
@@ -621,7 +621,7 @@ mod tests {
             reveal_fee: 0,
             tally_fee: 0,
             time_lock: 0,
-            pkh: [45; 20],
+            pkh: PublicKeyHash::default(),
```
lrubiorod: Use `..DataRequestOuput::default()`

node/src/actors/chain_manager/mod.rs
```diff
@@ -341,6 +351,37 @@ fn update_transaction_pool(transactions_pool: &mut TransactionsPool, transaction
...
+            match output {
+                Output::ValueTransfer(x) => {
+                    if x.pkh == own_pkh {
+                        own_utxos.insert(OutputPointer {
+                            transaction_id: transaction.hash(),
+                            output_index: i as u32,
+                        });
+                    }
+                }
+                Output::DataRequest(_) => {}
```
```suggestion
                _=> {}
```

node/src/actors/chain_manager/transaction_factory.rs
```diff
@@ -0,0 +1,618 @@
...
+    dr: DataRequestOutput,
+    fee: u64,
+    own_utxos: &HashSet<OutputPointer, S>,
+    own_pkh: PublicKeyHash,
+    all_utxos: &UnspentOutputsPool,
+) -> Result<TransactionBody, u64> {
+    build_vtt_inner(vec![], Some(dr), fee, own_utxos, own_pkh, all_utxos)
+}
+
+fn build_vtt_inner<S: std::hash::BuildHasher>(
```
lrubiorod: Comments??

node/src/actors/chain_manager/transaction_factory.rs
```diff
@@ -0,0 +1,618 @@
...
+        );
+        assert_eq!(
+            build_vtt(vec![], 1001, &own_utxos, own_pkh, &all_utxos),
+            Err(1000)
+        );
+        assert_eq!(
+            build_vtt(vec![pay_bob(500)], 600, &own_utxos, own_pkh, &all_utxos),
+            Err(1000)
+        );
+        assert_eq!(
```
lrubiorod: Delete redundant test

node/src/actors/chain_manager/transaction_factory.rs
```diff
@@ -0,0 +1,618 @@
...
+            pay_me(334),
+        ];
+        let (own_utxos, all_utxos) = build_utxo_set(outputs, None, vec![]);
+        assert_eq!(own_utxos.len(), 7);
+
+        let t1 = build_vtt(vec![pay_bob(1000)], 0, &own_utxos, own_pkh, &all_utxos).unwrap();
+        assert_eq!(outputs_sum_not_mine(&t1.outputs), 1000);
+
+        let t2 = build_vtt(vec![pay_bob(990)], 10, &own_utxos, own_pkh, &all_utxos).unwrap();
+
```
lrubiorod:
```suggestion
```

node/src/actors/chain_manager/mod.rs
lrubiorod: or if let

node/src/actors/chain_manager/actor.rs
```diff
@@ -177,4 +175,22 @@ impl ChainManager {
...
+        signature_mngr::pkh()
+            .into_actor(self)
+            .map_err(|e, _act, _ctx| {
+                error!(
+                    "Error while getting public key hash from signature manager: {}",
+                    e
+                );
+            })
+            .and_then(|res, act, _ctx| {
+                act.own_pkh = Some(res);
```
anler: *Suggestion*: : changing actor state inside a future is *malo*

node/src/actors/chain_manager/actor.rs
```diff
@@ -177,4 +175,22 @@ impl ChainManager {
             })
             .wait(ctx);
     }
+
+    /// Load public key hash from signature manager
+    fn get_pkh(&mut self, ctx: &mut Context<Self>) {
+        signature_mngr::pkh()
+            .into_actor(self)
```
anler: *Suggestion*: `map_error` could be moved on top of `into_actor` and you avoid declaring the `_act, _ctx` params

node/src/actors/chain_manager/actor.rs
```diff
@@ -177,4 +175,22 @@ impl ChainManager {
...
+                    "Error while getting public key hash from signature manager: {}",
+                    e
+                );
+            })
+            .and_then(|res, act, _ctx| {
+                act.own_pkh = Some(res);
+                debug!("Public key hash successfully loaded from signature manager");
+                actix::fut::ok(())
+            })
+            .wait(ctx);
```
anler: *Suggestion*: the wait can be avoided if *"in principle"* we remove the race condition of changing the actor state from inside the callback

node/src/actors/chain_manager/mod.rs
```diff
@@ -341,6 +351,37 @@ fn update_transaction_pool(transactions_pool: &mut TransactionsPool, transaction
     }
 }
 
+/// Update the list of unspent outputs that can be spent by this node.
+fn update_own_utxos(
+    own_utxos: &mut HashSet<OutputPointer>,
+    own_pkh: PublicKeyHash,
+    transactions: &[Transaction],
+) {
```
anler: *Suggestion*: I would prefer a function that filter all the transactions with output `own_pkh` and returns an iterator to those values, later we could use it to update the utxos. The idea is to keep the *side-effecting* as contained as possible

node/src/signature_mngr.rs
```diff
@@ -109,8 +121,8 @@ impl Actor for SignatureManager {
             })
             .map_err(|e| log::error!("Couldn't initialize Signature Manager: {}", e))
             .into_actor(self)
-            .map(|secret_key, _, ctx| {
-                ctx.notify(SetKey(secret_key));
+            .map(|secret_key, act, _ctx| {
+                act.set_key(secret_key);
```
anler: Why not use `ctx.notify` as before?

tmpolaczyk: When using notify, the actor can receive other messages before that notify, which leads to handling external messages in an uninitialized state. By calling `act.set_key()`, we guarantee a synchronous initialization.

data_structures/src/data_request.rs
```diff
@@ -412,6 +399,28 @@ impl DataRequestPool {
...
+                    output_index,
+                };
+                // A data request output should have a valid value transfer input
+                // Which we assume valid as it should have been already verified
+                // time_lock_epoch: The epoch during which we will start accepting
+                // commitments for this data request
+                // FIXME(#338): implement time lock
+                // An enhancement to the epoch manager would be a handler GetState which returns
+                // the needed constants to calculate the current epoch. This way we avoid all the
+                // calls to GetEpoch
```
aesedepece: Agreed.

node/src/actors/chain_manager/handlers.rs
```diff
@@ -471,3 +474,71 @@ impl Handler<PeersBeacons> for ChainManager {
...
+    fn handle(&mut self, msg: BuildVtt, ctx: &mut Self::Context) -> Self::Result {
+        match transaction_factory::build_vtt(
+            msg.vto,
+            msg.fee,
+            &self.chain_state.own_utxos,
+            self.own_pkh.unwrap(),
+            &self.chain_state.unspent_outputs_pool,
+        ) {
+            Err(total_value) => error!(
+                "Cannot build value transfer transaction with cost > {}",
```
aesedepece:
```suggestion
                "Cannot build a transaction transferring more value than the current available balance {} + {} > {}",
```

aesedepece: `msg.vto.value, msg.fee, total_value`

node/src/actors/chain_manager/handlers.rs
```diff
@@ -471,3 +474,71 @@ impl Handler<PeersBeacons> for ChainManager {
...
+    fn handle(&mut self, msg: BuildDrt, ctx: &mut Self::Context) -> Self::Result {
+        match transaction_factory::build_drt(
+            msg.dro,
+            msg.fee,
+            &self.chain_state.own_utxos,
+            self.own_pkh.unwrap(),
+            &self.chain_state.unspent_outputs_pool,
+        ) {
+            Err(total_value) => error!(
+                "Cannot build data request transaction with cost > {}",
```
aesedepece: I'd use a similar message here as in my previous suggestion.

node/src/actors/chain_manager/mod.rs
aesedepece: It looks to me like this should be affected by a diffing strategy too. Instead of mutating `own_utxos` here, we should accumulate `insert(output_pointer)` and `delete(output_pointer)` commands into a vector just as we are validating the transactions, with views to later apply those changes atomically when we consolidate the block.

node/src/signature_mngr.rs
```diff
@@ -158,3 +172,14 @@ impl Handler<Sign> for SignatureManager {
...
     }
 }
+
+impl Handler<GetPkh> for SignatureManager {
+    type Result = <GetPkh as Message>::Result;
+
+    fn handle(&mut self, _msg: GetPkh, _ctx: &mut Self::Context) -> Self::Result {
+        match self.keypair {
+            Some((_secret, public)) => Ok(PublicKeyHash::from_public_key(&public.into())),
+            None => bail!("Signature Manager cannot sign because it contains no key"),
```
aesedepece:
```suggestion
            None => bail!("Tried to retrieve the public key hash for node's main keypar from Signature Manager, but it contains none (looks like it was not initialized properly)"),
```

PR #573, opened 2019-04-10
Remove redundant code in tests

lrubiorod:
Created transaction_example and block_example methods
Added Default trait to some structs in chain.rs
Removed redundant code in tests
Modified codec log from debug to trace
Ignored test_run_retrieval_random_api() by connection problems

(Merge after #571)

### DIFF
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -225,7 +225,7 @@ impl Hashable for Transaction {
 }
 
 /// Block header structure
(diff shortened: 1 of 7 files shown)


schemas/witnet/witnet.proto
```diff
@@ -62,10 +62,7 @@ message Block {
         Hash hash_merkle_root = 3;
     }
     message LeadershipProof {
```
aesedepece: As I mentioned in #571, we will be redefining `LeadershipProof` as soon as we implement the VRF.

PR #577, opened 2019-04-11
Synchronization without cloning pools

anler:
## Changes
`validate_block` function now returns a *diff* of changes to apply to the utxo set instead of cloning the utxo set before validation. This has two implications:

1. When choosing the candidate: Each block candidate keeps its own diff, when one is chosen as the best candidate, its diff is consolidated into the chain state.
2. When synchronizing the chain: Each block in the batch of 500 blocks is validated, if it's valid, its diff is consolidated into the chain state. After the validation of the entire batch the chain is persisted into the storage. If for some reason one block it's not valid, the entire batch is cancelled and the chain state is restored to the last persisted state.

### DIFF
--- a/node/src/actors/chain_manager/handlers.rs
+++ b/node/src/actors/chain_manager/handlers.rs
@@ -190,7 +190,7 @@ impl Handler<AddBlocks> for ChainManager {
             StateMachine::WaitingConsensus => {}
             StateMachine::Synchronizing => {
                 if let Some(target_beacon) = self.target_beacon {
-                    let mut batch_succeded = true;
+                    let mut batch_succeeded = true;
                     for block in msg.blocks.iter() {
                         if let Err(e) = self.process_requested_block(ctx, block) {
                             log::error!("Error processing block: {}", e);
--- a/node/src/actors/chain_manager/handlers.rs
+++ b/node/src/actors/chain_manager/handlers.rs
@@ -205,14 +205,14 @@ impl Handler<AddBlocks> for ChainManager {
                         }
                     }
 
-                    if batch_succeded {
-                        self.persist_chain_state(ctx);
+                    if batch_succeeded {
                         self.persist_blocks_batch(ctx, msg.blocks, target_beacon);
                         let to_be_stored =
                             self.chain_state.data_request_pool.finished_data_requests();
                         to_be_stored.into_iter().for_each(|dr| {
                             self.persist_data_request(ctx, &dr);
                         });
+                        self.persist_chain_state(ctx);
                     }
 
                     let beacon = self.get_chain_beacon();
--- a/node/src/actors/chain_manager/handlers.rs
+++ b/node/src/actors/chain_manager/handlers.rs
@@ -5,7 +5,7 @@ use witnet_data_structures::{
     chain::{CheckpointBeacon, Epoch, Hashable, InventoryEntry, InventoryItem},
     error::ChainInfoError,
 };
-use witnet_validations::validations::validate_block;
+use witnet_validations::validations::{validate_block, validate_transaction, UtxoDiff};
 
 use super::{ChainManager, ChainManagerError, StateMachine};
 use crate::{
@@ -20,7 +20,6 @@ use crate::{
     utils::mode_consensus,
 };
 use std::collections::HashMap;
-use witnet_validations::validations::validate_transaction;
 
 ////////////////////////////////////////////////////////////////////////////////////////
 // ACTOR MESSAGE HANDLERS
@@ -89,13 +88,11 @@ impl Handler<EpochNotification<EveryEpochPayload>> for ChainManager {
                 if let (Some(current_epoch), Some(chain_info)) =
                     (self.current_epoch, self.chain_state.chain_info.as_ref())
                 {
-                    let candidates = self.candidates.clone();
-
                     // Decide the best candidate
                     // TODO: replace for loop with a try_fold
                     let mut chosen_candidate = None;
-                    for (key, block_candidate) in candidates {
-                        if let Some((chosen_key, _)) = chosen_candidate.clone() {
+                    for (key, block_candidate) in self.candidates.drain() {
+                        if let Some((chosen_key, _, _)) = chosen_candidate {
                             if chosen_key < key {
                                 // Ignore candidates with bigger hashes
                                 continue;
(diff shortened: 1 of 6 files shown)


validations/src/validations.rs
```diff
@@ -774,3 +753,89 @@ mod tests {
...
+    }
+
+    /// Record an insertion to perform on the utxo set
+    pub fn insert_utxo(&mut self, output_pointer: OutputPointer, output: Output) {
+        self.diff.utxos_to_add.insert(output_pointer, output);
+    }
+
+    /// Record a deletion to perform on the utxo set
+    pub fn remove_utxo(&mut self, output_pointer: OutputPointer) {
+        match self.diff.utxos_to_add.remove(&output_pointer) {
```
lrubiorod: clippy doesnt like it

```suggestion
        if let None = self.diff.utxos_to_add.remove(&output_pointer) {
   self.diff.utxos_to_remove.insert(output_pointer);
 }
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -278,54 +271,55 @@ impl ChainManager {
...
-                    ctx.address().do_send(AddTransaction {
-                        transaction: reveal,
-                    })
-                }
-                // Persist finished data requests into storage
-                let to_be_stored = self.chain_state.data_request_pool.finished_data_requests();
-                to_be_stored.into_iter().for_each(|dr| {
-                    self.persist_data_request(ctx, &dr);
-                    if info_flag {
+                // Insert candidate block into `block_chain` and persist it
```
```suggestion
                // Insert candidate block into `block_chain`  state
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -278,54 +271,55 @@ impl ChainManager {
...
+                self.chain_state.block_chain.insert(block_epoch, block_hash);
+
+                // Persist chain_info into storage
```
lrubiorod: Move this comment to its properly place, near to `self.persist_chain_state(ctx);` in line 316

node/src/actors/chain_manager/handlers.rs
```diff
@@ -89,13 +88,11 @@ impl Handler<EpochNotification<EveryEpochPayload>> for ChainManager {
...
                     (self.current_epoch, self.chain_state.chain_info.as_ref())
                 {
-                    let candidates = self.candidates.clone();
-
                     // Decide the best candidate
                     // TODO: replace for loop with a try_fold
                     let mut chosen_candidate = None;
-                    for (key, block_candidate) in candidates {
-                        if let Some((chosen_key, _)) = chosen_candidate.clone() {
+                    for (key, block_candidate) in self.candidates.drain() {
```
tmpolaczyk: Now that we are 
Since now we are draining the vector, in line 152 there is a redundant clear:
```rust
// Clear candidates
self.candidates.clear();
```

node/src/actors/chain_manager/handlers.rs
```diff
@@ -197,28 +189,41 @@ impl Handler<AddBlocks> for ChainManager {
         match self.sm_state {
             StateMachine::WaitingConsensus => {}
             StateMachine::Synchronizing => {
-                let old_chain_state = self.chain_state.clone();
                 if let Some(target_beacon) = self.target_beacon {
-                    for block in msg.blocks {
+                    let mut batch_succeded = true;
```
aesedepece:
```suggestion
                    let mut batch_succeeded = true;
```

node/src/actors/chain_manager/handlers.rs
```diff
@@ -197,28 +189,41 @@ impl Handler<AddBlocks> for ChainManager {
...
-                        .chain_state
-                        .chain_info
-                        .as_ref()
-                        .unwrap()
-                        .highest_block_checkpoint;
-                    if our_beacon == target_beacon {
+
+                    if batch_succeded {
+                        self.persist_chain_state(ctx);
+                        self.persist_blocks_batch(ctx, msg.blocks, target_beacon);
```
aesedepece: These two lines produce subsequent synchronous writes to the storage without atomicity. I'd suggest reversing their order because I prefer my storage to temporally have orphan blocks rather than perpetually missing blocks that are actually referred in the block chain.

node/src/actors/chain_manager/handlers.rs
```diff
@@ -197,28 +189,41 @@ impl Handler<AddBlocks> for ChainManager {
...
+                        let to_be_stored =
+                            self.chain_state.data_request_pool.finished_data_requests();
+                        to_be_stored.into_iter().for_each(|dr| {
+                            self.persist_data_request(ctx, &dr);
```
aesedepece: This may also be done before `persist_chain_state` for the same reasons.

node/src/actors/chain_manager/mod.rs
```diff
@@ -278,67 +281,88 @@ impl ChainManager {
...
                         show_info_tally(&self.chain_state.unspent_outputs_pool, dr, block_epoch);
-                    }
-                });
+                    });
 
-                if info_flag {
                     show_info_dr(&self.chain_state.data_request_pool, &block);
 
-                    debug!("{:?}", block);
+                    log::trace!("{:?}", block);
```
aesedepece: 💞

node/src/actors/chain_manager/mod.rs
```diff
@@ -278,67 +281,88 @@ impl ChainManager {
...
                 error!("No ChainInfo loaded in ChainManager");
             }
         }
     }
+
+    fn get_chain_beacon(&self) -> CheckpointBeacon {
+        self.chain_state
+            .chain_info
+            .as_ref()
+            .unwrap()
```
aesedepece: Can we avoid the `unwrap` and rather handle the `None` case where used?

node/src/actors/chain_manager/mod.rs
```diff
@@ -278,54 +271,55 @@ impl ChainManager {
...
+
+                    // Persist chain_info into storage
                     self.persist_chain_state(ctx);
-                }
 
-                // Send notification to JsonRpcServer
-                JsonRpcServer::from_registry().do_send(NewBlock { block })
+                    // Send notification to JsonRpcServer
+                    JsonRpcServer::from_registry().do_send(NewBlock {
+                        block: block.clone(),
```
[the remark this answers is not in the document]
anler: Unfortunately, it is. Currently the line of thought is: *consolidate_block* shouldn't take ownership of the block because when validating and consolidating block batches, we need to keep ownership of those blocks in order to persist them. The ideal would be to move this *notification-related/actor-coupled* code out of the consolidation function.

node/src/actors/chain_manager/mod.rs
anler: We could and I did previously, but I reverted the change because I didn't want to introduce too many changes. Problem is that the invariant for the actor's state and the state machine states is not that clear and there are a lot of unnecessary checks for optional values that inside the synchronization/candidate logic aren't optional anymore.

Instead of just changing this unwrap I would prefer to leave it as it is and refactor this actor in the future to avoid the unnecessary checks, e.g.: [let Some(...)](https://github.com/witnet/witnet-rust/pull/577/files/95916ff11b761ef16975ac9f9f31786a01f13414#diff-089d6d21f03cb529865367bff4779506R192), [match self.chain_state.chain_info.as_mut()](https://github.com/witnet/witnet-rust/blob/95916ff11b761ef16975ac9f9f31786a01f13414/node/src/actors/chain_manager/mod.rs#L274) and so on

validations/src/validations.rs
```diff
@@ -1,15 +1,17 @@
+use std::default::Default;
```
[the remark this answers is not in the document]
aesedepece: This needs to be addressed.

node/src/actors/chain_manager/handlers.rs
aesedepece: This needs to be addressed.

PR #586, opened 2019-04-15
Continuous integration and Justfile overhaul

### LINKED ISSUE #583: Build a docker container for continuous integration
We can build a docker container with all our compilation boilerplate already in place:
- `x86_64-unknown-linux-musl` Rust toolchain
- `clang`
- `libssl-dev`
- `protobuf-compiler`
- `librocksdb5.17`, as we may try to link it dinamically

Then we can tell Travis to checkout the container and run the tests inside it.

This would also allow us to run the tests locally from any platform in a consistent way.

aesedepece:
- Speeds up travis builds by improving caching and linking rocksdb dynamically
- Renames `just travis` to `just ci`
- Adds the possibility of passing flags to `just clippy`, `just fmt` and `just node`

Close #583

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,18 @@
-language: rust
-cache: cargo
-dist: trusty
+dist: xenial
 sudo: false
+language: rust
+
 rust:
   - stable
+
+# https://levans.fr/rust_travis_cache.html
+# Need to cache the whole `.cargo` directory to keep .crates.toml for `cargo update` to work...
+# ...but don't cache the cargo registry
+cache:
+  directories:
+    - /home/travis/.cargo
+before_cache:
+  - rm -rf /home/travis/.cargo/registry
 
 addons:
   apt:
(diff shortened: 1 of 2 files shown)


aesedepece:
@witnet/devs this would rename `just travis` as `just ci`

tmpolaczyk:
When running `just` with no arguments, it runs the first command from the `Justfile` (currently it installs clippy). I think it would be useful to have `ci` as the first command, for convenience.

aesedepece:
Agreed!

.travis.yml
```diff
@@ -20,15 +29,15 @@ addons:
       - gcc
       - binutils-dev
       - protobuf-compiler
+      - librocksdb-dev
 
 env:
   global:
+    - ROCKSDB_LIB_DIR=/usr/lib/ # Use dynamically linked RocksDB library to speed up builds
```
tmpolaczyk: Does this mean that end users will have to install rocksdb in order to run our binaries?

aesedepece: Not at all! Users can opt-in for dynamic linking of RocksDB by setting this environment variable, but the default is still local compilation of the C library + static linking. This line is only for Travis.

tmpolaczyk: I see, nice!

PR #588, opened 2019-04-17
Added log::info for block candidates

### LINKED ISSUE #568: Log our own block candidate proposals
It'd be nice if the `witnet-rust` logged an `info` trace every time we propose a block candidate. The logged message may namely contain the hash of the proposed block, which would facilitate telling if we got lucky and our proposed block candidate got consolidated.

gil-air-may:
Closes #568 

I have added a `log::info` on the Handler for AddCandidate.

Here's how it looks:

block candidate

### DIFF
--- a/node/src/actors/chain_manager/handlers.rs
+++ b/node/src/actors/chain_manager/handlers.rs
@@ -238,7 +238,7 @@ impl Handler<AddCandidates> for ChainManager {
     fn handle(&mut self, msg: AddCandidates, _ctx: &mut Context<Self>) {
         // AddCandidates is needed in all states
         for block in msg.blocks {
-            self.process_candidate(block)
+            self.process_candidate(block);
         }
     }
 }
--- a/node/src/actors/chain_manager/mining.rs
+++ b/node/src/actors/chain_manager/mining.rs
@@ -116,6 +116,12 @@ impl ChainManager {
                             Ok(_) => {
                                 // Send AddCandidates message to self
                                 // This will run all the validations again
+
+                                let block_hash = block.hash();
+                                log::info!(
+                                    "Proposed block candidate {}",
+                                    Yellow.bold().paint(block_hash.to_string())
+                                );
                                 act.handle(
                                     AddCandidates {
                                         blocks: vec![block],


node/src/actors/chain_manager/handlers.rs
```diff
@@ -238,7 +239,10 @@ impl Handler<AddCandidates> for ChainManager {
     fn handle(&mut self, msg: AddCandidates, _ctx: &mut Context<Self>) {
         // AddCandidates is needed in all states
         for block in msg.blocks {
-            self.process_candidate(block)
+            let block_hash = block.hash();
+            self.process_candidate(block);
+            info!("Block {} candidate proposed",
```
anler: We are trying to avoid using log functions without the crate prefix because it's easier to locate them with a simple grep: `log::`. So instead of using just `info!("...")` it would be better to use `log::info!("....")`.

gil-air-may: Hmm, I see. I will change that right away.

aesedepece:
Hi @gil-air-may, thanks for your PR!

It seems like your change makes the node produce the log message every time it sees a new candidate, which happens many times for every proposed block because of the way in which blocks are broadcast (a.k.a. gossiping). Also, other node's candidates are being logged.

The point of this issue is adding a log message only for our own block proposals. Therefore, the part of the code where the logging happens should probably be close to where mining happens, not block processing and validation.

Please let us know if you need more assistance finding where in the code base needs the log trace be placed. Or feel free to guess it by yourself! :wink:

gil-air-may:
Hey there @aesedepece! I'm new to this kind of system, so I guess that explains my oversight hehe. Anyway, the app seems really interesting and I enjoy the challenge. I will give it another shot later today.

gil-air-may:
Ok, I have stopped logging all the block's hashes that were being broadcasted. Right now, I believe the message appears exclusively for the node's proposed block. I tested out the changes running two nodes simultaneously.

Node 1 displays:

57604446_306342683395221_2576776800445136896_n

And Node 2 displays:
58420016_408391999952697_6239433943066607616_n

Let me know if you need more improvements :smile:

node/src/actors/chain_manager/mining.rs
```diff
@@ -116,6 +116,12 @@ impl ChainManager {
                             Ok(_) => {
                                 // Send AddCandidates message to self
                                 // This will run all the validations again
+
+                                let block_hash = block.hash();
+                                log::info!(
+                                    "Block {} candidate proposed",
```
aesedepece:
```suggestion
                                    "Proposed block candidate {}",
```

aesedepece:
Hey @gil-air-may commit messages need to abide by the [conventional commits specification](https://www.conventionalcommits.org/en/v1.0.0-beta.2/). A example for this issue could be:
```
feat(mining): add log::info for own block candidates
```
Btw, feel free to drop by the [Witnet community on Discord](https://discord.gg/X4uurfP) at any time to discuss any future contributions and learn more about the project :D

gil-air-may:
Squashed and renamed :+1:

aesedepece:
Merged as c0e5f912bded7c0119c35a34d01ecd6fc068eff1 :dancer:

PR #623, opened 2019-04-26
feat(storage_mngr): replace json with bincode for serialization

tmpolaczyk:
JSON was a temporary solution, and it does not work when maps have non-string keys.

[Bincode](https://github.com/TyOverby/bincode) is a very simple serializer where "the size of the encoded object will be the same or smaller than the size that the object takes up in memory in a running Rust program".

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -179,6 +179,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "bincode"
+version = "1.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "autocfg 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


tmpolaczyk:
It looks like something doesn't work when trying to read from storage:

```
[2019-04-26T09:27:10Z ERROR witnet_node::signature_mngr] Couldn't initialize Signature Manager: tag for enum is not valid, found 195
```

tmpolaczyk:
> It looks like something doesn't work when trying to read from storage:
> 
> ```
> [2019-04-26T09:27:10Z ERROR witnet_node::signature_mngr] Couldn't initialize Signature Manager: tag for enum is not valid, found 195
> ```

Fixed, the problem was that we were trying to deserialize everything as an `Option<T>`, with JSON it did work because `None` is just null, but in bincode `Option` is a tagged enum, so it was trying to parse the tag.

Issue #603, opened 2019-04-22
Create Sessions Manager + Session in wallet


mariocao:

anler:
Implemented in #648

Issue #602, opened 2019-04-22
Wallet actors hierarchy


mariocao:

anler:
Implemented in #648

PR #624, opened 2019-04-29
Add new `net` crate and refactor `wallet` crate to use it

anler:
# What

Related to Issues: #602 and #603

* Create a `witnet_net` crate of reusable networking components such as servers and clients, that can be used by the wallet and node independently.
* Refactor `witnet_wallet` to use the websockets server from `witnet_crate` separating its business logic from networking code

## Issue 602

After further investigation, implementing a tree of supervision in our actix application seems a bit too much and will add a lot of effort in our side since Actix in this regard is quite different than Akka.

## Issue 603

What's implemented here defers from the issue in that the approach used here is more close to the one taken by other libraries in the actix ecosystem such as [actix-web](https://github.com/actix/actix-web/) and [actix-net](https://github.com/actix/actix-net). I believe this approach where the library user is presented with a well-defined API oriented to combinators instead of actors greatly reduces the amount of code it needs to be written to reuse components.

## Usage
Run the wallet with:

```sh
cargo run -p witnet_wallet
```

and then in the javascript console of your web browser type:

```js
var sock= (() => { let s = new WebSocket('ws://localhost:3030');s.addEventListener('message', (e) => {  console.log('Rcv =>', e.data) });return s; })();

sock.send('{"jsonrpc":"2.0","method":"getWalletInfos","id":"1"}');
```

## Highlights
These are other things addressed by this PR.

### `witnet_wallet`
* Server and business logic is separated
* Wallet request handlers are released of the task of encoding the response as json
* Client used by handlers to forward requests to the node is simplified and its futures are scheduled in the same context as the handler

### `witnet_net`
* Gracefully handling of Unix signals
* Concurrent server: it will spawn as many threads as cores are available
* Websocket handlers can broadcast notifications without the need of other clients to be subscribed

## Further work

Work still pending that I think is best to address in a separate PR:
* Move Node client to `witnet_crate`
* Support different transports in the Node client
* Add state to the wallet handlers
* Make the `wallet` and `net` configurable through their API


Cargo.toml
```diff
@@ -11,7 +11,7 @@ description = "An in-progress open source implementation of the Witnet protocol
 edition = "2018"
 
 [workspace]
-members = ["config", "node", "crypto", "data_structures", "p2p", "storage", "wallet", "validations", "protected", "reputation"]
+members = ["config", "node", "crypto", "data_structures", "p2p", "storage", "wallet", "validations", "protected", "net"]
```
tmpolaczyk:
```suggestion
members = ["config", "node", "crypto", "data_structures", "p2p", "storage", "wallet", "validations", "protected", "reputation", "net"]
```

wallet/src/client/actor.rs
```diff
@@ -0,0 +1,96 @@
...
+}
+
+impl Supervised for JsonRpc {}
+
+impl SystemService for JsonRpc {}
+
+/// A JsonRPC request.
+pub struct Request {
+    pub method: String,
+    pub params: Option<rpc::Value>,
```
tmpolaczyk: This could be a `jsonrpc::Params` initialized to `None`.
```suggestion
    pub params: rpc::Params,
```

Then, the params method would be just

```rust
self.params = params;
self
```

wallet/src/response.rs
```diff
@@ -0,0 +1,9 @@
+//! Utilities related to JsonRPC responses.
+
+use futures::Future;
+use jsonrpc_core::Error;
+
+/// TODO: doc
+pub trait Response<I>: Future<Item = I, Error = Error> {}
```
tmpolaczyk: Good idea! However, keep in mind that the `fn a() -> impl Response<T>` syntax cannot be used when the function returns different types of futures. In that case, we need boxed futures as defined in `node/src/actors/jsonrpc/json_rpc_methods.rs`:

```
type JsonRpcResultAsync<T> = Box<dyn Future<Item = T, Error = jsonrpc_core::Error> + Send>;
```

anler: I believe the Box is not needed since you can always combine different futures into one of the same type.

wallet/src/client/actor.rs
anler: 👌nice one

wallet/src/response.rs
tmpolaczyk: Oh, I see, that's what the `Either` type is for: https://tokio.rs/docs/futures/combinators/#returning-from-multiple-branches

anler:
Superseded by #648

PR #627, opened 2019-04-29
[After rust1.35]fix(justfile): modify justfile to fix clippy

lrubiorod:


aesedepece:
This is to be merged as soon as Rust `1.35` is stabilized.

tmpolaczyk:
Rust 1.35 has just been stabilized

tmpolaczyk:
The build fails because the dev-dependency `just` used the deprecated syntax in their own repo. I think it should be ok to just remove the `just` dependency.

PR #642, opened 2019-05-08
Remove custom TryFrom and TryInto implementations

kinggoesgaming:
The local TryFrom and TryInto traits replaced with rust's actual traits, as they are now stable


kinggoesgaming:
Ignore the `This currently does not touch the try_from crate codebase` bit in the commit message.

kinggoesgaming:
I will create a new PR (that has all the changes consolated into one)

PR #646, opened 2019-05-09
Fix hash serialization in bincode

### LINKED ISSUE #643: Bincode deserialization fails with untagged enums
Reading a block from storage fails on master.

Bincode does not support untagged enums, like this one in `chain.rs`:

```rust
#[serde(untagged)]
enum HashSerializationHelper {
```

The `HashSerializationHelper` is a hack which allows us to use `"hash": "01321231...` in JSON but `[0x01, 0x32, 0x12, ...` in storage. In other words, we can choose to use the hash as a string or as a byte array. The problem is that bincode cannot distinguish between a string and a byte array, so it must be explicitly tagged. Removing the `#[serde(untagged)]` attribute would fix the issue, but the JSON would also be tagged, as `"hash": { "AsString": "01321231...`.

However, we could use the `is_human_readable()` serde method in order to implement a custom serialization. That method will return `true` when using JSON but `false` when using bincode, so we can implement a serializer and a deserializer for `Hash` which will always store it as a byte array in bincode, and always as a string in JSON. Therefore removing the `HashSerializationHelper`.

It would be nice to have a test to prevent this kind of errors in the future, but it isn't as simple as creating an empty block and storing it in the storage because for example an empty `Vec<Hash>` field would be serialized and deserialized with no problems. So we would need a sample block with all the possible kinds of transactions.

tmpolaczyk:
Close #643


data_structures/src/chain.rs
```diff
@@ -404,6 +386,28 @@ impl fmt::Display for PublicKeyHash {
     }
 }
 
+/// Error when parsing hash from string
+#[derive(Debug, Fail)]
+pub enum PublicKeyHashParseError {
+    #[fail(display = "Invalid hash length: expected 20 bytes but got {}", _0)]
```
aesedepece:
```suggestion
    #[fail(display = "Invalid PKH length: expected 20 bytes but got {}", _0)]
```

data_structures/src/error.rs
```diff
@@ -164,13 +165,13 @@ pub enum BlockError {
 
 #[derive(Debug, Fail)]
 pub enum OutputPointerParseError {
-    #[fail(display = "output pointer has an invalid length")]
-    InvalidHashLength,
+    #[fail(display = "failed to parse transaction hash: {}", _0)]
```
aesedepece:
```suggestion
    #[fail(display = "Failed to parse transaction hash: {}", _0)]
```

data_structures/src/error.rs
```diff
@@ -164,13 +165,13 @@ pub enum BlockError {
...
+    Hash(HashParseError),
     #[fail(
         display = "output pointer has the wrong format, expected '<transaction id>:<output index>'"
     )]
     MissingColon,
-    #[fail(display = "could not parse output index as an integer")]
+    #[fail(display = "could not parse output index as an integer: {}", _0)]
```
aesedepece:
```suggestion
    #[fail(display = "Could not parse output index as an integer: {}", _0)]
```

data_structures/src/lib.rs
```diff
@@ -24,5 +24,10 @@ pub mod error;
 /// Module containing data_request structures
 pub mod data_request;
 
+/// Serialization boilerplate to allow serializing some data structures as
+/// strings or bytes depending on the serializer.
+mod serialization_helpers;
+
+// TODO: tests should not in src/tests.rs, there already exists a tests/ folder
```
aesedepece:
```suggestion
// TODO: tests should not in src/tests.rs, there already exists a `tests/` folder
```

PR #648, opened 2019-05-09
Wallet persistence, state, and `getWalletInfos` handler

anler:
# Overview

This PR achieves three things:

1. Add an actor `Storage` that controls access to the wallet databases. This actor runs behind a `SyncArbiter` because its operations are blocking.
2. Add an actor `App` that contains the wallet API exposed through websockets. This actor will contain the state used across all handlers.
3. Implement the `getWalletInfos` handler.

## How to test

Open a browser console and paste this:

```js
var sock= (() => { let s = new WebSocket('ws://localhost:11212');s.addEventListener('message', (e) => { console.log('Rcv =>', e.data) });return s; })();
sock.send('{"jsonrpc":"2.0","method":"getWalletInfos","id":"1"}');
// Result: Rcv => {"jsonrpc":"2.0","result":[],"id":"1"}
```

## Log messages

If you run the wallet with `trace` level you will be able to see the received params and the returned response:

```
% cargo run -p witnet_wallet -- run
   Compiling witnet_wallet v0.2.0 (/home/anler/src/public/witnet-rust/wallet)
    Finished dev [unoptimized + debuginfo] target(s) in 24.06s
     Running `target/debug/witnet_wallet run`
[DEBUG] Loading config from `/home/anler/src/public/witnet-rust/witnet.toml`
[INFO ] Starting 1 workers
[INFO ] Starting server on 127.0.0.1:11212
[INFO ] Received connection from 127.0.0.1:37936
[TRACE] Received request: {"jsonrpc":"2.0","method":"getWalletInfos","id":"1"}
[DEBUG] Received request getWalletInfos(params: None)
[DEBUG] Sending response Array([])
[TRACE] Sending response: {"jsonrpc":"2.0","result":[],"id":"1"}
[INFO ] Closed connection to 127.0.0.1:37936
...
```


wallet/src/wallet.rs
```diff
@@ -0,0 +1,185 @@
+//! # Wallet related structs
```
tmpolaczyk: I guess this structs will be replaced by the handlers in `wallet/src/actors/app/handlers/`?

anler: Probably, I have kept them here for the meantime but when implementing the message for the handler we will probably move each one into the appropriate handler.

net/src/server/ws/actors/notifications.rs
```diff
@@ -0,0 +1,100 @@
...
+    fn handle(&mut self, msg: Subscribe, _ctx: &mut Self::Context) -> Self::Result {
+        self.subscribers.push(msg.0)
+    }
+}
+
+/// Refresh the list of subscriptions
+///
+/// It tells the [`Notifications`](Notifications) to replace its list of recipients by the received
+/// one.
+struct Refresh(Vec<Option<Recipient<Notify>>>);
```
aesedepece: What does `None` mean for this `Option`?

wallet/src/actors/storage/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+mod error;
+mod handlers;
+
+pub use error::Error;
+pub use handlers::*;
+
+/// Storage actor.
+pub struct Storage {
+    /// Path where to store the wallet database.
+    /// This is used during the wallet genration step.
```
aesedepece:
```suggestion
    /// This is used during the wallet generation step.
```

wallet/src/actors/storage/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+    // TODO: remove this allow attribute.
```
aesedepece: Does this still hold?

wallet/src/actors/storage/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+    #[allow(dead_code)]
+    wallet: Option<Result<rocksdb::DB, rocksdb::Error>>,
+}
+
+impl Storage {
+    /// Create a new instance of the Storage actor.
+    ///
+    /// It receives a path where to store:
+    ///
+    /// 1. The wallets database, which contains metadata about all crated wallets.
```
aesedepece:
```suggestion
    /// 1. The wallets database, which contains metadata about all created wallets.
```

net/src/server/ws/actors/notifications.rs
anler: It's kind of a __hole__ meaning that there was a recipient that is no longer reachable and should be ignored.

net/src/server/ws/mod.rs
```diff
@@ -0,0 +1,93 @@
...
+    ///
+    /// Accepts a factory function that will be used to create the websocket request-handlers. The
+    /// function receives a notification function that can be used by the handlers to send notifications
+    /// to all opened sockets. This function will block the current thread until the server is stopped.
+    /// The server can be configured with a builder-like pattern.
+    pub fn new(factory: F) -> Self {
+        let addr = net::SocketAddr::V4(net::SocketAddrV4::new(
+            net::Ipv4Addr::new(127, 0, 0, 1),
+            3200,
+        ));
```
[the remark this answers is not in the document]
anler: Yes, that address is the default one but we configure it through the `address()` method [here](https://github.com/witnet/witnet-rust/blob/158ac8ebaa2d4513467bb3db23b4b932b88bd657/wallet/src/lib.rs#L143)

wallet/src/actors/storage/mod.rs
anler: Yes, I will improve the description though

net/src/server/ws/actors/notifications.rs
aesedepece: Does it make sense to keep the hole in place? How sparse can the vector become over time?

tmpolaczyk:
The node must start before the wallet, otherwise I get a connection error.

PR #653, opened 2019-05-17
Segregate transaction types in block

### LINKED ISSUE #595: Segregate transactions in blocks

### LINKED ISSUE #632: Separate commits and reveals from `TransactionsPool`
Currently when mining a new block all the transactions are retrieved from the `TransactionsPool`. This forces us to add extra validations when dealing with commits and reveals.

It would be nice to separate commits and reveals from the `TransactionsPool`, maybe into the `DataRequestPool`, or a structure closely related to it. This `DataRequestTransactionsPool` (temporary name) should serve as a helper structure, and should not be persisted with the chain state (just like the `TransactionsPool`). Data request transactions can still be stored in the `TransactionsPool` because their validation rules are the same as value transfer transactions

`insert_commit`:
When a commit is received from the network, aside from the usual validations, we also should check that:

* The pointed data request is valid and in commit stage

`insert_reveal`:
When a reveal is received from the network, we should check that:

* The pointed data request is valid and in reveal stage
* There exists an equivalent commit hash from a commitment in a consolidated block

All of the above validations could be done on insertion (when the transaction is first received from the network), or on get (when we discovered eligibility to mine a block and need all the commits and reveals). Since the validations are relatively cheap, "on insertion" seems a better strategy. This makes the `get` methods simpler to implement:

`get_commits`:

* For each data request in commit stage, check that there are enough commits (if there are more commits that expected, randomly choose some of them), and return all the valid commit transactions.

`get_reveals`:

* Since reveals have been validated at this point, the only problem is missing reveals. What is the minimum number of reveals for a data request to be considered completed? If the minimum is one, we can just return all the known reveals.

The `get_commits` and `get_reveals` methods should not mutate the pool state, instead the pool should be updated on block consolidation. It should:
* Drop all the commits (because commits expire after one epoch)
* Drop reveals from data requests that are no longer in reveal stage. In other words, drop reveals for data requests which have reveals in the consolidated block.

lrubiorod:
Modified TransactionsPool to manage different types of Transaction
Modified Block to segregate different types of Transaction
Created BlockTransactions and BlockMerkleTrees structs
Modified validations, mining and consolidating processes to apply these changes
Updated protobuf schema

Close #595 
Close #632


validations/src/validations.rs
```diff
@@ -645,12 +626,34 @@ where
...
-    merkle_tree == merkle_tree_root(transactions)
+    // Compute `hash_merkle_root` and build block header
+    let mint_hash_merkle_root = merkle_tree_root(&[block.txns.mint.clone()]);
+    let vt_hash_merkle_root = merkle_tree_root(&block.txns.value_transfer_txns);
+    let dr_hash_merkle_root = merkle_tree_root(&block.txns.data_request_txns);
+    let commit_hash_merkle_root = merkle_tree_root(&block.txns.commit_txns);
+    let reveal_hash_merkle_root = merkle_tree_root(&block.txns.reveal_txns);
+    let tally_hash_merkle_root = merkle_tree_root(&block.txns.tally_txns);
+
+    (mint_hash_merkle_root == block.block_header.merkle_roots.mint_hash_merkle_root)
```
tmpolaczyk: You could create a new `BlockMerkleRoots` and just compare it with `block.block_header.merkle_roots`

validations/src/validations.rs
```diff
@@ -416,161 +355,203 @@ pub fn validate_transaction_signature(
...
     // Validate Merkle Root
-    let Hash::SHA256(mr) = block.block_header.hash_merkle_root;
-    if mt.root() != Sha256(mr) {
+    let Hash::SHA256(vt_mr) = block.block_header.merkle_roots.vt_hash_merkle_root;
+    let Hash::SHA256(dr_mr) = block.block_header.merkle_roots.dr_hash_merkle_root;
+    let Hash::SHA256(co_mr) = block.block_header.merkle_roots.commit_hash_merkle_root;
+    let Hash::SHA256(re_mr) = block.block_header.merkle_roots.reveal_hash_merkle_root;
+    let Hash::SHA256(ta_mr) = block.block_header.merkle_roots.tally_hash_merkle_root;
+
+    if (mint_hash_merkle_root != block.block_header.merkle_roots.mint_hash_merkle_root)
```
tmpolaczyk: You could create a new `BlockMerkleRoots` and just compare it with `block.block_header.merkle_roots`

data_structures/src/chain.rs
```diff
@@ -130,7 +130,44 @@ pub struct Block {
...
-    pub txns: Vec<Transaction>,
+    pub txns: BlockTransactions,
+}
+/// Block transactions
+#[derive(Debug, Default, Eq, PartialEq, Clone, Serialize, Deserialize, ProtobufConvert)]
+#[protobuf_convert(pb = "witnet::Block_BlockTransactions")]
+pub struct BlockTransactions {
+    /// Mint transaction,
+    pub mint: MintTransaction,
+    /// A non-empty list of signed value transfer transactions
```
```suggestion
    /// A list of signed value transfer transactions
```

data_structures/src/chain.rs
```diff
@@ -576,7 +634,10 @@ impl TransactionsPool {
     /// ```
     pub fn new() -> Self {
         TransactionsPool {
```
```suggestion
        TransactionsPool::default()
```

data_structures/src/chain.rs
```diff
@@ -553,16 +607,20 @@ pub struct RADDeliver {
...
 /// transactions sorted from by transactions with bigger fees to
 /// transactions with smaller fees.
 #[derive(Debug, Default, Clone)]
 pub struct TransactionsPool {
-    transactions: HashMap<Hash, WeightedTransaction>,
+    vt_transactions: HashMap<Hash, WeightedVTTransaction>,
     sorted_index: BTreeSet<WeightedHash>,
+    // Currently transactions related with data requests don't use weight
+    dr_transactions: HashMap<Hash, DRTransaction>,
+    co_transactions: HashMap<Hash, HashMap<Hash, CommitTransaction>>,
```
tmpolaczyk: Add comment:

```rust
// A map of `data_request_hash` to a map of `commit_hash` to `commit_transaction`
```

data_structures/src/chain.rs
```diff
@@ -639,64 +706,180 @@ impl TransactionsPool {
...
+            .co_transactions
+            .iter_mut()
+            .map(|(dr_pointer, commits)| {
+                let mut commits_vec = vec![];
+
+                if let Some(dr_output) = dr_pool.get_dr_output(&dr_pointer) {
+                    let n_commits = dr_output.witnesses as usize;
+
+                    if commits.len() >= n_commits {
+                        for (i, (_h, c)) in commits.drain().enumerate() {
```
```suggestion
                        for (_h, c) in commits.drain().take(n_commits) {
```

tmpolaczyk: Or even better, something like:

```suggestion
let commits_vec = commits.drain().take(n_commits)).collect();
```

data_structures/src/data_request.rs
```diff
@@ -498,19 +495,18 @@ mod tests {
     }
 
     fn from_tally_to_storage(
-        epoch: u32,
+        _epoch: u32,
```
tmpolaczyk: Remove this parameter if it is not used anymore
```suggestion
```

data_structures/src/chain.rs
```diff
@@ -163,8 +200,25 @@ pub struct BlockHeader {
...
-    pub hash_merkle_root: Hash,
+    /// A 256-bit hashes based on all of the transactions committed to this block
+    pub merkle_roots: BlockMerkleRoots,
+}
+/// Block merkle tree roots
+#[derive(Debug, Eq, PartialEq, Clone, Serialize, Deserialize, ProtobufConvert, Default)]
+#[protobuf_convert(pb = "witnet::Block_BlockHeader_BlockMerkleRoots")]
+pub struct BlockMerkleRoots {
+    /// A 256-bit hash based on the mint transaction committed to this block
+    pub mint_hash_merkle_root: Hash,
```
aesedepece: This is interesting, but I'm undecided if strictly necessary. In any case, it's not a merkle root but a plain hash, so we should name it accordingly :wink:

PR #660, opened 2019-05-23
feat(wallet): define jsonrpc-errors

anler:
# Custom Error Codes

This idea is based in [Ethereum's JSON-RPC implementation](https://github.com/ethereum/wiki/wiki/JSON-RPC-Error-Codes-Improvement-Proposal)

| Code | Error           | Description                                                                  |
| -----| ------          | ------                                                                       |
| 1    | Execution error | There was an internal error. The data field will contain additional details. |
| 2    | Node error      | There was an error when communicating with the node.                         |

## Execution Error

The execution error will contain additional error fields in the `data` attribute to further explain
what went wrong. So far there's only one catch-all case defined:

| Code      | Error         | Description     |
| --------- | ------        | ------          |
| 100       | Generic Error | There an error. |

The idea is to augment this list with the error codes we find useful.


wallet/src/error.rs
```diff
@@ -26,3 +26,11 @@ impl fmt::Display for Error {
         }
     }
 }
+
+impl Into<Value> for Error {
+    fn into(self) -> Value {
+        match self {
+            err => json!({ "code": 100, "message": format!("{}", err) }),
```
tmpolaczyk:
```suggestion
            err => json!({ "code": self.code(), "message": format!("{}", err) }),
```

Instead of hardcoding the error codes as constants in a separate file, I would just write a error to code conversion function with a giant match of all the possible errors:

```rust
impl Error {
    fn code(&self) -> i64 {
        match self {
            Error::Execution(..) => 1
            _ => 100
        }
    }
}
```

PR #661, opened 2019-05-24
fix(mining): keep mining a block after tally errors

### LINKED ISSUE #631: Handle errors in tally mining
Currently, if any consensus script fails, mining fails, because we use join_all to collect all the futures and join_all stops on the first error.
This can be quickly fixed by mapping Ok(T) => Ok(Some(T)) and Err(_) => Ok(None), and then applying a flat_map.

When a consensus script fails, that error should be reported as the result of the data request.
Otherwise, when the error is not a radon error (for example an actix mailbox error), mining should continue without including this tally.

tmpolaczyk:
Fix #631

### DIFF
--- a/node/src/actors/chain_manager/mining.rs
+++ b/node/src/actors/chain_manager/mining.rs
@@ -291,70 +291,92 @@ impl ChainManager {
     ) -> impl Future<Item = Vec<TallyTransaction>, Error = ()> {
         let data_request_pool = &self.chain_state.data_request_pool;
 
-        // Include Tally transactions, one for each data request in tally stage
-        let mut future_tally_transactions = vec![];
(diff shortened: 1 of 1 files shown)


node/src/actors/chain_manager/mining.rs
```diff
@@ -291,70 +291,86 @@ impl ChainManager {
...
+                                ),
+                            );
 
-                    futures::future::ok(tally)
+                            futures::future::ok(tally)
+                        })
+                        // This future should always return Ok because join_all short-circuits on the
+                        // first Err, and we want to keep creating tallies after the first error
+                        // Map Result<T, E> to Result<Option<T>, ()>
+                        .then(|x| futures::future::ok(x.ok()))
```
lrubiorod: Specify in the comments the future RADON error handling

PR #664, opened 2019-06-02
Subscribe Wallet to Node updates

anler:
# Main changes

This PR is divided into three major changes:
1. Migration to `jsonrpc_ws_server` and websockets client implementaiton in `witnet_net`
2. Subscription to node updates from the wallet
3. Subscription to wallet updates from the client

## 1. Migration to `jsonrpc_ws_server` and websockets client implementaiton in `witnet_net`

Integrating `jsonrpc_core` crate and the Websockets server from actix-web was kind of hard, so we backed-up to use `jsonrpc_ws_server` crate.

`witnet_net::client::tcp::jsonrpc` has been added as a reusable client for JsonRPC over TCP.

## 2.  Subscription to node updates from the wallet

When the wallet starts, it will try to subscribe to new-blocks notifications from the node. If the node isn't running, it will try in the next request involving the node. Currently, the notifications received from the node are sent to the client as they are, this will change in the future though.

## 3. Subscription to wallet updates from the client

The wallet keeps a list of maximum 10 subscribers for real time updates. Whenever the wallet receives updates from the node, it will notify the clients about those updates. In order to subscribe to wallet updates, the client needs to issue the following request:

```json
{"jsonrpc":"2.0","method":"subscribeNotifications" ,"id":"1"}
```

and, in order to unsubscribe:

```json
{"jsonrpc":"2.0","method":"unsubscribeNotifications" ,"id":"1"}
```

# Other changes

The wallet will handle `SIGINT` and `SIGTERM` signals and perform a graceful shutdown.

# How to test

In order to test the whole thing:
1. Run a node with mining enable
2. Run the wallet with the appropiate node url: `cargo run -p witnet_wallet -- run --node-url '...'`
3. Connect to the wallet websockets server (by default it runs at `ws://127.0.0.1:11212`
4. Send the subscribe request
5. Wait for new blocks to be notified


aesedepece:
I'm trying your example and it looks to me like it's always lagging 1 epoch behind :open_mouth:

anler:
@aesedepece I have added logs to trace the received-notification's checkpoint:

image

 I have correlated those logs with the node logs and it seems it's working correctly :thinking:

PR #671, opened 2019-06-03
Testnet 3 release candidate

aesedepece:
- updates the IP of the bootstrap node
- upgrade consensus constants and their documentation
- bumps the versions of all the components from `0.2.x` to `0.3.0`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1404,7 +1404,7 @@ dependencies = [
 
 [[package]]
 name = "partial_struct"
-version = "0.2.1"
+version = "0.3.0"
 dependencies = [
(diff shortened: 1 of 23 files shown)


anler:
@aesedepece I was thinking, is it really necessary to update all sub-crates to the same version? One alternative would be to leave any sub-crate to evolve separately as [actix-web](https://github.com/actix/actix-web) does

aesedepece:
@anler don't really have a strong opinion on that. Maybe we can continue like this until `1.0.0` (mainnet), then make a decision. It doesn't really matter that much, anyway, as none of the sub-crates is published on crates.io and they are loaded locally.

PR #677, opened 2019-06-05
Handle bugs produced by empty epochs

### LINKED ISSUE #674: Handle epochs with no block
To ensure correct behavior when an epoch has not consolidated a block, we must:

- [ ] Update `ActiveReputationSet` with an empty vector.
- [ ] Clear commits from `TransactionsPool`.

### LINKED ISSUE #675: Add change in TallyTransaction in case of less reveals than expected
Exist the calculate_tally_change function in data_request.rs to calculate if there are change for the data request creator but it is not used in mining

lrubiorod:
Modified log level of synced state to info
Skipped clippy cognitive_complexity warnings
Updated commits every epoch and reveals every consolidated block
Handled ARS queue in epochs without blocks
Added tally change in tally mining

Close #674 
Close #675

### DIFF
--- a/node/src/actors/chain_manager/handlers.rs
+++ b/node/src/actors/chain_manager/handlers.rs
@@ -1,5 +1,5 @@
 use actix::{fut::WrapFuture, prelude::*};
-use log::{debug, error, warn};
+use log;
 
 use witnet_data_structures::{
     chain::{
@@ -43,7 +43,7 @@ impl Handler<EpochNotification<EpochPayload>> for ChainManager {
     type Result = ();
 
     fn handle(&mut self, msg: EpochNotification<EpochPayload>, ctx: &mut Context<Self>) {
-        debug!("Epoch notification received {:?}", msg.checkpoint);
+        log::debug!("Epoch notification received {:?}", msg.checkpoint);
 
         match self.sm_state {
             StateMachine::WaitingConsensus => {}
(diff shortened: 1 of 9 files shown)


node/src/actors/chain_manager/mod.rs
```diff
@@ -651,7 +653,7 @@ fn update_reputation(
 
     // Update active reputation set
     // Add block miner pkh to active identities
-    rep_eng.ars.push_activity(revealers.chain(vec![miner_pkh]));
+    rep_eng.update_ars(block_epoch, revealers.chain(vec![miner_pkh]));
```
tmpolaczyk: Since the `ars` is only updated here, it can lead to incorrect results. Consider the case when we are in epoch 100 and there are 40 epochs with no block and the `ars` has a maximum size of 40. The set of active identities should be empty, but it won't be updated until after processing the block for epoch 140. Therefore in the reputation validations for block 140 we will count identities which have participated in epochs 60-99 as active, when clearly they weren't active because in epochs 100-139 there were no blocks.

lrubiorod: It is also updated in ChainManager/handler.rs 138, in case of an empty epoch

node/src/actors/chain_manager/handlers.rs
```diff
@@ -24,6 +24,7 @@ use crate::{
     },
     utils::mode_consensus,
 };
+use witnet_data_structures::chain::ChainState;
```
tmpolaczyk: Badly formatted import

reputation/src/lib.rs
```diff
@@ -5,10 +5,12 @@
...
-#![deny(missing_docs)]
 
 pub mod trs;
 pub use trs::TotalReputationSet;
 
 pub mod ars;
 pub use ars::ActiveReputationSet;
+
+/// Module containing error definitions
+pub mod error;
```
tmpolaczyk: Since you are creating a error module, could you move the errors from `trs.rs` there? (`NonSortedAlpha` and `RepError`)

PR #678, opened 2019-06-05
100% validation tests

tmpolaczyk:
Close #522 

Missing validations which depend on #685:

* Commits in block: try to build a block with more commits than needed / less commits than needed / some duplicate commits.
* Reveals in block: try to build a block with more reveals than needed / less reveals than needed / some duplicate reveals.
* Tallies in block: try to insert two tallies for the same data request, try to reward the same witness more than one for one data request, try to change the witness reward.
* Block consolidation: valid blocks must not cause any errors in the consolidation process.

PR #679, opened 2019-06-06
Miscellaneous fixes

### LINKED ISSUE #650: Modify MintTransaction
Modify MintTransaction outputs field:

outputs: Vec<ValueTransferOutput> -> output: ValueTransferOutput

tmpolaczyk:
These are some changes that I made while writing the tests in #678 

- Fix #650: mint transaction should only have one output
- Fix VRF false positives
- Fix DataRequestEligibilityClaim using wrong epoch in proof message
- Fix DataRequest fee calculation
- Add validation for VRF Proof PKH == signature PKH


validations/src/validations.rs
```diff
@@ -566,17 +575,13 @@ pub fn validate_block_transactions(
     let ta_hash_merkle_root = ta_mt.root();
 
     // Validate mint
-    validate_mint_transaction(
-        &block.txns.mint,
-        total_fee,
-        block_reward(block.block_header.beacon.checkpoint),
-    )?;
+    validate_mint_transaction(&block.txns.mint, total_fee, beacon.checkpoint)?;
```
lrubiorod: change variable name beacon for block_beacon

data_structures/src/error.rs
```diff
@@ -60,7 +60,7 @@ pub enum TransactionError {
         dr_value, witnesses
     )]
     InvalidDataRequestValue { dr_value: i64, witnesses: i64 },
-    #[fail(display = "Data Request witnesses number is not enough")]
+    #[fail(display = "There needs to be at least 1 witness in a data request")]
```
lrubiorod: It is not defined that 1 witness is the minimum

PR #683, opened 2019-06-07
fix(node): tag some TODOs to created issues

lrubiorod:


node/src/actors/chain_manager/mining.rs
```diff
@@ -72,18 +72,14 @@ impl ChainManager {
             _ => return,
         };
 
-        if beacon.checkpoint > current_epoch {
+        if beacon.checkpoint >= current_epoch {
```
aesedepece: Undocumented edit. What are you trying to conceal? :smirk:

tmpolaczyk: He's trying to introduce a backdoor :scream:

lrubiorod: Chain checkpoint beacon is the checkpoint beacon of the last consolidated block. Due to consolidate a block from epoch N is done in epoch N+1, our chain beacon always come from the past epoch. A chain beacon with the current epoch is the same error if it is come from the future

tmpolaczyk: Then put this change in a separate commit :P

node/src/actors/chain_manager/mining.rs
```diff
@@ -72,18 +72,14 @@ impl ChainManager {
...
             // We got a block from the future
             error!(
                 "The current highest checkpoint beacon is from the future ({:?} > {:?})",
```
tmpolaczyk:
```suggestion
                "The current highest checkpoint beacon is from the future ({:?} >= {:?})",
```

And add a comment explaining why >= is the same as >

PR #690, opened 2019-06-11
Create wallet

### LINKED ISSUE #615: Implement a method in the wallet api to create a wallet

anler:
# What

Implement the *createWallet* API endpoint. This endpoint will:

* Generate a Master HD-Wallet key from the given seed source
* Store the wallet data encrypted with a key derived from the given password using `pbkdf2`

### Request

```json
{
  "id": "<request id>",
  "params": {
    "seedSource": {
      "data": "<mnemonics value> | <xprv value>",
      "source": "<mnemonics> | <xprv>"
    },
    "password": "<password>",
    "caption": "<description for the wallet>"
  },
  "method": "createWallet",
  "jsonrpc": "2.0"
}
```

### Success Response

A wallet id looks like this: `8d5a9fbb4ba3b84f3ec4e27f65b3fc442ec835ad91a1280997617b3030b2c396`.

```json
{
  "id": "<requet id>",
  "result": {
    "walletId": "<wallet id>"
  },
  "jsonrpc": "2.0"
}
```

## Other changes

* Created a `crypto` actor that performs the generation of the keys
* Created a `storage` module out of the `actors` module, with generic, actor-independent code
* Use RocksDB's `WriteBatch` to perform writes atomically
* Use RocksDB's [write operator](https://github.com/facebook/rocksdb/wiki/Merge-Operator) to efficiently update the list of existing wallet ids
* Add `serde` feature flag to `crypto` crate to add serialize/deserialize support to keys
* Implement new type `ProtectedString` in `protected` crate, this is similar to `Protected` type but its `as_ref()` method returns a `&str`
* Refactor `crypto` and `node` crates to use `ProtectedString` type for password
* Add support for a `[log]` section in the config file to control application log-level
* On startup, show the log level configured and from where (config, env variable, flag) read that value

## Screenshot

image

Closes #615


protected/src/serde.rs
```diff
@@ -0,0 +1,39 @@
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
```
tmpolaczyk: There is no need to implement Serialize and Deserialize manually, you can add a conditional attribute to the struct definition which will serialize the struct by converting it to another type and serializing that:

```rust
#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
#[cfg_attr(feature = "serde", serde(from = "Vec<u8>", into = "Vec<u8>"))]
pub struct Protected(Memzero<Vec<u8>>); 
```

And that will give you the wanted behaviour, as long as `Protected` implements `From<Vec<u8>>`, `Clone` and `Into<Vec<u8>>`.

wallet/src/actors/storage/mod.rs
```diff
@@ -2,64 +2,97 @@
...
-        let wallets = rocksdb::DB::open(&options, db_path.join("wallets"));
+    pub fn new(db: Arc<rocksdb::DB>) -> Self {
+        Self { db }
+    }
 
-        Self {
-            db_path,
-            wallets,
-            wallet: None,
+    pub fn get_wallet_infos(&self) -> Result<Vec<wallet::WalletInfo>, error::Error> {
```
tmpolaczyk: What will happen when we have more than one key in storage? What about a generic private getter and public getters:

```rust
    pub fn get_wallet_infos(&self) -> Result<Vec<wallet::WalletInfo>, error::Error> { 
        self.get("wallet-infos")
    }
```

```rust
    fn get<V>(&self, key: &str) -> Result<V, error::Error> {
        let result = self
            .db
            .get(key)
            .map_err(error::Error::DbGetFailed)?;

        match result {
            Some(db_vec) => {
                let value =
                    deserialize(db_vec.as_ref()).map_err(error::Error::DeserializeFailed)?;
                Ok(value)
            }
            None => Ok(Vec::new()),
        }
    }
```

wallet/src/api/routes.rs
```diff
@@ -1,12 +1,32 @@
 use actix::prelude::*;
+use failure::Fail;
 use futures::{future, Future};
-use jsonrpc_core::{Middleware, Params};
+use jsonrpc_core::{self as rpc, Middleware, Params};
 use jsonrpc_pubsub::{PubSubHandler, PubSubMetadata};
-use serde_json::{json, to_value};
+use serde_json::{self as json, json, to_value};
```
tmpolaczyk: Please avoid ambiguous module renames.
```suggestion
use serde_json::{json, to_value};
```

witnet.toml
```diff
@@ -11,3 +11,6 @@ db_path = ".witnet-rust-testnet-3"
 
 [consensus_constants]
 checkpoint_zero_timestamp = 1559347200
+
+[log]
+level = "debug"
```
tmpolaczyk: I think the default should be "info"

wallet/src/actors/storage/mod.rs
anler: I had it generic in the beginning but since depending on the key we will use encryption or not, I decided to strive for a simpler to understand solution like this. Also the internal type `DB` is already generic, so I think we would benefit more of a [Repository Pattern](https://martinfowler.com/eaaCatalog/repository.html) here

witnet.toml
anler: This is kind of the defaults for development, `bootstrap_peers_period_seconds` is set to `15` for example, the defaults in the crate config use `info`

wallet/src/api/routes.rs
anler: Let's discuss this in another channel

protected/src/serde.rs
anler: I think I won't address this, because it requires [`Into<Vec<u8>> for Protected`](https://serde.rs/container-attrs.html#into) and that's something I'm trying to avoid since I want to keep ownership of those bytes to later zero-them. I mean, I actually removed the `ProtectedString::to_string`, so if the user really wants to *unprotect* those bytes it has to call `protected.as_ref().to_vec()` or `protected_str.as_ref().to_string()`.

wallet/src/api/routes.rs
anler: Ok, you won :laughing:

wallet/src/actors/storage/mod.rs
anler: I now understand what you meant, I added a `wallet/src/storage` module with all the generic operations.

tmpolaczyk:
I get an error after restarting the wallet:

```
> {"jsonrpc": "2.0", "id": 1, "method": "getWalletInfos"}
< {"jsonrpc":"2.0","result":{"infos":[{"caption":"<description for the wallet>","id":"6c344625884c2f910065ab170dc18ad3cbbc03c7234507c7c22dbd78e3b26667"},{"caption":"<description for the wallet 2>","id":"bb6fddfe0f23d16cee84bd99fcae587edf556c67063392c2e1ee6c57d132af30"}]},"id":1}
disconnected (code: 1006)
$ wscat --connect 127.0.0.1:11212
connected (press CTRL+C to quit)
> {"jsonrpc": "2.0", "id": 1, "method": "getWalletInfos"}
< {"jsonrpc":"2.0","error":{"code":1,"message":"Execution Error.","data":"db key not found"},"id":1}
> { "id": "<request id>", "params": { "seedSource": { "data": "exotic demand way fatigue skull poverty happy divide scrub seed jeans novel", "source": "mnemonics" }, "password": "1", "caption": "<description for the wallet>" }, "method": "createWallet", "jsonrpc": "2.0" }
< {"jsonrpc":"2.0","result":{"walletId":"6c344625884c2f910065ab170dc18ad3cbbc03c7234507c7c22dbd78e3b26667"},"id":"<request id>"}
> {"jsonrpc": "2.0", "id": 1, "method": "getWalletInfos"}
< {"jsonrpc":"2.0","error":{"code":1,"message":"Execution Error.","data":"db key not found"},"id":1}
```

PR #691, opened 2019-06-12
refactor(node): run storage in sync-arbiter thread

anler:
Refactor storage manager to run the actor (which is blocking) behind a `SyncArbiter`.


aesedepece:
Awesome. I'll test this on the raspi later today.

PR #692, opened 2019-06-12
refactor(cli): rename run command to server

anler:
Rename wallet and node `run` commands to `server`, but keep `run` as an alias, so this change shouldn't break anything.

### DIFF
--- a/src/cli/node/with_node.rs
+++ b/src/cli/node/with_node.rs
@@ -59,7 +59,7 @@ pub fn exec_cmd(command: Command, mut config: Config) -> Result<(), failure::Err
 
 #[derive(Debug, StructOpt)]
 pub enum Command {
-    #[structopt(name = "run", about = "Run the Witnet server.")]
+    #[structopt(name = "server", about = "Run a Witnet node server.", alias = "run")]
     Run(ConfigParams),
     #[structopt(
         name = "raw",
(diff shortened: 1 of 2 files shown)


src/cli/node/with_node.rs
```diff
@@ -59,7 +59,7 @@ pub fn exec_cmd(command: Command, mut config: Config) -> Result<(), failure::Err
 
 #[derive(Debug, StructOpt)]
 pub enum Command {
-    #[structopt(name = "run", about = "Run the Witnet server.")]
+    #[structopt(name = "server", about = "Run the Witnet server.", alias = "run")]
```
aesedepece:
```suggestion
    #[structopt(name = "server", about = "Run a Witnet node server.", alias = "run")]
```

src/cli/wallet/with_wallet.rs
```diff
@@ -29,7 +29,11 @@ pub fn exec_cmd(command: Command, mut config: Config) -> Result<(), failure::Err
 
 #[derive(Debug, StructOpt)]
 pub enum Command {
-    #[structopt(name = "run", about = "Run the wallet websockets server.")]
+    #[structopt(
+        name = "server",
+        about = "Run the wallet websockets server.",
```
aesedepece: How about this?
```suggestion
        about = "Run a wallet server exposing a websockets API.",
```

aesedepece:
Merged as 8870b4196b5ace2588cffcdabd7f9d6316c51899.

PR #694, opened 2019-06-12
fix(node): fix bug in reputation synchronization

lrubiorod:
Fixed a bug produced by updating active reputation system before than expected

### DIFF
--- a/node/src/actors/chain_manager/handlers.rs
+++ b/node/src/actors/chain_manager/handlers.rs
@@ -97,12 +97,6 @@ impl Handler<EpochNotification<EveryEpochPayload>> for ChainManager {
                     reputation_engine: Some(ref mut rep_engine),
                     ..
                 } => {
-                    // Update reputation in possible empty epochs
-                    let expected_ars_epoch = current_epoch - 1;
-                    if let Err(e) = rep_engine.ars.update_empty(expected_ars_epoch) {
-                        log::error!("Error updating empty reputation: {}", e);
-                    }
-
                     // Decide the best candidate
                     // TODO: replace for loop with a try_fold
                     let mut chosen_candidate = None;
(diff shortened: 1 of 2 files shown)


validations/src/validations.rs
```diff
@@ -724,7 +724,7 @@ pub fn validate_block(
     utxo_set: &UnspentOutputsPool,
     data_request_pool: &DataRequestPool,
     vrf: &mut VrfCtx,
-    rep_eng: &ReputationEngine,
```
tmpolaczyk: The `validate_block` function should not mutate anything, we need to find another place to update the reputation engine.

lrubiorod: The other option could be to put poe validation out of validate_block function, and update reputation engine before to call it

tmpolaczyk: (1/2) But why is this update needed? Isn't enough to update the reputation engine...

node/src/actors/chain_manager/handlers.rs
```diff
@@ -135,13 +129,14 @@ impl Handler<EpochNotification<EveryEpochPayload>> for ChainManager {
...
+                        let previous_epoch = msg.checkpoint - 1;
                         log::warn!(
                             "There was no valid block candidate to consolidate for epoch {}",
-                            msg.checkpoint - 1
+                            previous_epoch
                         );
 
                         // Update ActiveReputationSet in case of epochs without blocks
-                        if let Err(e) = rep_engine.ars.update(vec![], expected_ars_epoch) {
+                        if let Err(e) = rep_engine.ars.update(vec![], previous_epoch) {
```
tmpolaczyk: (2/2) ...here

lrubiorod: the problem are empty epochs while we are consolidating old blocks. For example:
We have a very few probability to mine, and there are only one miner now. After some empty epochs, the reputation engine is updated and now that miner has 100% of the probability and mine a block. If later other node has to synchronize that chain, if we dont update the reputation engine before to check the poe, this node keeps the few probability until it consolidate a new block, so the PoE that the other node show (f555... for example) is not valid for it, and reject a valid block thinking that it is not

tmpolaczyk: So it is a problem while synchronizing? Then we can handle it in `AddBlocks`

lrubiorod: :thinking: good point, I am going to try

PR #698, opened 2019-06-16
Unlock wallet

anler:
# What

Implement the *unlockWallet* API endpoint. This enpoint will:

* Use the provided password to derive the key used to encrypt/decrypt the stored password
* Use the key to read the wallet contents and keep it open for the current session

### Request

The `seesionId` param is used to associate the unlocked wallet with the given session. After that, every time the client wants to interact with this wallet it will have to specify its `sessionId` as well as the `walletId`.

```json
{
  "id": "<request id>",
  "params": {
    "password": "<wallet password>",
    "sessionId": "<session id>",
    "walletId": "<wallet id>"
  },
  "method": "unlockWallet",
  "jsonrpc": "2.0"
}
```

### Success Response

If the unlock wallet succeeds it will send an empty ok-response.

```json
{
  "id": "1",
  "result": {
    "unlockedWalletId":"6c344625884c2f910065ab170dc18ad3cbbc03c7234507c7c22dbd78e3b26667"
  },
  "jsonrpc": "2.0"
}
```

## Example: Unlock a wallet used by another session

If the wallet has been unlocked by another session, trying to unlock it again will append the session id to the internal list of sessions associated to an unlocked wallet.

### DIFF
--- a/wallet/src/actors/app/mod.rs
+++ b/wallet/src/actors/app/mod.rs
@@ -8,7 +8,7 @@ use failure::Error;
 use futures::future;
 use jsonrpc_core as rpc;
 use jsonrpc_pubsub as pubsub;
-use serde_json::{self as json, json};
+use serde_json::json;
 
 use witnet_net::client::tcp::{jsonrpc as rpc_client, JsonRpcClient};
 use witnet_protected::ProtectedString;
@@ -82,7 +82,7 @@ impl App {
         &mut self,
         method: String,
         params: rpc::Params,
-    ) -> ResponseFuture<json::Value, Error> {
+    ) -> ResponseFuture<serde_json::Value, Error> {
         match &self.node_client {
             Some(addr) => {
                 let req = rpc_client::Request::method(method)
--- a/wallet/src/actors/app/mod.rs
+++ b/wallet/src/actors/app/mod.rs
@@ -1,6 +1,7 @@
 //! # Application actor.
 //!
 //! See [`App`](App) actor for more information.
+use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 
 use actix::prelude::*;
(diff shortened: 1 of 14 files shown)


wallet/src/storage/error.rs
```diff
@@ -0,0 +1,25 @@
...
+    #[fail(display = "couldn't open database file")]
+    OpenDbFailed(#[cause] rocksdb::Error),
+    #[fail(display = "database operation failed: {}", _0)]
+    DbOpFailed(#[cause] rocksdb::Error),
+    #[fail(display = "db key not found")]
+    DbKeyNotFound,
+    #[fail(display = "cipher operation failed: {}", _0)]
+    CipherOpFailed(#[cause] cipher::Error),
+    #[fail(display = "could not find a wallet with the given id")]
+    UnknonwWalletId,
```
tmpolaczyk:
```suggestion
    UnknownWalletId,
```

wallet/src/actors/app/error.rs
```diff
@@ -24,4 +24,6 @@ pub enum Error {
     StorageCommFailed(#[cause] actix::MailboxError),
     #[fail(display = "could not communicate with cryptographic engine")]
     CryptoCommFailed(#[cause] actix::MailboxError),
+    #[fail(display = "the wallet is already bein used by another session")]
```
tmpolaczyk:
```suggestion
    #[fail(display = "the wallet is already being used by another session")]
```

wallet/src/actors/app/mod.rs
```diff
@@ -34,13 +35,42 @@ pub struct App {
...
+    unlocked_wallets: HashMap<wallet::WalletId, wallet::SessionId>,
 }
 
+// let result = if self.opened_wallets.borrow().iter().any(|id_| id_ == id) {
+//     Err(storage::Error::WalletAlreadyOpenend(id.to_string()))
+// } else {
+
+// };
+
+// result
```
tmpolaczyk: I guess this comment was meant to be removed?

tmpolaczyk:
When a session disconnects with an unlocked wallet, that wallet becomes unaccessible. I guess this will be fixed when a `lockWallet` method is implemented?

tmpolaczyk:
I think the design would be better if the `forceOwnership` flag didn't exist. Wallets should lock when a session disconnects.

anler:
I'm going to remove it and let a wallet be used by multiple clients. Detecting session disconnects is not possible currently since JsonRpc is a stateless protocol, also, we want to be able to keep the wallets opened so the can keep up to date with the network.

PR #718, opened 2019-06-20
Proofs of inclusion for transactions in blocks

### LINKED ISSUE #708: Proof-of-inclusion generation function
TBFE

tmpolaczyk:
Close #708 

In order to build a proof of inclusion, we need the entire merkle tree, not just the root. This PR adds a new merkle tree implementation which stores all the intermediate nodes.

Missing: 
* improve documentation

### DIFF
--- a/crypto/src/merkle.rs
+++ b/crypto/src/merkle.rs
@@ -165,6 +165,178 @@ impl<T: Copy> ProgressiveMerkleTree<T> {
     }
 }
 
+/// Full merkle tree with all the intermediate nodes, used to generate
+/// inclusion proofs
+#[derive(Debug)]
+pub struct FullMerkleTree<T: 'static> {
+    nodes: Vec<Vec<T>>,
+    hash_concat: fn(T, T) -> T,
+    empty_hash: &'static T,
+}
+
+impl FullMerkleTree<Sha256> {
+    /// Merkle tree using SHA256 function
+    pub fn sha256(leaves: Vec<Sha256>) -> Self {
+        Self::new(leaves, sha256_concat, &EMPTY_SHA256)
+    }
+}
+
+impl<T: Copy + PartialEq> FullMerkleTree<T> {
+    /// New full merkle tree
+    pub fn new(leaves: Vec<T>, hash_concat: fn(T, T) -> T, empty_hash: &'static T) -> Self {
+        let nodes = FullMerkleTree::build(leaves, hash_concat);
+        Self {
+            nodes,
+            hash_concat,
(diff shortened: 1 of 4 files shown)


tmpolaczyk:
Initially I was going to use this crate: https://crates.io/crates/merkle-cbt, but it turns out that their merkle-tree building strategy is different from ours: we group the leaves on the left side while they scramble the leaves in order to make the tree more balanced.

So for a tree with 9 leaves, we would only need 1 hash to prove inclusion of the last leaf but 4 hashes to prove inclusion of all any other leaf, while their tree will be balanced so they need 4 hashes to prove inclusion of the last two leaves and 3 hashes to prove inclusion of the other leaves.

```
Our merkle tree with 9 nodes:

          ^
      ^     |
   ^    ^   |
 ^  ^  ^  ^ |
12,34,56,78,9
```

```
CBMT with 9 nodes:
       ^
    ^    ^
  ^  ^  ^  ^
 ^1,23,45,67
89
```

Speaking in numbers, in the case of a tree with 9 leaves, we have an average hashes per proof of `(8 * 4 + 1) / 9 == 3.66` while they have `(7 * 3 + 2 * 4) / 9 = 3.22`. Similarly, in the case of 1025 leaves:

```py
>>> (1024*11+1) / 1025
10.990243902439024
>>> (1023*10+2*11) / 1025
10.001951219512195
```

That's the worst case scenario, we basically do an extra hash, while in the best case scenario where the tree has 2^n leaves, both implementations are equivalent.

The only disadvantage to their method is that it cannot be used to build growable merkle-trees: the list of leaves must be passed at once because the layout of the tree depends on the number of leaves, so leaves cannot be added on demand without recalculating the entire tree.

Since we still are in time to change our merkle tree implementation, I think it's worth discussing pros and cons, and whether this trade-off makes given some estimated number of transactions per block.

crypto/src/merkle.rs
```diff
@@ -165,6 +165,158 @@ impl<T: Copy> ProgressiveMerkleTree<T> {
...
+        self.nodes.last().map(|x| x[0]).unwrap_or(*self.empty_hash)
+    }
+
+    /// Get the nodes
+    pub fn nodes(&self) -> &[Vec<T>] {
+        &self.nodes
+    }
+
+    /// Create inclusion proof for element at index
+    pub fn inclusion_proof(&self, index: usize) -> Option<InclusionProof<T>> {
```
lrubiorod: Please, include comments to explain the code

lrubiorod:
Include estimation in issue #708

PR #720, opened 2019-06-21
refactor(data_structures): remove pkh from DataRequestOutput and add …

### LINKED ISSUE #715: Remove PKH from data request outputs
TBFE

lrubiorod:
…it in DataRequestState

Close #715


data_structures/src/data_request.rs
```diff
@@ -359,7 +375,7 @@ mod tests {
         let dr_pointer = dr_transaction.hash();
 
         let mut p = DataRequestPool::default();
-        p.process_data_request(&dr_transaction, epoch);
+        let _a = p.process_data_request(&dr_transaction, epoch);
```
tmpolaczyk: The "unused result" warning means that the function could fail, so don't ignore that error

```suggestion
        p.process_data_request(&dr_transaction, epoch).unwrap();
```

tmpolaczyk:
Can you also remove the `pkh` field from the example data requests in `examples/`?

PR #721, opened 2019-06-24
Modify hash method for dr and tally and include them in the PoI

### LINKED ISSUE #706: Change generation of data request transaction identifiers
TBFE

### LINKED ISSUE #707: Change generation of tally transaction identifiers
TBFE

### LINKED ISSUE #709: Proof-of-inclusion method for data request transactions
For the witnet bridge, we need to prove that a data request transaction contains a certain data request output. A pretty straightforward way to do it, is to define the data request transaction hash as

    H(H(inputs_outputs) || H(dr_output))

(where || means concatenation). That is, the hash of a data request transaction is the hash of the concatenation of the hash of the inputs and outputs of the data request, and the hash of the data request output. Therefore, in order to prove that a data request transaction DRT whose hash is HDRT contains a data request output DRO, we just need to provide a hash HIO such that H(HIO || H(DRO)) == HDRT

### LINKED ISSUE #710: Proof-of-inclusion method for tally transactions
TBFE

lrubiorod:

Close #706 
Close #707 
Close #709 
Close #710 

Modified methods to create the hash of DRTransaction and TallyTransaction:
--> DRTransaction hash = hash( hash(dr_output bytes) + hash(dr_transaction bytes) )
--> TallyTransaction hash = hash( hash( hash(dr_pointer bytes) + hash(tally_result bytes) ) + hash(tally_transaction bytes) )

The idea to use the total transaction bytes instead of only the rest of fields is avoid to implement new types of messages in Protobuff that represents a vector ("repeated") of messages, so the bytes of dr_output, dr_pointer and tally_result are included twice (inside and outside transaction)

The idea to concatenate tally_result and dr_pointer hashes in TallyTransaction hash creation is to have only one level of hashes in the proof of inclusion of TallyTransaction

### DIFF
--- a/data_structures/src/transaction.rs
+++ b/data_structures/src/transaction.rs
@@ -10,10 +10,7 @@ use crate::{
 };
 use protobuf::Message;
 use std::cell::Cell;
-use witnet_crypto::{
-    hash::{calculate_sha256, Sha256},
-    merkle::FullMerkleTree,
-};
+use witnet_crypto::{hash::calculate_sha256, merkle::FullMerkleTree};
 
 pub trait MemoizedHashable {
     fn hashable_bytes(&self) -> Vec<u8>;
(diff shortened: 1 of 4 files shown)


data_structures/src/transaction.rs
```diff
@@ -321,6 +339,20 @@ impl TallyTransaction {
         }
     }
 
+    /// Specified data to be divided in a new level in the proof of inclusion
+    /// In this case data = Hash( Hash(dr_pointer) + Hash(tally) )
```
tmpolaczyk:
```suggestion
    /// In this case data = Hash( dr_pointer || tally )
```

PR #723, opened 2019-06-27
Improve API error messages

anler:
# What

This PR improves (and fixes in a lot of cases) the error messages returned by the API.

## Better `Invalid Params` error

If the params of the request aren't valid, now along the response we send a `data` attribute containing a link to the wiki describing the schema for those params, e.g.:

```json
{
  "code": -32602,
  "message": "Invalid params: invalid type: null, expected struct CreateMnemonicsRequest.",
  "data": {
    "schema": "https://github.com/witnet/witnet-rust/wiki/Create-Mnemonics"
  }
}
```

## Better validation errors

Proper validation errors has been implemented for requests that are syntactically correct. For example, in the `createMnemonics` method, the client must specify how many words the sentence should contain in the `length` field, but if it's not one of the valid values, a concise response is sent back indicating about the error:

### Request

```json
{
  "jsonrpc": "2.0",
  "id": "1",
  "method": "createMnemonics",
  "params": {
    "length": 9
  }
}
```

### Response

```json
{
  "jsonrpc": "2.0",
  "error": {
    "code": 400,
    "message": "Validation Error",
    "data": [
      [
        "length",
        "Invalid Mnemonics Length. Must be 12, 15, 18, 21 or 24"
      ]
    ]
  },
  "id": "1"
}
```
**Note**: The *data* attribute contains a list of tuples of the form: `[<request field>, <error msg>]`. A list of tuples is used because is easier to generate generically from Rust code.

## Other improvements

- Now a `[rocksdb]` section can be added to the config and set some rocksdb-related settings like: **increase_parallelism**, **create_if_missing**, **compaction_readahead_size**, **use_fsync**, and **enable_statistics**.
- The `[wallet]` section of the config has been extended with the following settings: **db_path**, **db_file_name**, **db_encrypt_hash_iterations**, **db_encrypt_iv_length**, **db_encrypt_salt_length**, **seed_password**, **master_key_salt**, **id_hash_iterations**, and **id_hash_function**.
- Reorganization of the code in general
- Created a [wiki](https://github.com/witnet/witnet-rust/wiki/Wallet) documenting the API of the wallet
- Refactored the [Unlock Wallet](https://github.com/witnet/witnet-rust/wiki/Unlock-Wallet) API to support  sessions that expire


wallet/src/actors/crypto/mod.rs
```diff
@@ -73,6 +78,24 @@ impl Crypto {
...
+                let id_bytes = pbkdf2_sha256(
+                    password.as_ref(),
+                    self.master_key_salt.as_ref(),
+                    self.id_hash_iterations,
+                );
+
+                hex::encode(id_bytes)
+            }
+        }
+    }
```
anler: @mariocao @aesedepece @girazoki 📣️ Please, can you review this function? This is the function used to generate session ids for unlocked wallets. Here `key` is the [Key](https://github.com/witnet/witnet-rust/pull/723/files#diff-a4ea0d2d0f3ed43e241630042951988aR93) generated from the password used to encrypt/decrypt the wallet in the storage.

wallet/src/actors/crypto/mod.rs
```diff
@@ -1,61 +1,95 @@
...
+    pub fn gen_session_id(&self, key: &wallet::Key) -> String {
+        match self.id_hash_function {
+            HashFunction::Sha256 => {
+                let rand_bytes: [u8; 32] = self.rng.borrow_mut().gen();
+                let password =
+                    [key.secret.as_ref(), key.salt.as_ref(), rand_bytes.as_ref()].concat();
+                let id_bytes = pbkdf2_sha256(
+                    password.as_ref(),
+                    self.master_key_salt.as_ref(),
+                    self.id_hash_iterations,
```
anler: @mariocao @aesedepece @girazoki 📣️ Please, can you review this function? This is the function used to generate session ids for unlocked wallets. Here `key` is the Key generated from the password used to encrypt/decrypt the wallet in the storage.

aesedepece: Even if it looks overkill at first sight, I like the idea of using a password-based key derivation function so the logging-in mechanism is rate limited so as to deter bruteforce attacks.

In terms of security / privacy, my intuition is that this approach is correct. As a matter of fact, with a password that is partially or totally secret, a single hash should be enough to avoid recovery of its "ingredients". Of course, in this case a bruteforce attack may be quite feasible as the length of the ingredients is known, hence the need for thousands of PBKDF rounds. Deniability is not a problem in this case either (if someone guesses the ingredients of the password, being able to map that to a session ID does no additional harm).

@girazoki @mariocao have your say!

wallet/src/actors/app/mod.rs
```diff
@@ -164,92 +190,105 @@ impl App {
...
-                Box::new(fut::ok(()))
+        match self.wallet_keys.get(&wallet_id).cloned() {
+            Some(wallet_key) => {
+                let f = self
+                    .crypto
+                    .send(crypto::GenSessionId(wallet_key.clone()))
+                    .map_err(app::Error::CryptoFailed)
+                    .into_actor(self)
+                    .and_then(|id, slf, _ctx| {
+                        log::debug!("Wallet already unlocked by another session.");
```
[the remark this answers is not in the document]
anler: About this change, I'm not a big fan of logging sensible information (if someone gets the session id it will have access to that session's wallets), so I would prefer leaving it as: `log::debug!("Wallet `{}` was already unlocked another session", wallet_id);`, but what do you think?

PR #745, opened 2019-07-04
feat(p2p): modify Peers to use Bitcoin bucketing

### LINKED ISSUE #732: Buckets data structure

### LINKED ISSUE #737: Modify GetPeers method in PeersManager to support bucketing

### LINKED ISSUE #738: Add a method in PeersManager for trying peer addresses

lrubiorod:
Modified Peers structure and methods to implement Bitcoin bucketing with 2 buckets (tried and new)
Modified get_random_peer method in Peers that choose a peer randomly from each bucket
Modified Peers tests
Added src_address field in AddPeers message
Created a new method in PeersManager to import peers from a Peers structure
Created a new method in PeersManager, `try_peer`, to check if a peer is good initializing a handshake protocol

Close #732 
Close #737
Close #738

### DIFF
--- a/p2p/src/peers/mod.rs
+++ b/p2p/src/peers/mod.rs
@@ -1,85 +1,281 @@
 //! Library for managing a list of available peers
 
-use serde::{Deserialize, Serialize};
-
-use std::collections::HashMap;
-use std::net::SocketAddr;
-
 use rand::{thread_rng, Rng};
+use serde::{Deserialize, Serialize};
+use std::{collections::HashMap, net::SocketAddr};
 
+use witnet_crypto::hash::calculate_sha256;
 use witnet_util::timestamp::get_timestamp;
 
 /// Peer information being used while listing available Witnet peers
 #[derive(Serialize, Deserialize)]
 struct PeerInfo {
     address: SocketAddr,
-    _timestamp: i64,
+    timestamp: i64,
 }
 
 /// Peers TBD
 #[derive(Default, Serialize, Deserialize)]
 pub struct Peers {
-    /// Server sessions
-    peers: HashMap<SocketAddr, PeerInfo>,
+    /// Bucket for tried addresses
+    tried_bucket: HashMap<u16, PeerInfo>,
+    /// Bucket for new addresses
+    new_bucket: HashMap<u16, PeerInfo>,
+    /// Nonce value
+    sk: u64,
 }
 
 impl Peers {
-    /// Add multiple peer addresses and save timestamp
+    /// Create a new instance of Peers
+    pub fn new() -> Self {
+        Peers {
+            sk: thread_rng().gen(),
+            tried_bucket: HashMap::new(),
+            new_bucket: HashMap::new(),
+        }
+    }
+
+    /// Algorithm to calculate index for the new addresses buckets
+    pub fn new_bucket_index(&self, socket_addr: &SocketAddr, src_socket_addr: &SocketAddr) -> u16 {
+        let (_, group, host_id) = split_socket_addresses(socket_addr);
+        let (_, src_group, _) = split_socket_addresses(src_socket_addr);
+
+        calculate_index_for_new(self.sk, &src_group, &group, &host_id)
+    }
+
+    /// Algorithm to calculate index for the tried addresses buckets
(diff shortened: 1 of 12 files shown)


p2p/src/peers/mod.rs
```diff
@@ -1,85 +1,257 @@
...
+
+    let data = [&[sk], group, &[i]].concat();
+    let data_hash = calculate_sha256(&data);
+    let bucket = u16::from(data_hash.0[31]) % 64;
+
+    let data = [&[sk], host_id, &[i]].concat();
+    let data_hash = calculate_sha256(&data);
+    let slot = u16::from(data_hash.0[31]) % 64;
+
+    bucket * slot
```
tmpolaczyk: I think this should be something like:

```suggestion
    bucket * NUM_SLOTS + slot
```

Otherwise when the bucket is 0 the index is always 0, the slot is ignored.

p2p/src/peers/mod.rs
```diff
@@ -1,85 +1,271 @@
...
+
+        calculate_index_for_tried(self.sk, &ip, &group, &host_id)
+    }
+
+    /// Algorithm to calculate index for the new addresses buckets
+    fn new_bucket_index(&self, socket_addr: &SocketAddr, src_socket_addr: &SocketAddr) -> u16 {
+        let (_, group, host_id) = split_socket_addresses(socket_addr);
+        let (_, src_group, _) = split_socket_addresses(src_socket_addr);
+
+        calculate_index_for_new(self.sk, &src_group, &group, &host_id)
```
tmpolaczyk: I see that the `index_for_new` depends on the source socket address: the address of the peers that sent us the list of peers. What happens when two different peers send us the same peer? Will it be stored twice in different buckets? Or should be check if we already had this peer in new. And what about if we already have this peer in tried?

p2p/src/peers/mod.rs
```diff
@@ -1,85 +1,271 @@
...
 #[derive(Default, Serialize, Deserialize)]
 pub struct Peers {
-    /// Server sessions
-    peers: HashMap<SocketAddr, PeerInfo>,
+    /// Bucket for tried addresses
+    tried_bucket: HashMap<u16, PeerInfo>,
+    /// Bucket for new addresses
+    new_bucket: HashMap<u16, PeerInfo>,
+    /// Nonce value
+    sk: u8,
```
aesedepece: @girazoki something to say about this `u8`?

p2p/src/peers/mod.rs
lrubiorod: If 2 different peers send us the same peer will be saved in a different bucket. @girazoki explained me that this feature is important to have redundancy of good peers against possible peers provided by a bad peer. Also, if you do not know the other peer sources is impossible to check if this peer exist in new. Respect to check in tried, @girazoki say that the new and tried buckets has to be independents, so we do not need to check if it is in tried

p2p/src/peers/mod.rs
lrubiorod: @girazoki say that better use a u64

p2p/src/peers/mod.rs
```diff
@@ -1,85 +1,275 @@
...
+    /// Nonce value
+    sk: u64,
 }
 
 impl Peers {
-    /// Add multiple peer addresses and save timestamp
+    /// Create a new instance of Peers
+    pub fn new() -> Self {
+        Peers {
+            sk: thread_rng().gen_range(0, u64::max_value()),
```
tmpolaczyk: You can just do:
```suggestion
            sk: thread_rng().gen(),
```

p2p/src/peers/mod.rs
```diff
@@ -1,85 +1,282 @@
...
+            .collect();
+
+        Ok(result)
     }
 
-    /// Remove a peer given an address
+    /// Add multiple peer addresses and save timestamp in the tried addresses bucket
+    /// If an address did already exist, it will be checked and gets overwritten
+    /// in case of it would not be valid
+    /// Returns all the overwritten or rejected addresses
```
tmpolaczyk: Shouldn't peers be checked before getting overwritten? This implementation will always overwrite the peers in tried, is that the expected behavior or will it change in the future?

PR #746, opened 2019-07-05
Add consensus parameter threshold in our consensus method

### LINKED ISSUE #730: Revisit chain tip consensus using C parametric consensus requirement

### LINKED ISSUE #731: Add C parametric consensus requirement in configuration

lrubiorod:
Added consensus_c parameter in config
Modified mode_consensus method to use this parameter
Added test to check it

Close #730 
Close #731

### DIFF
--- a/config/src/defaults.rs
+++ b/config/src/defaults.rs
@@ -65,6 +65,11 @@ pub trait Defaults {
         Duration::from_secs(5)
     }
 
+    /// Constant to specify when consensus is achieved (in %)
+    fn connections_consensus_c(&self) -> u32 {
+        70
+    }
+
     /// Timestamp at the start of epoch 0
     fn consensus_constants_checkpoint_zero_timestamp(&self) -> i64;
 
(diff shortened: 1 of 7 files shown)


config/src/defaults.rs
```diff
@@ -135,6 +135,11 @@ pub trait Defaults {
         0.5
     }
 
+    /// Default period between epochs
```
tmpolaczyk:
```suggestion
    /// Constant to specify when consensus is achieved (in %)
```

schemas/witnet/witnet.proto
```diff
@@ -280,6 +280,7 @@ message ConsensusConstants {
     uint32 reputation_issuance = 7;
     uint32 reputation_issuance_stop = 8;
     double reputation_penalization_factor = 9;
+    uint32 consensus_c = 10;
```
tmpolaczyk: Despite its name, this constant is not part of the consensus parameters: this means that two peers with a different `consensus_c` should be able to peer without any problems. Therefore, `consensus_c` should be stored in a different data structure, and should be loaded from config on every node startup.

PR #750, opened 2019-07-08
JSONRPC: status method for node

### LINKED ISSUE #727: Add `status` method to NodeAPI
* Beacon tip
* Synchronized?
* Number of peers

tmpolaczyk:
Close #727 

* Beacon tip
* Synchronized?
* Number of peers

Sample call:

```
{"jsonrpc":"2.0","id":1,"method":"status"}
```
```
{"jsonrpc":"2.0","result":{"chainBeacon":{"checkpoint":576448,"hashPrevBlock":"eb1a106824538b226454423d7e988b0ec72ce74b9b28f5d0252de2381d41d405"},"numPeersInbound":1,"numPeersOutbound":1,"synchronized":true},"id":1}
```

### DIFF
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -115,6 +115,7 @@ pub struct ConsensusConstants {
     Copy, Clone, Debug, Default, Eq, Hash, PartialEq, Serialize, Deserialize, ProtobufConvert,
 )]
 #[protobuf_convert(pb = "witnet::CheckpointBeacon")]
+#[serde(rename_all = "camelCase")]
 pub struct CheckpointBeacon {
     /// The serial number for an epoch
     pub checkpoint: Epoch,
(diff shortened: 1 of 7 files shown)


aesedepece:
Let's make sure we use `camelCase` for everything in our APIs.

tmpolaczyk:
I changed the `CheckpointBeacon` serialization to use `camelCase`, but there are still many JSON-RPC methods that expect `snake_case` inputs. For example in `buildDataRequest` the `DataRequestOutput` has `snake_case` fields, and the `inventory` method accepts blocks and transactions in the default serialization format.

In order to make all of our APIs `camelCase`, we would need to add the `#[serde(rename_all = "camelCase")]` attribute to all of our serializable types. That would be a pretty automatic task, but it would break most of the usages of our APIs, including the data requests in the `examples/` folder, and the wallet-node integration.

PR #763, opened 2019-07-23
Migrate from MessagePack to CBOR

aesedepece:
This is needed for https://github.com/witnet/witnet-ethereum-bridge/issues/9


aesedepece:
Rebased on top of #778

PR #779, opened 2019-08-02
fix(node): add address from config after load peer from storage

lrubiorod:
Fixed a bug that avoid the peer address from config when load peers from storage


tmpolaczyk:
Merged as 1ac9065abe08968f6b9b3921499874fed3bcb341

PR #798, opened 2019-09-02
Radon operators

### LINKED ISSUE #794: Radon 1/2

lrubiorod:
Implement some RADON operators.

Close #794


rad/src/operators/float.rs
```diff
@@ -42,3 +59,183 @@ pub fn less_than(input: &RadonFloat, args: &[Value]) -> Result<RadonBoolean, Rad
...
+    let wrong_args = || RadError::WrongArguments {
+        input_type: RadonFloat::radon_type_name(),
+        operator: "Power".to_string(),
+        args: args.to_vec(),
+    };
+
+    let arg = args.first().ok_or_else(wrong_args)?.to_owned();
+    let exp = from_value::<i32>(arg).map_err(|_| wrong_args())?;
+
+    Ok(RadonFloat::from(input.value().powi(exp)))
```
tmpolaczyk: Why is the argument an integer? You could use floating point exponentiation with `powf` here.

rad/src/operators/float.rs
```diff
@@ -42,3 +59,183 @@ pub fn less_than(input: &RadonFloat, args: &[Value]) -> Result<RadonBoolean, Rad
...
+fn test_float_negate() {
+    let positive_integer = RadonFloat::from(10.0);
+    let negative_integer = RadonFloat::from(-10.0);
+
+    assert_eq!(negate(&positive_integer), negative_integer);
+    assert_eq!(negate(&negative_integer), positive_integer);
+}
+
+#[test]
+fn test_float_modulo() {
```
tmpolaczyk: For the sake of documentation, could you also test the result of `(-1.0) % 2.0` here?

rad/src/operators/integer.rs
```diff
@@ -0,0 +1,245 @@
...
+pub fn modulo(input: &RadonInteger, args: &[Value]) -> Result<RadonInteger, RadError> {
+    let wrong_args = || RadError::WrongArguments {
+        input_type: RadonInteger::radon_type_name(),
+        operator: "Modulo".to_string(),
+        args: args.to_vec(),
+    };
+
+    let arg = args.first().ok_or_else(wrong_args)?.to_owned();
+    let modulo = from_value::<i128>(arg).map_err(|_| wrong_args())?;
+    Ok(RadonInteger::from(input.value() % modulo))
```
tmpolaczyk: Actually, in Rust `%` is the remainder operator, not the modulo operator. The main difference is what happens with negative numbers: `-1 mod 2 = 1` but `-1 rem 2 = -1`. I assume we want to keep the standard behavior, so there is a method which can be used as the modulo operator: `rem_euclid`.

Additionally, performing a modulo with 0 will panic at runtime, so we should use the checked variant of the method:
```suggestion
    Ok(RadonInteger::from(input.value().checked_rem_euclid(modulo)))
```

rad/src/operators/integer.rs
```diff
@@ -0,0 +1,245 @@
...
+        "Overflow error".to_string(),
+    );
+}
+
+#[test]
+fn test_integer_modulo() {
+    let rad_int = RadonInteger::from(10);
+    let value = Value::Integer(3);
+
+    assert_eq!(modulo(&rad_int, &[value]).unwrap(), RadonInteger::from(1));
```
tmpolaczyk: Same as with floats, could you add more examples with negative numbers? Here is a nice discussion with examples: https://internals.rust-lang.org/t/mathematical-modulo-operator/5952

rad/src/operators/string.rs
```diff
@@ -25,6 +34,30 @@ pub fn to_float(input: &RadonString) -> Result<RadonFloat, RadError> {
...
+}
+
+pub fn to_bool(input: &RadonString) -> Result<RadonBoolean, RadError> {
+    bool::from_str(&input.value())
+        .map(RadonBoolean::from)
+        .map_err(Into::into)
+}
+
+pub fn length(input: &RadonString) -> RadonInteger {
+    RadonInteger::from(input.value().len() as i128)
```
tmpolaczyk: I think clippy prefers `i128::from(x)`, but `impl From<usize> for i128` does not exist. There is a `TryFrom` implementation but I'm not sure if we should handle errors or just unwrap.

```suggestion
    RadonInteger::from(i128::try_from(input.value().len()).unwrap())
```

examples/bitcoin_price.json
```diff
@@ -1 +1 @@
...
+{"jsonrpc":"2.0","method":"sendRequest","id":1,"params":{"dro":{"data_request":{"not_before":0,"retrieve":[{"kind":"HTTP-GET","url":"https://api.coindesk.com/v1/bpi/currentprice.json","script":[136, 24, 69, 24, 116, 130, 24, 97, 99, 98, 112, 105, 24, 116, 130, 24, 97, 99, 86, 83, 68, 24, 116, 130, 24, 97, 106, 114, 97, 116, 101, 95, 102, 108, 111, 97, 116, 24, 114]}],"aggregate":{"script":[129, 130, 24, 87, 3]},"consensus":{"script":[129, 130, 24, 87, 3]},"deliver":[{"kind":"HTTP-GET","url":"https://hooks.zapier.com/hooks/catch/3860543/l2awcd/"},{"kind":"HTTP-GET","url":"https://hooks.zapier.com/hooks/catch/3860543/l1awcw/"}]},"value":1002,"witnesses":2,"backup_witnesses":1,"commit_fee":0,"reveal_fee":0,"tally_fee":0,"time_lock":0},"fee":0}}
```
tmpolaczyk:
```suggestion
{"jsonrpc":"2.0","method":"sendRequest","id":1,"params":{"dro":{"data_request":{"not_before":0,"retrieve":[{"kind":"HTTP-GET","url":"https://api.coindesk.com/v1/bpi/currentprice.json","script":[136, 24, 69, 24, 116, 130, 24, 97, 99, 98, 112, 105, 24, 116, 130, 24, 97, 99, 85, 83, 68, 24, 116, 130, 24, 97, 106, 114, 97, 116, 101, 95, 102, 108, 111, 97, 116, 24, 114]}],"aggregate":{"script":[129, 130, 24, 87, 3]},"consensus":{"script":[129, 130, 24, 87, 3]},"deliver":[{"kind":"HTTP-GET","url":"https://hooks.zapier.com/hooks/catch/3860543/l2awcd/"},{"kind":"HTTP-GET","url":"https://hooks.zapier.com/hooks/catch/3860543/l1awcw/"}]},"value":1002,"witnesses":2,"backup_witnesses":1,"commit_fee":0,"reveal_fee":0,"tally_fee":0,"time_lock":0},"fee":0}}
```

PR #811, opened 2019-09-09
feat(p2p): implement feeler periodically function

### LINKED ISSUE #736: "Feeler" worker that curates peer addresses and puts good ones in the tried bucket

lrubiorod:
Close #736

### DIFF
--- a/config/src/defaults.rs
+++ b/config/src/defaults.rs
@@ -60,6 +60,12 @@ pub trait Defaults {
         Duration::from_secs(30)
     }
 
+    /// Default period for trying connecting to recently discovered peer addresses
+    /// in search for quality peers
+    fn connections_feeler_peers_period(&self) -> Duration {
+        Duration::from_secs(2)
+    }
+
     /// Default handshake timeout
     fn connections_handshake_timeout(&self) -> Duration {
         Duration::from_secs(5)
--- a/node/src/actors/connections_manager/mod.rs
+++ b/node/src/actors/connections_manager/mod.rs
@@ -70,6 +70,7 @@ impl ConnectionsManager {
     /// Method to process resolver ConnectAddr response
     fn process_connect_addr_response(
         response: Result<ResolverResult, MailboxError>,
+        feeler_flag: bool,
     ) -> actix::fut::FutureResult<(), (), Self> {
         // Process the Result<ResolverResult, MailboxError>
         match response {
@@ -93,7 +94,17 @@ impl ConnectionsManager {
                             });
 
                         // Request the creation of a new session actor from connection
-                        ConnectionsManager::request_session_creation(stream, SessionType::Outbound);
+                        if feeler_flag {
+                            ConnectionsManager::request_session_creation(
+                                stream,
+                                SessionType::Feeler,
+                            );
+                        } else {
+                            ConnectionsManager::request_session_creation(
+                                stream,
+                                SessionType::Outbound,
+                            );
+                        }
 
                         actix::fut::ok(())
                     }
--- a/node/src/actors/connections_manager/mod.rs
+++ b/node/src/actors/connections_manager/mod.rs
@@ -70,7 +70,7 @@ impl ConnectionsManager {
     /// Method to process resolver ConnectAddr response
     fn process_connect_addr_response(
         response: Result<ResolverResult, MailboxError>,
-        feeler_flag: bool,
+        session_type: SessionType,
     ) -> actix::fut::FutureResult<(), (), Self> {
         // Process the Result<ResolverResult, MailboxError>
         match response {
@@ -94,17 +94,7 @@ impl ConnectionsManager {
                             });
 
                         // Request the creation of a new session actor from connection
-                        if feeler_flag {
-                            ConnectionsManager::request_session_creation(
-                                stream,
-                                SessionType::Feeler,
-                            );
-                        } else {
-                            ConnectionsManager::request_session_creation(
-                                stream,
-                                SessionType::Outbound,
-                            );
-                        }
+                        ConnectionsManager::request_session_creation(stream, session_type);
 
                         actix::fut::ok(())
                     }
--- a/node/src/actors/peers_manager/actor.rs
+++ b/node/src/actors/peers_manager/actor.rs
@@ -27,12 +27,16 @@ impl Actor for PeersManager {
 
                 // Get server address
                 let server_addr = config.connections.server_addr;
+                act.peers.set_server(server_addr);
 
                 // Get bucketing update period
                 act.bucketing_update_period = config.connections.bucketing_update_period;
 
                 // Get handshake time_out
                 act.handshake_timeout = config.connections.handshake_timeout;
+
+                // Get feeler period
+                let feeler_peers_period = config.connections.feeler_peers_period;
 
                 // Add all peers
                 info!(
@@ -47,7 +51,7 @@ impl Actor for PeersManager {
                 storage_mngr::get::<_, Peers>(&PEERS_KEY)
                     .into_actor(act)
                     .map_err(|e, _, _| error!("Couldn't get peers from storage: {}", e))
-                    .and_then(move |peers_from_storage, act, _| {
+                    .and_then(move |peers_from_storage, act, _ctx| {
                         // peers_from_storage can be None if the storage does not contain that key
                         if let Some(peers_from_storage) = peers_from_storage {
                             // Add all the peers from storage
@@ -61,6 +65,9 @@ impl Actor for PeersManager {
 
                 // Start the storage peers process on SessionsManager start
                 act.persist_peers(ctx, storage_peers_period);
+
+                // Start the feeleer peers process on SessionsManager start
+                act.feeler(ctx, feeler_peers_period);
 
                 fut::ok(())
             })
--- a/node/src/actors/peers_manager/mod.rs
+++ b/node/src/actors/peers_manager/mod.rs
@@ -112,6 +112,18 @@ impl PeersManager {
             _ => {}
         }
     }
+
+    /// Method to try peers periodically to move peers from new to tried
+    pub fn feeler(&mut self, ctx: &mut Context<Self>, feeler_peers_period: Duration) {
+        // Schedule the discovery_peers with a given period
+        ctx.run_later(feeler_peers_period, move |act, ctx| {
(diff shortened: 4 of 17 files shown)


config/src/defaults.rs
```diff
@@ -60,6 +60,11 @@ pub trait Defaults {
         Duration::from_secs(30)
     }
 
+    /// Default period for discovering peers
```
aesedepece: This could be a more accurate description:
```suggestion
    /// Default period for trying connecting to recently discovered peer addresses in search for quality peers
```
@mariocao and @girazoki please confirm I got it right.

config/src/defaults.rs
```diff
@@ -60,6 +60,11 @@ pub trait Defaults {
...
+    fn connections_feeler_peers_period(&self) -> Duration {
+        Duration::from_secs(2)
```
aesedepece: Any rationale behind the default value?

node/src/actors/peers_manager/actor.rs
```diff
@@ -34,6 +34,9 @@ impl Actor for PeersManager {
                 // Get handshake time_out
                 act.handshake_timeout = config.connections.handshake_timeout;
 
+                // Get peers
```
aesedepece: I don't think this comment describes what the next line is doing, hehehe.

p2p/src/peers/mod.rs
```diff
@@ -158,6 +158,16 @@ impl Peers {
             .collect()
     }
 
+    /// Remove a peer given an index from new addresses bucket
+    /// Returns the removed addresses
+    pub fn remove_from_new_with_index(&mut self, indexes: &[u16]) -> Vec<SocketAddr> {
```
aesedepece: Are we certain this will always succeed and not remove the wrong entry?

p2p/src/peers/mod.rs
```diff
@@ -179,6 +189,17 @@ impl Peers {
...
+    pub fn get_new_random(&self) -> Option<(u16, SocketAddr)> {
+        // Random index with range [0, len) of the peers vector
+        let index = thread_rng().gen_range(0, self.new_bucket.len());
+
+        self.new_bucket
+            .iter()
+            .nth(index)
+            .map(|(k, v)| (*k, v.address.to_owned()))
+    }
+
```
aesedepece: Please note the existence of this:
https://docs.rs/rand/0.7.0/rand/seq/trait.IteratorRandom.html

config/src/defaults.rs
tmpolaczyk: It's probably too low?

p2p/src/peers/mod.rs
tmpolaczyk: We do the operation of "chose a random element from a hashmap" many times, perhaps we can create an issue for this.

config/src/defaults.rs
lrubiorod: It has to be as soon as possible, but speaking with @girazoki it has to be similar to handshake time. So finally we speak to use about 1 or 2 seconds

node/src/actors/peers_manager/actor.rs
lrubiorod: :man_facepalming:

node/src/actors/peers_manager/mod.rs
```diff
@@ -112,6 +112,18 @@ impl PeersManager {
             _ => {}
         }
     }
+
+    /// Method to try peers periodically to move peers from new to tried
+    pub fn feeler(&mut self, ctx: &mut Context<Self>, feeler_peers_period: Duration) {
+        // Schedule the discovery_peers with a given period
+        ctx.run_later(feeler_peers_period, move |act, ctx| {
+            if let Some((key, peer)) = act.peers.get_new_random() {
+                act.peers.remove_from_new_with_index(&[key]);
```
tmpolaczyk: So if the `try_peer` fails, this peer is removed from new. Is that the expected behavior?

lrubiorod: if try_peer fails means that it is not a useful peer

tmpolaczyk:
So the feeler basically just tries to connect to a random peer from "new", and if the handshake succeeds, that peer is inserted into "tried".

Does that new connection consolidate into a full session? Is it affected by the `outbound_limit`? What happens with peers that are currently connected to us? Will we open a new connection?

And most importantly, how do we test this?

p2p/src/peers/mod.rs
```diff
@@ -109,6 +142,8 @@ impl Peers {
             })
             .collect();
 
+        log::debug!("Added new peers: \n{}", self);
```
tmpolaczyk: Shall we change all this log messages to trace?

lrubiorod: yep, i think so

p2p/src/peers/mod.rs
lrubiorod: oks, we can include in a new issue

p2p/src/peers/mod.rs
lrubiorod: Due that new bucket needs the information about the source address, it is difficult to remove without that information. So using the index we can remove directly from the hashmap.
The other solution could be to have 2 structs for PeerInfo, one without source for the tried bucket and another with source for the new bucket

p2p/src/peers/mod.rs
```diff
@@ -92,9 +110,24 @@ impl Peers {
         // Note: if the peer address exists, the peer info will be overwritten
         let result = addrs
             .into_iter()
-            // Filter out unspecified addresses (aka 0.0.0.0)
-            .filter(|address| !address.ip().is_unspecified())
             .filter_map(|address| {
+                // Filter out unspecified addresses (aka 0.0.0.0), and the server address
+                if address.ip().is_unspecified()
+                    || self.is_server_address(&address).is_none()
+                    || self.is_server_address(&address).unwrap()
```
aesedepece:
```suggestion
                    || self.is_server_address(&address).unwrap_or(false)
```

p2p/src/peers/mod.rs
```diff
@@ -92,9 +110,24 @@ impl Peers {
...
+                {
+                    return None;
+                }
+
+                let index = self.tried_bucket_index(&address);
+                let elem = self.tried_bucket.get(&index);
+
+                // If the index point to the same address that it is already
+                // in tried, we don't include in new bucket
+                if elem.is_some() && (elem.unwrap().address == address) {
```
aesedepece:
```suggestion
                if let Some(tried_address) = elem {
                    if address == tried_address {
                        return None
```

p2p/src/peers/mod.rs
```diff
@@ -92,9 +110,24 @@ impl Peers {
         // Note: if the peer address exists, the peer info will be overwritten
         let result = addrs
             .into_iter()
-            // Filter out unspecified addresses (aka 0.0.0.0)
-            .filter(|address| !address.ip().is_unspecified())
             .filter_map(|address| {
+                // Filter out unspecified addresses (aka 0.0.0.0), and the server address
```
aesedepece: May I suggest giving this function a single point of exit?

p2p/src/peers/mod.rs
aesedepece: Shall we take the chance to apply the change here as an example that others can follow?

p2p/src/peers/mod.rs
lrubiorod: to only have a single point of exit that you suggest after, it is more easy to evaluate this expression

lrubiorod:
What changes are you saying? About randomizing hasmaps? If yes, those changes comes in the next PR

node/src/actors/sessions_manager/handlers.rs
```diff
@@ -96,23 +97,29 @@ impl Handler<Unregister> for SessionsManager {
...
-                "Session (type {:?}) unregistered for peer {}",
-                msg.session_type, msg.address
-            ),
-            Err(error) => error!(
-                "Error while unregistering peer {} (session type {:?}): {}",
-                msg.address, msg.session_type, error
-            ),
+        // First evaluate Feeler case
+        if msg.session_type == SessionType::Feeler {
+            // Feeler sessions are not working with SessionsManager
```
aesedepece:
```suggestion
            // Feeler sessions should not be managed by `SessionsManager`
```

p2p/src/error.rs
```diff
@@ -17,6 +17,9 @@ pub enum SessionsError {
     /// Errors when updating sessions
     #[fail(display = "Is not an outbound consolidated peer")]
     NotOutboundConsolidatedPeer,
+    /// Errors when using SessionType::Feeler sessions in SessionsManager
+    #[fail(display = "Peer provided from feeler has been used in SessionsManager")]
```
aesedepece:
```suggestion
    #[fail(display = "SessionsManager was told to manage a Feeler session. The session will be ignored because Feeler sessions should not be managed.")]
```

p2p/src/peers/mod.rs
aesedepece: Ping.

p2p/src/sessions/mod.rs
```diff
@@ -89,16 +91,17 @@ where
         &mut self,
         session_type: SessionType,
         status: SessionStatus,
-    ) -> &mut BoundedSessions<T> {
+    ) -> Result<&mut BoundedSessions<T>, failure::Error> {
```
aesedepece: This makes me happy.

node/src/actors/connections_manager/mod.rs
```diff
@@ -70,6 +70,7 @@ impl ConnectionsManager {
     /// Method to process resolver ConnectAddr response
     fn process_connect_addr_response(
         response: Result<ResolverResult, MailboxError>,
+        feeler_flag: bool,
```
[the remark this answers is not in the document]
tmpolaczyk: +1, `foo(SessionType::Feeler)` is always a better API than `foo(true)`

lrubiorod:
https://github.com/witnet/witnet-rust/pull/811#discussion_r325634544

I have created another PR which rebase this one and use the random functions for hashmaps. Would you like that added that PR to this one?

tmpolaczyk:
This PR includes some changes which may impede initializing the bootstrap node, as now the nodes cannot use themselves as peers. Therefore, we would need two bootstrap nodes.

PR #820, opened 2019-09-16
Use Bech32 in node

### LINKED ISSUE #819: Use Bech32 in node

tmpolaczyk:
Close #819 

This PR changes the display and serialization implementation of `PublicKeyHash` to always use Bech32 instead of the hex representation we used until now.

This includes logs, CLI commands and JSON-RPC commands: everything uses Bech32.

The implementation uses a static variable to get the current environment (mainnet/testnet). This is because the Bech32 encoding prefix changes with the environment: "wit" for mainnet and "twit" for testnet. This may be a problem, because now everything that uses the `witnet_data_structures` crate must properly initialize the environment. To reduce impact on end users, the default environment is set to mainnet, and it has to be explicitly set to testnet by the initialization code. For example, the `witnet` command sets the environment right after reading the configuration. A benefit of this is that mainnet blocks and transactions will be rejected by any testnet client, as the JSON serialization changes.


tmpolaczyk:
I have some ideas to minimize potential errors by forgetting to manually set the environment:

* Create a `load_config` function in `witnet_config`, and set the environment there. Then encourage using that function everywhere. This won't help in some cases, for example the bridge doesn't use `witnet_config`.
* Add a flag to check that `get_environment` is never used before `set_environment` at runtime. This implies a unknown overhead, probably atomic store every time we serialize a PKH, because we need a flag to check if the `get_environment` was already called.
* Instead of initializing the environment to mainnet, initialize it to `None`. This will force users to initialize it because we can show a better error: "Please set the environment". In the current implementation the only error will be "Address is for different environment: prefix \"wit\" is not valid for testnet". Maybe we can change it to something like "Node running in testnet environment but address uses incompatible \"wit\" prefix".

PR #822, opened 2019-09-17
Uniquely index transactions

anler:
## What

* Add unit tests for `index_transactions` method
* Uniquely index transactions, that is, if a block is indexed twice its transactions aren't indexed/duplicated anymore in the database
* Add a new method `contains` to `Database` interface

### DIFF
--- a/wallet/src/repository/wallet/mod.rs
+++ b/wallet/src/repository/wallet/mod.rs
@@ -240,7 +240,7 @@ where
         let value = self.db.get(&keys::transaction_value(account, index))?;
         let kind = self.db.get(&keys::transaction_type(account, index))?;
         let timestamp = self.db.get(&keys::transaction_timestamp(account, index))?;
-        let hash = self.db.get(&keys::transaction_hash(account, index))?;
+        let hash: Vec<u8> = self.db.get(&keys::transaction_hash(account, index))?;
         let label = self.db.get_opt(&keys::transaction_label(account, index))?;
         let fee = self.db.get_opt(&keys::transaction_fee(account, index))?;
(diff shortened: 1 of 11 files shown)


wallet/src/repository/wallet/mod.rs
```diff
@@ -514,6 +517,7 @@ where
             batch.put(keys::account_balance(account), new_balance)?;
             batch.put(keys::account_utxo_set(account), db_utxo_set)?;
             batch.put(keys::transaction_next_id(account), txn_next_id)?;
+            batch.put(keys::transactions_index(txn_hash), ())?;
```
tmpolaczyk: Could you add a comment here?

```
// Mark transaction as indexed so we don't accidentally index it twice
```

anler: Now I'm associating the transaction id to its hash, so I believe the comment is no longer needed

wallet/src/repository/wallet/tests/factories.rs
```diff
@@ -0,0 +1,129 @@
...
+        )
+        .unwrap();
+
+    let wallet = Wallet::unlock(db.clone(), params, engine).unwrap();
+
+    (wallet, db)
+}
+
+pub fn pkh() -> types::PublicKeyHash {
+    let bytes: [u8; 20] = rand::random();
```
tmpolaczyk: Non-deterministic tests? :eyes:

anler: hahaha not really, these are pretty common for generating random data for tests, think of
[faker](https://github.com/paraseba/faker) or its Rust equivalent [faker-rs](https://github.com/cksac/fake-rs)

PR #825, opened 2019-09-18
Review protobuff schema

### LINKED ISSUE #823: Harmonize protobuff schema

lrubiorod:
Removed Ping Pong messages
Removed Error, DataRequest and DataResult
Removed RADDelivery
Renamed consensus by tally in DataRequestOutput
Renamed not_before by time_lock and removed previous time_lock in DataRequestOutput
Included time_lock in VTTransaction

*** Validation of VTT time_lock  (#824) will be added in other PR to ease work with wallet and sheikah
Close #823


tmpolaczyk:
I think it's confusing to call `time_lock` a field which:

* In `DRTransaction`s, the transaction will be included into a block as soon as possible, but commits will only be accepted after `time_lock.
* In `VTTransaction`s, the transaction will only be included into a block after the `time_lock`.

How about we rename the data request `time_lock` to `commit_time_lock`?

lrubiorod:
> I think it's confusing to call `time_lock` a field which:
> 
>     * In `DRTransaction`s, the transaction will be included into a block as soon as possible, but commits will only be accepted after `time_lock`.
> 
>     * In `VTTransaction`s, the transaction will only be included into a block after the `time_lock`.
> 
> 
> How about we rename the data request `time_lock` to `commit_time_lock`?

I agree with you, but in case of VTT these transactions will be added to a block but it can not be included as a UTXO to spend after time_lock (so I am thinking that it would difficult the index utxo issue)

tmpolaczyk:
> I agree with you, but in case of VTT these transactions will be added to a block but it can not be included as a UTXO to spend after time_lock (so I am thinking that it would difficult the index utxo issue)

True, I edited my comment. So `time_lock` means "do not accept transactions which reference this transaction until time T". In that case I am OK with calling both `time_lock`.

data_structures/src/chain.rs
```diff
@@ -635,6 +635,7 @@ impl Input {
 pub struct ValueTransferOutput {
     pub pkh: PublicKeyHash,
     pub value: u64,
+    pub time_lock: u64,
```
```suggestion
    /// Do not accept transactions which reference this transaction until timestamp
    pub time_lock: u64,
```

Same for data requests

data_structures/src/chain.rs
```diff
@@ -761,11 +762,10 @@ impl Default for RADType {
 #[derive(Debug, Eq, PartialEq, Clone, Serialize, Deserialize, ProtobufConvert, Hash, Default)]
 #[protobuf_convert(pb = "witnet::DataRequestOutput_RADRequest", crate = "crate")]
 pub struct RADRequest {
-    pub not_before: u64,
+    pub time_lock: u64,
```
```suggestion
    /// Do not accept commits for this data request until timestamp
    pub time_lock: u64,
```

data_structures/src/chain.rs
aesedepece: For VTTs, the most accurate description would be:
> The value attached to a time-locked output cannot be spent before the specified timestamp. That is, they cannot be used as an input in any transaction of a subsequent block for an epoch whose opening timestamp predates the time lock.

data_structures/src/chain.rs
aesedepece: > Commitments for this request will not be accepted in any block proposed for an epoch whose opening timestamp predates the specified time lock. This effectively prevents a request from being processed before a specific future point in time.

PR #826, opened 2019-09-18
Refactor validations crate

anler:
## What

1. Create new function `validate_data_request_output` which is used by the wallet
2. Add more descriptive errors: `NoReward` (*The fees of the data request is greater that its value*) and `NonUniformReward` (*The reward is not the same for each witness*)
3. Move tests inside the crate since they are unit and not integration tests.

### DIFF
--- a/data_structures/src/error.rs
+++ b/data_structures/src/error.rs
@@ -29,6 +29,8 @@ pub enum BuildersError {
 pub enum TransactionError {
     #[fail(display = "The transaction is invalid")]
     NotValidTransaction,
+    #[fail(display = "Sum of fees overflows")]
+    FeeOverflow,
     /// The transaction creates value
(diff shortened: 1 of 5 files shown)


data_structures/src/error.rs
```diff
@@ -58,13 +60,16 @@ pub enum TransactionError {
         change, expected_change
     )]
     InvalidTallyChange { change: u64, expected_change: u64 },
-    #[fail(display = "Invalid Data Request reward: {}", reward)]
-    InvalidDataRequestReward { reward: i64 },
     #[fail(
-        display = "Invalid Data Request reward ({}) for this number of witnesses ({})",
-        dr_value, witnesses
+        display = "The fees of the data request: {}, is greater that its value: {}",
```
lrubiorod:
```suggestion
        display = "The fees of the data request: {}, is greater or equal than its value: {}",
```

PR #832, opened 2019-09-20
Fix balances

anler:
## What

When creating VTT and Data Request transactions the balance was not being updated accordingly. This PR fixes that.

### DIFF
--- a/wallet/src/repository/wallet/mod.rs
+++ b/wallet/src/repository/wallet/mod.rs
@@ -37,8 +37,20 @@ where
             .unwrap_or_else(|| vec![account]);
 
         let transaction_next_id = db.get_or_default(&keys::transaction_next_id(account))?;
-        let balance = db.get_or_default(&keys::account_balance(account))?;
-        let utxo_set = db.get_or_default(&keys::account_utxo_set(account))?;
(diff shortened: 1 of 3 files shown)


wallet/src/repository/wallet/mod.rs
```diff
@@ -546,14 +562,20 @@ where
...
 
             let extended_sign_key =
                 parent_key.derive(&self.engine, &types::KeyPath::default().index(index))?;
 
             payment = payment
                 .checked_add(key_balance.amount)
                 .ok_or_else(|| Error::TransactionValueOverflow)?;
+            balance = balance
+                .checked_sub(key_balance.amount)
+                .ok_or_else(|| Error::BalanceUnderflow)?;
```
tmpolaczyk: Should the balance be updated when the transaction is created, or when the transaction is included in a block?

Note that in the future we will have different types of balance: confirmed and pending.

anler: Since the used UTXOs cannot be used to create other transactions, in practice the balance has dropped down, so I would leave it as it is at the moment

anler:
Superseded by #846

PR #836, opened 2019-09-24
feat(radon): implement average mean for RadonArray

lrubiorod:


rad/src/operators/array.rs
```diff
@@ -58,6 +58,36 @@ pub fn map(input: &RadonArray, args: &[Value]) -> Result<RadonTypes, RadError> {
...
+
+                if prev_len.is_none() {
+                    for sub_item in rad_value.value().into_iter() {
+                        v.push(vec![sub_item]);
+                    }
+                    prev_len = Some(sub_value_len);
+                } else if prev_len == Some(sub_value_len) {
+                    for (i, sub_item) in rad_value.value().into_iter().enumerate() {
+                        let v_aux = v.get_mut(i).unwrap();
+                        v_aux.push(sub_item);
```
tmpolaczyk:
```suggestion
                        v[i].push(sub_item);
```

rad/src/operators/array.rs
```diff
@@ -58,6 +58,36 @@ pub fn map(input: &RadonArray, args: &[Value]) -> Result<RadonTypes, RadError> {
...
+    let mut prev_len = None;
+    for item in input.value() {
+        match item {
+            RadonTypes::Array(rad_value) => {
+                let sub_value = rad_value.value();
+                let sub_value_len = sub_value.len();
+
+                if prev_len.is_none() {
+                    for sub_item in rad_value.value().into_iter() {
+                        v.push(vec![sub_item]);
```
tmpolaczyk: You could optimize this by preallocating a vector with capacity for `input.value().len()` elements, but it's a very small optimization.

rad/src/operators/array.rs
```diff
@@ -58,6 +58,36 @@ pub fn map(input: &RadonArray, args: &[Value]) -> Result<RadonTypes, RadError> {
     Ok(RadonArray::from(result).into())
 }
 
+pub fn transpose(input: &RadonArray) -> Result<Vec<Vec<RadonTypes>>, RadError> {
```
tmpolaczyk: If the input is `&RadonArray`, shouldn't the output be `Result<RadonArray, RadError>`?

tmpolaczyk: And add tests please, especially for the error cases

rad/src/reducers/average.rs
```diff
@@ -1,24 +1,167 @@
...
+    // Float case
+    if inner_type == discriminant(&RadonTypes::Float(RadonFloat::default())) {
+        let sum = value.iter().try_fold(0f64, |sum, item| match item {
+            RadonTypes::Float(f64_value) => Ok(sum + f64_value.value()),
+            _ => Err(RadError::MismatchingTypes),
         });
+        let sum = sum?;
+
+        // Divide sum by the count of numeric values that were summed
+        let mean_value: f64 = sum.div(value_len as f64);
```
tmpolaczyk: What should happen if `value_len == 0`?

Currently, it returns `NaN`. Add test:

```rust
#[test]
fn test_operate_reduce_mean_empty() {
    let input = RadonArray::from(vec![]);
    let output = mean(&input).unwrap();
    assert_eq!(
        output,
        RadonTypes::from(RadonFloat::from(std::f64::NAN))
    );
}
```

rad/src/error.rs
```diff
@@ -117,6 +117,12 @@ pub enum RadError {
     /// Overflow error
     #[fail(display = "Overflow error")]
     Overflow,
+    /// Mismatching types
+    #[fail(display = "Mismatching types")]
```
aesedepece: This error message is lacking context. We should print both values for debugging purposes.

rad/src/error.rs
```diff
@@ -117,6 +117,12 @@ pub enum RadError {
...
+    MismatchingTypes,
+    /// Arrays to be reduced have different sizes
+    #[fail(display = "Arrays to be reduced have different sizes")]
```
aesedepece: - For context, e should explicitly state which reducer thrown the error.
- We can print the sizes here as `(x != y)`

rad/src/types/array.rs
```diff
@@ -31,6 +34,26 @@ impl RadonArray {
         self.inner_type
     }
 
+    pub fn inner_type_name(&self) -> String {
```
aesedepece: Wow this is wild. There's no reason for calling `self.inner_type()` all the time.

PR #837, opened 2019-09-24
feat(rad): add array sort operator

girazoki:
This PR adds sort functionality for array types (for integers and strings)

### DIFF
--- a/rad/src/error.rs
+++ b/rad/src/error.rs
@@ -77,6 +77,18 @@ pub enum RadError {
         reducer, inner_type
     )]
     UnsupportedReducer { inner_type: String, reducer: String },
+    /// The sort operator is not implemented for non-string arrays
+    #[fail(
+        display = "Sort is not supported for Array with inner type `{}`",
+        inner_type
+    )]
+    UnsupportedSortOp { inner_type: String },
+    /// The operator is not implemented for non-homogeneous arrays
+    #[fail(
+        display = "`{}` is not supported for Array with non homogeneous types",
+        operator
+    )]
+    UnsupportedOpNonHomogeneous { operator: String },
     /// There was a tie after applying the mode reducer
     #[fail(
         display = "There was a tie after applying the mode reducer on values: `{:?}`",
--- a/rad/src/error.rs
+++ b/rad/src/error.rs
@@ -79,13 +79,13 @@ pub enum RadError {
     UnsupportedReducer { inner_type: String, reducer: String },
     /// The sort operator is not implemented for non-string arrays
     #[fail(
-        display = "Sort is not supported for Array with inner type `{}`",
+        display = "ArraySort is not supported for RadonArray with inner type `{}`",
         inner_type
     )]
     UnsupportedSortOp { inner_type: String },
     /// The operator is not implemented for non-homogeneous arrays
     #[fail(
-        display = "`{}` is not supported for Array with non homogeneous types",
+        display = "`{}` is not supported for RadonArray with non homogeneous types",
         operator
     )]
     UnsupportedOpNonHomogeneous { operator: String },
--- a/rad/src/operators/array.rs
+++ b/rad/src/operators/array.rs
@@ -1,15 +1,11 @@
 use crate::error::RadError;
 use crate::reducers::{self, RadonReducers};
 use crate::script::{execute_radon_script, unpack_radon_call};
-use crate::types::{
-    array::RadonArray, integer::RadonInteger, string::RadonString, RadonType, RadonTypes,
-};
-use itertools::Itertools;
+use crate::types::{array::RadonArray, integer::RadonInteger, RadonType, RadonTypes};
 
 use num_traits::FromPrimitive;
 use serde_cbor::value::{from_value, Value};
 use std::clone::Clone;
-use std::mem::discriminant;
 
 pub fn count(input: &RadonArray) -> RadonInteger {
     RadonInteger::from(input.value().len() as i128)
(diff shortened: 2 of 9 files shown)


rad/src/operators/array.rs
```diff
@@ -58,6 +62,44 @@ pub fn map(input: &RadonArray, args: &[Value]) -> Result<RadonTypes, RadError> {
...
+    let disc = input.inner_type();
+    let value = input.value();
+
+    let mut result = Vec::new();
+    // Distinguish each discriminant case
+    if disc == discriminant(&RadonTypes::from(RadonString::default())) {
+        result = value
+            .iter()
+            .cloned()
+            .sorted_by(|a, b| match (a, b) {
```
tmpolaczyk: There is a `sorted_by_key` function which is more clear in this case, you just need to return the inner string.

```suggestion
            .sorted_by_key(|v| match v {
```

rad/src/operators/array.rs
```diff
@@ -58,6 +62,44 @@ pub fn map(input: &RadonArray, args: &[Value]) -> Result<RadonTypes, RadError> {
...
+                (RadonTypes::Integer(a), RadonTypes::Integer(b)) => a.cmp(&b),
+                _ => unreachable!(),
+            })
+            .collect();
+    } else {
+        Err(RadError::UnsupportedSortOp {
+            inner_type: value[0].clone().radon_type_name(),
```
tmpolaczyk: Maybe this will panic if `value` is empty? Add a test for sorting an empty array, it should return an empty array.

rad/src/error.rs
```diff
@@ -71,6 +71,15 @@ pub enum RadError {
...
     UnsupportedReducer { inner_type: String, reducer: String },
+    /// The sort operator is not implemented for non-string arrays
+    #[fail(
+        display = "Sort is not supported for Array with inner type `{}`",
+        inner_type
+    )]
+    UnsupportedSortOp { inner_type: String },
+    /// The operator is not implemented for non-homogeneous arrays
+    #[fail(display = "Operator is not supported for Array with non homogeneous types")]
+    UnsupportedOpNonHomogeneous,
```
aesedepece: Let's take the operator as an string argument and print it here.

rad/src/operators/array.rs
tmpolaczyk: It turns out this is cannot be done because of lifetimes...

rad/src/error.rs
```diff
@@ -77,6 +77,18 @@ pub enum RadError {
         reducer, inner_type
     )]
     UnsupportedReducer { inner_type: String, reducer: String },
+    /// The sort operator is not implemented for non-string arrays
+    #[fail(
+        display = "Sort is not supported for Array with inner type `{}`",
```
aesedepece:
```suggestion
        display = "ArraySort is not supported for RadonArray with inner type `{}`",
```

rad/src/error.rs
```diff
@@ -77,6 +77,18 @@ pub enum RadError {
...
+        inner_type
+    )]
+    UnsupportedSortOp { inner_type: String },
+    /// The operator is not implemented for non-homogeneous arrays
+    #[fail(
+        display = "`{}` is not supported for Array with non homogeneous types",
```
aesedepece:
```suggestion
        display = "`{}` is not supported for RadonArray with non homogeneous types",
```

rad/src/operators/array.rs
```diff
@@ -81,6 +85,50 @@ pub fn filter(input: &RadonArray, args: &[Value]) -> Result<RadonTypes, RadError
...
 
+pub fn sort(input: &RadonArray) -> Result<RadonArray, RadError> {
+    // if input is empty, return the array
+    if input.value().is_empty() {
+        return Ok(input.clone());
+    }
+    // Sort not applicable if not homogeneous
+    if !input.is_homogeneous() {
+        Err(RadError::UnsupportedOpNonHomogeneous {
+            operator: "Sort".to_string(),
```
aesedepece:
```suggestion
            operator: "ArraySort".to_string(),
```

rad/src/operators/array.rs
```diff
@@ -81,6 +81,53 @@ pub fn filter(input: &RadonArray, args: &[Value]) -> Result<RadonTypes, RadError
     Ok(RadonArray::from(result).into())
 }
 
+pub fn sort(input: &RadonArray, args: &[Value]) -> Result<RadonArray, RadError> {
+    let input_value = input.value();
+    let mapped_array = map(input, args)?;
+    let mapped_array = match mapped_array {
+        RadonTypes::Array(x) => x,
+        _ => Err(RadError::UnsupportedSortOp {
```
tmpolaczyk: This is unreachable as long as map(array) returns an array.

rad/src/operators/array.rs
```diff
@@ -229,3 +276,326 @@ fn test_filter_negative() {
...
+    let output = sort(&input, &script).unwrap_err();
+
+    assert_eq!(
+        output.to_string(),
+        "ArraySort is not supported for RadonArray with inner type `RadonFloat`"
+    )
+}
+
+#[test]
+fn test_operate_sort_string_2_arrays() {
```
tmpolaczyk:
```suggestion
fn test_sort_string_2_arrays() {
```

tmpolaczyk: I think `operate` is used for tests which directly use the `operate()` function, can you remove it from all this test names?

PR #838, opened 2019-09-25
feat(radon): add RadonArray filter operator

mariocao:


rad/src/error.rs
```diff
@@ -28,6 +28,12 @@ pub enum RadError {
     /// The given key is not present in a RadonMap
     #[fail(display = "Failed to get key `{}` from RadonMap", key)]
     MapKeyNotFound { key: String },
+    /// The given subscript does not return RadonBoolean in an ArrayFilter
+    #[fail(
+        display = "ArrayFilter subscript output was not RadonBoolean (was `{}`)",
+        key
+    )]
+    ArrayFilterWrongSubscript { key: String },
```
tmpolaczyk:
```suggestion
    ArrayFilterWrongSubscript { value: String },
```

PR #842, opened 2019-09-26
feat(radon): include RadonInteger in AverageMean reducer

lrubiorod:


rad/src/reducers/average.rs
```diff
@@ -1,3 +1,4 @@
+use crate::types::integer::RadonInteger;
```
tmpolaczyk: fmt

tmpolaczyk:
Please add a test for mean of mixed types: mean(1, 1.0), mean(1.0, 1) and mean("foo", 1), etc.

PR #844, opened 2019-09-27
Add send-request command to node

tmpolaczyk:
With option to try serialized data requests locally:

    cargo run -- node send-request --run --hex=0aa0...

Additionally, renamed run_consensus to run_tally for consistency.


aesedepece:
This is great. Only objection from my side is about naming. I would favor `send-request` in accordance with the latest naming ("witnet requests" instead of "data requests").

PR #847, opened 2019-10-02
Update wallet details

anler:
## What

Add new endpoint `updateWallet` which can be used to modify a wallet's name and/or caption.

### Request

```json
{
  "params": {
    "caption": "My testing wallet",
    "name": "Testing",
    "sessionId": "273518861f799d2f6cd82caba7d2cd6d416dffbf4ab8c7a2550c31695557ac09",
    "walletId": "6c344625884c2f910065ab170dc18ad3cbbc03c7234507c7c22dbd78e3b26667"
  },
  "method": "updateWallet",
  "id": "1",
  "jsonrpc": "2.0"
}
```

### Response

```json
{
  "id": "1",
  "jsonrpc": "2.0",
  "result": true
}
```

### DIFF
--- a/crypto/src/key.rs
+++ b/crypto/src/key.rs
@@ -198,9 +198,9 @@ impl ExtendedSK {
         let (hrp, data) = bech32::decode(slip32).map_err(KeyError::deserialization_err)?;
 
         if hrp.as_str() != "xprv" {
-            Err(KeyError::Deserialization(failure::format_err!(
+            return Err(KeyError::Deserialization(failure::format_err!(
                 "does not start with xprv"
-            )))?
+            )));
         }
 
         let bytes: Vec<u8> =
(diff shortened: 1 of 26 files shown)


crypto/src/key.rs
```diff
@@ -139,6 +193,85 @@ impl ExtendedSK {
         }
     }
 
+    /// Create a new extended secret key from the given slip32-encoded string.
+    pub fn from_slip32(slip32: &str) -> Result<(Self, KeyPath), KeyError> {
+        let (hrp, data) = bech32::decode(slip32).map_err(KeyError::deserialization_err)?;
+
+        if hrp.as_str() != "xprv" {
+            return Err(KeyError::Deserialization(failure::format_err!(
+                "does not start with xprv"
```
aesedepece:
```suggestion
                "prefix is not \"xprv\""
```

PR #858, opened 2019-10-28
Implement reveal delay to avoid malleability

lrubiorod:


aesedepece:
Generally speaking, this looks great, and I'm specially surprised of the code surface of the change, which is indeed much smaller than we expected.

My only objection is about naming, sorry for nitpicking. I'd rather go with "reveal rounds" instead of "reveal stages".

tmpolaczyk:
What about the minimum and maximum values for `reveal_stages`?

In the current implementation the minimum is 1. We may raise it to 2 in the future because that's way more safe. Setting `reveal_stages` to 0 is currently accepted and treated the same as 1. Should we add a validation which makes data requests with `reveal_stages: 0` invalid? Or should we just make 0 mean minimum? Or maybe rename the field to `extra_reveal_stages` so 0 means minimum and 1 means minimum + 1?

As per the maximum, I don't see any big problems with allowing data requests to stay in reveal stage for long periods of time, but I also don't see why would anyone want to do that.

lrubiorod:
I like the idea to use this parameter as an extra rounds instead of number of rounds. So in case of 0, there are not extra reveal rounds required. I dont  expecified a maximum and minimum (so it means that it is one). So... would be there a extra validation for tally in case of not all reveals and not spent all the rounds specified by the request?

tmpolaczyk:
True, we need validations to ensure:
* not accepting incomplete tallies before the reveal round limit
* not accepting blocks which should have a tally but the miner forgot to include it, to enforce the reveal limit
* not accepting blocks with duplicate reveals (the same reveal in different blocks)

tmpolaczyk:
The data requests in `examples/` need to be updated with the new `extra_reveal_rounds` field

tmpolaczyk:
Are we sure that all our documentation is updated and says that commit_fee is the fee per commit and reveal fee is the fee per reveal?

lrubiorod:
> Are we sure that all our documentation is updated and says that commit_fee is the fee per commit and reveal fee is the fee per reveal?

Define "sure" :P

PR #860, opened 2019-10-31
Remove sleep 5 and improve PeersBeacons logic

tmpolaczyk:


tmpolaczyk:
In `impl Handler<PeersBeacons> for ChainManager` we check if we are in consensus with our outbound peers.

Before this PR, if one peer failed to send us a `CheckpointBeacon`, the `PeersBeacons` message was not sent and `ChainManager` went into `WaitingConsensus` state on the next epoch.

After this PR, if one peer fails to send us a `CheckpointBeacon`, we set it to `None` and send all the beacons to `ChainManager`. Now the consensus function operates on `Option<CheckpointBeacon>`, so it can count missing beacons. Therefore, there is consensus as long as there is a majority of peers that agree on the beacon.

PR #861, opened 2019-11-05
test: parse and run data requests in examples/ folder

tmpolaczyk:
This ensures that our examples always work.

* Added a test which iterates over the json files in `examples/` and tries to parse them as data requests.
* Added tests which execute the request radon scripts locally, with fake input data, but they have to be added manually.
* To avoid untested data requests, there is a hardcoded list of filenames and if that list does not match the filenames in `examples/`, the test fails.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ env:
     - RUST_BACKTRACE="1"
 
 before_script:
-  - just || curl -LSfs https://japaric.github.io/trust/install.sh | sh -s -- --git casey/just --target x86_64-unknown-linux-musl --to ~/.cargo/bin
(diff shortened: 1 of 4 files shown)


tests/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+#[derive(Debug, Deserialize)]
+struct JsonRpcRequest<'a, T> {
+    jsonrpc: &'a str,
+    method: &'a str,
+    id: Id<'a>,
+    params: T,
+}
+
+/*
+#[test]
```
lrubiorod: Should we remove these comments?

tmpolaczyk:
This PR used to fail in travis with a linker error related to rocksdb. The cause of that issue is unclear, but the fix is to remove the `tests/lib.rs` file, and move the tests somewhere else.

PR #863, opened 2019-11-06
Travis linker error

tmpolaczyk:
This is not the problem that made #861 to fail, but obviously is nice to have.


tmpolaczyk:
Superseded by #861

PR #866, opened 2019-11-07
feat: make nightly tests use an initialized storage

aesedepece:
This changes the nightly tests on GitHub Actions so it uses an initialized storage that is always the same.

### DIFF
--- a/.github/workflows/midnight.yml
+++ b/.github/workflows/midnight.yml
@@ -1,9 +1,6 @@
 name: Check every midnight
 
 on:
-  push:
-    branches:
-      - "*"
   schedule:
     - cron:  '0 0 * * *'
 


tmpolaczyk:
It didn't work

> 2019-11-07T10:18:30.9746246Z [36mnode_1    |[0m [2019-11-07T10:18:29Z INFO  witnet_node::signature_mngr] Generating and persisting a new master key for this node

aesedepece:
This is frozen until release of Testnet-5.1

aesedepece:
image

.github/workflows/midnight.yml
```diff
@@ -31,6 +31,11 @@ jobs:
           sudo apt-get install -y g++-5 cmake libcurl4-openssl-dev libelf-dev libdw-dev cmake gcc binutils-dev protobuf-compiler librocksdb-dev
           just || curl -LSfs https://japaric.github.io/trust/install.sh | sh -s -- --git casey/just --target x86_64-unknown-linux-musl --to ~/.cargo/bin
 
+      - name: Load persistent storage
+        run: |
+          curl -L https://github.com/witnet/witnet-rust/releases/download/0.5.0-rc1/witnet-rust-testnet-5-tests-storage.tar.gz --output ./storage.tar.gz
```
tmpolaczyk: This will always download the storage for version 0.5.0-rc1, is that what we want?

aesedepece: It's just that the storage file is attached to that specific release.
If next testnets are hard-forked from the genesis, we will need to produce a new clean storage, upload it somewhere else an update the URL here, but that's OK-ish for now I guess.

PR #867, opened 2019-11-07
feat(validations): add validations to duplicated commmits, reveals an…

### LINKED ISSUE #862: Validate the case of duplicated commits or reveals in a same block

lrubiorod:
…d tallies

Close #862

### DIFF
--- a/validations/src/tests.rs
+++ b/validations/src/tests.rs
@@ -2380,6 +2380,7 @@ fn reveal_valid_commitment() {
         error.unwrap_err().downcast::<TransactionError>().unwrap(),
         TransactionError::DuplicatedReveal {
             pkh: public_key.pkh(),
+            dr_pointer,
         }
     );
 }
@@ -2465,7 +2466,9 @@ fn dr_pool_with_dr_in_tally_stage_2_reveals(
     let dr_output = DataRequestOutput {
         witnesses: 2,
         reveal_fee: 50,
-        value: 1100,
+        commit_fee: 50,
+        tally_fee: 100,
+        value: 1300,
         data_request: example_data_request(),
         ..DataRequestOutput::default()
     };
@@ -2974,8 +2977,14 @@ static MILLION_TX_OUTPUT: &str =
 
 static LAST_BLOCK_HASH: &str = "62adde3e36db3f22774cc255215b2833575f66bf2204011f80c03d34c7c9ea41";
 
-fn test_block<F: FnMut(&mut Block) -> bool>(mut mut_block: F) -> Result<(), failure::Error> {
-    let dr_pool = DataRequestPool::default();
+fn test_block<F: FnMut(&mut Block) -> bool>(mut_block: F) -> Result<(), failure::Error> {
+    test_block_with_drpool(mut_block, DataRequestPool::default())
+}
+
+fn test_block_with_drpool<F: FnMut(&mut Block) -> bool>(
+    mut mut_block: F,
+    dr_pool: DataRequestPool,
+) -> Result<(), failure::Error> {
     let vrf = &mut VrfCtx::secp256k1().unwrap();
     let rep_eng = ReputationEngine::new(100);
     let mut utxo_set = UnspentOutputsPool::default();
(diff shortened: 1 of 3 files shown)


validations/src/validations.rs
```diff
@@ -736,7 +736,15 @@ pub fn validate_block_transactions(
     let mut co_mt = ProgressiveMerkleTree::sha256();
     let mut commits_number = HashMap::new();
     let block_beacon = block.block_header.beacon;
+    let mut commit_hs = HashSet::new();
     for transaction in &block.txns.commit_txns {
+        if commit_hs.contains(&transaction.hash()) {
```
tmpolaczyk: Instead of checking the transaction hash, this should check `(dr_pointer, pkh)`, to ensure that there is only one commit per pkh per data request.

tmpolaczyk: Same with reveals and tallies.

validations/src/tests.rs
```diff
@@ -3053,6 +3056,87 @@ fn test_block<F: FnMut(&mut Block) -> bool>(mut mut_block: F) -> Result<(), fail
     .map(|_| ())
```
tmpolaczyk: You can refactor the `test_block` function into:

    test_block_with_drpool(mut_block, DataRequestPool::default())

So we avoid some duplicated code

validations/src/tests.rs
```diff
@@ -3361,6 +3445,254 @@ fn block_add_vtt_but_dont_update_merkle_tree() {
...
+    // Insert valid proof
+    let mut cb = CommitTransactionBody::default();
+    cb.dr_pointer = dr_hash;
+    cb.proof =
+        DataRequestEligibilityClaim::create(vrf, &secret_key, commit_beacon, dr_hash).unwrap();
+    // Sign commitment
+    let cs = sign_t(&cb);
+    let c_tx = CommitTransaction::new(cb, vec![cs]);
+
+    let mut block_hash = Hash::default();
```
tmpolaczyk: I think this is unused now?

validations/src/tests.rs
```diff
@@ -3361,6 +3445,254 @@ fn block_add_vtt_but_dont_update_merkle_tree() {
...
+    let x = test_block_with_drpool(
+        |b| {
+            // We include a duplicate tally
+            b.txns.commit_txns.push(c_tx.clone());
+            b.txns.commit_txns.push(c_tx.clone());
```
tmpolaczyk: Aside from pushing the same commit twice, could you also test two different commits from the same pkh?

    let mut cb2 = cb.clone();
    cb2.commitment = Hash::SHA256([1; 32]);
    let cs = sign_t(&cb2);
    let c2_tx = CommitTransaction::new(cb2, vec![cs2]);

validations/src/tests.rs
```diff
@@ -3361,6 +3445,254 @@ fn block_add_vtt_but_dont_update_merkle_tree() {
...
+    };
+    let tally_transaction = TallyTransaction::new(dr_pointer, tally_value.clone(), vec![vt0, vt1]);
+
+    let mut block_hash = Hash::default();
+
+    let x = test_block_with_drpool(
+        |b| {
+            // We include a duplicate tally
+            b.txns.tally_txns.push(tally_transaction.clone());
+            b.txns.tally_txns.push(tally_transaction.clone());
```
tmpolaczyk: Same as with commits, I think it's possible to create a valid tally with a different hash, for example by reordering the transactions:

    let tally_transaction2 = TallyTransaction::new(dr_pointer, tally_value.clone(), vec![vt1, vt0]);

validations/src/validations.rs
```diff
@@ -746,6 +747,14 @@ pub fn validate_block_transactions(
             epoch,
             epoch_constants,
         )?;
+
+        // Validation for only one commit for pkh/data request in a block
+        let pkh = transaction.body.proof.proof.pkh();
+        if commit_hs.contains(&(dr_pointer, pkh)) {
```
tmpolaczyk: You can use `!commit_hs.insert()` here, because

> If the set did have this value present, false is returned.

validations/src/tests.rs
```diff
@@ -3361,6 +3370,253 @@ fn block_add_vtt_but_dont_update_merkle_tree() {
...
+    cb.proof =
+        DataRequestEligibilityClaim::create(vrf, &secret_key, commit_beacon, dr_hash).unwrap();
+    // Sign commitment
+    let cs = sign_t(&cb);
+    let c_tx = CommitTransaction::new(cb.clone(), vec![cs]);
+
+    let mut cb2 = cb;
+    cb2.commitment = Hash::SHA256([1; 32]);
+    let cs2 = sign_t(&cb2);
+    let c2_tx = CommitTransaction::new(cb2, vec![cs2]);
```
tmpolaczyk: Can you add

    assert_ne!(c_tx.hash(), c2_tx.hash());

Just to be sure?

validations/src/validations.rs
```diff
@@ -736,6 +736,7 @@ pub fn validate_block_transactions(
     let mut co_mt = ProgressiveMerkleTree::sha256();
     let mut commits_number = HashMap::new();
     let block_beacon = block.block_header.beacon;
+    let mut commit_hs = HashSet::new();
```
tmpolaczyk:
```suggestion
    let mut commit_hs = HashSet::with_capacity(block.txns.commit_txns.len());
```

PR #879, opened 2019-11-12
Test add transaction

### LINKED ISSUE #875: Add tests to transaction handlers

tmpolaczyk:
Add generic `TransactionPool::contains` method.
Changed the return type of `commit_contains` and `reveal_contains` from `bool` to `Result<bool, TransactionError>`, to handle the case where there are two commits with different hash but equal pkh for the same data request. We only accept the first one and ignore the rest as duplicates. Same with reveals.
This fixes an issue where one commit can overwrite another and we enter an infinite loop since the received commit is always considered new.

Move `AddTransaction` logic from handler to its own method in `chain_manager/mod.rs`
And move generic transaction validation function to validations crate.
Add tests for handling commits and reveals with no signatures.
Add tests for handling invalid transaction types.

Close #875

### DIFF
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -21,6 +21,7 @@ use witnet_reputation::{ActiveReputationSet, TotalReputationSet};
 use witnet_util::parser::parse_hex;
 
 use crate::chain::Signature::Secp256k1;
+use crate::error::TransactionError;
 use crate::{
     data_request::DataRequestPool,
     error::{
@@ -923,6 +924,38 @@ impl TransactionsPool {
         self.re_transactions.clear();
     }
 
+    /// Returns `Ok(true)` if the pool already contains this transaction.
+    /// Returns an error if:
+    /// * The transaction is of an invalid type (mint or tally)
+    /// * The commit transaction has the same data request pointer and pkh as
+    /// an existing transaction, but different hash.
+    /// * The reveal transaction has the same data request pointer and pkh as
+    /// an existing transaction, but different hash.
+    pub fn contains(&self, transaction: &Transaction) -> Result<bool, TransactionError> {
+        let tx_hash = transaction.hash();
+
+        match transaction {
+            Transaction::ValueTransfer(_vt) => Ok(self.vt_contains(&tx_hash)),
+            Transaction::DataRequest(_drt) => Ok(self.dr_contains(&tx_hash)),
+            Transaction::Commit(ct) => {
+                let dr_pointer = ct.body.dr_pointer;
+                let pkh = ct.body.proof.proof.pkh();
+
+                self.commit_contains(&dr_pointer, &pkh, &tx_hash)
+            }
+            Transaction::Reveal(rt) => {
+                let dr_pointer = rt.body.dr_pointer;
+                let pkh = rt.body.pkh;
+
+                self.reveal_contains(&dr_pointer, &pkh, &tx_hash)
+            }
+            // Tally and mint transaction only exist inside blocks, it should
+            // be impossible for nodes to broadcast these kinds of transactions.
+            Transaction::Tally(_tt) => Err(TransactionError::NotValidTransaction),
+            Transaction::Mint(_mt) => Err(TransactionError::NotValidTransaction),
+        }
+    }
+
     /// Returns `true` if the pool contains a value transfer
     /// transaction for the specified hash.
     ///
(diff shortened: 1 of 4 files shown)


data_structures/src/chain.rs
```diff
@@ -923,6 +923,34 @@ impl TransactionsPool {
...
+    pub fn contains(&self, transaction: &Transaction) -> bool {
+        let tx_hash = transaction.hash();
+
+        match transaction {
+            Transaction::ValueTransfer(_vt) => self.vt_contains(&tx_hash),
+            Transaction::DataRequest(_drt) => self.dr_contains(&tx_hash),
+            Transaction::Commit(ct) => {
+                let dr_pointer = ct.body.dr_pointer;
+
+                if let Some(pkh) = &ct.signatures.get(0).map(|x| x.public_key.pkh()) {
```
tmpolaczyk: Maybe it would be better to use:

    let pkh = ct.body.proof.proof.pkh();

lrubiorod: it should be the same

data_structures/src/chain.rs
```diff
@@ -2034,4 +2064,137 @@ mod tests {
...
+        assert_eq!(transactions_pool.contains(&transaction), false);
+    }
+
+    #[test]
+    fn transactions_pool_contains_commit_same_pkh() {
+        let c1 = Hash::SHA256([1; 32]);
+        let c2 = Hash::SHA256([2; 32]);
+        let t1 = Transaction::Commit(CommitTransaction {
+            body: CommitTransactionBody::new(Default::default(), c1, Default::default()),
+            signatures: vec![KeyedSignature {
```
lrubiorod:
```suggestion
            signatures: vec![KeyedSignature::default()]
```

data_structures/src/chain.rs
```diff
@@ -2034,4 +2064,137 @@ mod tests {
...
+                signature: Signature::default(),
+                public_key: Default::default(),
+            }],
+        });
+        let t2 = Transaction::Commit(CommitTransaction {
+            body: CommitTransactionBody::new(Default::default(), c2, Default::default()),
+            signatures: vec![KeyedSignature {
```
lrubiorod: same here

data_structures/src/chain.rs
lrubiorod: the same in the other tests

node/src/actors/chain_manager/mod.rs
```diff
@@ -769,3 +840,133 @@ fn show_info_dr(data_request_pool: &DataRequestPool, block: &Block) {
         );
     }
 }
+
+#[cfg(test)]
+mod tests {
+    pub use super::*;
+    use witnet_data_structures::chain::{
```
lrubiorod: format imports?

node/src/actors/chain_manager/mod.rs
```diff
@@ -769,3 +840,133 @@ fn show_info_dr(data_request_pool: &DataRequestPool, block: &Block) {
...
+        cm.sm_state = StateMachine::Synced;
+        cm.current_epoch = Some(0);
+        cm.epoch_constants = Some(EpochConstants {
+            checkpoint_zero_timestamp: 0,
+            checkpoints_period: 1,
+        });
+        cm.vrf_ctx = Some(VrfCtx::secp256k1().unwrap());
+        cm.chain_state.chain_info = Some(ChainInfo {
+            environment: Environment::Testnet1,
+            consensus_constants: ConsensusConstants {
```
lrubiorod: should we add Default trait for ConsensusConstants?

node/src/actors/chain_manager/mod.rs
```diff
@@ -769,3 +840,133 @@ fn show_info_dr(data_request_pool: &DataRequestPool, block: &Block) {
...
+        cm.chain_state.reputation_engine = Some(ReputationEngine::new(100));
+
+        cm
+    }
+
+    #[test]
+    fn add_commit_no_signatures() {
+        let mut cm = empty_chain_manager();
+        let msg = AddTransaction {
+            transaction: Transaction::Commit(CommitTransaction {
```
```suggestion
            transaction: Transaction::Commit(CommitTransaction::default())
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -769,3 +840,133 @@ fn show_info_dr(data_request_pool: &DataRequestPool, block: &Block) {
...
+                hash: Default::default(),
+            }
+        );
+    }
+
+    #[test]
+    fn add_reveal_no_signatures() {
+        let mut cm = empty_chain_manager();
+        let msg = AddTransaction {
+            transaction: Transaction::Reveal(RevealTransaction {
```
```suggestion
            transaction: Transaction::Reveal(RevealTransaction::default())
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -769,3 +840,133 @@ fn show_info_dr(data_request_pool: &DataRequestPool, block: &Block) {
...
+            TransactionError::NotValidTransaction
+        );
+    }
+
+    #[test]
+    fn add_tally_transaction() {
+        // Tally transactions when received outside of a block are ignored
+        let mut cm = empty_chain_manager();
+        let msg = AddTransaction {
+            transaction: Transaction::Tally(TallyTransaction::new(
```
```suggestion
            transaction: Transaction::Tally(TallyTransaction::default())
```

node/src/actors/chain_manager/mod.rs
tmpolaczyk: Probably no because the values are different in testnet and mainnet. But the default values are stored in [config/src/defaults.rs](https://github.com/witnet/witnet-rust/blob/master/config/src/defaults.rs), so it should be possible to create methods like `ConsensusConstants::testnet` or `ConsensusConstants::mainnet`.

Issue #869, opened 2019-11-11
Implement out-of-consensus witnessing slashing


mariocao:
Also known as "Punish liars"

aesedepece:
This is already unlocked as stage 3 from #880 is complete.

PR #880, opened 2019-11-12
Add support for RADON errors

### LINKED ISSUE #873: Implement RADON errors
@aesedepece says:

> ## Problem statement
> - There are some circumpstances under which a request may not be resolvable by the Witnet network.
> - Those situations may be caused most times by poorly designed requests.
> - Even if designing the requests properly is absolutely up to the requester, we need to provide reasonable UX for handling those situations and recovering from them (which may be paramount for the security and liveness of consumer smart contracts).
>
> ## Current state
> - The `Witnet.Result` Solidity library expects `witnet-rust` to produce errors in the form of `CBOR` positive integers that are tagged with the `37` identifier, which tells that they need to be handled as an error code, not a `RadonInteger` result.
> - However, `witnet-rust` is not producing error codes yet. Upon runtime failures, it will fail silently with no further action.
>
> ## Proposed changes
> - Define all the possible error codes, which need to be standardized at the protocol level as it may be potentially consensus-critical (e.g. 100% agreement of witnesses on the fact that an API failed).
> - Change the general exceptions handler in the `witnet_rad` executor / engine so that it returns a `37`-tagged CBOR value containing the correct error code.
> - Ideally, come with a solution for running integration tests that prove full compatibility of errors produced by `witnet-rust` and errors decoded by the Solidity library.

aesedepece:
This will fix #873 (you can find a proper description of the issue right there).

I'm approaching this issue in 4 different stages
1. make `witnet_rad` public methods use `RadonTypes` instead of `Vec<u8>` and adapt all dependent crates so they use the new API.
2. implement `Report` type and make rad executor methods use it.
3. modify `Operable` trait so it has a contextful method for some operators to put metadata into a structure that can be later used for constructing the `Report`.
4. put errors into commit / reveal / tally transactions.


aesedepece:
Only stages 1-2-3 blocked #869 and those are already complete :muscle:

aesedepece:
Yet to be done:
- Be able to decode CBOR errors (`37`-tagged) into `RadResult`.
- Fix all `FIXME`s
- Fix all broken tests
- Add new tests

data_structures/src/radon_error.rs
```diff
@@ -0,0 +1,147 @@
...
+#[repr(u8)]
+/// List of RADON-level errors.
+/// **WARNING: these codes are consensus-critical.** They can be renamed but they cannot be
+/// re-assigned without causing a non-backwards-compatible protocol upgrade.
+pub enum RadonErrors {
+    /// Unknown error. Something went really bad!
+    Unknown = 0x00,
+    // Script format errors
+    /// At least one of the source scripts is not a valid CBOR-encoded value.
+    SourceScriptNotCBOR = 0x01,
```
tmpolaczyk: How about unifying this with `SourceScriptNotArray`?

```suggestion
    SourceScriptNotCBORArray = 0x01,
```

data_structures/src/radon_error.rs
```diff
@@ -0,0 +1,147 @@
...
+    /// The CBOR value decoded from a source script is not an Array.
+    SourceScriptNotArray = 0x2,
```
lrubiorod:
```suggestion
    SourceScriptNotArray = 0x02,
```
TOC

data_structures/src/radon_error.rs
```diff
@@ -0,0 +1,147 @@
...
+    /// The Array value decoded form a source script is not a valid RADON script.
+    SourceScriptNotRADON = 0x3,
```
lrubiorod:
```suggestion
    SourceScriptNotRADON = 0x03,
```
Same :P

wallet/src/repository/wallet/mod.rs
```diff
@@ -810,8 +810,7 @@ where
 fn convert_block_epoch_to_timestamp(epoch: Epoch) -> i64 {
     // FIXME: we need EpochConstants to convert between epochs and timestamps
     // In the meanwhile, just return the epoch as the timestamp
-
-    epoch as i64
+    i64::from(epoch)
```
tmpolaczyk: Strange, clippy didn't complain to me. Have you run `rustup update`?

data_structures/src/data_request.rs
```diff
@@ -367,14 +377,19 @@ pub fn create_tally(
         outputs.push(vt_output_change);
     }
 
-    TallyTransaction::new(dr_pointer, consensus_tally, outputs)
+    // FIXME: make this factory return a Result, as this line could easily fail and right now we are
```
lrubiorod: Now or later?

tmpolaczyk:
Commit 8ee96da 
feat: add support for `RadonArray::flatten`

should be in a different PR, unless it is blocking something related to this PR.

data_structures/src/data_request.rs
```diff
@@ -1,15 +1,19 @@
-use log::{debug, info};
 use std::collections::{BTreeMap, HashMap, HashSet};
+use std::convert::TryFrom;
 
+use log::{debug, info};
+use serde::{Deserialize, Serialize};
+
+use crate::radon_report::TypeLike;
```
tmpolaczyk: Formatting

data_structures/src/data_request.rs
```diff
@@ -367,14 +377,19 @@ pub fn create_tally(
         outputs.push(vt_output_change);
     }
 
-    TallyTransaction::new(dr_pointer, consensus_tally, outputs)
+    // FIXME: make this factory return a Result, as this line could easily fail and right now we are
+    //  ignoring failures and putting empty vectors (which will fail when decoded)
+    let tally_bytes = Vec::try_from(report).unwrap_or_default();
```
tmpolaczyk: Why not fail immediately?

```suggestion
    let tally_bytes = Vec::try_from(report).unwrap();
```

data_structures/src/radon_report.rs
```diff
@@ -0,0 +1,183 @@
...
+    /// A positional vector of "truthers" and "liars", i.e. reveals that passed all the filters vs.
+    /// those which were filtered out.
+    /// This follows a reverse logic: `false` is truth and `true` is lie.
+    liars: Vec<bool>,
+    /// Proportion between total reveals and "truthers" count:
+    /// `reveals.len() / liars.iter().filter(std::not::Ops).count()`
+    consensus: f32,
+}
+
+#[test]
```
tmpolaczyk: Put tests inside the `mod tests` boilerplate

node/src/actors/chain_manager/mining.rs
```diff
@@ -368,8 +373,14 @@ impl ChainManager {
                 .map(|(dr_pointer, reveals, dr_state)| {
                     debug!("Building tally for data request {}", dr_pointer);
 
-                    let results: Vec<Vec<u8>> =
-                        reveals.iter().map(|r| r.body.reveal.clone()).collect();
+                    // FIXME: parallelize decoding of reveals.
+                    // FIXME: do not ignore failed decoding of faulty reveals.
```
tmpolaczyk: Do we have any tests for non-deserializable reveals?

node/src/actors/rad_manager/handlers.rs
```diff
@@ -1,31 +1,45 @@
 //! Message handlers for `RadManager`
+
 use actix::{Handler, Message};
-use std::convert::TryFrom;
+use rayon::iter::{IntoParallelRefIterator, ParallelIterator};
```
tmpolaczyk: You probably want

```suggestion
use rayon::prelude::*;
```

rad/src/operators/array.rs
```diff
@@ -1,15 +1,18 @@
+use std::clone::Clone;
+use std::convert::TryFrom;
+
+use rayon::iter::{IntoParallelRefIterator, ParallelIterator};
+use serde_cbor::value::{from_value, Value};
+
+use crate::script::execute_contextfree_radon_script;
```
tmpolaczyk: fmt

rad/src/operators/array.rs
```diff
@@ -91,7 +99,7 @@ pub fn filter(input: &RadonArray, args: &[Value]) -> Result<RadonTypes, RadError
     let first_arg = args.get(0).ok_or_else(wrong_args)?;
     match first_arg {
         Value::Integer(arg) => {
-            let filter_code = RadonFilters::from_i128(*arg).ok_or_else(wrong_args)?;
+            let filter_code = RadonFilters::try_from(*arg as u8).map_err(|_| wrong_args())?;
```
tmpolaczyk: The conversion to `u8` should also be a try_from:
```suggestion
            let filter_code = RadonFilters::try_from(u8::try_from(*arg)?).map_err(|_| wrong_args())?;
```

data_structures/src/radon_error.rs
```diff
@@ -0,0 +1,147 @@
...
+    /// Allow CBOR encoding of `RadonError` structures.
+    pub fn encode(&self) -> Result<Vec<u8>, IE> {
+        let mut encoder = GenericEncoder::new(Cursor::new(Vec::new()));
+        encoder.value(&Value::from(self))?;
+
+        Ok(encoder.into_inner().into_writer().into_inner())
+    }
+}
+
+// FIXME: merge `RadonError::encode` into this.
```
lrubiorod: now or later

rad/src/types/mod.rs
```diff
@@ -156,51 +190,138 @@ impl TryFrom<Value> for RadonTypes {
             Value::Map(_) => RadonMap::try_from(value).map(Into::into),
             Value::Text(_) => RadonString::try_from(value).map(Into::into),
             Value::Integer(_) => RadonInteger::try_from(value).map(Into::into),
-            _ => Ok(RadonBytes::from(value).into()),
+            Value::Bytes(_) => Ok(RadonTypes::from(RadonBytes::from(value))),
+            Value::Null => Err(RadError::Decode {
+                from: String::from("serde_cbor::Value::Null"),
+                to: String::from("RadonTypes"),
+            }),
+            Value::__Hidden => Err(RadError::Decode {
```
tmpolaczyk: I think you are not supposed to use that variant, and instead use a wildcard `_`. This allows for updates of the cbor library not breaking our builds.

```suggestion
            _ => Err(RadError::Decode {
```

data_structures/src/radon_report.rs
```diff
@@ -0,0 +1,183 @@
...
+    pub completion_time: Option<Instant>,
+    pub start_time: Option<Instant>,
+    pub script_index: Option<u8>,
+}
+
+/// Implementation of convenience methods for `ReportContext`
+impl ReportContext {
+    /// Set start time.
+    pub fn start(&mut self) {
+        self.start_time = Some(Instant::now());
```
lrubiorod: should we use NTP time?

tmpolaczyk: No, this is only used to calculate a duration. We don't care about the exact start time.

data_structures/src/radon_report.rs
```diff
@@ -0,0 +1,183 @@
...
+
+/// Tally-specific metadata structure.
+#[derive(Clone, Debug, Default)]
+pub struct TallyMetaData {
+    /// A positional vector of "truthers" and "liars", i.e. reveals that passed all the filters vs.
+    /// those which were filtered out.
+    /// This follows a reverse logic: `false` is truth and `true` is lie.
+    liars: Vec<bool>,
+    /// Proportion between total reveals and "truthers" count:
+    /// `reveals.len() / liars.iter().filter(std::not::Ops).count()`
```
lrubiorod:
```suggestion
    /// `liars.iter().filter(std::not::Ops).count() / reveals.len() `
```

node/src/actors/chain_manager/mining.rs
tmpolaczyk: No, we do not have tests. Created #894

node/src/actors/chain_manager/mining.rs
```diff
@@ -368,8 +374,22 @@ impl ChainManager {
...
-                    let results: Vec<Vec<u8>> =
-                        reveals.iter().map(|r| r.body.reveal.clone()).collect();
+                    // Ignore reveals that could not be decoded.
+                    //  Those reveal that cannot be decoded are most likely malformed and therefore
+                    //  their authors are punished in the same way as non-revealers.
+                    let results: Vec<RadonTypes> = reveals.clone()
+                        .into_par_iter()
+                        .filter_map(|reveal_tx| {
+                            let slice = reveal_tx.body.reveal.as_slice();
+                            match RadonTypes::try_from(slice) {
```
tmpolaczyk: This logic should be inside `RadManager`. Maybe the `RunTally` message can accept the input as `Vec<u8>` and do the decoding?

aesedepece:
Merge request @witnet/devs

rad/src/types/mod.rs
```diff
@@ -325,3 +327,21 @@ impl TryFrom<&cbor::value::Value> for RadonTypes {
         }
     }
 }
+
+/// Decode a vector of instances of RadonTypes from an already mapped iterator that yields
+/// `(&[u8], &T)`. The `err_action` argument allows the caller of this function to decide whether
+/// it should act in a lossy way, i.e. ignoring items that cannot be decoded or replacing them with
+/// default values.
+pub fn serial_iter_decode<T>(
+    iter: MappedU8Iter<T>,
```
tmpolaczyk: You should avoid using concrete iterator types and instead use a generic `I: IntoIterator<Item=(&[u8], &T)>`

PR #883, opened 2019-11-14
feat(utils): use NTP timestamps to avoid sync problems

### LINKED ISSUE #874: Implement NTP (Network Time Protocol) synchronization

lrubiorod:
Close #874

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -103,6 +103,11 @@ pub struct Config {
     #[partial_struct(ty = "PartialLog")]
     #[partial_struct(serde(default))]
     pub log: Log,
+
+    /// Ntp-related configuration
+    #[partial_struct(ty = "PartialNtp")]
+    #[partial_struct(serde(default))]
+    pub ntp: Ntp,
 }
 
 /// Log-specific configuration.
@@ -276,6 +281,20 @@ pub struct Mining {
     pub enabled: bool,
 }
 
+/// NTP-related configuration
+#[derive(PartialStruct, Debug, Clone, PartialEq)]
+#[partial_struct(derive(Deserialize, Default, Debug, Clone, PartialEq))]
+pub struct Ntp {
+    /// Period that indicate the validity of a ntp timestamp
+    pub update_period: Duration,
+
+    /// Server to query ntp information
+    pub servers: Vec<String>,
+
+    /// Enable NTP for clock synchronization
+    pub enabled: bool,
+}
+
 impl Config {
     pub fn from_partial(config: &PartialConfig) -> Self {
         let defaults: &dyn Defaults = match config.environment {
@@ -317,6 +336,7 @@ impl Config {
             mining: Mining::from_partial(&config.mining, defaults),
             wallet: Wallet::from_partial(&config.wallet, defaults),
             rocksdb: Rocksdb::from_partial(&config.rocksdb, defaults),
+            ntp: Ntp::from_partial(&config.ntp, defaults),
         }
     }
 }
@@ -472,6 +492,25 @@ impl Mining {
                 .enabled
                 .to_owned()
                 .unwrap_or_else(|| defaults.mining_enabled()),
+        }
+    }
+}
+
+impl Ntp {
+    pub fn from_partial(config: &PartialNtp, defaults: &dyn Defaults) -> Self {
+        Ntp {
+            update_period: config
+                .update_period
+                .to_owned()
+                .unwrap_or_else(|| defaults.ntp_update_period()),
+            servers: config
+                .servers
+                .clone()
+                .unwrap_or_else(|| defaults.ntp_server()),
+            enabled: config
+                .enabled
+                .to_owned()
+                .unwrap_or_else(|| defaults.ntp_enabled()),
         }
     }
 }
--- a/config/src/defaults.rs
+++ b/config/src/defaults.rs
@@ -204,6 +204,23 @@ pub trait Defaults {
     fn rocksdb_enable_statistics(&self) -> bool {
         false
     }
+
+    fn ntp_update_period(&self) -> Duration {
+        Duration::from_secs(600)
+    }
+
+    fn ntp_server(&self) -> Vec<String> {
+        vec![
+            "0.pool.ntp.org:123".to_string(),
+            "1.pool.ntp.org:123".to_string(),
+            "2.pool.ntp.org:123".to_string(),
+            "3.pool.ntp.org:123".to_string(),
+        ]
+    }
+
+    fn ntp_enabled(&self) -> bool {
+        true
+    }
 }
 
 /// Struct that will implement all the mainnet defaults
--- a/node/src/actors/epoch_manager/mod.rs
+++ b/node/src/actors/epoch_manager/mod.rs
@@ -11,7 +11,9 @@ use witnet_data_structures::{
     chain::{Epoch, EpochConstants},
     error::EpochCalculationError,
 };
-use witnet_util::timestamp::{duration_between_timestamps, get_timestamp, get_timestamp_nanos};
+use witnet_util::timestamp::{
+    duration_between_timestamps, get_timestamp, get_timestamp_nanos, update_global_timestamp,
+};
 
 use crate::actors::messages::{EpochNotification, EpochResult};
 use crate::config_mngr;
@@ -126,6 +128,13 @@ impl EpochManager {
                 // Start checkpoint monitoring process
                 actor.checkpoint_monitor(ctx);
 
+                // Start ntp update process
+                if config.ntp.enabled {
+                    let ntp_addr = config.ntp.servers[0].clone();
+                    update_global_timestamp(ntp_addr.as_str());
+                    actor.update_ntp_timestamp(ctx, config.ntp.update_period, ntp_addr);
+                }
+
                 fut::ok(())
             })
             .map_err(|err, _, _| {
@@ -165,9 +174,14 @@ impl EpochManager {
                     Err(_) => return,
                 };
 
+                let last_checked_epoch = act.last_checked_epoch.unwrap_or(0);
+
                 // Send message to actors which subscribed to all epochs
-                for subscription in &mut act.subscriptions_all {
-                    subscription.send_notification(current_epoch);
+                if current_epoch > last_checked_epoch {
+                    for subscription in &mut act.subscriptions_all {
+                        // Only send new epoch notification
+                        subscription.send_notification(current_epoch);
+                    }
                 }
 
                 // Get all the checkpoints that had some subscription but were skipped for some
@@ -175,7 +189,7 @@ impl EpochManager {
                 // resources to execute in time...)
                 let epoch_checkpoints: Vec<_> = act
                     .subscriptions_epoch
-                    .range(act.last_checked_epoch.unwrap_or(0)..=current_epoch)
+                    .range(last_checked_epoch..=current_epoch)
                     .map(|(k, _v)| *k)
                     .collect();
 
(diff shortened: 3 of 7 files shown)


util/Cargo.toml
```diff
@@ -8,3 +8,11 @@ workspace = ".."
 [dependencies]
 chrono = "0.4.9"
 failure = "0.1.6"
+lazy_static = "1.4.0"
+log = "0.4.8"
+ntp = {git="https://github.com/JeffBelgum/ntp/"}
```
tmpolaczyk: Fork?

util/Cargo.toml
```diff
@@ -8,3 +8,11 @@ workspace = ".."
...
+time = "0.1.42"
+
+[dependencies.serde]
```
tmpolaczyk: I prefer the alternative syntax:

    serde = { version = "1.0.101", features = ["derive"] }

util/src/timestamp.rs
```diff
@@ -1,22 +1,120 @@
...
     let utc: DateTime<Utc> = Utc::now();
 
-    // Return number of non-leap seconds since Unix epoch and the number of nanoseconds since the last second boundary
-    (utc.timestamp(), utc.timestamp_subsec_nanos())
+    let utc_secs = utc.timestamp();
+    let utc_subsec_nanos = utc.timestamp_subsec_nanos();
+
+    let mut old_ts = get_global_timestamp();
+
+    if utc_secs - old_ts.last_ntp.0 > old_ts.update_period {
```
tmpolaczyk: I would add a comment, something like

    // If the last request to a NTP server was more than update_period seconds ago,
    // query the server for a new timestamp

tmpolaczyk: And I think all the logic inside this "if" should be a separate function, for example `update_global_timestamp()`.

util/src/timestamp.rs
```diff
@@ -1,22 +1,120 @@
...
+    let utc_dur = Duration::new(utc_secs as u64, utc_subsec_nanos);
+    let result = if old_ts.bigger {
+        utc_dur.checked_add(old_ts.diff)
+    } else {
+        utc_dur.checked_sub(old_ts.diff)
+    };
+
+    match result {
+        Some(x) => (x.as_secs() as i64, x.subsec_nanos()),
+        None => (utc_secs, utc_subsec_nanos),
```
tmpolaczyk: This `None` means that there has been an overflow while calculating the timestamp. I think it is better to panic than to return a potentially invalid timestamp.

util/src/timestamp.rs
```diff
@@ -1,22 +1,120 @@
...
+    }
+}
+
+fn local_time(timestamp: ntp::protocol::TimestampFormat) -> chrono::DateTime<chrono::Local> {
+    let unix_time = ntp::unix_time::Instant::from(timestamp);
+    chrono::Local.timestamp(unix_time.secs(), unix_time.subsec_nanos() as _)
+}
+/// Get NTP timestamp from an addr specified
+pub fn get_timestamp_ntp(addr: &str) -> Option<(i64, u32)> {
+    ntp::request(addr)
```
tmpolaczyk: What's the timeout for this request?

lrubiorod: 5 secs

tmpolaczyk:
I think it would be better to move the NTP updating mechanism outside of the get_timestamp function. The main issue with the current approach is the timeout: a call to get_timestamp can block for up to 5 seconds.

We could have a periodic callback, for example in EpochManager, which would call update_global_timestamp. This way all the calls to get_timestamp have more or less the same overhead. It would be really nice if that function was async, so that it doesn't block the executor thread, but the NTP crate used here does not allow that. I don't know if there are other crates with that feature, but we could create a new issue for replacing the NTP request with an async one.

config/src/config.rs
```diff
@@ -276,6 +281,17 @@ pub struct Mining {
...
 
+/// Ntp-related configuration
+#[derive(PartialStruct, Debug, Clone, PartialEq)]
+#[partial_struct(derive(Deserialize, Default, Debug, Clone, PartialEq))]
+pub struct Ntp {
+    /// Period that indicate the validity of a ntp timestamp
+    pub update_period: i64,
+
+    /// Server to query ntp information
+    pub server: String,
```
tmpolaczyk: Even if it is not implemented, the idea is to support multiple NTP servers. We can make this option an array and just use the first one for the moment.

```suggestion
    pub servers: Vec<String>,
```

node/src/actors/epoch_manager/mod.rs
```diff
@@ -126,6 +129,13 @@ impl EpochManager {
                 // Start checkpoint monitoring process
                 actor.checkpoint_monitor(ctx);
 
+                // Start ntp update process
+                let mut global_ts = get_global_timestamp();
+                global_ts.server = config.ntp.server.clone();
+                set_global_timestamp(global_ts);
```
tmpolaczyk: There is a race condition here because the global timestamp could change between the call to get and the call to set (if a different thread calls set).

This set should only modify global_ts.server, not the entire structure. So instead of `set`, a better API would be `get_mut`.

node/src/actors/epoch_manager/mod.rs
```diff
@@ -165,17 +175,22 @@ impl EpochManager {
                     Err(_) => return,
                 };
 
+                let last_checked_epoch = act.last_checked_epoch.unwrap_or(0);
+
                 // Send message to actors which subscribed to all epochs
                 for subscription in &mut act.subscriptions_all {
-                    subscription.send_notification(current_epoch);
+                    if current_epoch > last_checked_epoch {
```
tmpolaczyk: This if can go outside the for.

util/src/timestamp.rs
```diff
@@ -1,22 +1,106 @@
...
+            old_ts.bigger = false;
+        }
+    } else {
+        log::warn!("Update global timestamp failed. Ntp request not achieve");
+    }
+}
+
+fn local_time(timestamp: ntp::protocol::TimestampFormat) -> chrono::DateTime<chrono::Local> {
+    let unix_time = ntp::unix_time::Instant::from(timestamp);
+    chrono::Local.timestamp(unix_time.secs(), unix_time.subsec_nanos() as _)
```
tmpolaczyk:
```suggestion
    chrono::Local.timestamp(unix_time.secs(), unix_time.subsec_nanos() as u32)
```

config/src/config.rs
```diff
@@ -276,6 +281,17 @@ pub struct Mining {
     pub enabled: bool,
 }
 
+/// Ntp-related configuration
```
tmpolaczyk:
```suggestion
/// NTP-related configuration
```

And we should probably explain somewhere what is NTP and why it's useful.

config/src/config.rs
```diff
@@ -276,6 +281,17 @@ pub struct Mining {
...
+#[derive(PartialStruct, Debug, Clone, PartialEq)]
+#[partial_struct(derive(Deserialize, Default, Debug, Clone, PartialEq))]
+pub struct Ntp {
+    /// Period that indicate the validity of a ntp timestamp
+    pub update_period: i64,
+
+    /// Server to query ntp information
+    pub server: Vec<String>,
```
tmpolaczyk:
```suggestion
    pub servers: Vec<String>,
```

config/src/config.rs
```diff
@@ -276,6 +281,17 @@ pub struct Mining {
     pub enabled: bool,
 }
 
+/// Ntp-related configuration
+#[derive(PartialStruct, Debug, Clone, PartialEq)]
+#[partial_struct(derive(Deserialize, Default, Debug, Clone, PartialEq))]
+pub struct Ntp {
+    /// Period that indicate the validity of a ntp timestamp
+    pub update_period: i64,
```
tmpolaczyk: Maybe change this field from i64 to Duration, for consistency? But I see that there are other fields which represent a time duration but are i64s...

    #[partial_struct(serde(
        default,
        deserialize_with = "from_secs",
        rename = "feeler_peers_period_seconds"
    ))]
    pub feeler_peers_period: Duration,

config/src/defaults.rs
```diff
@@ -204,6 +204,14 @@ pub trait Defaults {
     fn rocksdb_enable_statistics(&self) -> bool {
         false
     }
+
+    fn ntp_update_period(&self) -> i64 {
+        600
+    }
+
+    fn ntp_server(&self) -> Vec<String> {
+        vec!["0.pool.ntp.org:123".to_string()]
```
tmpolaczyk: It is recommended to use at least 4 NTP servers, let's add a few more.

node/src/actors/epoch_manager/mod.rs
```diff
@@ -126,6 +129,14 @@ impl EpochManager {
                 // Start checkpoint monitoring process
                 actor.checkpoint_monitor(ctx);
 
+                // Start ntp update process
+                {
+                    let mut global_ts = get_mut_global_timestamp();
+                    global_ts.server = config.ntp.server[0].clone();
```
tmpolaczyk: This will panic if there are no NTP servers in the config. Maybe in that case we should disable NTP?

util/src/timestamp.rs
```diff
@@ -1,22 +1,106 @@
...
+/// NTP Timestamp information
+#[derive(Debug, Default, Serialize, Deserialize, PartialEq, Clone)]
+pub struct TimestampNTP {
+    /// Difference between NTP and system timestamp
+    pub ntp_diff: Duration,
+    /// Flag to indicate if NTP is bigger or smaller
+    /// than system timestamp
+    pub bigger: bool,
+    /// Address to require NTP timestamp information
+    pub server: String,
```
tmpolaczyk: Maybe we can remove this field from here and add `server` as a parameter to `update_global_timestamp`?

config/src/defaults.rs
lrubiorod: current library can not manage more than one library

config/src/config.rs
lrubiorod: in witnet.io documentation?

node/src/actors/epoch_manager/mod.rs
lrubiorod: If there are not ntp servers in config file, default settings will be used

tmpolaczyk: I mean if someone explicitly sets

    servers = []

lrubiorod: Why would do that?

config/src/config.rs
```diff
@@ -276,6 +281,20 @@ pub struct Mining {
...
+#[derive(PartialStruct, Debug, Clone, PartialEq)]
+#[partial_struct(derive(Deserialize, Default, Debug, Clone, PartialEq))]
+pub struct Ntp {
+    /// Period that indicate the validity of a ntp timestamp
+    pub update_period: Duration,
+
+    /// Server to query ntp information
+    pub servers: Vec<String>,
+
+    /// Binary flag telling whether to enable the NTP Timestamp or not
```
tmpolaczyk:
```suggestion
    /// Enable NTP for clock synchronization?
```

node/src/actors/epoch_manager/mod.rs
```diff
@@ -209,6 +223,19 @@ impl EpochManager {
             },
         );
     }
+
+    /// Method to monitor checkpoints and execute some actions on each
+    fn update_ntp_timestamp(&self, ctx: &mut Context<Self>, period: Duration, addr: String) {
+        // Wait until next checkpoint to execute the periodic function
+        ctx.run_later(period, move |act, ctx| {
+            // Get current epoch
```
tmpolaczyk:
```suggestion
```

util/Cargo.toml
```diff
@@ -8,3 +8,8 @@ workspace = ".."
 [dependencies]
 chrono = "0.4.9"
 failure = "0.1.6"
+lazy_static = "1.4.0"
+log = "0.4.8"
+ntp = {git="https://github.com/witnet/ntp/"}
```
tmpolaczyk: Whitespace
```suggestion
ntp = { git = "https://github.com/witnet/ntp" }
```

config/src/config.rs
tmpolaczyk: Ok, created https://github.com/witnet/documentation/issues/20

PR #886, opened 2019-11-19
feat(session): do not allow node synchronization between nodes with d…

### LINKED ISSUE #399: During handshake, peers should verify their respective clocks
In order to avoid synchronization problems with peers, peers should check the `timestamp` field to assess  clock drifts between peers (e.g. clock drifts should not exceed 5 seconds).

Note: max drift allowed should be defined as a configuration parameter (in millis)

lrubiorod:
…ifferent current epoch

Close #399

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -200,6 +200,10 @@ pub struct Connections {
     ))]
     pub handshake_timeout: Duration,
 
+    /// Handshake maximum timestamp difference in seconds
+    /// Set to 0 to disable timestamp comparison in handshake
+    pub handshake_max_ts_diff: i64,
+
     /// Number of seconds before giving up waiting for requested blocks
     pub blocks_timeout: i64,
 
@@ -441,6 +445,10 @@ impl Connections {
             handshake_timeout: config
                 .handshake_timeout
                 .unwrap_or_else(|| defaults.connections_handshake_timeout()),
+            handshake_max_ts_diff: config
+                .handshake_max_ts_diff
+                .to_owned()
+                .unwrap_or_else(|| defaults.connections_handshake_max_ts_diff()),
             blocks_timeout: config
                 .blocks_timeout
                 .to_owned()
@@ -727,6 +735,7 @@ mod tests {
             discovery_peers_period: Some(Duration::from_secs(100)),
             feeler_peers_period: Some(Duration::from_secs(1)),
             handshake_timeout: Some(Duration::from_secs(3)),
+            handshake_max_ts_diff: Some(17),
             blocks_timeout: Some(5),
             consensus_c: Some(51),
             bucketing_update_period: Some(200),
@@ -743,6 +752,7 @@ mod tests {
         assert_eq!(config.feeler_peers_period, Duration::from_secs(1));
         assert_eq!(config.handshake_timeout, Duration::from_secs(3));
         assert_eq!(config.blocks_timeout, 5);
+        assert_eq!(config.handshake_max_ts_diff, 17);
         assert_eq!(config.consensus_c, 51);
         assert_eq!(config.bucketing_update_period, 200);
     }
--- a/node/src/actors/session/handlers.rs
+++ b/node/src/actors/session/handlers.rs
@@ -65,7 +65,7 @@ impl Handler<EpochNotification<EveryEpochPayload>> for Session {
 
     fn handle(&mut self, msg: EpochNotification<EveryEpochPayload>, ctx: &mut Context<Self>) {
         debug!("Periodic epoch notification received {:?}", msg.checkpoint);
-        self.current_epoch = Some(msg.checkpoint);
+        self.current_epoch = msg.checkpoint;
 
         let now = get_timestamp();
         if self.blocks_timestamp != 0 && now - self.blocks_timestamp > self.blocks_timeout {
@@ -116,14 +116,35 @@ impl StreamHandler<BytesMut, Error> for Session {
                     (
                         _,
                         SessionStatus::Unconsolidated,
-                        Command::Version(Version { sender_address, .. }),
+                        Command::Version(Version {
+                            sender_address,
+                            timestamp,
+                            ..
+                        }),
                     ) => {
-                        let msgs = handshake_version(self, &sender_address);
-                        for msg in msgs {
-                            self.send_message(msg);
-                        }
+                        let received_ts = timestamp;
+                        let current_ts = get_timestamp();
 
-                        try_consolidate_session(self, ctx);
+                        if self.handshake_max_ts_diff != 0
+                            && (current_ts - received_ts).abs() > self.handshake_max_ts_diff
+                        {
+                            log::warn!(
+                                "Dropping peer because their timestamp is different from ours\
+                                 ({:+} seconds), current timestamp: {}",
+                                (received_ts - current_ts),
+                                current_ts,
+                            );
+
+                            // Stop this session
+                            ctx.stop();
+                        } else {
+                            let msgs = handshake_version(self, &sender_address);
+                            for msg in msgs {
+                                self.send_message(msg);
+                            }
+
+                            try_consolidate_session(self, ctx);
+                        }
                     }
                     (_, SessionStatus::Unconsolidated, Command::Verack(_)) => {
                         handshake_verack(self);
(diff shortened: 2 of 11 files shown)


aesedepece:
Don't you think this will cause a small percentage of peers with perfectly synchronized clocks to be dropped unfairly just because the `Version` message was sent just before the checkpoint but received right after it?

aesedepece:
Under the assumption of 30s epochs and [~135ms average latency each way](https://arxiv.org/pdf/1801.03998.pdf) plus a small computation overhead, we can expect only 1 in 100-200 handshakes to be dropped.

I know that 0.5%-1% false negatives sounds negligible, but this rate may skyrocket though if actix gets in the middle, storage locks our threads for too long or many other reasons.

node/src/actors/peers_manager/actor.rs
```diff
@@ -67,7 +68,11 @@ impl Actor for PeersManager {
                 act.persist_peers(ctx, storage_peers_period);
 
                 // Start the feeleer peers process on SessionsManager start
-                act.feeler(ctx, feeler_peers_period);
+                act.feeler(
+                    ctx,
+                    feeler_peers_period,
+                    Duration::from_secs(config.consensus_constants.checkpoints_period as u64),
```
tmpolaczyk: Why is this offset needed?

aesedepece: Same question here :thinking:

lrubiorod: > negligible

Another way is to check if there are a difference of more than one epochs to avoid this

lrubiorod:
https://github.com/witnet/witnet-rust/pull/886#issuecomment-555920281

Another way is to check if there are a difference of more than one epochs to avoid this

aesedepece:
Another option would be exchanging the current NTP time instead of the epoch number and dropping the connection if the difference with a peer is more than a threshold set in the configuration (in that way, you don't punish handshakes that happen at the end of an epoch).

node/src/actors/peers_manager/actor.rs
lrubiorod: Some problems appears when we try to send messages from a session which have not current_epoch updated (and it has a None). So this offset is to create a wait at the beginning, when the node is starting, to do not use the feeler and the bootstrap methods until have an updated Epoch in current_epoch state

tmpolaczyk: So we wouldn't need that if we had a synchronous way to get the current epoch at the beginning? Maybe we could use `EpochConstants::epoch_at(get_timestamp())`, and get `EpochConstants` directly from config.

config/src/config.rs
```diff
@@ -200,6 +200,9 @@ pub struct Connections {
     ))]
     pub handshake_timeout: Duration,
 
+    /// Handshake maximum timestamp difference
```
tmpolaczyk:
```suggestion
    /// Handshake maximum timestamp difference in seconds
```

tmpolaczyk:
```suggestion
    /// Set to 0 to disable timestamp comparison in handshake
```

node/src/actors/session/handlers.rs
```diff
@@ -116,14 +116,35 @@ impl StreamHandler<BytesMut, Error> for Session {
...
-                        for msg in msgs {
-                            self.send_message(msg);
+                        let current_ts = get_timestamp();
+                        let received_ts = timestamp;
+
+                        if self.handshake_max_ts_diff != 0
+                            && (current_ts - received_ts).abs() > self.handshake_max_ts_diff
+                        {
+                            log::warn!(
+                                "Invalid timestamp in Handshake Version message received.\n\
```
tmpolaczyk: It would be nice to see the difference in the log:

```suggestion
                                "Dropping peer because their timestamp is too far away (±{} seconds)
```

"too far away" may not be the best words, maybe "because the difference between timestamps is too big" or something.

node/src/actors/session/handlers.rs
```diff
@@ -572,7 +593,7 @@ fn handshake_version(session: &mut Session, sender_address: &Address) -> Vec<Wit
             session.magic_number,
             session.server_addr,
             session.remote_addr,
-            0,
+            session.current_epoch,
```
tmpolaczyk: But current_epoch is not used anymore, right? We use the timestamp instead. Can we remove the `current_epoch` field from the `Version` message?

node/src/actors/session/mod.rs
```diff
@@ -86,19 +86,22 @@ pub struct Session {
...
+    handshake_max_ts_diff: i64,
 }
 
 /// Session helper methods
 impl Session {
     /// Method to create a new session
     pub fn new(
-        server_addr: SocketAddr,
-        remote_addr: SocketAddr,
+        (server_addr, remote_addr): (SocketAddr, SocketAddr),
```
tmpolaczyk: Hacks

p2p/src/sessions/mod.rs
```diff
@@ -117,8 +120,9 @@ where
             .set_limit(outbound_consolidated_limit);
     }
     /// Method to set the handshake timeout
-    pub fn set_handshake_timeout(&mut self, handshake_timeout: Duration) {
+    pub fn set_handshake_options(&mut self, handshake_timeout: Duration, ts_diff: i64) {
```
tmpolaczyk: I would prefer to have two different setters instead: `set_handshake_timeout` and `set_handshake_max_ts_diff`

node/src/actors/session/handlers.rs
lrubiorod: we always had a useless current_epoch in the Version :P

node/src/actors/session/mod.rs
lrubiorod: clippy things

tmpolaczyk: We can just disable that lint globally, I think that's better than doing these hacks all the time.

node/src/actors/session/handlers.rs
tmpolaczyk: New issue?

lrubiorod: created #888 issue for that

node/src/actors/session/handlers.rs
```diff
@@ -116,14 +116,36 @@ impl StreamHandler<BytesMut, Error> for Session {
...
+                        let current_ts = get_timestamp();
+                        let received_ts = timestamp;
+
+                        if self.handshake_max_ts_diff != 0
+                            && (current_ts - received_ts).abs() > self.handshake_max_ts_diff
+                        {
+                            log::warn!(
+                                "Dropping peer because their timestamp is too big (±{} seconds).\n\
+                                 Current timestamp: {}\nReceived timestamp: {}",
+                                self.handshake_max_ts_diff,
```
tmpolaczyk: I meant to use actual difference here:

```suggestion
                                (current-ts - received_ts),
```

lrubiorod: sure, I am sleepy today >_<

node/src/actors/session/handlers.rs
```diff
@@ -116,14 +116,33 @@ impl StreamHandler<BytesMut, Error> for Session {
...
-                        for msg in msgs {
-                            self.send_message(msg);
+                        let current_ts = get_timestamp();
+                        let received_ts = timestamp;
+
+                        if self.handshake_max_ts_diff != 0
+                            && (current_ts - received_ts).abs() > self.handshake_max_ts_diff
+                        {
+                            log::warn!(
+                                "Dropping peer because their timestamp is too big (±{} seconds",
```
tmpolaczyk: How about we show the sign of the difference, and the local timestamp:

```suggestion
                                "Dropping peer because their timestamp is different from ours ({:+} seconds), current timestamp: {}",
                                (received_ts - current_ts), current_ts,
```

tmpolaczyk:
There's a strange issue with travis:

Failed build:
https://travis-ci.com/witnet/witnet-rust/builds/137571494#L232

Fixed build after that, running [literally the same code](https://github.com/witnet/witnet-rust/compare/2369ae43d00f931b7970911518c83343f4536752..0569e52dde1571fda636bb2f070a0e84d1dbaa51):
https://travis-ci.com/witnet/witnet-rust/builds/137577352#L234

It looks like the failed build does not use the custom apt sources specified in .travis.yml. Why?

Well, in a completely different PR I remove these custom sources and update the ubuntu version. You can see that that results here:
https://travis-ci.com/witnet/witnet-rust/builds/137548235#L171

So my guess is that it's an issue with some travis cache.

PR #887, opened 2019-11-20
chore: update rocksdb to 0.13

### LINKED ISSUE #791: Unpin rocksdb version 0.12.2
Right now we cannot use the latest version of the rocksdb crate (0.12.3) because of a bug which breaks cross-compilation (https://github.com/rust-rocksdb/rust-rocksdb/issues/322). That bug is fixed in the master branch, but it is not released yet. So when the rocksdb crate releases a new version, we should check that it compiles on all our supported architectures and unpin the version.

tmpolaczyk:
It looks like it finally compiles in arm

Close #791

### DIFF
--- a/.github/workflows/midnight.yml
+++ b/.github/workflows/midnight.yml
@@ -26,7 +26,6 @@ jobs:
 
       - name: Install environment dependencies
         run: |
-          sudo add-apt-repository -y ppa:maarten-fonville/protobuf
           sudo apt-get update -y -qq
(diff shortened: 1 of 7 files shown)


tmpolaczyk:
There is a linker error because the version installed in travis is not compatible...

tmpolaczyk:
So it turns out that the issue with #861 that prevented us from having tests in a specific folder could be fixed by updating the ubuntu version in travis

PR #895, opened 2019-11-26
Improve RADON executor

tmpolaczyk:
This PR makes `run_retrieval` async, meaning that now we can run all the retrievals in parallel.


rad/src/lib.rs
```diff
@@ -32,13 +32,17 @@ pub fn run_retrieval_with_data(retrieve: &RADRetrieve, response: String) -> Resu
...
     match retrieve.kind {
         RADType::HttpGet => {
-            let response = reqwest::get(&retrieve.url)
-                .map_err(RadError::from)?
+            let response = reqwest::Client::builder()
+                .build()?
+                .get(&retrieve.url)
+                .send()
+                .await?
+                // TODO: replace with .bytes() and let RADON handle the encoding?
```
tmpolaczyk: The `reqwest` crate supports many custom encodings by respecting the `Content-Type` HTTP header. It also supports a default encoding with `text_with_charset`, although the encoding specified in the header has priority.

We can change `.text()` to `.bytes()` and let RADON do the decoding, but we lose the information from the HTTP headers.

tmpolaczyk:
In this PR I change the library we use to perform HTTP GET requests, from [reqwest](https://github.com/seanmonstar/reqwest) to [surf](https://github.com/http-rs/surf). The reason is that the newest version of reqwest is not compatible with the old version of tokio, which we are forced to use because actix hasn't been updated yet.

This is a very small change (~20 lines of code), so we can go back to using reqwest at any time, or use a different library altogether. I have not explored any alternative async HTTP request libraries. I think the best decision would be to just use curl, although surf is just a wrapper around another library which is a wrapper around libcurl.

We could continue using an old version of reqwest that does not confict with actix. The only downside is that that would make `witnet_rad` depend on futures 0.1 and the old tokio, making it cumbersome to use from new projects which are using std futures.

The problem with surf is that it is very immature compared to reqwest: many methods panic instead of returning results, I found issues when querying certain servers, and they are actively adding features so the API could change at any time.

So we can change back to reqwest after actix updates their tokio dependency and we refactor all of our code to the new actix version.

The good news is that except for some libraries which depend on a specific tokio version, it's very easy to integrate code which uses the new std futures in actix.

node/src/actors/rad_manager/handlers.rs
```diff
@@ -1,45 +1,83 @@
...
-
-use witnet_rad as rad;
+use actix::{Handler, Message, ResponseFuture};
 
 use crate::actors::messages::{ResolveRA, RunTally};
+use futures::Future;
+use tokio::util::FutureExt;
+use witnet_data_structures::radon_report::RadonReport;
+use witnet_rad as rad;
+use witnet_rad::error::RadError;
```
lrubiorod: fmt

node/src/actors/rad_manager/handlers.rs
```diff
@@ -1,45 +1,83 @@
...
-                    .ok()
-            })
-            .collect();
-
-        // Perform aggregation on the values that made it to the output vector after applying the
-        // source scripts (aka _normalization scripts_ in the original whitepaper) and filtering out
-        // failures.
-        rad::run_aggregation_report(retrieve_responses, &aggregator)
+        let timeout = msg.timeout;
+        let fut03 = async {
```
lrubiorod: comment about what is fut03?

src/cli/node/json_rpc_client.rs
```diff
@@ -214,10 +214,13 @@ pub fn send_vtt(
 }
 
 fn run_dr_locally(dr: &DataRequestOutput) -> Result<RadonTypes, failure::Error> {
+    let run_retrieval_blocking =
+        |retrieve| futures03::executor::block_on(witnet_rad::run_retrieval(retrieve));
```
lrubiorod: Could you add a comment explaining why block in this case?

wallet/src/actors/worker/methods.rs
```diff
@@ -20,10 +20,13 @@ impl Worker {
     }
 
     pub fn run_rad_request(&self, request: types::RADRequest) -> Result<types::RadonTypes> {
+        let run_retrieval_blocking =
+            |retrieve| futures03::executor::block_on(witnet_rad::run_retrieval(retrieve));
```
lrubiorod: same comment here

PR #900, opened 2019-12-03
Liars punishment

### LINKED ISSUE #869: Implement out-of-consensus witnessing slashing

lrubiorod:
Close #869

### DIFF
--- a/data_structures/src/data_request.rs
+++ b/data_structures/src/data_request.rs
@@ -4,14 +4,13 @@ use std::convert::TryFrom;
 use log::{debug, info};
 use serde::{Deserialize, Serialize};
 
-use crate::radon_report::TypeLike;
 use crate::{
     chain::{
         DataRequestOutput, DataRequestReport, DataRequestStage, DataRequestState, Epoch,
         EpochConstants, Hash, Hashable, PublicKeyHash, ValueTransferOutput,
     },
     error::{DataRequestError, TransactionError},
-    radon_report::RadonReport,
+    radon_report::{RadonReport, Stage, TypeLike},
     transaction::{CommitTransaction, DRTransaction, RevealTransaction, TallyTransaction},
 };
 
@@ -48,9 +47,12 @@ impl DataRequestPool {
 
     /// Get all reveals related to a `DataRequestOuput`
     pub fn get_reveals(&self, dr_pointer: &Hash) -> Option<Vec<&RevealTransaction>> {
-        self.data_request_pool
-            .get(dr_pointer)
-            .map(|dr_state| dr_state.info.reveals.values().collect())
+        self.data_request_pool.get(dr_pointer).map(|dr_state| {
+            let mut reveals: Vec<&RevealTransaction> = dr_state.info.reveals.values().collect();
+            reveals.sort_unstable_by_key(|reveal| reveal.body.pkh);
+
+            reveals
+        })
     }
 
     /// Insert a reveal transaction into the pool
@@ -64,7 +66,10 @@ impl DataRequestPool {
             .iter()
             .filter_map(|(dr_pointer, dr_state)| {
                 if let DataRequestStage::TALLY = dr_state.stage {
-                    let reveals = dr_state.info.reveals.values().cloned().collect();
+                    let mut reveals: Vec<RevealTransaction> =
+                        dr_state.info.reveals.values().cloned().collect();
+                    reveals.sort_unstable_by_key(|reveal| reveal.body.pkh);
+
                     Some((*dr_pointer, reveals))
                 } else {
                     None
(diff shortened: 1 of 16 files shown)


data_structures/src/data_request.rs
```diff
@@ -47,10 +46,18 @@ impl DataRequestPool {
...
-            .map(|dr_state| dr_state.info.reveals.values().collect())
+    pub fn get_reveals(&self, dr_pointer: &Hash) -> Option<Vec<RevealTransaction>> {
+        self.data_request_pool.get(dr_pointer).map(|dr_state| {
+            let mut pkhs: Vec<PublicKeyHash> = dr_state.info.reveals.keys().cloned().collect();
+
+            pkhs.sort();
+            let reveals: Vec<RevealTransaction> = pkhs
+                .iter()
+                .map(|pkh| dr_state.info.reveals.get(pkh).unwrap().clone())
+                .collect();
```
tmpolaczyk: You can avoid calling collect twice because the reveal transactions already include the pkh, so you can sort by key:

```rust
    pub fn get_reveals(&self, dr_pointer: &Hash) -> Option<Vec<&RevealTransaction>> {
        self.data_request_pool.get(dr_pointer).map(|dr_state| {
            let mut reveals: Vec<&RevealTransaction> =
                dr_state.info.reveals.values().collect();
            reveals.sort_unstable_by_key(|reveal| reveal.body.pkh);

            reveals
        })
    }
```

tmpolaczyk: But why are we sorting by pkh in the first place? Maybe we can leave this function like it was before, and sort the reveals where we actually need them to be sorted.

data_structures/src/data_request.rs
```diff
@@ -338,48 +351,62 @@ pub fn create_tally<RT>(
...
-                let vt_output = ValueTransferOutput {
-                    pkh: reveal.body.pkh,
-                    value: reveal_reward,
-                    time_lock: 0,
-                };
-                Some(vt_output)
-            } else {
-                // TODO: Penalize dishonest nodes
-                None
+    if let Stage::Tally(tally_metadata) = report.metadata.clone() {
```
tmpolaczyk: You don't need this clone

```suggestion
    if let Stage::Tally(tally_metadata) = &report.metadata {
```

data_structures/src/data_request.rs
```diff
@@ -338,48 +351,62 @@ pub fn create_tally<RT>(
...
+        let reveal_reward = calculate_dr_vt_reward(dr_output);
+
+        let liars = tally_metadata.liars;
```
tmpolaczyk:
```suggestion
        let liars = &tally_metadata.liars;
```

data_structures/src/radon_report.rs
```diff
@@ -143,20 +143,59 @@ pub struct TallyMetaData {
...
     /// This follows a reverse logic: `false` is truth and `true` is lie.
-    liars: Vec<bool>,
+    pub liars: Vec<bool>,
     /// Proportion between total reveals and "truthers" count:
     /// `liars.iter().filter(std::ops::Not).count() / reveals.len()`
     consensus: f32,
 }
 
+impl TallyMetaData {
+    /// Update liars vector
```
tmpolaczyk: What does that mean?

tmpolaczyk:
Add tests and documentation about what exactly is a "lie".

node/src/actors/chain_manager/mod.rs
```diff
@@ -542,26 +542,17 @@ impl ReputationInfo {
...
-                    // FIXME: restore this clause once `true_revealer` is unmocked.
-                    /*if true_revealer(reveal_tx, tally) {
-                        0
-                    } else {
-                        1
-                    }*/
-                    0
-                })
-                .unwrap_or(1);
+            let liar = if tally_vts.contains(pkh) { 0 } else { 1 };
```
tmpolaczyk: Add comment:
```suggestion
            // If the identity behaved honestly, it must have received a reward.
            // So we know which identities lied because they do not have a vtt transaction in the tally.
            let liar = if tally_vts.contains(pkh) { 0 } else { 1 };
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -542,26 +542,17 @@ impl ReputationInfo {
...
         let replication_factor = dr_state.data_request.witnesses;
         self.alpha_diff += Alpha(u32::from(replication_factor));
 
-        let _tally = &tally_transaction.tally;
-
+        let mut tally_vts = HashSet::new();
+        // TODO: Review case of data requester is also a witness
+        for vt in &tally_transaction.outputs {
+            tally_vts.insert(vt.pkh);
+        }
```
tmpolaczyk: You can be functional

```rust
// Set of pkhs which were rewarded in the tally transaction
let tally_vts: HashSet<_> = tally_transaction.outputs.iter().map(|vt| vt.pkh).collect();
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -542,26 +542,17 @@ impl ReputationInfo {
...
         let dr_state = &data_request_pool.data_request_pool[&dr_pointer];
-        let reveals = &dr_state.info.reveals;
         let commits = &dr_state.info.commits;
         let replication_factor = dr_state.data_request.witnesses;
         self.alpha_diff += Alpha(u32::from(replication_factor));
 
-        let _tally = &tally_transaction.tally;
-
+        let mut tally_vts = HashSet::new();
+        // TODO: Review case of data requester is also a witness
```
tmpolaczyk: With the current implementation, the data requester can lie with no penalization if the tally transaction contains a change output. It would be nice to have a test for this case.

data_structures/src/data_request.rs
lrubiorod: in both places where use these functions we need that are sorted so... make sense for me that return the result sorted. This sorting is needed to apply tally scripts and punish positions

data_structures/src/radon_report.rs
```diff
@@ -183,3 +222,24 @@ mod tests {
         assert_eq!(encoded, expected);
     }
 }
+
+#[test]
+fn test_update_liars() {
```
tmpolaczyk: Move this test inside tests module

data_structures/src/radon_report.rs
```diff
@@ -143,20 +143,59 @@ pub struct TallyMetaData {
...
+    /// Update liars vector
+    /// new_liars length has to be less than false elements in liars
+    pub fn update_liars(&mut self, new_liars: Vec<bool>) {
+        if self.liars.is_empty() {
+            self.liars = new_liars;
+        } else {
+            let mut vec = vec![];
+            let mut new_iter = new_liars.iter();
+
+            // TODO: Improve code
```
tmpolaczyk: How about

```rust
            for liar in &mut self.liars {
                if !*liar {
                    *liar = *new_iter.next().unwrap();
                }
            }
```

node/src/actors/chain_manager/mod.rs
lrubiorod: Yes, it happens. But you say a test for check that this bug happens or doesnt happen?

tmpolaczyk: Ideally I think we should fix that so it does not happen. But I don't know how straightforward would that be.

tmpolaczyk: Created #903

PR #910, opened 2019-12-10
Forbid dangerous operators in tally stage

### LINKED ISSUE #904: Forbid dangerous RADON opcodes in Tally scripts

tmpolaczyk:
The following operators are now forbidden from being used during the tally stage:

* All the array operators except `ArrayFilter` and `ArrayReduce`

This does not affect subscripts, only top-level operators, so for example `ArrayGet` can still be used inside an `ArrayFilter`.

Important change: reducers lose the ability to modify the liars vector. If a user wants to punish all the reveals different from the mode, they should use a filter instead.

Another important change: the liars vector is now frozen after the first reduce. This prevents an issue where running filters after a call to reduce resulted in an invalid liars vector. It also makes a lot of sense: we only want to penalize liars based on the values of the input array. Since a reduce consumes the input array, it does not make sense to penalize after that.

Close #904

### DIFF
--- a/rad/src/operators/mod.rs
+++ b/rad/src/operators/mod.rs
@@ -163,7 +163,12 @@ pub fn identity(input: RadonTypes) -> Result<RadonTypes, RadError> {
 pub fn check_valid_operator_for_tally_stage(call: &RadonCall) -> Result<(), RadError> {
     // List of forbidden operators.
     // When this list grows and performance starts to be a concern, use a bitset instead
-    const FORBIDDEN_IN_TALLY: [RadonOpCodes; 2] = [RadonOpCodes::ArraySort, RadonOpCodes::ArrayGet];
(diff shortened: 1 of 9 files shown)


rad/src/operators/mod.rs
```diff
@@ -152,6 +152,24 @@ pub fn identity(input: RadonTypes) -> Result<RadonTypes, RadError> {
     Ok(input)
 }
 
+/// Check if the operator can be used in a tally script.
+///
+/// Some operators cannot be used in the tally stage because they are considered to be
+/// "consensus unsafe".
+///
+/// For example, sorting the input is forbidden.
```
tmpolaczyk: I would like to explain why, @aesedepece is something like this ok?

Because the tally result should not depend on the order, so sorting it is a logic error. Sorting tally inputs could produce mistakes in reputation redistribution.

lrubiorod:
This PR is a temporal solution until #912 would be implemented

PR #914, opened 2019-12-16
feat(docker): add `curl` as a dependency for `debug-run` image

aesedepece:
- This was needed after replacing `reqwest` with `surf`
- All other docker images already had `curl` in place
- The upgraded image is already available on Docker Hub

~~DISCLAIMER: this commit is unsigned. I'll sign it before merging (also: it is OK to have the merger sign it in my behalf)~~


aesedepece:
Signed!

PR #913, opened 2019-12-13
RadonBytes/RadonMixed refactor Part2

lrubiorod:
Close #907 
Close #908 
Close #916

PR #922, opened 2019-12-19
feat: autogenerate tests in examples/

### LINKED ISSUE #916: Automatize data request json creation process in /examples
It is very annoying to modify the data request script bytes manually

tmpolaczyk:
Close #916

Autogenerate but not overwrite: the files still need to be updated manually, but now it's just one copy paste.


lrubiorod:
Added to #913 PR

PR #928, opened 2019-12-23
 Implement get (commits, revels) for TransactionsPool

### LINKED ISSUE #918: Implement get (commits, revels) for TransactionsPool
Since #917, the `TransactionsPool::get` method can be used to get transactions from the in-memory transactions pool. However, it only works on value transfer and data request transactions, due to the internal structure of the `TransactionsPool`.

We need some sort of index to be able to query all kinds of transactions by hash efficiently.

tmpolaczyk:
Close #918.

This will allow `getTransactions` method to work with unconfirmed commits and reveals.

But since commits and reveals are cleared every epoch, we cannot use this method to see commits or reveals that were not accepted into a block.

### DIFF
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -1403,7 +1403,6 @@ impl TransactionsPool {
     }
 
     /// Get transaction by hash
-    // FIXME(#918): only vtt and drt are supported, commits and reveals are not indexed by hash
     pub fn get(&self, hash: &Hash) -> Option<Transaction> {
         self.vt_transactions
             .get(hash)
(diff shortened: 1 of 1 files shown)


node/src/actors/chain_manager/mod.rs
```diff
@@ -615,6 +615,9 @@ fn update_pools(
         }
     }
 
+    // Remove commits because they expire every epoch
+    transactions_pool.clear_commits();
```
lrubiorod: This line is only running during block consolidation, so if there are not blocks in a epoch, it could be produce that there are commits from different epochs

PR #931, opened 2019-12-23
Batch storage writing API

### LINKED ISSUE #919: Add batch writing API to StorageManager
Since #917, there is an `AddItems` message that allows writing blocks and transactions to the storage in batch. However, it is implemented as a for loop that sends `AddItem` messages, and that's very sub-optimal. But in order to implement it the right way, we need a `put_batch` method in storage manager.

tmpolaczyk:
Close #919

This PR introduces a new `put_batch` method in `storage_mngr`. Internally that's implemented as `Storage::put` in a for loop, but this way we save all the overhead of message passing, since that for loop happens in a different thread.


aesedepece:
Merged as c411864295c0a23a1918692a3353ee7da9c0e7c2.

PR #946, opened 2020-01-08
create mode filter

### LINKED ISSUE #932: Create RADON Mode filter
It filter all the results that are not equal to the mode, and it punish in case of tally

lrubiorod:
Fix #932


aesedepece:
TDD, huh? :smirk:

rad/src/filters/mode.rs
```diff
@@ -0,0 +1,208 @@
...
+            })
+            .collect();
+
+        Ok(output_i128)
+    }
+
+    #[test]
+    fn test_filter_mode_empty() {
+        let input = vec![];
+        let expected = RadError::ModeEmpty;
```
tmpolaczyk: A filter applied to an empty array could just return an empty array, and the error would naturally appear when trying to reduce an empty array.

lrubiorod: The error is gonna appear anyway, so I think is better to handle as soon as possible

PR #948, opened 2020-01-09
Refactor dr.value to dr.witness_reward

### LINKED ISSUE #915: [Pseudo-WIP] Explicit request rewards
```
Pseudo-WIP: #915
Title: Explicit request rewards
Status: Draft
Type: Standards Track
```

## Abstract

This is a proposal for improvement of the way in which the rewards for witnessing nodes are denominated in Witnet requests. It aims to improve the user experience of developers who write or edit requests by hand, and in general to enhance the ease of understanding of the protocol's data structures.

## Motivation

Witnet allows parametrization of many of the economic incentives that affect the life cycle of your requests. Namely, those incentives are:

- `request_fee`: the amount of wit tokens that will be earned by the Witnet miner that publishes your request in a block.
- `reward`: the amount of wit tokens that every each of the Witnet nodes assigned to your request will earn if they honestly fulfill their commitments and reveals.
- `commit_fee`: the amount of wit tokens that will be earned by Witnet miners for each each valid commitment transaction they include in a block.
- `reveal_fee`: the amount of wit tokens that will be earned by Witnet miners for each valid reveal transaction they include in a block.
- `tally_fee`: the amount of wit tokens that will be earned by the Witnet miner that publishes in a block the tally of all the reveal transactions related to your request.

As explained in the protocol documentation, the cost model for requests is straightforward:
```
request_fee + witnesses * (reward + commit_fee + reveal_fee) + tally_fee
```

However, the current implementation of the `DataRequestOutput` data structure is missing the `reward` field and uses a `value` field whose value needs to satisfy the following constraint:

```
value > witnesses * (reward + commit_fee + reveal_fee) + tally_fee
```

The rationale for why the `value` exists at all is because `DataRequestOutput` and `ValueTransferOutput` were a single data structure that at some point in the past.

We have often found ourselves failing to produce valid requests when fine-tuning the [JSON examples][examples] from the [witnet-rust] repository. If the value constraint is not satisfied, the node shows a warning which explains that the remainder from subtracting the fees from `value` is not divisible by the number of witnesses.

The lack of explicitness when denominating witnessing fees is therefore producing two adverse effects:
- Poor user experience, as the incentive for each witness is not explicitly stated.
- An additional integer division check, i.e. redundant code.
- Additional logic in the [witnet-requests-js] library, which uses the same denominations as explained in the documentation and then computes the `value` field on the fly when compiling a request.

## Specification

This proposal entails the following changes:
- Removing the `value` field from the `DataRequestOutput` structure.
- Adding a `witness_reward` to the `DataRequestOutput` structure.
- Removing the `value / witnesses` integer division validation.
- Modifying the cost computation function so it follows the cost model below.
- Adapting all `DataRequestOutput` and `DataRequest` factories.
- Adapting all JS libraries, including [witnet-requests-js] and [Sheikah].
- Adapting all examples in the [witnet-rust] repository.

### Proposed Witnet requests cost model
```
request_fee + witnesses * (reward + commit_fee + reveal_fee) + tally_fee
```

### Consensus
This proposal is a breaking change to consensus and thus can only be adopted either through a soft or hard fork. As long as we are in rebootable testnet stage, the latter way is suggested.

### Examples
This proposal renders all JSON examples outdated and not valid under the new cost model and validation rules.

Examples using [witnet-requests-js] are unaffected.

### Libraries and clients
The [witnet-requests-js] library needs to adopt the specified changes and a new release needs to be published. The [truffle-box] will need to target the latest version of [witnet-requests-js]. Projects using either the truffle box or the witnet-requests-js library will directly need to update their dependencies but their content will remain unaffected.

[Sheikah] has no support for deploying requests yet, so it is unaffected.

### Documentation
The new specification has no impact on documentation, as it follows what it is already documented.

## Acknowledgements
This proposal emerged after a convo with @lrubiorod, who detailed the current situation and gave feedback on the specification found herein.

[examples]: https://github.com/witnet/witnet-rust/tree/master/examples
[witnet-rust]: https://github.com/witnet/witnet-requests-js
[witnet-requests-js]: https://github.com/witnet/witnet-requests-js
[truffle-box]: https://github.com/witnet/truffle-box
[Sheikah]: https://github.com/witnet/witnet-requests-js

tmpolaczyk:
Close #915 

This is a breaking change, as it changes the protocol schema


node/src/actors/chain_manager/transaction_factory.rs
```diff
@@ -317,10 +326,13 @@ mod tests {
 
     fn outputs_sum(transaction: &Transaction) -> u64 {
         match transaction {
-            Transaction::ValueTransfer(tx) => tx.body.outputs.iter().map(|o| o.value).sum(),
-            Transaction::DataRequest(tx) => {
-                tx.body.outputs.iter().map(|x| x.value).sum::<u64>() + tx.body.dr_output.value
+            Transaction::ValueTransfer(vt_tx) => {
+                transaction_outputs_sum(&vt_tx.body.outputs).unwrap()
```
lrubiorod: Comments about why unwrap is safe here

node/src/actors/chain_manager/transaction_factory.rs
```diff
@@ -317,10 +326,13 @@ mod tests {
...
             }
+            Transaction::DataRequest(dr_tx) => transaction_outputs_sum(&dr_tx.body.outputs)
+                .unwrap()
+                .checked_add(dr_tx.body.dr_output.checked_total_value().unwrap())
+                .unwrap(),
```
lrubiorod: Comments about why unwrap is safe here

validations/src/tests.rs
```diff
@@ -1155,242 +1302,116 @@ fn data_request_witnesses_1() {
...
-    let x = test_drtx(DataRequestOutput {
-        value: 1000,
-        commit_fee: 126,
-        reveal_fee: 126,
-        tally_fee: 502,
-        witnesses: 2,
-        min_consensus_percentage: 51,
-        data_request,
-        ..DataRequestOutput::default()
+        tally_fee: u64::max_value(),
```
lrubiorod: Is it not the same than before?

PR #949, opened 2020-01-09
fix: implement send transaction method

mariocao:
This PR fixes the send transaction API method and refactors the `sendVtt` and `sendDataRequest` methods into a single one called `sendTransaction`.

The JsonRPC `sendTransaction` method expects the following parameters:

```json
{
    "walletId": <walletId>,
    "sessionId": <sessionId>,
    "transactionId": <transactionId>,
}
```

Co-authored-by: Gorka Irazoqui <gorka.irazoki@gmail.com>


wallet/src/actors/app/routes.rs
```diff
@@ -78,6 +78,7 @@ pub fn connect_routes<T, S>(
         ("subscribeNotifications", {
             let addr = api.clone();
             move |params: Params, _meta, subscriber: Subscriber| {
+                log::trace!("------------> subscription!! ");
```
tmpolaczyk: I guess this can be removed?

mariocao: Indeed, that should not be there

lrubiorod:
rebase with master. There are breaking changes merged

PR #952, opened 2020-01-09
fix(data_structures): update reveal stage on zero reveals

### LINKED ISSUE #945: Fix missing reveals remediation
While creating tally transactions nodes should take into account the `extra_reveal_rounds` parameter in order to avoid DoS from committers.

In case of 0 revealers and the extra reveal rounds are over, the tally should be resolved as `RadonError`.

tmpolaczyk:
Fix #945

### DIFF
--- a/rad/src/types/mod.rs
+++ b/rad/src/types/mod.rs
@@ -296,6 +296,12 @@ impl TryFrom<&cbor::value::Value> for RadonTypes {
                                                 }
                                             }
                                         }
+                                        RadonErrors::NoReveals => {
+                                            return Err(RadError::NoReveals);
+                                        }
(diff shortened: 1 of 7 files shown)


rad/src/types/mod.rs
```diff
@@ -296,6 +296,12 @@ impl TryFrom<&cbor::value::Value> for RadonTypes {
                                                 }
                                             }
                                         }
+                                        RadonErrors::NoReveals => {
+                                            return Err(RadError::NoReveals);
+                                        }
+                                        RadonErrors::Unknown => {
```
lrubiorod: It is not necessary now

PR #955, opened 2020-01-14
fix(wallet): re-implement index transactions

### LINKED ISSUE #956: Re-implement index transactions
Re-implement the index transaction business logic under the following assumptions:

- `createVtt` request does not persist the transactions
- `createVtt` request does not modify the UTXO set
- `createVtt` API method will return a signed transaction
- Sheikah will use `sendTransaction` API method to trigger a broadcast of the attached transaction towards the node.

Therefore, the index transaction logic will be (for now) completely stateless regarding any pending transaction or blocked/used UTXO. UTXOs and balances will be updated once transactions are indexed every new block.

Things to address in future issues:

- When Sheikah confirms a transaction (using `sendTransaction`), the wallet should mark the used UTXOs as `used` (attaching a `timestamp`) and the transaction as `pending`.
- When Sheikah requests the transaction list, the wallet will return also the pending transactions.
- When Sheikah requests the balance (or is notified), the wallet will return also the `available balance` and the `unconfirmed balance` (taking into account the pending UTXOs and transactions).
- Wallet will implement some logic to support the recovery of `blocked` UTXOs in case a pending transaction is never included in the chain.

mariocao:
This PR:

- re-implements the index transactions business logic
- adds key to DBKeyNotFound error

Closes #956

### DIFF
--- a/wallet/src/model.rs
+++ b/wallet/src/model.rs
@@ -69,19 +69,14 @@ where
 #[derive(Debug, Serialize)]
 pub struct Transaction {
     pub hex_hash: String,
-    pub value: u64,
-    pub entry: TransactionEntry,
+    /// Transaction value from the wallet perspective: `value = own_outputs - own_inputs`
+    /// - A positive value means that the wallet received WITs from others.
+    /// - A negative value means that the wallet sent WITs to others.
+    pub value: i64,
     pub kind: TransactionType,
-    pub label: Option<String>,
     pub fee: Option<u64>,
     pub block: Option<BlockInfo>,
     pub timestamp: i64,
-}
-
-#[derive(Debug, Serialize, Deserialize)]
-pub enum TransactionEntry {
-    Debit,
-    Credit,
 }
 
 #[derive(Debug, Serialize, Deserialize)]
--- a/wallet/src/repository/wallet/mod.rs
+++ b/wallet/src/repository/wallet/mod.rs
@@ -1,13 +1,12 @@
 use std::sync::RwLock;
 
 use super::*;
-use crate::types::Hashable as _;
 use crate::{
     account, constants,
     db::{Database, WriteBatch as _},
     model,
     params::Params,
-    types,
+    types::{self, Hashable as _},
 };
 
 mod state;
@@ -15,7 +14,11 @@ mod state;
 mod tests;
 
 use state::State;
-use witnet_data_structures::chain::{Environment, Epoch};
+use std::convert::TryFrom;
+use witnet_data_structures::{
+    chain::{Environment, Epoch},
+    transaction::VTTransactionBody,
+};
 
 pub struct Wallet<T> {
     db: T,
@@ -250,20 +253,16 @@ where
     /// Get a transaction if exists.
     pub fn get_transaction(&self, account: u32, index: u32) -> Result<model::Transaction> {
         let value = self.db.get(&keys::transaction_value(account, index))?;
-        let entry = self.db.get(&keys::transaction_entry(account, index))?;
         let kind = self.db.get(&keys::transaction_type(account, index))?;
         let timestamp = self.db.get(&keys::transaction_timestamp(account, index))?;
         let hash: Vec<u8> = self.db.get(&keys::transaction_hash(account, index))?;
-        let label = self.db.get_opt(&keys::transaction_label(account, index))?;
         let fee = self.db.get_opt(&keys::transaction_fee(account, index))?;
         let block = self.db.get_opt(&keys::transaction_block(account, index))?;
 
         Ok(model::Transaction {
             value,
-            entry,
             kind,
             hex_hash: hex::encode(hash),
-            label,
             fee,
             block,
             timestamp,
@@ -314,38 +313,21 @@ where
         block: &model::BlockInfo,
         txns: &[types::VTTransactionBody],
     ) -> Result<()> {
-        // FIXME: Handle multiple accounts when indexing transactions
-        let account = 0;
         let mut state = self.state.write()?;
 
         for txn in txns {
             let hash = txn.hash().as_ref().to_vec();
-
-            match self.db.get_opt(&keys::transactions_index(&hash))? {
-                Some(txn_id) => {
-                    self.db
-                        .put(&keys::transaction_block(account, txn_id), block)?;
-                }
-                None => {
-                    for input in &txn.inputs {
-                        self._index_transaction_input(&mut state, &hash, &input, block)?;
-                    }
-
-                    for (index, output) in txn.outputs.iter().enumerate() {
-                        let out_ptr = model::OutPtr {
-                            txn_hash: hash.clone(),
-                            output_index: index as u32,
-                        };
-                        self._index_transaction_output(&mut state, &hash, out_ptr, &output, block)?;
-                    }
-                }
+            match self
+                .db
+                .get_opt::<_, u32>(&keys::transactions_index(&hash))?
+            {
+                None => self._index_transaction(&mut state, &hash, txn, block)?,
+                Some(_) => log::warn!(
+                    "The transaction {} already exists in the database",
+                    txn.hash()
+                ),
             }
         }
-
-        // TODO: persist changes to db
-        // self.db.write(batch)?;
-
-        // TODO: persist new values to state
 
         Ok(())
     }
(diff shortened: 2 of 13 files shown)


wallet/src/model.rs
```diff
@@ -69,8 +69,7 @@ where
 #[derive(Debug, Serialize)]
 pub struct Transaction {
     pub hex_hash: String,
-    pub value: u64,
-    pub entry: TransactionEntry,
+    pub value: i64,
```
tmpolaczyk: Why is this now a `i64`? A `Transaction` value cannot be negative

wallet/src/repository/wallet/mod.rs
```diff
@@ -315,38 +316,19 @@ where
         txns: &[types::VTTransactionBody],
     ) -> Result<()> {
         // FIXME: Handle multiple accounts when indexing transactions
```
tmpolaczyk: Is multi-account support still a goal? If not, remove this comment.

wallet/src/repository/wallet/mod.rs
```diff
@@ -315,38 +316,19 @@ where
...
-                            txn_hash: hash.clone(),
-                            output_index: index as u32,
-                        };
-                        self._index_transaction_output(&mut state, &hash, out_ptr, &output, block)?;
-                    }
-                }
+            if self
+                .db
+                .get_opt::<_, u32>(&keys::transactions_index(&hash))?
+                .is_none()
```
tmpolaczyk: Add comments about why are we checking that this value is `None`, and explain what happens when the value is `Some` (it is ignored?)

wallet/src/repository/wallet/mod.rs
```diff
@@ -648,116 +570,96 @@ where
...
-
-            self.db.write(batch)?;
-
-            state.transaction_next_id = txn_next_id;
-            state.balance = new_balance;
-            state.utxo_set.remove(&out_ptr);
+        let mut db_utxo_set: model::UtxoSet = self
+            .db
+            .get(&keys::account_utxo_set(account))
+            .unwrap_or_default();
```
tmpolaczyk: `unwrap_or_default` ignores any errors and creates a new variable. Maybe it would be better to use `get_opt(...)?.unwrap_or_default()`, so errors are handled with `?`.

wallet/src/model.rs
mariocao: We got rid of the type `Entry` which was `Debit` or `Credit`, which now can be derived by Sheikah from its sign.

wallet/src/repository/wallet/mod.rs
```diff
@@ -648,116 +570,96 @@ where
...
+
+        let mut input_amount: u64 = 0;
+        for input in txn.inputs.iter() {
+            let out_ptr: model::OutPtr = input.output_pointer().into();
+
+            if let Some(model::KeyBalance { amount, .. }) = state.utxo_set.get(&out_ptr).cloned() {
```
tmpolaczyk: `map.get()` followed with `map.remove()` can be simplified as just `map.remove()`, because remove returns the value, if any.

```suggestion
            if let Some(model::KeyBalance { amount, .. }) = state.utxo_set.remove(&out_ptr) {
```

wallet/src/repository/wallet/mod.rs
```diff
@@ -648,116 +570,96 @@ where
...
-            batch.put(keys::transaction_next_id(account), txn_next_id)?;
-            batch.put(keys::transactions_index(txn_hash), txn_id)?;
-            batch.put(keys::transaction_hash(account, txn_id), txn_hash)?;
-
-            self.db.write(batch)?;
-
-            state.transaction_next_id = txn_next_id;
-            state.balance = new_balance;
-            state.utxo_set.remove(&out_ptr);
+        let mut db_utxo_set: model::UtxoSet = self
```
tmpolaczyk: What's the difference between `db_utxo_set` and `state.utxo_set`?

wallet/src/repository/wallet/mod.rs
mariocao: It was something from the inherited code. The `unwrap_or_default` is used when the UXTO set was never initialized and an empty one is created.

wallet/src/repository/wallet/mod.rs
mariocao: State is memory, DB is persisted in the database. Please take into account that many things you are commenting are not decisions of @lrubiorod and I.

mariocao: use `wallet --help` :rofl:

wallet/src/repository/wallet/mod.rs
```diff
@@ -648,116 +570,96 @@ where
...
+                    amount: output.value,
+                };
+
+                let out_ptr = model::OutPtr {
+                    txn_hash: Vec::from(txn_hash),
+                    output_index: index as u32,
+                };
+
+                match db_utxo_set.insert(out_ptr.clone(), key_balance.clone()) {
+                    Some(x) if x == key_balance => {}
```
tmpolaczyk: When can this branch happen? When indexing the same transaction twice? Or when some transaction does not change the account balance?

wallet/src/repository/wallet/mod.rs
```diff
@@ -648,116 +570,96 @@ where
...
+        }
 
-            batch.put(&keys::transaction_value(account, txn_id), amount)?;
-            batch.put(
-                keys::transaction_entry(account, txn_id),
-                model::TransactionEntry::Credit,
-            )?;
+        let txn_balance = (output_amount as i128)
+            .checked_sub(input_amount as i128)
+            .ok_or_else(|| Error::TransactionBalanceUnderflow)?;
```
tmpolaczyk: If `output_amount` is a `u64` and `input_amount` is a `u64`, this error can never happen. But you can leave it to prevent future errors.

wallet/src/repository/wallet/mod.rs
```diff
@@ -648,116 +570,96 @@ where
...
+
+        let txn_balance =
+            i64::try_from(txn_balance).map_err(|_| Error::TransactionValueOverflow)?;
```
tmpolaczyk: Shouldn't this be transaction balance?

```suggestion
            i64::try_from(txn_balance).map_err(|_| Error::TransactionBalanceOverflow)?;
```

wallet/src/repository/wallet/mod.rs
```diff
@@ -648,116 +570,96 @@ where
...
+            let new_balance = if txn_balance > 0 {
+                state
+                    .balance
+                    .checked_add(txn_balance.abs() as u64)
+                    .ok_or_else(|| Error::TransactionBalanceOverflow)?
+            } else {
+                state
+                    .balance
+                    .checked_sub(txn_balance.abs() as u64)
+                    .ok_or_else(|| Error::TransactionBalanceUnderflow)?
```
tmpolaczyk: I don't think this is a good error name. Is this the error that appears when trying to create a transaction with more value than the account balance? If so, it needs to be more descriptive. In the node that error is called `NoMoney`, and provides context about the account balance and the transaction value.

If this error should never happen (and only appears when the storage is corrupted or something), then I'm fine with renaming it as:

```suggestion
                    .ok_or_else(|| Error::TransactionBalanceNegative)?
```

wallet/src/model.rs
tmpolaczyk: Then this `value` is the change of the account balance? Add comment :)

wallet/src/repository/wallet/mod.rs
mariocao: If `Some` we will log a warning because the transaction is already in our database. This should never happen. :man_facepalming:

wallet/src/repository/wallet/tests/mod.rs
```diff
@@ -793,7 +752,7 @@ fn test_index_transaction_vtt_created_by_wallet() {
 
     let a_block = factories::BlockInfo::default().create();
     let our_address = wallet.gen_external_address(None).unwrap();
-    let their_address = wallet.gen_external_address(None).unwrap();
+    let their_pkh = PublicKeyHash::from_str("wit1mx47n6zp4y574a3xvjt90kefc7lggvgw0v39mf").unwrap();
```
tmpolaczyk: Why is this?

mariocao: I changed it to `factories::pkh()`. The problem before was that the generated `pkh` is derived from the same wallet, so that the `index_transactions` was not working as the test was intended.

PR #959, opened 2020-01-15
Support interception, serialization and commitment of all `RadonError`s

### LINKED ISSUE #939: Support interception, serialization and commitment of all `RadonError`s
Currently, only HTTP Status errors are committed by node operators.

### LINKED ISSUE #963: Tally scripts should be re-intercepted into a catch-all RadonError
## Description
Under the current business logic, any `RadError`s raised during execution of Radon scripts are matched against a "interception" matrix that turns consensus-critical errors into actual `RadonTypes` that can be committed/revealed/tallied.

Note however that there's no guarantees for the interception matrix to actually cover all consensus-critical `RadError`s — some of them may inadvertently not be intercepted.

That's an issue for the Tally stage. Failure to produce a valid tally from the existing reveals stops the node from producing a block, and may easily lead to a situation in which no more blocks are ever appended to the chain.

## Proposed solution
When producing tallies, we need to re-intercept the output of the interceptor by turning any `RadError` that may have "survived" interception into a `RadonError::TallyError` or similar.

This solution should make the node a whole lot more robust, as this is the final missing piece to ensure that no valid Witnet request gets ever stuck without an eventual tally after a discrete time.

tmpolaczyk:
Close #939
Close #963 

This implements a bidirectional conversion between `RadonErrors` and `RadError`. This way we avoid duplicating the serialization logic. Basically, all `RadonErrors` must be convertible to `RadError`, and if that `RadError` has any extra arguments, these arguments are encoded into the CBOR array. Detailed explanation:

`RadonError` is the type reported by nodes when a data request has some error. Most `RadonTypes` are encoded as their CBOR equivalent: for example `RadonInteger` is encoded as `Value::Integer` and `RadonString` is `Value::Text`. But there is no `Value::Error`, so the encoding of `RadonError` is a bit different.

A `RadonError` is serialized as a tagged CBOR array with tag id 39:

```rust
CborValue::Tagged(Tag::of(39), Box::new(CborValue::Array(values)))
```

Where `values` is an array where the first element is the error code, and the other elements are extra arguments (depending on the error code, most of the errors do not have any extra arguments).

All the possible error codes are defined as `RadonErrors`.

The `values` array is created using the `ErrorLike` trait:

```rust
impl ErrorLike for RadError {
    fn encode_cbor_array(&self) -> Vec<SerdeCborValue> {
        self.try_into_cbor_array().unwrap()
    }
}
```

`try_into_cbor_array` performs the serialization. The error arguments are serialized using `serde_cbor`, so anything that implements `Serialize` works. One important exception is anything related to `RadonTypes`. Because of the special serialization of `RadonError`, `RadonTypes` cannot implement `Serialize`. But for example the `ModeTie` error has one argument `values: RadonArray`. The current workaround is to first convert the `RadonArray` to bytes, and then serialize that bytes. This results in CBOR inside CBOR, we must deserialize that twice, but it works. Some alternatives are replacing `RadonArray` with `Vec<Values>`, or making that argument optional, or not supporting `ModeTie` errors.

Deserialization is performed in `try_from_kind_and_cbor_args`. It's pretty straight-forward given the serialization. Everything is tested so errors like "forgetting to implement serialize when adding a new `RadonErrors`" or copy-paste errors can be easily avoided.

### DIFF
--- a/node/src/actors/chain_manager/mining.rs
+++ b/node/src/actors/chain_manager/mining.rs
@@ -289,7 +289,6 @@ impl ChainManager {
                     rad_manager_addr
                         .send(ResolveRA {
                             rad_request,
-                            dr_pointer,
                             timeout: data_request_timeout,
                         })
                         .map(move |result| match result {
@@ -387,10 +386,7 @@ impl ChainManager {
                             &slice,
                             e
                         );
-                            Some(
-                                RadonReport::from_result(Err(e), &ReportContext::default())
-                                    .unwrap(),
-                            )
+                            Some(RadonReport::from_result(Err(e), &ReportContext::default()))
                         },
                     );
 
@@ -406,16 +402,13 @@ impl ChainManager {
                         })
                         .then(|result| match result {
                             // If the result of `RunTally` is `Ok`, it will be published as tally
-                            Ok(Ok(value)) => futures::future::ok(value),
+                            Ok(Some(value)) => futures::future::ok(value),
                             // If the result of `RunTally` is `Err`, we ignore this data request.
                             // If a data request has an invalid tally script, it will never resolve.
                             // The Radon engine should return `Ok(RadonError)` for errors which should
                             // be published as the result of the data request, or we could use a
                             // special radon value to indicate "generic error"
-                            Ok(Err(e)) => {
-                                log::warn!("Couldn't run tally: {}", e);
-                                futures::future::err(())
-                            }
+                            Ok(None) => unreachable!("Couldn't run tally"),
                             Err(e) => {
                                 log::error!("Couldn't run tally: {}", e);
                                 futures::future::err(())
--- a/node/src/actors/chain_manager/mining.rs
+++ b/node/src/actors/chain_manager/mining.rs
@@ -13,14 +13,13 @@ use witnet_data_structures::{
         Hashable, PublicKeyHash, TransactionsPool, UnspentOutputsPool, ValueTransferOutput,
     },
     data_request::{create_tally, DataRequestPool},
-    radon_report::{RadonReport, ReportContext},
     transaction::{
         CommitTransaction, CommitTransactionBody, MintTransaction, RevealTransaction,
         RevealTransactionBody, TallyTransaction,
     },
     vrf::{BlockEligibilityClaim, DataRequestEligibilityClaim, VrfMessage},
 };
-use witnet_rad::{error::RadError, types::serial_iter_decode};
+use witnet_rad::types::serial_iter_decode;
 use witnet_validations::validations::{
     block_reward, calculate_randpoe_threshold, calculate_reppoe_threshold, dr_transaction_fee,
     merkle_tree_root, update_utxo_diff, validate_block, vt_transaction_fee, UtxoDiff,
@@ -378,16 +377,7 @@ impl ChainManager {
                     let reports = serial_iter_decode(
                         &mut reveals
                             .iter()
-                            .map(|reveal_tx| (reveal_tx.body.reveal.as_slice(), reveal_tx)),
-                        |e: RadError, slice: &[u8], reveal_tx: &RevealTransaction| {
-                            log::warn!(
-                            "Could not decode reveal from {:?} (revealed bytes were `{:?}`): {:?}",
-                            reveal_tx,
-                            &slice,
-                            e
-                        );
-                            Some(RadonReport::from_result(Err(e), &ReportContext::default()))
-                        },
+                            .map(|reveal_tx| reveal_tx.body.reveal.as_slice()),
                     );
 
                     let min_consensus_ratio =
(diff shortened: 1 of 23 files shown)


aesedepece:
Please consider that `RadonError` is a consensus-critical data structure. It needs to be uniform across multiple implementations.

The most extensive specification for it can be found at https://github.com/witnet/witnet-ethereum-bridge/blob/master/contracts/Witnet.sol#L22:

```solidity
  enum ErrorCodes {
    // 0x00: Unknown error. Something went really bad!
    Unknown,
    // Script format errors
    /// 0x01: At least one of the source scripts is not a valid CBOR-encoded value.
    SourceScriptNotCBOR,
    /// 0x02: The CBOR value decoded from a source script is not an Array.
    SourceScriptNotArray,
    /// 0x03: The Array value decoded form a source script is not a valid RADON script.
    SourceScriptNotRADON,
    // Complexity errors
    /// 0x10: The request contains too many sources.
    RequestTooManySources,
    /// 0x11: The script contains too many calls.
    ScriptTooManyCalls,
    // Operator errors
    /// 0x20: The operator does not exist.
    UnsupportedOperator,
    // Retrieval-specific errors
    /// 0x30: At least one of the sources could not be retrieved, but returned HTTP error.
    HTTP,
    /// 0x31: Retrieval of at least one of the sources timed out.
    RetrievalTimeout,
    // Math errors
    /// 0x40: Math operator caused an underflow.
    Underflow,
    /// 0x41: Math operator caused an overflow.
    Overflow,
    /// 0x42: Tried to divide by zero.
    DivisionByZero,
  }
```

node/src/actors/chain_manager/mining.rs
```diff
@@ -406,16 +392,13 @@ impl ChainManager {
...
                             // If the result of `RunTally` is `Err`, we ignore this data request.
                             // If a data request has an invalid tally script, it will never resolve.
                             // The Radon engine should return `Ok(RadonError)` for errors which should
                             // be published as the result of the data request, or we could use a
                             // special radon value to indicate "generic error"
-                            Ok(Err(e)) => {
-                                log::warn!("Couldn't run tally: {}", e);
-                                futures::future::err(())
-                            }
+                            Ok(None) => unreachable!("Couldn't run tally"),
```
lrubiorod: If None is unreachable, why use an Option?

tmpolaczyk: Actix doesn't like handlers returning `RadonReport` :man_shrugging:

rad/src/types/mod.rs
```diff
@@ -5,7 +5,7 @@ use std::{
 };
 
 use cbor::value::Value as CborValue;
-use serde::Serialize;
+use log::warn;
```
lrubiorod:
```suggestion
use log;
```
and later use log::warn!

rad/src/error.rs
```diff
@@ -253,11 +277,223 @@ pub enum RadError {
...
+        message, inner
+    )]
+    UnhandledIntercept {
+        inner: Option<Box<RadError>>,
+        message: Option<String>,
+    },
+}
+
+impl RadError {
+    pub fn try_from_kind_and_cbor_args(
```
aesedepece: As discussed, this is worth being separated from `RadError` and directly returning a `RadonError`. That way, the `match kind` would be much simpler, as it would only need to cover those `RadonErrors` for which we expect / support argument decoding.

rad/src/error.rs
```diff
@@ -360,3 +587,125 @@ impl TryFrom<Result<RadonTypes, RadError>> for RadonTypes {
...
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::types::integer::RadonInteger;
+    use num_enum::TryFromPrimitive;
+    use serde_cbor::Value;
+
+    fn rad_error_example(radon_errors: RadonErrors) -> RadError {
```
aesedepece: This is cool!

rad/src/types/mod.rs
tmpolaczyk: Fun fact: `use log;` doesn't do anything because dependencies are already imported. The code will work just as well without it.

rad/src/error.rs
```diff
@@ -253,11 +277,251 @@ pub enum RadError {
         achieved, required
     )]
     InsufficientConsensus { achieved: f64, required: f64 },
+    /// The request contains too many sources.
+    #[fail(display = "The request contains too many sources")]
+    RequestTooManySources,
```
tmpolaczyk: Do we want to add arguments to this errors? For example `RequestTooManySources { num_sources: u8 }`.

lrubiorod: Should be great, like num_sources and limit_sources

tmpolaczyk:
Some unhandled errors:

```
RadonError(UnhandledIntercept { inner: None, message: Some("inner: HttpOther { message: \"couldn\\\'t resolve host name\" }") })
```

Do we want to unify `HttpOther` with `HTTPError`? This particular error was "couldn't resolve host name", that's probably a DNS error that can happen if the network is down.

rad/src/error.rs
tmpolaczyk: Oh nevermind, they are not implemented yet. The arguments will be added when they are implemented

rad/src/error.rs
aesedepece: image

PR #960, opened 2020-01-15
feat(node): avoid commits after mining block in an epoch

lrubiorod:


node/src/actors/chain_manager/mod.rs
```diff
@@ -494,6 +501,16 @@ impl ChainManager {
             self.epoch_constants,
             self.vrf_ctx.as_mut(),
         ) {
+            if let Transaction::Commit(_commit) = &msg.transaction {
+                let timestamp_mining = epoch_constants
+                    .block_mining_timestamp(current_epoch)
+                    .unwrap();
+
+                if timestamp_now > timestamp_mining {
+                    return Err(ChainManagerError::NoCommitsAllowed.into());
```
tmpolaczyk:
```suggestion
                    let e = ChainManagerError::TooLateToCommit;
                    log::debug!("{}", e);
                    return Err(e.into());
```

PR #961, opened 2020-01-15
fix(config): reduce timeout duration to increase performance

lrubiorod:


lrubiorod:
> _**Make your commitments in 3 seconds or go to hell!!!**_

Now the timeout is in 15 seconds (half testnet epoch), but the problem is that when you commit the RetrieveTimeout error, rest of nodes are in mining block mode, so that commit will be deleted in the next epoch. Your commitment has to be created and distributed by all the nodes before than half epoch ( or less if NTP is not properly working)

tmpolaczyk:
Probably `2/5` of an epoch should be enough?

PR #974, opened 2020-01-22
feat(node): improve NoMoneyError message to show available and total …

### LINKED ISSUE #969: Misleading information while creating Value Transfer Transactions
Context: Value Transfer transactions require available UXTO to be used as inputs.

Problem: when a node operator creates several Value Transfer Transactions (VTT), the node will block UTXOs. Even with a greater balance than the sum of the VTTs to be created it is possible not to be able to created all of them because there are not enough UTXOs to block per each VTT.

Example:

 - 1 UTXO of 10 kWITS and the operator wants to create 2 VTTs of value 5 kWITs.
 - The node will be able only to create 1 VTT of value 5 kWITS because the UTXO will already be blocked.
 - Obviously, once the 1st VTT is included in a block, the second VTT can be created.

lrubiorod:
…balance

Close #969

### DIFF
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -6,7 +6,7 @@ use secp256k1::{
 };
 use serde::{Deserialize, Serialize};
 use std::{
-    collections::{BTreeMap, BTreeSet, HashMap, HashSet},
+    collections::{BTreeMap, BTreeSet, HashMap},
     convert::{TryFrom, TryInto},
     fmt,
     ops::{AddAssign, SubAssign},
@@ -1755,7 +1755,8 @@ pub struct ChainState {
     /// List of consolidated blocks by epoch
     pub block_chain: Blockchain,
     /// List of unspent outputs that can be spent by this node
-    pub own_utxos: HashSet<OutputPointer>,
+    /// Those UTXOs have a timestamp value to avoid double spending
+    pub own_utxos: HashMap<OutputPointer, u64>,
     /// Reputation engine
     pub reputation_engine: Option<ReputationEngine>,
 }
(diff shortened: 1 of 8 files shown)


data_structures/src/chain.rs
```diff
@@ -1755,7 +1755,7 @@ pub struct ChainState {
     /// List of consolidated blocks by epoch
     pub block_chain: Blockchain,
     /// List of unspent outputs that can be spent by this node
-    pub own_utxos: HashSet<OutputPointer>,
+    pub own_utxos: HashMap<OutputPointer, u64>,
```
tmpolaczyk: Document what's this `u64`

data_structures/src/error.rs
```diff
@@ -170,6 +170,20 @@ pub enum TransactionError {
         value
     )]
     InvalidMinConsensus { value: u32 },
+    /// Error when there is not enough balance to create a transaction
+    #[fail(
+        display = "Cannot build a transaction. Transaction value more than available balance:\
```
tmpolaczyk:
```suggestion
        display = "Cannot build a transaction. Transaction value is greater than available balance:\
```

data_structures/src/error.rs
```diff
@@ -170,6 +170,20 @@ pub enum TransactionError {
...
+                   (Total Balance:{}, Available Balance:{}, Transaction value:{})",
+        total_balance, available_balance, transaction_value
+    )]
+    NoMoney {
+        total_balance: u64,
+        available_balance: u64,
+        transaction_value: u64,
+    },
+    /// Zero amount specified
+    #[fail(display = "Amount required has to be more than zero")]
```
tmpolaczyk:
```suggestion
    #[fail(display = "Cannot build transaction with zero value")]
```

PR #976, opened 2020-01-22
feat(storage): create dynamic storage keys

### LINKED ISSUE #970: Migrable testnet
The node should allow the operator to connect to different testnets (with different consensus constants) without requiring to remove the DB data. This will allow the operator to maintain its identity (i.e. private keys) while trying different testnet versions.

lrubiorod:
Close #970

### DIFF
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -20,12 +20,12 @@ use witnet_protected::Protected;
 use witnet_reputation::{ActiveReputationSet, TotalReputationSet};
 use witnet_util::parser::parse_hex;
 
-use crate::chain::Signature::Secp256k1;
-use crate::error::TransactionError;
 use crate::{
+    chain::Signature::Secp256k1,
     data_request::DataRequestPool,
     error::{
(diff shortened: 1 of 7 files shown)


data_structures/src/chain.rs
```diff
@@ -138,6 +138,13 @@ pub struct ConsensusConstants {
     pub reputation_penalization_factor: f64,
 }
 
+impl ConsensusConstants {
+    pub fn get_magic(&self) -> u16 {
```
aesedepece: Make sure all usages of this logic are replaced with a call to this method.

node/src/actors/peers_manager/actor.rs
```diff
@@ -48,7 +46,11 @@ impl Actor for PeersManager {
                     Err(e) => error!("Error when adding peer addresses from config: {}", e),
                 }
 
-                storage_mngr::get::<_, Peers>(&PEERS_KEY)
+                let consensus_constants = (&config.consensus_constants).clone();
```
tmpolaczyk: This clone is not needed, you can call `get_magic` directly.

PR #983, opened 2020-01-23
feat(data_request): resend reveal transactions until included in a bl…

lrubiorod:
…ock or TallyStage

### DIFF
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -1698,12 +1698,10 @@ impl DataRequestState {
         }
     }
 
-    /// Advance to the next stage, returning true on success.
+    /// Advance to the next stage.
     /// Since the data requests are updated by looking at the transactions from a valid block,
     /// the only issue would be that there were no commits in that block.
-    pub fn update_stage(&mut self) -> bool {
-        let old_stage = self.stage;
-
+    pub fn update_stage(&mut self) {
         self.stage = match self.stage {
             DataRequestStage::COMMIT => {
                 if self.info.commits.is_empty() {
@@ -1730,8 +1728,6 @@ impl DataRequestState {
                 }
             }
         };
-
-        self.stage != old_stage
     }
 }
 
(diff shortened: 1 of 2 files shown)


data_structures/src/data_request.rs
```diff
@@ -204,47 +204,56 @@ impl DataRequestPool {
...
-                                    dr_pointer
+                            if let Some(transaction) = waiting_for_reveal.get(dr_pointer) {
+                                // We submitted a commit for this data request!
+                                // But has it been included into the block?
+                                let pkh = PublicKeyHash::from_public_key(
+                                    &transaction.signatures[0].public_key,
                                 );
-                                debug!(
-                                    "Commit with pkh ({}) removed from the list of commits waiting \
+                                if dr_state.info.commits.contains_key(&pkh) && !dr_state.info.reveals.contains_key(&pkh){
```
tmpolaczyk: This `if` should be split into two, otherwise the message in the `else` branch is not correct

data_structures/src/data_request.rs
```diff
@@ -204,47 +204,53 @@ impl DataRequestPool {
                 // We can notify the user that a data request from "my_claims" is available
                 // for reveal.
                 if dr_state.update_stage() {
```
tmpolaczyk: `update_stage` only returns true when the stage changes

tmpolaczyk: I think you can just remove the `if` and everything will work as expected

```suggestion
                dr_state.update_stage();
```

data_structures/src/chain.rs
```diff
@@ -1701,9 +1701,7 @@ impl DataRequestState {
     /// Advance to the next stage, returning true on success.
```
```suggestion
    /// Advance to the next stage.
```

data_structures/src/data_request.rs
```diff
@@ -223,29 +224,33 @@ impl DataRequestPool {
...
-                                return Some(transaction);
-                            } else {
+                                // until it would be included in a block.
+                                return Some(transaction.clone());
+                            } else if !dr_state.info.commits.contains_key(&pkh) {
                                 info!(
                                     "The sent commit transaction has not been \
                                      selected to be part of the data request {:?}",
                                     dr_pointer
                                 );
```
tmpolaczyk: We can remove the reveal from `waiting_for_reveal` here, is it worth it?

PR #984, opened 2020-01-24
feat: reuse crypto engine context

tmpolaczyk:
This improves the node performance.

Tests pass but some functions have too many arguments


aesedepece:
Looks nifty.

What functions do consider to have to many arguments? `validate_vt_transaction` ?

tmpolaczyk:
I had to add `#[allow(clippy::too_many_arguments)]` to `validate_commit_transaction`, but this PR adds one argument more to all the validations that deal with signatures.

PR #985, opened 2020-01-27
Fix update pools also in WaitingConsensus Stage

lrubiorod:


data_structures/src/error.rs
```diff
@@ -215,10 +215,14 @@ pub enum BlockError {
     #[fail(display = "The block has an invalid PoE")]
     NotValidPoe,
     #[fail(
-        display = "The block eligibility claim VRF proof hash is greater than the target hash: {} > {}",
-        vrf_hash, target_hash
+        display = "The block {} eligibility claim VRF proof hash is greater than the target hash: {} > {}",
+        block_hash, vrf_hash, target_hash
```
tmpolaczyk: All the errors from `BlockError` should have the block hash. Wouldn't it be better to always print the block hash when printing a `BlockError`?

Or maybe use some failure context magic: https://docs.rs/failure/0.1.6/failure/trait.ResultExt.html#tymethod.with_context

PR #994, opened 2020-01-30
fix(rad): ensure current reducers abide by `Fn(Array<T>) -> T`

### LINKED ISSUE #989: Provide type-explicit reducers
## Issue description

As per recent changes in the data structures of the protocol, the aggregation and tally stages no longer rely on RADON scripts but rather appoint one pre-defined reducer and many pre-defined filters.

As a consequence, we lost the ability to force the output type of a stage to the one of our convenience. For example, a request whose source type is `Integer` currently results in a tally of type `Float` if a `AverageMean` reducer is used anywhere in the pipeline.

### Approach A
One easy fix would be for reducers to honor the input type, i.e. `AverageMean` would return `Float` when applied on `Array<Float>` and `Integer` when the input is `Array<Integer>`.

Note however that having `Integer` sources but wanting a `Float` tally is perfectly legit, and is indeed a behavior that we can foresee many users to expect.

Under that scenario, if an user wants a `Float` average from `Integer` sources, the source scripts themselves should cast the integer to `Float`, so `AverageMean` deals with them accordingly instead of rounding them up or down at the end.

The main merit for this approach is that no consensus-critical data structures need to be changed, and it only affects the business logic in the node, but nothing else in any other component.

### Approach B
A different approach is splitting some reducers into multiple identifiers so we are explicit about the expected output type, e.g. `AverageMeanInteger` vs. `AverageMeanFloat`. 

The main merit for this approach is explicitness.

## Proposal

I'm not settled on recommending one approach or the other yet. Let's gather feedback and discuss.

aesedepece:
Fix #989

### DIFF
--- a/rad/src/filters/deviation.rs
+++ b/rad/src/filters/deviation.rs
@@ -135,7 +135,8 @@ fn boolean_standard_filter(input: &RadonArray, sigmas_float: f64) -> Result<Rado
     match value.first() {
         None => Ok(input.clone()),
         Some(RadonTypes::Float(_)) => {
-            let mean = reducers::average::mean(input)?;
+            let mean =
+                reducers::average::mean(input, reducers::average::MeanReturnPolicy::ReturnFloat)?;
             let mean_float = RadonFloat::try_from(mean)?;
             let std_dev = reducers::deviation::standard(input)?;
             let std_dev_float = RadonFloat::try_from(std_dev)?;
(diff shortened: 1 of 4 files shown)


tmpolaczyk:
The current approach always converts integers to `f64`, which means we lose precision when the sum of the inputs goes above Number.MAX_SAFE_INTEGER:

```rust
    #[test]
    fn test_reduce_average_mean_integer_big() {
        let input = &RadonArray::from(vec![
            RadonInteger::from(1_111_111_111_111_111_111).into(),
            RadonInteger::from(3_333_333_333_333_333_333).into(),
        ]);
        let args = &[Value::Integer(RadonReducers::AverageMean as i128)];
        let expected = RadonTypes::Integer(RadonInteger::from(2_222_222_222_222_222_222));

        let output = reduce(input, args).unwrap();

        assert_eq!(output, expected);
    }
```

```
thread 'reducers::average::tests::test_reduce_average_mean_integer_big' panicked at 'assertion failed: `(left == right)`
  left: `Integer(RadonInteger { value: 2222222222222222336 })`,
 right: `Integer(RadonInteger { value: 2222222222222222222 })`', rad/src/reducers/average.rs:269:9
```

tmpolaczyk:
I would prefer to use an enum instead of `strict: bool`, this way we can more easily update the mean function with more rounding modes (floor, ceil) or special behaviors in general.

```rust
enum XX {
    ReturnFloat,
    RoundToInteger,
}
```

PR #996, opened 2020-01-31
chore: do not strip release binaries

tmpolaczyk:
This should add some information to backtraces, but I'm not sure how much of an impact does this make.
Binary size before strip: 31 MB
Binary size after strip: 24 MB


tmpolaczyk:
This is an alternative to #995

aesedepece:
What do you have against strippers? :see_no_evil:

PR #997, opened 2020-01-31
Implement a try data request in the rad library

mariocao:
This PR adds a new function to try data request given a `RadRequest` in order to be used in the wallet and possibly in the future in the CLI.

### DIFF
--- a/wallet/src/actors/worker/methods.rs
+++ b/wallet/src/actors/worker/methods.rs
@@ -1,5 +1,4 @@
 use jsonrpc_core as rpc;
-use rayon::prelude::*;
 use serde_json::json;
 
 use super::*;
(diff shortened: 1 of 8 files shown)


wallet/src/actors/worker/methods.rs
```diff
@@ -19,26 +18,11 @@ impl Worker {
         })
     }
 
-    pub fn run_rad_request(&self, request: types::RADRequest) -> Result<types::RadonTypes> {
-        // Block on data request retrieval because the wallet was designed with a blocking
-        // run_retrieval in mind.
-        // This can be made non-blocking by returning a future here and updating
```
tmpolaczyk: You can leave this part of comment (last 2 lines) as it is still relevant

aesedepece: Looks to me like the the latest changes doesn't really match the suggestion :sweat_smile:

PR #1010, opened 2020-02-04
Mining rf

### LINKED ISSUE #1000: Implement mining replication and backup factor
In order to reduce the number of epochs without a block, the block mining algorithm should be updated.

Old formula: `VRF / 2^256 < (R) / (N_ARS)` with R = 1
New formula: `VRF / 2^256 < (R) / (N_ARS)` with R = 8

However, the mechanism to select the best block changes. First, we find the block valid under R = 4 which has the most associated reputation. If there is no valid block, we try again with R = 5. Then try again with R = 6, R = 7, and finally R = 8.

tmpolaczyk:
Close #1000


lrubiorod:
Should be show (in trace), the hashes for the different slots when we mine??

PR #1013, opened 2020-02-05
fix(session): add hashset for pending transactions to avoid verify a …

lrubiorod:
…transaction more than once

### DIFF
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -6,7 +6,7 @@ use secp256k1::{
 };
 use serde::{Deserialize, Serialize};
 use std::{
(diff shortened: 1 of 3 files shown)


data_structures/src/chain.rs
```diff
@@ -924,6 +924,8 @@ pub struct TransactionsPool {
     re_hash_index: HashMap<Hash, RevealTransaction>,
     // A map of `data_request_hash` to a map of `reveal_pkh` to `reveal_transaction_hash`
     re_transactions: HashMap<Hash, HashMap<PublicKeyHash, Hash>>,
+    // A hashset of pending transactions hashes
```
```suggestion
    // A hashset of recently recieved transactions hashes
    // Used to avoid validating the same transaction more than once
```

PR #1016, opened 2020-02-06
feat(validations): do not allow commits from the data requester

lrubiorod:


data_structures/src/error.rs
```diff
@@ -81,6 +81,9 @@ pub enum TransactionError {
     /// Signature not found
     #[fail(display = "Transaction signature not found")]
     SignatureNotFound,
+    /// Commitment from data requester
+    #[fail(display = "Commitment has the same PublicKeyHash than Data Request")]
```
tmpolaczyk:
```suggestion
    #[fail(display = "The creator of the Data Request cannot participate in it by sending commits")]
```

node/src/actors/chain_manager/mining.rs
```diff
@@ -229,10 +229,20 @@ impl ChainManager {
 
         for (dr_pointer, data_request_output) in dr_pointers.into_iter().filter_map(|dr_pointer| {
             // Filter data requests that are not in data_request_pool
-            self.chain_state
+            if let Some(dr_state) = self
+                .chain_state
                 .data_request_pool
-                .get_dr_output(&dr_pointer)
-                .map(|data_request_output| (dr_pointer, data_request_output))
+                .data_request_pool
```
tmpolaczyk: Why not leave it as before?

```suggestion
                .get_dr_output(&dr_pointer)
```

tmpolaczyk:
Add a test that checks for this error

node/src/actors/chain_manager/mining.rs
lrubiorod: because I need the pkh of the data request. And this field is in DataRequestState but not in DataRequestOutput

node/src/actors/chain_manager/mining.rs
```diff
@@ -229,10 +229,20 @@ impl ChainManager {
...
+                .get(&dr_pointer)
+            {
+                if dr_state.pkh == own_pkh {
+                    None
```
tmpolaczyk: Add comment:
```suggestion
                    // Skip commits from the creator of the data request, as they cannot be included
                    None
```

PR #1018, opened 2020-02-07
feat(wallet): make index transactions generic

mariocao:
This PR abstracts the index transaction functions to be generic to any `Transaction`. This PR already extends the index transactions to support `DRTransactions`. With that change, deployed transactions (using a wallet's UTXO) are already identified and indexed by the wallet.

### DIFF
--- a/wallet/src/repository/wallet/tests/factories.rs
+++ b/wallet/src/repository/wallet/tests/factories.rs
@@ -58,6 +58,13 @@ pub fn transaction_id() -> types::TransactionId {
     types::TransactionId::SHA256(bytes)
 }
 
+pub fn vtt_from_body(body: types::VTTransactionBody) -> types::Transaction {
+    types::Transaction::ValueTransfer(VTTransaction {
+        body,
+        signatures: vec![],
+    })
+}
+
 #[derive(Default)]
 pub struct Input {
     transaction_id: Option<types::TransactionId>,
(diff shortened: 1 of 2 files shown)


wallet/src/repository/wallet/mod.rs
```diff
@@ -15,10 +16,14 @@ mod tests;
...
 use state::State;
 use std::convert::TryFrom;
-use witnet_data_structures::{
-    chain::{Environment, Epoch},
-    transaction::VTTransactionBody,
-};
+use witnet_data_structures::chain::{Environment, Epoch};
+
+/// Internal structure used to gather state mutations while indexing block transactions
+struct TransactionMutation {
```
tmpolaczyk: This looks vaguely similar to https://github.com/witnet/witnet-rust/blob/631dc672e62eee8866fa62306f5b8054ed5a6e55/validations/src/validations.rs#L1552-L1556

wallet/src/repository/wallet/mod.rs
```diff
@@ -514,123 +520,172 @@ where
...
+                &txn.body.inputs,
+                &txn.body.outputs,
+            )?,
+            _ => return Err(Error::UnsupportedTransactionType(format!("{:?}", txn))),
+        };
 
-        let txn_balance =
-            i64::try_from(txn_balance).map_err(|_| Error::TransactionBalanceOverflow)?;
+        // If balance changes, then update memory state and DB
+        if transaction_mutation.txn_balance != 0 {
```
tmpolaczyk: Is this correct? What if the transaction has zero balance but spends some UTXOs? For example sending a transaction to ourselves.

mariocao: Good catch!

PR #1022, opened 2020-02-10
Memohash sync

aesedepece:
Forced implementation of `Sync` for `MemoHash`.

This is needed because `MemoHash::hash` is of type `Cell`, which in turn has constraint `!Sync`.

Putting `Cell` here gave us memoization through interior mutability. However, it prevented us
from sharing any structure containing a MemoHash from being shared across threads.

This implementation restores the possibility of sharing this structure across threads with
limited impact to thread safety, as the `Cell` itself is only set through calling
`<T: MemoizedHashable>::hash`, which is a pure method. Therefore the maximum impact is a race
condition in the case that `MemoHash::get` is called concurrently.

### DIFF
--- a/data_structures/src/transaction.rs
+++ b/data_structures/src/transaction.rs
@@ -59,6 +59,42 @@ pub enum Transaction {
     Mint(MintTransaction),
 }
 
+impl From<VTTransaction> for Transaction {
+    fn from(transaction: VTTransaction) -> Self {
+        Self::ValueTransfer(transaction)
+    }
+}
+
+impl From<DRTransaction> for Transaction {
+    fn from(transaction: DRTransaction) -> Self {
+        Self::DataRequest(transaction)
+    }
+}
+
+impl From<CommitTransaction> for Transaction {
+    fn from(transaction: CommitTransaction) -> Self {
+        Self::Commit(transaction)
+    }
+}
+
+impl From<RevealTransaction> for Transaction {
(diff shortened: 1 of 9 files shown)


tmpolaczyk:
Not safu

```rust
    #[test]
    fn memohash_sync() {
        let hash0 = Arc::new(VTTransaction::default()).hash();
        for _ in 0..1_000_000 {
            let vt_tx = Arc::new(VTTransaction::default());

            let mut hashes = vec![];
            for _ in 0..8 {
                let vt_tx = Arc::clone(&vt_tx);

                let t = std::thread::spawn(move || {
                    vt_tx.hash()
                });
                hashes.push(t);
            }

            for h in hashes {
                let h = h.join().unwrap();
                assert_eq!(h, hash0);
            }
        }
    }
```

```
thread 'transaction::tests::memohash_sync' panicked at 'assertion failed: `(left == right)`
  left: `e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b828`,
 right: `e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855`', data_structures/src/transaction.rs:629:17
```

Solution: replace `Cell` with `RwLock`, or find a crate that provides safe and fast memoization.

aesedepece:
I withdraw my proposal and close this for the time being.

PR #1025, opened 2020-02-11
fix(logger): set logger as restrictive as given level option

Tommytrg:
Set logger as restrictive as given level option.

### DIFF
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -75,7 +75,7 @@ fn init_logger(opts: LogOptions) {
             None
         })
         .format_module_path(opts.module_path)
-        .filter_level(log::LevelFilter::Info)
+        .filter_level(opts.level)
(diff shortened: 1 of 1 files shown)


tmpolaczyk:
Unfortunately this also affects logs from external crates, such as `tokio_reactor`,

```
[2020-02-11T14:50:38Z INFO  witnet_node::actors::peers_manager::actor] Adding the following peer addresses from config: [V4(127.0.0.1:21337)]
[2020-02-11T14:50:38Z INFO  witnet_node::actors::connections_manager] P2P server has been started at V4(0.0.0.0:21337)
[2020-02-11T14:50:38Z DEBUG tokio_reactor] adding I/O source: 0
[2020-02-11T14:50:38Z DEBUG tokio_reactor::registration] scheduling Read for: 0
```

PR #1030, opened 2020-02-14
Penalize data requester

lrubiorod:


validations/src/validations.rs
```diff
@@ -767,6 +759,22 @@ pub fn validate_tally_transaction<'a>(
         num_commits,
     )?;
 
+    if pkh_liars.len() != ta_tx.out_of_consensus.len() {
+        return Err(TransactionError::MismatchingOutOfConsensusNumber.into());
+    }
+
+    let tally_liars: HashSet<_> = ta_tx.out_of_consensus.iter().collect();
```
tmpolaczyk: You can simplify this comparison by sorting the two vectors and comparing for equality, you don't need a hashset.

node/src/actors/chain_manager/mod.rs
```diff
@@ -721,12 +721,16 @@ impl ReputationInfo {
...
 
         // Set of pkhs which were rewarded in the tally transaction
-        // TODO: Review case of data requester is also a witness
         let tally_vts: HashSet<_> = tally_transaction.outputs.iter().map(|vt| vt.pkh).collect();
+        let tally_liars: HashSet<_> = tally_transaction.out_of_consensus.iter().collect();
         for pkh in commits.keys() {
             // If the identity behaved honestly, it must have received a reward.
             // So we know which identities lied because they do not have a vtt transaction in the tally.
-            let liar = if tally_vts.contains(pkh) { 0 } else { 1 };
+            let liar = if tally_vts.contains(pkh) && !tally_liars.contains(pkh) {
```
tmpolaczyk: Isn't the first check redundant?

```suggestion
            let liar = if !tally_liars.contains(pkh) {
```

validations/src/tests.rs
```diff
@@ -15,9 +15,11 @@ use witnet_protected::Protected;
 use witnet_rad::{error::RadError, reducers::RadonReducers};
 
 use crate::validations::*;
+use witnet_rad::filters::RadonFilters;
```
tmpolaczyk: fmt

node/src/actors/chain_manager/mod.rs
lrubiorod: non revealers has to be punished too

tmpolaczyk: I thought they were also inserted into `tally_liars`, doesn't that make more sense?

PR #1037, opened 2020-02-18
misc bridge fixes

tmpolaczyk:
This makes a few breaking changes to the bridge configuration file.

`enable_block_relay` has been separated into two different flags:

```toml
# Enable block relay from witnet to ethereum, relay only new blocks
# (blocks that were recently consolidated)
enable_block_relay_new_blocks = true
# Enable block relay from witnet to ethereum, relay only old blocks
# (old blocks that were never posted to the block relay, but contain a posted tally)
enable_block_relay_old_blocks = true
```

And now the gas limits can be set in the config file:

```toml
# Gas limits for some methods.
# To let the client estimate, comment out the methods
[gas_limits]
claim_data_requests = 500000
post_data_request = 1000000
post_new_block = 200000
report_data_request_inclusion = 200000
report_result = 200000
```

### DIFF
--- a/bridges/ethereum/src/actors/block_relay_and_poi.rs
+++ b/bridges/ethereum/src/actors/block_relay_and_poi.rs
@@ -34,7 +34,7 @@ pub fn block_relay_and_poi(
             let block_relay_contract = eth_state.block_relay_contract.clone();
             let config2 = config.clone();
 
-            let (block, _is_new_block) = match msg {
+            let (block, is_new_block) = match msg {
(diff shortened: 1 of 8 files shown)


aesedepece:
I'm favorable to merging this and trying it out for some days in a private testnet-7 alike network.

tmpolaczyk:
If the brige fails, just revert commit 98f59a54f5e52a9b1178085e8b5b53a8eed6858a

PR #1038, opened 2020-02-19
Export private key

### LINKED ISSUE #1033: Export node private key

tmpolaczyk:
Close #1033 

This PR adds the ability to export the private key associated to the identity used by the node. A new `exportPrivateKey` command can be used to print that private key:

```
$ cargo run -- node masterKeyExport
Private key for pkh twit1cpnnz00l0lg84mn6t8wsq855f88nx5rw6tkfkn:
xprv1qru3g4rsugt5edkkhpycn5xnxctdwd3zr6m0re294u6urwvzjcexzqya0njnwklpgs8t8ffu5cx3evlc72n5e7tyfyxmg4aaxjt30ls49c4mrp5y
```

Or write the private key to a file inside the storage dir specified in the config:
```
$ cargo run -- node masterKeyExport --write
Private key written to /home/t/witnet-rust/.witnet-rust-testnet-6/private_key_twit1cqhaxkxatx5uf7kr2qe9772qa2hqdp5705jwx4.txt
```

In the future, the `masterKeyImport` command will be used to restore the identity.

### DIFF
--- a/src/cli/node/with_node.rs
+++ b/src/cli/node/with_node.rs
@@ -1,5 +1,4 @@
-use std::net::SocketAddr;
-use std::time::Duration;
+use std::{net::SocketAddr, path::PathBuf, time::Duration};
 
 use structopt::StructOpt;
 
@@ -81,6 +80,24 @@ pub fn exec_cmd(command: Command, mut config: Config) -> Result<(), failure::Err
                 })
                 .expect("Error setting handler for both SIGINT (Ctrl+C) and SIGTERM (kill)");
             })
+        }
+        Command::MasterKeyExport {
+            node,
+            write,
+            write_to,
+        } => {
+            let write_to_path = match (write, write_to) {
+                // Don't write
+                (false, None) => None,
+                // Write to default storage path
+                (true, None) => Some(config.storage.db_path),
+                // Write to custom path
+                (_, Some(path)) => Some(path),
+            };
+            rpc::master_key_export(
+                node.unwrap_or(config.jsonrpc.server_address),
+                write_to_path.as_ref().map(|x| x.as_path()),
+            )
         }
     }
 }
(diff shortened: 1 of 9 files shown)


tmpolaczyk:
Should we add a warning that this command will send the private key over an unencrypted JSON-RPC connection?

lrubiorod:
> Should we add a warning that this command will send the private key over an unencrypted JSON-RPC connection?

where?

lrubiorod:
why allow --pkh option if we can only request our pkh?

tmpolaczyk:
> why allow --pkh option if we can only request our pkh?

This way you don't accidentally export a different private key if for example running two nodes. Also most other clients have an "address" argument in their export private key method, because they support more than one key per node.

lrubiorod:
Those are the beginnings of 10 exported master keys:
xprv1qqjwjc...
xprv1qqzzlx...
xprv1qpdfk4...
xprv1qqaxhw...
xprv1qql4ge...
xprv1qzvql5...
xprv1qr5p2j...
xprv1qqnlkp...
xprv1qqzkem...
xprv1qqwr6w...

Is it ok that after xprv1 prefix, there always be a q or most of the time qq?

tmpolaczyk:
> Is it ok that after xprv1 prefix, there always be a q or most of the time qq?

This is the "master key" prefix, there is nothing we can do to change that.

aesedepece:
I'd suggest renaming this method to `masterKeyExport` for two reasons:
- Being explicit about the fact that what will get exported is the master key of the node.
- Making sure that both import / export methods are listed consecutively in the list of supported methods.

aesedepece:
UX wise, that'd remove the need for the `--pkh` argument too, IMO.

src/cli/node/with_node.rs
```diff
@@ -1,4 +1,5 @@
 use std::net::SocketAddr;
+use std::path::PathBuf;
```
lrubiorod: fmt

aesedepece:
👏👏👏👏👏

PR #1040, opened 2020-02-19
Add option to easily send value to a new wallet

### LINKED ISSUE #1029: Implement a method to create a prefilled wallet
As a user running a Witnet node in testnet, I want be able to create a prefilled wallet with created addresses and wits on it.

tmpolaczyk:
Close #1029 

Add optional field `prefill` to `unlock_wallet`

```
{
  "wallet_id": "...",
  "password": "...",
  "prefill": [1000, 2000, 3000]
}
```

This will generate 3 new addresses and send 1000 nanowits to the first address, 2000 to the second and 3000 to the third.

In case of error (if the node is not synced or it does not have any balance), the addresses are generated but will not receive any wits.

### DIFF
--- a/wallet/src/actors/app/handlers/close_session.rs
+++ b/wallet/src/actors/app/handlers/close_session.rs
@@ -5,7 +5,6 @@ use crate::actors::app;
 use crate::types;
 
 #[derive(Debug, Serialize, Deserialize)]
-#[serde(rename_all = "camelCase")]
 pub struct CloseSessionRequest {
     pub(crate) session_id: types::SessionId,
 }
--- a/wallet/src/actors/app/handlers/create_data_req.rs
+++ b/wallet/src/actors/app/handlers/create_data_req.rs
@@ -6,7 +6,6 @@ use crate::types;
 use crate::types::{Hashable as _, ProtobufConvert as _};
 
 #[derive(Debug, Deserialize)]
-#[serde(rename_all = "camelCase")]
 pub struct CreateDataReqRequest {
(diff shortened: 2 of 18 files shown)


aesedepece:
Thinking of mainnet stage, I think this feature should be reliant on a
configuration setting that enables or disables it. Otherwise, the security
implications of this are just outrageous!

tmpolaczyk:
This PR is a simple wrapper around the `sendValue` JSON-RPC node method, so it would be easier to add an option to disable dangerous methods when running a public node.

aesedepece:
Great. That's exactly what I meant. Thanks!

tmpolaczyk:
Ok, would you classify a hypothetical `exportPrivateKey` method as "dangerous" or as "shouldn't even exist"?

aesedepece:
Good point. Take Geth for a example: once you enable JSONRPC, you need to
specify which features do you want to expose through it. Sending funds or
exporting keys are totally the kind of features you don't want to expose by
default ;)

aesedepece:
Is this the final PR?

tmpolaczyk:
Yes, we are waiting for the next witnet release because sheikah uses the witnet binary for testing.

aesedepece:
But they get the binary from the releases page, right? This can be merged safely, doesn't it?

tmpolaczyk:
According to @Tommytrg it's better if the API of the binary is the same as the API of the master branch.

aesedepece:
@Tommytrg please explain yourself :pray:

PR #1042, opened 2020-02-19
[wip] Genesis block

lrubiorod:


aesedepece:
For security reasons, I'm against making the node download the genesis
block from the wild Internet.

I'd rather make it read it from the filesystem, and distribute the JSON
along the binary, so that people can look into it before running the node,
instead of blindly building a chain on top of an unknown genesis block 🤷‍♂️

tmpolaczyk:
Did a few improvements on this on a different branch: https://github.com/tmpolaczyk/witnet-rust/tree/genesis_block_1

Now if two nodes build the genesis block they can sync together, although I haven't tested if the UTXOS from the genesis block can be spent, and genesis block validation is not implemented.

lrubiorod:
Continue in https://github.com/tmpolaczyk/witnet-rust/tree/genesis_block_1

PR #1048, opened 2020-02-27
feat(wallet): implement epoch to timestamp conversion

tmpolaczyk:
The wallet now reads the consensus constants from the config file (or uses the defaults if they are not present). These constants include the epoch zero timestamp and the checkpoints period, which are used to convert from epoch number to timestamp.


tmpolaczyk:
Merged

PR #1060, opened 2020-03-03
Import master key

### LINKED ISSUE #1034: Import node private key

tmpolaczyk:
Close #1034 

The idea is to be able to import the master key when starting the node for the first time:

```
$ cargo run -- node run --master-key-import private_key_twit1wpmdccr3rn5zw8wuppu4wsg34ul9yvl6awt2q5.txt 
```

The current implementation requires users to delete the storage in order to import a different master key. This avoids any problems that could arise from people accidentally overwriting their private key and losing their mined wits. Also, it's significantly simpler to implement, as the current code assumes that the node master key never changes in many places.

This is the error that appears in the logs when trying to import a master key without deleting the storage:
```
[2020-03-04T11:47:04Z ERROR witnet_node::signature_mngr] Failed to configure master key: Tried to overwrite node master key with a different one.
    Node pkh:     twit19ddjj2w8vv98mzdujkzgw36kqtpnt3l22r9dj9
    Imported pkh: twit1wpmdccr3rn5zw8wuppu4wsg34ul9yvl6awt2q5
    
    In order to import a different master key, you need to export the current master key and delete the storage
Error: Non-zero exit code: 1
```

In order to be able to import a different master key without deleting the storage, we would need to re-index the list `own_utxos` with the new pkh. Otherwise only the imported reputation would be correct, the balance would be 0. That would still require stopping the node, though.

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -261,6 +261,10 @@ pub struct Storage {
     /// Path to the directory that will contain the database. Used
     /// only if backend is RocksDB.
     pub db_path: PathBuf,
+    /// Path to the master key to import when initializing the node
+    #[partial_struct(skip)]
+    #[partial_struct(serde(default))]
+    pub master_key_import_path: Option<PathBuf>,
 }
 
 fn as_protected_string<'de, D>(deserializer: D) -> Result<Option<Protected>, D::Error>
@@ -518,6 +522,7 @@ impl Storage {
                 .db_path
                 .to_owned()
                 .unwrap_or_else(|| defaults.storage_db_path()),
+            master_key_import_path: config.master_key_import_path.clone(),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


config/src/config.rs
```diff
@@ -113,6 +113,10 @@ pub struct Config {
     #[partial_struct(ty = "PartialMempool")]
     #[partial_struct(serde(default))]
     pub mempool: Mempool,
+
+    // TODO: where to keep this?
+    #[partial_struct(serde(default))]
+    pub master_key_import: Option<PathBuf>,
```
lrubiorod:
```suggestion
    pub master_key_import_path: Option<PathBuf>,
```

config/src/config.rs
```diff
@@ -113,6 +113,10 @@ pub struct Config {
     #[partial_struct(ty = "PartialMempool")]
     #[partial_struct(serde(default))]
     pub mempool: Mempool,
+
+    // TODO: where to keep this?
```
lrubiorod: Storage?

node/src/signature_mngr.rs
```diff
@@ -6,11 +6,13 @@
 use actix::prelude::*;
 use failure;
 use failure::bail;
+use failure::format_err;
```
lrubiorod: fmt

PR #1061, opened 2020-03-03
probability to mine a block

### LINKED ISSUE #1035: Log eligibility probability

lrubiorod:
Close #1035

### DIFF
--- a/node/src/actors/chain_manager/mining.rs
+++ b/node/src/actors/chain_manager/mining.rs
@@ -1,4 +1,4 @@
-use std::convert::TryFrom;
+use std::{cmp::Ordering, convert::TryFrom};
 
 use actix::{
     ActorFuture, AsyncContext, Context, ContextFutureSpawner, Handler, SystemService, WrapFuture,
@@ -9,7 +9,8 @@ use futures::future::{join_all, Future};
 use witnet_data_structures::{
     chain::{
         Block, BlockHeader, BlockMerkleRoots, BlockTransactions, CheckpointBeacon, EpochConstants,
-        Hashable, PublicKeyHash, TransactionsPool, UnspentOutputsPool, ValueTransferOutput,
+        Hashable, PublicKeyHash, ReputationEngine, TransactionsPool, UnspentOutputsPool,
+        ValueTransferOutput,
     },
     data_request::{create_tally, DataRequestPool},
     radon_report::{RadonReport, ReportContext},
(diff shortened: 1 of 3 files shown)


node/src/actors/chain_manager/mining.rs
```diff
@@ -136,6 +145,11 @@ impl ChainManager {
                         Yellow.bold().paint("[Mining]"),
                         Yellow.bold().paint(beacon.checkpoint.to_string())
                     );
+                    let mining_prob =
+                        calculate_mining_probability(&rep_engine, own_pkh, mining_rf, mining_bf);
+                    // Discount the already reached probability
+                    let mining_prob = mining_prob / probability * 100.0;
```
tmpolaczyk: This division is no incorrect, right?

```suggestion
```

tmpolaczyk: Although I think I would prefer `calculate_mining_probability` to return a number between 0.0 and 1.0, and multiply by 100 here before printing.

lrubiorod: I was thinking the same before, but it was to not change the other code. Anyways, I will change

node/src/actors/chain_manager/mining.rs
```diff
@@ -625,6 +639,97 @@ fn build_block(
     (block_header, txns)
 }
 
+fn internal_calculate_mining_probability(
+    rf: f64,
+    n: f64,
+    first_time: bool,
```
tmpolaczyk: If you pass a boolean argument to the function, it's like having two different functions. If the code can be simplified to calculate this argument inside then great, otherwise just split it into two functions to improve clarity.

lrubiorod: The idea is differing the main prob and the added prob when using bf

node/src/actors/chain_manager/mining.rs
```diff
@@ -625,6 +639,94 @@ fn build_block(
...
+        // we obtain the maximum probability divided in the nodes we share the same reputation
+        1.0 / (equal as f64)
+    } else if rf > n && greater > 0 {
+        // In case of replication factor exceed the active node number and not being the most reputed
+        // we obtain the minimum probability
+        0.0
+    } else {
+        let mut aux = internal_calculate_mining_probability(rf, n as f64, rf, greater, equal, less);
+        let mut k = rf + 1;
+        while k <= bf {
```
tmpolaczyk: To avoid the break on the next line:

```suggestion
        while k <= bf && k <= n {
```

node/src/actors/chain_manager/mining.rs
```diff
@@ -136,6 +145,11 @@ impl ChainManager {
                         Yellow.bold().paint("[Mining]"),
                         Yellow.bold().paint(beacon.checkpoint.to_string())
                     );
+                    let mining_prob =
+                        calculate_mining_probability(&rep_engine, own_pkh, mining_rf, mining_bf);
+                    // Discount the already reached probability
+                    let mining_prob = mining_prob / probability * 100.0;
+                    log::info!("Probability to mine a block: {:.6}%", mining_prob);
```
tmpolaczyk: Maybe say that this is the probability that the created block will be the consolidated block:

```suggestion
                    log::info!("Probability that the mined block will be selected: {:.6}%", mining_prob);
```

Issue #1047, opened 2020-02-26
Use `.witnet` as storage path by default


aesedepece:
## Description

In the past, we used to sandbox the storage of different testnet versions by using different paths for RocksDB.

Since chain data and identities are now sandboxed internally to the database (see #970, #976), we can stop changing paths. 

This would be a final change: storage paths would not need to be updated anymore, not even past the mainnet token generation event.

I'm suggesting using `.witnet` as the default path storage.

## Changes

This likely implies the following changes:
- Modifying `witnet.toml` so that the storage path reads `.witnet` instead of `.witnet-testnet-6`
- Modifying the defaults in the `witnet_config` sub-crate to reflect the same change.

## Impact

- The first testnet version that would make use of this change could be Testnet 7.
- Existing docker containers will likely change identities. Identities can be preserved by means of any of these:
  - editing the volume binding for the storage so that the old storage is bound into `.witnet` instead of `.witnet-testnet-6`
  - the key is exported and re-imported
- Existing non-docker nodes will likely change identities. Identities can be preserved by means of any of these:
  - renaming `.witnet-testnet-6` as `.witnet`
  - modifying `witnet.toml` so that storage path reads `.witnet` instead of `.witnet-testnet-6`
  - the key is exported and re-imported
- New docker containers or nodes are not affected by this change. Only caution is that the documentation needs to be updated so the recommended commands mention `.witnet` instead of `.witnet-testnet-6`.

aesedepece:
As a side note, we could take the chance to rename the `testnet-3` configuration environment as simply `testnet`.

tmpolaczyk:
> As a side note, we could take the chance to rename the `testnet-3` configuration environment as simply `testnet`.

In a separate commit please

aesedepece:
@witnet/devs Are we in time for including this into the final `0.7.0` release?

The idea of updating the documentation every time we make a new release is quite unamusing to me.

tmpolaczyk:
We can just edit the `witnet.toml` that's shipped with the release and change the defaults later.

Actually, the config crate sets the default to `.witnet-rust-testnet-3` :see_no_evil:

aesedepece:
@tmpolaczyk you want to work on it or is it OK for me to take it?

tmpolaczyk:
Go ahead

aesedepece:
This is solved on master.

PR #1069, opened 2020-03-10
chore(TOML): use `.witnet` as storage path + set log level to INFO

aesedepece:
Partially solving #1047

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3900,7 +3900,7 @@ dependencies = [
 
 [[package]]
 name = "witnet"
-version = "0.6.2"
+version = "0.7.0"
 dependencies = [
  "bytecount 0.6.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "ctrlc 3.1.3 (registry+https://github.com/rust-lang/crates.io-index)",
--- a/node/src/actors/chain_manager/mining.rs
+++ b/node/src/actors/chain_manager/mining.rs
@@ -1,4 +1,4 @@
-use std::convert::TryFrom;
+use std::{cmp::Ordering, convert::TryFrom};
 
 use actix::{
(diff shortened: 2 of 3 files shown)


witnet.toml
```diff
@@ -29,7 +29,7 @@ genesis_path = "genesis_block.json"
 genesis_mining = false
```
aesedepece: @witnet/devs should this be set to `true`? What's the use of this flag at all?

tmpolaczyk: If set to true, nodes will read the `genesis_block.json` file and try to create a genesis block. If set to false, nodes will just download the genesis block from other peers.

In order to change it to true we need to bundle the `genesis_block.json` file or otherwise the docker image won't work.

I think issue #1065 will remove this flag entirely making true the default.

tmpolaczyk:
Why change the default log level?

witnet.toml
lrubiorod: yes, I am doing that. I will upload a PR soon

aesedepece:
> Why change the default log level?

Isn't it a bit overwhelming for newcomers?

witnet.toml
aesedepece: Great. Excited about #1065 !

tmpolaczyk:
> > Why change the default log level?
> 
> Isn't it a bit overwhelming for newcomers?

I don't know about that, but this will reduce the quality of the issues reported by node operators, as we will have to explicitly ask for "debug" logs.

aesedepece:
> I don't know about that, but this will reduce the quality of the issues reported by node operators, as we will have to explicitly ask for "debug" logs.

Ok, let's leave it as it is for the time being. I committed too impulsively. I'll withdraw my latest commit. Thanks!

PR #1071, opened 2020-03-11
fix(genesis): remove mining_genesis flag to avoid genesis inconsistence

### LINKED ISSUE #1064: Minor issue with genesis block creation: nodes cannot sync until a new block is mined
When bootstrap nodes mine the genesis block but do not mine any other blocks, other nodes that did not mine the genesis block cannot sync. This is because the beacon sent by the bootstrap nodes has `checkpoint: 0, hash: genesis_hash` which is considered out of consensus by the nodes that are on `checkpoint: 0, hash: bootstrap_hash`.

As soon as some node joins the network with `genesis_mining = true` and starts mining, the bootstrap nodes consolidate that block and all the nodes can sync.

Solution: add a special case to the `PeersBeacons` handler in `ChainManager`, similar to the special case when the consensus is that there is no genesis block. This should allow nodes to "sync" to the genesis hash.

### LINKED ISSUE #1065: Make genesis block compulsory

lrubiorod:
Close #1064 
Close #1065

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -301,8 +301,6 @@ pub struct Mining {
         rename = "data_request_timeout_seconds"
     ))]
     pub data_request_timeout: Duration,
-    /// Binary flag whether to mine genesis block or not
-    pub genesis_mining: bool,
     /// Genesis block path
     pub genesis_path: String,
 }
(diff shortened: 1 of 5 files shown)


genesis_block.json
```diff
@@ -0,0 +1,28 @@
+{
```
tmpolaczyk: Do we want this file in this repo?

lrubiorod: it everyone has to have the genesis block, it is not easier to have the genesis in the repo?

lrubiorod: what do you thin @aesedepece

aesedepece:
It has its own repo, right?

tmpolaczyk:
Ok so if we store the genesis block in a different repo, then when running a node from source users need to perform an extra step before being able to connect to the testnet. Where do we document this?

PR #1072, opened 2020-03-11
feat: add option for disabling sensitive JSON-RPC methods

### LINKED ISSUE #1050: Add configuration option for disabling sensitive JSONRPC methods
## Description

As briefly discussed in #1040, there exist methods whose current security assumptions only hold true if the node is executed in a private environment, and it is "personal", i.e. not shared by multiple users.

These methods are, at least:

- `getBalance` without `--pkh` flag
- `getReputation` without `--pkh` flag
- `masterKeyExport`
- `masterKeyImport`
- `send`
- `show-config`

## Proposal

- All this method should be switchable on or off through a single configuration setting.
- The setting can exist inside the `[jsonrpc]` section of `witnet.toml`, and be named `enable_sensistive_methods` (feel free to come up with a better name).
- The default value for `enable_sensitive_methods` should be `true`.

## Consequences
- This is no breaking change.
- This change should be reflected in the documentation.

## Side notes

The rationale for calling it `enable...` instead of `disable...` is because we already have other settings that start with `enable`, and it is better UX not to mix positive and negative settings.

tmpolaczyk:
Close #1050 

Now, users can set `enable_sensitive_methods = false` to disable the following JSON-RPC methods:

```
"createVRF",
"getPkh",
"getPublicKey",
"masterKeyExport",
"sendRequest",
"sendValue",
"sign",
```

Trying to use one of these methods when it has been disabled will result in a custom "unauthorized" error:

```
 $ cargo run -- -c witnet_01.toml node getPkh
    Finished dev [unoptimized + debuginfo] target(s) in 0.20s
     Running `target/debug/witnet -c witnet_01.toml node getPkh`
Loading config from: witnet_01.toml
Setting log level to: ERROR, source: Config
Error: ServerError { code: -32603, message: "Method getPkh not allowed while node setting json_rpc.enable_sensitive_methods is set to false" }
```

The default behavior is the same as before (all methods enabled), this is not a breaking change.

Unresolved issues:

* Print warning when starting the JSON-RPC server with `enable_sensitive_methods = true` and listening to a non-local address?
* Alternative names: `expose_private_methods`, `enable_personal_methods`, `public_mode`, `allow_insecure_methods`

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -284,6 +284,8 @@ pub struct JsonRPC {
     /// JSON-RPC server address, that is, the socket address (interface ip and
     /// port) for the JSON-RPC server
     pub server_address: SocketAddr,
+    /// Enable methods not suitable for shared nodes
+    pub enable_sensitive_methods: bool,
 }
 
 /// Mining-related configuration
@@ -532,6 +534,10 @@ impl JsonRPC {
                 .server_address
                 .to_owned()
                 .unwrap_or_else(|| defaults.jsonrpc_server_address()),
+            enable_sensitive_methods: config
+                .enable_sensitive_methods
+                .to_owned()
+                .unwrap_or_else(|| defaults.jsonrpc_enable_sensitive_methods()),
         }
     }
 }
@@ -834,6 +840,7 @@ mod tests {
         let partial_config = PartialJsonRPC {
             enabled: None,
             server_address: Some(addr),
+            enable_sensitive_methods: None,
         };
         let config = JsonRPC::from_partial(&partial_config, &Testnet);
 
--- a/config/src/defaults.rs
+++ b/config/src/defaults.rs
@@ -109,6 +109,11 @@ pub trait Defaults {
     /// Default JSON-RPC server addr
     fn jsonrpc_server_address(&self) -> SocketAddr;
 
+    /// JSON-RPC sensitive methods enabled by default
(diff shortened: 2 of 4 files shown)


lrubiorod:
Instead of hide sensitive methods, it wouldnt be better to request a password?

tmpolaczyk:
> Instead of hide sensitive methods, it wouldnt be better to request a password?

Good idea!

However, since the JSON-RPC connection is not encrypted, that wouldn't be secure. If we switch to JSON-RPC over TLS, then we can do that.

lrubiorod:
in the issue #1050 getReputation and getBalance is requested as a sensitive methods

tmpolaczyk:
> in the issue #1050 getReputation and getBalance is requested as a sensitive methods

Yes, that refers to calling them with no arguments, as they default to use the node pkh. However that is implemented at the CLI level, the CLI method `getBalance` calls `getPkh`.

So running `witnet node getBalance` results in:

    Error: ServerError { code: -32601, message: "Method not found" }

(We should probably improve that error message to give context around which method was not found)

But running `witnet node getBalance --pkh=twit1xwn...` works

aesedepece:
An ideal error message would read something in the lines of:

> Method `{}` not allowed while node setting `json_rpc.enable_sensitive_methods` is set to `false`.

PR #1074, opened 2020-03-12
fix: rename testnet3 environment to testnet

### LINKED ISSUE #1073: Rename "testnet3" environment to just "testnet"

tmpolaczyk:
Close #1073 

This can break some users configuration files: if they had explicitly set `environment = "testnet-3"`, they will see this error:

```
Error: Error parsing config file: unknown variant `testnet-3`, expected `mainnet` or `testnet` for key `environment` at line 40 column 1
```

But the default configuration file does not have that field, so hopefully this doesn't cause many problems.

This also changes the default storage path from `.witnet-rust-testnet-3` to `.witnet`. So users that have not set any `db_path` in the configuration file will have to manually rename that folder. But since the default configuration file now explicitly sets `db_path = ".witnet"`, this shouldn't be a common problem.


aesedepece:
Awesome.

Why `.witnet-rust-testnet` and not `.witnet` as it is in the TOML file in master?

PR #1081, opened 2020-03-13
to ease operating a node

bertux:
by using docker-compose to put all the Docker options in a file instead of a long command


aesedepece:
Merged as 36945e41638901159cc2b6255cefa466bd088d28. Thanks @bertux, that was a quality contribution!!!!

PR #1083, opened 2020-03-14
fix #954 Unused field `epoch` in `MintTransaction`

### LINKED ISSUE #954: Unused field `epoch` in `MintTransaction`
The `epoch` field in `MintTransaction` must always be equal to the block epoch. This field is only used in validations, where it raises a `BlockError::InvalidMintEpoch` if it is different from the block epoch:

https://github.com/witnet/witnet-rust/blob/170617a8b78aae5ab86c140bd324a4ddae4d5209/validations/src/validations.rs#L137

Since we already need the block epoch to validate the mint transaction, this field can be removed without breaking anything, and we can assume that the mint epoch is always equal to the block epoch.

Removing this field will make the `MintTransaction` just a `ValueTransferOutput`, with the fields `pkh`, `value` and `time_lock`.

```proto
message MintTransaction {
    fixed32 epoch = 1;
    ValueTransferOutput output = 2;
}
```

Steps:

* Make sure that we actually want to remove it
* Remove `epoch` field from `MintTransaction` in `witnet.proto`
* Remove `epoch` field from `MintTransaction` in `transactions.rs`
* Remove error `BlockError::InvalidMintEpoch` as it is now impossible
* Fix tests

xiaods:
fix #954

### DIFF
--- a/schemas/witnet/witnet.proto
+++ b/schemas/witnet/witnet.proto
@@ -200,6 +200,7 @@ message TallyTransaction {
 }
 
 message MintTransaction {
+    // The epoch field is used to ensure that each MintTransaction has a unique hash
     fixed32 epoch = 1;
     ValueTransferOutput output = 2;
 }
--- a/validations/src/tests.rs
+++ b/validations/src/tests.rs
@@ -649,7 +649,7 @@ fn vtt_two_inputs_one_signature() {
     };
     let utxo_pool = build_utxo_set_with_mint(vec![vto_21, vto_13], None, vec![]);
     let utxo_diff = UtxoDiff::new(&utxo_pool);
-    let vti0 = Input::new(utxo_pool.iter().nth(0).unwrap().0.clone());
+    let vti0 = Input::new(utxo_pool.iter().next().unwrap().0.clone());
     let vti1 = Input::new(utxo_pool.iter().nth(1).unwrap().0.clone());
 
     let vto0 = ValueTransferOutput {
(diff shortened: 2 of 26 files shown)


validations/src/tests.rs
```diff
@@ -1,4 +1,4 @@
-use std::sync::atomic::{AtomicU32, Ordering};
+//use std::sync::atomic::AtomicU32;
```
lrubiorod: it should be removed

validations/src/tests.rs
```diff
@@ -87,19 +87,18 @@ fn sign_t3<H: Hashable>(tx: &H) -> KeyedSignature {
 }
 
 // Counter used to prevent creating two transactions with the same hash
-static TX_COUNTER: AtomicU32 = AtomicU32::new(0);
+// static TX_COUNTER: AtomicU32 = AtomicU32::new(0);
```
lrubiorod: here too

tmpolaczyk: Oh yeah, now I remember. The epoch field is used to ensure that every `MintTransaction` has a unique hash, because otherwise when one node mines two empty blocks, the hash of the mint transaction will be the same in this two blocks, as it only depends on pkh and value.

schemas/witnet/witnet.proto
```diff
@@ -200,8 +200,7 @@ message TallyTransaction {
 }
 
 message MintTransaction {
-    fixed32 epoch = 1;
-    ValueTransferOutput output = 2;
+    ValueTransferOutput output = 1;
```
tmpolaczyk: So we have realized that the epoch field is actually useful, and shouldn't be removed. :sweat_smile: 

So just revert everything and add a comment to the schema, and we can mark the issue as fixed:

```suggestion
    // The epoch field is used to ensure that each MintTransaction has a unique hash
    fixed32 epoch = 1;
    ValueTransferOutput output = 2;
```

xiaods: done

aesedepece:
Merged in as 7488adcd75bf03de44523c61d8e3a53cfa7cb623.
Thanks @xiaods!

Issue #673, opened 2019-06-04
Hash::from_str silently accepts 32.5-byte inputs


tmpolaczyk:
The current implementation of `Hash::from_str` should only accept 32-byte values (64 characters). However, it also accepts 32.5-byte values (65 characters), ignoring the last character.

Sample test:

```rust
    #[test]
    fn hash_from_str() {
        // 32 bytes
        let a = Hash::from_str("1111111111111111111111111111111111111111111111111111111111111111");
        assert!(a.is_ok());

        // 32.5 bytes
        let b = Hash::from_str("11111111111111111111111111111111111111111111111111111111111111112");
        assert!(b.is_err());

        // 33 bytes
        let c =
            Hash::from_str("111111111111111111111111111111111111111111111111111111111111111122");
        assert!(c.is_err());
    }
```

This problem arises from the usage of the [`parse_hex` function from `witnet_util`](https://github.com/witnet/witnet-rust/blob/2cd5b6c0347df297013b8088ca49b723c50bf3f5/util/src/parser.rs#L1).

The best solution in my opinion is to remove that function and instead use the [`hex` crate](https://crates.io/crates/hex), which provides a really nice `decode_to_slice` method. Using that method, parsing a hash would be as simple as:

```rust
// s: &str
let mut h = [0; 32];
hex::decode_to_slice(s, &mut h)?;
Ok(Hash::SHA256(h))
```

This also simplifies the `HashParseError` type, which should only be a wrapper to the `FromHexError` error from the `hex` crate (we cannot use that error directly because it does not implement `Fail`).

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 0.043 ETH (5.0 USD @ $116.22/ETH)  attached to it as part of the Witnet fund.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/witnet/witnet-rust/673/4112).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/witnet/witnet-rust/673/4112).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $57,203.89 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
[...]

PR #1084, opened 2020-03-15
Issue 673

kiminuo:
@tmpolaczyk Extremely minimal fix for #673. It would require more work and some guidance on how to fix it properly / more in-depth.


data_structures/src/chain.rs
```diff
@@ -706,13 +706,14 @@ impl FromStr for Hash {
     type Err = HashParseError;
```
KiChjang: I think we should remove `HashParseError` altogether and substitute it with `hex::FromHexError`, that way we don't need to convert error types back and forth.

data_structures/src/chain.rs
```diff
@@ -2578,6 +2579,22 @@ mod tests {
...
 
+    #[test]
+    fn hash_from_str() {
+        // 32 bytes
+        let a = Hash::from_str("1111111111111111111111111111111111111111111111111111111111111111");
+        assert!(a.is_ok());
+
+        // 32.5 bytes
+        let b = Hash::from_str("11111111111111111111111111111111111111111111111111111111111111112");
+        assert!(b.is_err());
```
KiChjang: If we do decide to use `FromHexError` directly, then we can even assert equality like so:

```suggestion
        assert_eq!(b, Err(FromHexError::InvalidStringLength));
```

data_structures/src/chain.rs
kiminuo: Yes, that's my sentiment too.

tmpolaczyk: Makes sense, I think the only problem is that `hex::FromHexError` does not implement `Fail` and may not integrate well with our error system. So it will be easier to just define:

```rust
#[derive(Debug, Fail)]
pub enum HashParseError {
    #[fail(display = "Failed to parse hex: {}", _0)]
    Hex(#[cause] hex::FromHexError),
}
```

And add `.map_err(HashParseError::Hex)` everywhere.

tmpolaczyk:
Thanks for the PR! Could you also delete this file as it should no longer be needed:

https://github.com/witnet/witnet-rust/blob/23e746c1b10a19a679e67477aaf973fb935b474c/util/src/parser.rs

kiminuo:
@tmpolaczyk Actually, `parse_hex` is also present here https://github.com/witnet/witnet-rust/blob/35f9a8e1f40495a44b3264d85ea0c8806c65e6c4/data_structures/src/chain.rs#L710

I'll sign in to Discord and I'll try to contact you there, if you don't mind.

tmpolaczyk:
@kiminuo sure, go ahead

aesedepece:
Merged as 96592dd9ae8d492f05323bc633433552daa13396.

Welcome to the Witnet dev community @kiminuo and @KiChjang !

PR #1090, opened 2020-03-17
feat(wallet): add sign data JSON-RPC method

### LINKED ISSUE #1067: Implement message signing method
The wallet should provide a method to be consume by clients (e.g. Sheikah) in order to support generic message signing. One example use case scenario will be the token claiming process.

mariocao:
This JSON-RPC method `sign_data` is used to sign a generic message by using the wallet's master key.

## Request Params

```
{
  "sessionId": "<session id>",
  "walletId": "<wallet id>",
  "data": "<message to be hashed and signed>",
  "extended_pk": "<true in order to include extended public key>"
}
```

`extended_pk` flag is set to `true` the result will include the extended public key by including the `chaincode` field. This field is considered sensitive because with an extended public key it is possible to derive non-hardened child public keys. 

## Response

```
{
  "chaincode": "<hex string>",
  "public_key": "<hex string in compressed format>",
  "signature": "<hex string>"
}
```

Closes #1067


wallet/src/actors/app/handlers/sign_data.rs
```diff
@@ -0,0 +1,28 @@
...
+use serde::{Deserialize, Serialize};
+
+use crate::actors::app;
+use crate::{model, types};
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct SignDataRequest {
+    session_id: types::SessionId,
+    wallet_id: String,
+    data: String,
```
tmpolaczyk: Add comment explaining that the data is a string which will be hashed.

mariocao: Wasn't it clear enough? :rofl: 
I will include it. :+1:

tmpolaczyk: Ah, sorry, I meant a comment above the field:
```suggestion
    // Message to be signed
    data: String,
```

wallet/src/model.rs
```diff
@@ -48,6 +48,13 @@ pub struct Balance {
     pub total: String,
 }
 
+#[derive(Debug, Serialize)]
+pub struct ExtendedKeyedSignature {
+    pub signature: String,
+    pub public_key: String,
+    pub chaincode: String,
```
tmpolaczyk: When `extended_pk` is set to false, this field is serialized as `"chaincode":""`. It is possible to not serialize it by using a serde attribute:

```suggestion
    #[serde(skip_serializing_if = "is_default")]
    pub chaincode: String,
```

But you need to define the `is_default` function:

```rust
fn is_default<T: Default + PartialEq>(t: &T) -> bool {
    t == &T::default()
}
```

aesedepece:
Merged as 38838dfb76081df7587028afbb1473146934d523 :heart_decoration: :

PR #1092, opened 2020-03-18
feat(config): increase default values of some consensus constants

lrubiorod:
block weight to 150K
activity period to 2000 epochs
alpha clock to 20000 witnesses acts  

This PR is related to https://github.com/witnet/research/issues/94

### DIFF
--- a/config/src/defaults.rs
+++ b/config/src/defaults.rs
@@ -133,8 +133,7 @@ pub trait Defaults {
     }
 
     fn consensus_constants_max_block_weight(&self) -> u32 {
-        // TODO: Replace  with real max_block_weight value used in mainnet
-        10_000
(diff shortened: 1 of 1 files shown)


config/src/defaults.rs
```diff
@@ -133,8 +133,7 @@ pub trait Defaults {
     }
 
     fn consensus_constants_max_block_weight(&self) -> u32 {
-        // TODO: Replace  with real max_block_weight value used in mainnet
-        10_000
+        150_000
```
aesedepece: I am a man of powers of two:
```suggestion
        153_600
```

aesedepece:
Shall we do a follow-up PR with checkpoint period set to 45?

Issue #1098, opened 2020-03-18
Assessment: specify a limit of data requests or sources to resolve


lrubiorod:
When a node tries to request to too many sources, it implies that its EpochManager messages could be delayed and obtain epochs bigger and lower that the specified time. It could produce forks, or sending useless messages trough the network, lose peers and even panics.

The idea is creating a config parameter which specifies the maximum number of data requests or data requests sources that a node could solve without problems. That parameter could be modified by the node operator according to the system used.

This proposal could help to the reputation will be more distributed.

aesedepece:
I'll be working on this.

aesedepece:
```
[2020-03-18T15:50:30Z INFO  witnet_node::actors::chain_manager::mining] [Mining] Discovered eligibility for mining a data request b095e8bf0d81dd2a8065a0a0b315f510682b992043c46f0b80328abde33b1201 for epoch #14501
[2020-03-18T15:50:30Z INFO  witnet_node::actors::chain_manager::mining] [Mining] Refrained from resolving data request b095e8bf0d81dd2a8065a0a0b315f510682b992043c46f0b80328abde33b1201 for epoch #14501 because it contains 2 sources, which added to the sources that have already been retrieved 0 would total 2 retrievals, which exceed current limit (0)
```

PR #1099, opened 2020-03-18
Limit the number of retrieved sources per epoch

aesedepece:
This PR introduces a configuration entry, `data_request_max_retrievals_per_epoch`, that enables setting a limit on the number of sources that can be retrieved in a single epoch.

The motivation for this change is reducing the likelihood for a node to be de-synced because of some retrievals draining way too much resources.

Solve #1098

### DIFF
--- a/config/src/defaults.rs
+++ b/config/src/defaults.rs
@@ -124,6 +124,12 @@ pub trait Defaults {
         Duration::from_secs(2)
     }
 
+    /// Set the limit of retrievals per epoch to 65_535.
+    /// This in practice equals no limit enforcement.
+    fn mining_data_request_max_retrievals_per_epoch(&self) -> u16 {
+        core::u16::MAX
+    }
+
     /// Genesis block path, "./genesis_block.json" by default
     fn mining_genesis_path(&self) -> String {
         "genesis_block.json".to_string()
--- a/node/src/actors/chain_manager/mining.rs
+++ b/node/src/actors/chain_manager/mining.rs
@@ -1,4 +1,11 @@
-use std::{cmp::Ordering, convert::TryFrom, sync::atomic::Ordering::Relaxed};
+use std::{
+    cmp::Ordering,
+    convert::TryFrom,
+    sync::{
+        atomic::{self, AtomicU16},
+        Arc,
+    },
+};
 
 use actix::{
     ActorFuture, AsyncContext, Context, ContextFutureSpawner, Handler, SystemService, WrapFuture,
@@ -261,7 +268,7 @@ impl ChainManager {
 
         // `current_retrieval_count` keeps track of how many sources are being retrieved in this
         // epoch by using a reference-counted atomic counter that can be read and updated safely.
-        let current_retrieval_count = std::sync::Arc::new(std::sync::atomic::AtomicU16::new(0u16));
+        let current_retrieval_count = Arc::new(AtomicU16::new(0u16));
         let maximum_retrieval_count = self.data_request_max_retrievals_per_epoch;
 
         for (dr_pointer, data_request_output) in dr_pointers.into_iter().filter_map(|dr_pointer| {
@@ -283,8 +290,8 @@ impl ChainManager {
                 calculate_reppoe_threshold(rep_eng, &own_pkh, num_witnesses + num_backup_witnesses);
 
             // Grab a reference to `current_retrieval_count`
-            let cloned_retrieval_count = std::sync::Arc::clone(&current_retrieval_count);
-            let additional_retrieval_count =
+            let cloned_retrieval_count = Arc::clone(&current_retrieval_count);
+            let added_retrieval_count =
                 u16::try_from(data_request_output.data_request.retrieve.len())
                     .unwrap_or(core::u16::MAX);
 
@@ -328,9 +335,10 @@ impl ChainManager {
                 // Refrain from trying to resolve any more requests if we have already hit the limit
                 // of retrievals per epoch.
                 .and_then(move |vrf_proof| {
-                    let local_retrieval_count = cloned_retrieval_count.load(Relaxed);
+                    let mut start_retrieval_count = cloned_retrieval_count.load(atomic::Ordering::Relaxed);
+                    let mut final_retrieval_count = start_retrieval_count.saturating_add(added_retrieval_count);
 
-                    if local_retrieval_count + additional_retrieval_count > maximum_retrieval_count {
+                    if final_retrieval_count > maximum_retrieval_count {
                         log::info!("{} Refrained from resolving data request {} for epoch #{} because it contains {} \
                         sources, which added to the sources that have already been retrieved ({}) would total {} \
                         retrievals, which exceed current limit per epoch ({}). This limit exists for performance and \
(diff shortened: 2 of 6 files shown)


aesedepece:
`data_request_max_retrievals_per_epoch` has initially been set to `30` just to stay on the safe side, but further assessment on what's the sweet point for this value could be necessary.

node/src/actors/chain_manager/mining.rs
```diff
@@ -313,6 +324,36 @@ impl ChainManager {
...
+                            Yellow.bold().paint(local_retrieval_count.to_string()),
+                            Yellow.bold().paint((local_retrieval_count + additional_retrieval_count).to_string()),
+                            Yellow.bold().paint(maximum_retrieval_count.to_string())
+                        );
+
+                        Err(())
+                    } else {
+                        // Update `current_retrieval_count` thanks to interior mutability of the
+                        // `cloned_retrieval_count` reference.
+                        cloned_retrieval_count.fetch_add(additional_retrieval_count, Relaxed);
```
tmpolaczyk: There is a race condition here, a different thread could update the value between the load and the store. The simplest way to fix it is to use [compare_and_swap](doc.rust-lang.org/std/sync/atomic/struct.AtomicU16.html#method.compare_and_swap) and retry on error

node/src/actors/chain_manager/mining.rs
```diff
@@ -313,6 +324,36 @@ impl ChainManager {
                     }
                 })
                 .flatten()
+                // Refrain from trying to resolve any more requests if we have already hit the limit
+                // of retrievals per epoch.
+                .and_then(move |vrf_proof| {
+                    let local_retrieval_count = cloned_retrieval_count.load(Relaxed);
+
+                    if local_retrieval_count + additional_retrieval_count > maximum_retrieval_count {
```
tmpolaczyk: This operation can overflow a `u16`. To fix, cast everything to `u32`, or change the order of the operations:
```suggestion
                    if local_retrieval_count > maximum_retrieval_count - additional_retrieval_count {
```

node/src/actors/chain_manager/mining.rs
aesedepece: That's an interesting point. I willingly ignored the potential race condition under the assumption that everything works in the same thread, that `current_retrieval_count` is local to the `try_mine_data_request` function and that `try_mine_data_request` is only called once per epoch.

I'll try to incorporate your suggestion just in case my assumptions are proven wrong in the future :+1:

aesedepece: Wait a minute, the whole point of using `AtomicU16` is not having to deal with data races:

```rust
pub fn fetch_add(&self, val: $int_type, order: Ordering) -> $int_type {
    // SAFETY: data races are prevented by atomic intrinsics.
    unsafe { atomic_add(self.v.get(), val, order) }
}
```

tmpolaczyk: Oh sorry, I was referring to a race condition between this load: `cloned_retrieval_count.load(Relaxed)` and this store: `cloned_retrieval_count.fetch_add`

aesedepece: Updated. Feedback please!

node/src/actors/chain_manager/mining.rs
```diff
@@ -314,6 +325,47 @@ impl ChainManager {
...
+                        // to guarantee addition and prevent potential race conditions in a concurrent
+                        // scenario.
+                        loop {
+                            let internal_retrieval_count = cloned_retrieval_count.compare_and_swap(local_retrieval_count, local_retrieval_count + additional_retrieval_count, Relaxed);
+                            if internal_retrieval_count == local_retrieval_count {
+                                // The counter update was updated successfully, we can move on.
+                                break;
+                            } else {
+                                // The counter was updated somewhere else, addition must be retried.
+                                local_retrieval_count = internal_retrieval_count;
```
tmpolaczyk: Here we have to check ` if local_retrieval_count + additional_retrieval_count > maximum_retrieval_count` again

node/src/actors/chain_manager/mining.rs
tmpolaczyk: This is still not fixed? You can also used `checked_add` and return `true` on overflow

node/src/actors/chain_manager/mining.rs
```diff
@@ -259,6 +259,11 @@ impl ChainManager {
             num_active_identities,
         );
 
+        // `current_retrieval_count` keeps track of how many sources are being retrieved in this
+        // epoch by using a reference-counted atomic counter that can be read and updated safely.
+        let current_retrieval_count = std::sync::Arc::new(std::sync::atomic::AtomicU16::new(0u16));
```
tmpolaczyk: Could you import `Arc` and `AtomicU16`?

```suggestion
        let current_retrieval_count = Arc::new(AtomicU16::new(0u16));
```

node/src/actors/chain_manager/mining.rs
```diff
@@ -1,4 +1,11 @@
-use std::{cmp::Ordering, convert::TryFrom};
+use std::{
+    cmp::Ordering,
+    convert::TryFrom,
+    sync::{
+        atomic::{AtomicU16, Ordering::Relaxed},
```
tmpolaczyk: Better to use `Ordering::Relaxed` in the code

```suggestion
        atomic::{AtomicU16, Ordering},
```

node/src/actors/chain_manager/mining.rs
```diff
@@ -314,6 +332,53 @@ impl ChainManager {
                     }
                 })
                 .flatten()
+                // Refrain from trying to resolve any more requests if we have already hit the limit
+                // of retrievals per epoch.
+                .and_then(move |vrf_proof| {
+                    let mut start_retrieval_count = cloned_retrieval_count.load(Relaxed);
+                    let mut final_retrieval_count = start_retrieval_count.saturating_add(added_retrieval_count);
+
+                    if final_retrieval_count > maximum_retrieval_count {
```
tmpolaczyk: Ok, so if `maximum_retrieval_count` is `u16::MAX`, then this branch is unreachable and there is literally no limit. If that's the expected behavior add a short comment somewhere.

node/src/actors/chain_manager/mining.rs
```diff
@@ -314,6 +332,53 @@ impl ChainManager {
...
+                        log::info!("{} Refrained from resolving data request {} for epoch #{} because it contains {} \
```
tmpolaczyk: This line is very helpful but it may be too long for the logs. I was thinking about printing the full help only once and a short line afterwards, but that's just adding complexity for little benefit. Not sure what to do about that.

config/src/defaults.rs
```diff
@@ -127,6 +127,12 @@ pub trait Defaults {
         ))
     }
 
+    /// Set the limit of retrievals per epoch to 65_535.
+    /// This in practice equals no limit enforcement.
+    fn mining_data_request_max_retrievals_per_epoch(&self) -> u16 {
+        core::u16::MAX
+    }
```
aesedepece: @tmpolaczyk there it is :rofl:

PR #1101, opened 2020-03-19
Make retrieval timeout be expressed in milliseconds

aesedepece:
Using milliseconds gives finer granularity and adapts better to reality, as most APIs should normally return a result or an error under 1 second.

The default is changed to 1/30th of the epoch length, in line with #1099.

The `witnet.toml` config file is also changed so it uses `1000` milliseconds.

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -219,6 +219,16 @@ pub struct Connections {
     pub bucketing_update_period: i64,
 }
 
+fn from_millis<'de, D>(deserializer: D) -> Result<Option<Duration>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    Ok(match u64::deserialize(deserializer) {
+        Ok(secs) => Some(Duration::from_millis(secs)),
+        Err(_) => None,
+    })
+}
+
 fn from_secs<'de, D>(deserializer: D) -> Result<Option<Duration>, D::Error>
 where
     D: Deserializer<'de>,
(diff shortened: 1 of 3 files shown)


config/src/config.rs
```diff
@@ -299,8 +309,8 @@ pub struct Mining {
     /// Set to 0 to disable timeouts.
     #[partial_struct(serde(
         default,
-        deserialize_with = "from_secs",
-        rename = "data_request_timeout_seconds"
+        deserialize_with = "from_millis",
+        rename = "data_request_timeout_millis"
```
tmpolaczyk: `millis` sounds strange to me
```suggestion
        rename = "data_request_timeout_ms"
```

aesedepece: It doesn't feel natural to me either, but it is the convention that `Duration` is using.

Anyway, I support the idea that the convention of a library can affect our code for the sake of ease of understanding — but it should never affect the configuration files, where we need to use whatever gives better UX.

My only objection to `data_request_timeout_ms` is that using an acronym here (`ms`) sort of feels out of place among the rest of settings, which use full magnitude names like `seconds`.

witnet.toml
```diff
@@ -26,7 +26,7 @@ update_period_seconds = 8000000
 
 [mining]
 enabled = true
-data_request_timeout_seconds = 5
+data_request_timeout_milliseconds = 1000
```
tmpolaczyk: 1 second timeout may be too aggressive, thoughts @lrubiorod?

lrubiorod: I was using 3 seconds in my branch :P

aesedepece: What about `1500`? I'm ready to haggle every single millisecond!

aesedepece: Some input:
- At `1000`, the BTC price example resolves to `Timeout` in around ~3% of the attempts.
- At `2000`, the sample example resolves to `Timeout` in a solid 0% of the attempts.

PR #1102, opened 2020-03-19
Include commit fees in tally change

lrubiorod:


lrubiorod:
It is solved in #1105

Issue #695, opened 2019-06-13
Refactor tally validation to use `create_tally`


tmpolaczyk:
The `validate_tally_transaction` function implemented in `validations/src/validations.rs` is getting a bit complex. We could remove most of its logic if instead of checking for the multiple conditions for a tally to be valid, we just create a new tally and compare it with the one in the block. The `create_tally` function in `data_structures/src/data_request.rs` is pretty simple, so it shouldn't need any refactoring. The only think that could change is the value transfer outputs order, so we would have to insert them into a hashmap before comparing. If we compare the fields separately we can even maintain the custom errors.

PR #1105, opened 2020-03-19
fix(validations): include commit fees in the change VT in tally trans…

lrubiorod:
…action

### DIFF
--- a/validations/src/validations.rs
+++ b/validations/src/validations.rs
@@ -778,6 +778,7 @@ pub fn validate_tally_transaction<'a>(
     let tally_stage = &dr_output.data_request.tally;
     let non_error_min = f64::from(dr_output.min_consensus_percentage) / 100.0;
     let commit_pkhs: Vec<&PublicKeyHash> = dr_state.info.commits.keys().collect();
+    let commit_length = commit_pkhs.len();
 
     let dishonest_pkhs = validate_consensus(
         reveal_txns,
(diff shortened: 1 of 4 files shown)


validations/src/validations.rs
```diff
@@ -813,32 +820,45 @@ pub fn validate_tally_outputs<S: ::std::hash::BuildHasher>(
...
-    let reveal_reward = dr_state.data_request.witness_reward;
+    let witness_reward = dr_state.data_request.witness_reward;
+    let reveal_fee = dr_state.data_request.reveal_fee;
+    let commit_fee = dr_state.data_request.commit_fee;
     for (i, output) in ta_tx.outputs.iter().enumerate() {
         if change_required && i == ta_tx.outputs.len() - 1 && output.pkh == dr_state.pkh {
-            let expected_tally_change = reveal_reward * dishonest_len as u64
-                + dr_state.data_request.reveal_fee * (witnesses - n_reveals) as u64;
+            let expected_tally_change = if n_reveals == 0 {
+                witnesses as u64 * (witness_reward + reveal_fee + commit_fee)
```
tmpolaczyk: Is this correct when `n_reveals == 0` but `n_commits > 0`?

lrubiorod: O_O my mistake, I modified in other part but forgot here!! Thanks!!

tmpolaczyk:
When #695?

lrubiorod:
> When #695?

The problem with that issue is that we lose information about what happens inside, we only know that the tally created is different from the tally validated, but not why

PR #1120, opened 2020-03-23
Store coin age in UTXO set 

### LINKED ISSUE #1110: [Collaterals]: Unspent Transaction Outputs Set (UTXO Set)

tmpolaczyk:
Close #1110

The coin age can be calculated using the `included_in_block_number` method of `UnspentOutputsPool`. As the name implies, the coin age is measured in the number of blocks, not the number of epochs.

This changes the chain state serialization, so old nodes must delete the storage.

### DIFF
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -1942,7 +1942,12 @@ pub enum DataRequestStage {
 /// Unspent Outputs Pool
 #[derive(Clone, Debug, Default, Eq, PartialEq, Serialize, Deserialize)]
 pub struct UnspentOutputsPool {
+    /// Map of output pointer to value transfer output
     map: HashMap<OutputPointer, ValueTransferOutput>,
+    /// Map of transaction hash to a tuple of:
+    /// * The epoch of the block that included the transaction
+    /// * A reference count, used to keep this map clear after removing transactions
+    transaction_epoch: HashMap<Hash, (Epoch, u32)>,
 }
 
 impl UnspentOutputsPool {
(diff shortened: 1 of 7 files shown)


data_structures/src/chain.rs
```diff
@@ -1939,7 +1939,98 @@ pub enum DataRequestStage {
...
+
+    pub fn iter<'a>(
+        &'a self,
+    ) -> impl Iterator<Item = (&'a OutputPointer, &'a ValueTransferOutput)> + 'a {
+        self.map.iter()
+    }
+
+    /// Returns the epoch of the block that included the transaction referenced
+    /// by this OutputPointer. The difference between that epoch and the
+    /// current epoch is the "coin age".
```
aesedepece: That description doesn't exactly match the specification:
> Entries of the UTXO set need to be tagged with a reference to the block in which they were created so as to track how "old" are they in terms of **how many blocks have been consolidated into the block chain since their creation**.

We are in time to discuss the merits of one option vs. the other.

aesedepece: Any @witnet/devs want to chime in on this?

tmpolaczyk: I implemented both options so we don't lose much time after deciding the correct apporach.

The main arguments for each side are:

* Block number is consistent with the implementation of `extra_commit_rounds` and `extra_reveal_rounds`
* Epoch number can be used to calculate exactly when the UTXO will be available

aesedepece: Just for the record, before resolving the conversation — the final implementation **DOES match** the specification as per https://github.com/witnet/WIPs/pull/18.

data_structures/src/chain.rs
```diff
@@ -1939,7 +1939,102 @@ pub enum DataRequestStage {
...
+
+    pub fn iter(&self) -> std::collections::hash_map::Iter<OutputPointer, ValueTransferOutput> {
+        self.map.iter()
+    }
+
+    /// Returns the number of the block that included the transaction referenced
+    /// by this OutputPointer. The difference between that number and the
+    /// current number of consolidated blocks is the "coin age".
+    pub fn included_in_block_number(&self, k: &OutputPointer) -> Option<Epoch> {
+        if !self.map.contains_key(k) {
```
aesedepece: Please honor [rule #8](https://www.robertsrulesofcoders.com/)

tmpolaczyk: The only rules I follow are the clippy lints

aesedepece: Lol. Did it complain about this?

tmpolaczyk: No, and I think that lint does not exist. Should be easy to implement, maybe you can open them an issue :laughing:

PR #1124, opened 2020-03-24
Add collateral to CommitTransaction

### LINKED ISSUE #1112: [Collaterals]: Commitment Transactions

tmpolaczyk:
Close #1112

### DIFF
--- a/data_structures/src/error.rs
+++ b/data_structures/src/error.rs
@@ -212,19 +212,19 @@ pub enum TransactionError {
         input_value, output_value
     )]
     NegativeCollateral { input_value: u64, output_value: u64 },
-    /// Not enough collateral in commit transaction
+    /// Incorrect collateral in commit transaction
     #[fail(
-        display = "Not enough collateral. Expected: {}, found: {}",
+        display = "Incorrect collateral. Expected: {}, found: {}",
         expected, found
     )]
-    NotEnoughCollateral { expected: u64, found: u64 },
+    IncorrectCollateral { expected: u64, found: u64 },
     /// Collateral in commit transaction is not mature enough
     #[fail(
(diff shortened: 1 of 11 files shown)


data_structures/src/error.rs
```diff
@@ -200,12 +200,44 @@ pub enum TransactionError {
...
         display = "The specified collateral ({} nwits), is less than the minimum required ({} nwits)",
         value, min
     )]
     InvalidCollateral { value: u64, min: u64 },
+    /// Negative collateral in commit transaction
+    #[fail(
+        display = "Tried to create value in commit collateral. Input value: {}, output value: {}",
+        input_value, output_value
+    )]
+    NegativeCollateral { input_value: u64, output_value: u64 },
```
lrubiorod: Why dont use the IncorrectCollateral error for this case?

validations/src/tests.rs
```diff
@@ -5071,6 +5525,38 @@ fn block_add_2_vtt_same_input() {
...
+        let vt_body = VTTransactionBody::new(vec![Input::new(output1_pointer); 2], vec![vto0]);
+        let vts = sign_t(&vt_body);
+        let vt_tx1 = VTTransaction::new(vt_body, vec![vts; 2]);
+
+        (
+            BlockTransactions {
+                value_transfer_txns: vec![vt_tx1],
+                ..BlockTransactions::default()
+            },
+            // Why + 1?
```
lrubiorod: O_O

data_structures/src/error.rs
tmpolaczyk: Because the collateral value is negative, and the `IncorrectCollateral` error expects positive values.

PR #1125, opened 2020-03-25
Add collateral field in DataRequestOutput

### LINKED ISSUE #1111: [Collaterals]: Data Request Transactions

lrubiorod:
Close #1111


data_structures/src/chain.rs
```diff
@@ -904,6 +910,8 @@ pub struct DataRequestOutput {
     // >50 because simple majority
     // <100 because a 100% consensus encourages to commit a RadError for free
     pub min_consensus_percentage: u32,
+    // This field must be > collateral_minimum
```
tmpolaczyk: Do we allow collaterals with exactly the minimum value?

```suggestion
    // This field must be >= collateral_minimum
```

node/src/actors/chain_manager/actor.rs
```diff
@@ -61,7 +61,13 @@ impl ChainManager {
             })
             .and_then(|config, act, _ctx| {
                 let consensus_constants = config.consensus_constants.clone();
+
+                // Set ChainManager parameters related with the consensus constants
```
tmpolaczyk: Actually, consensus constants can be accessed as


    self
        .chain_state
        .chain_info
        .as_ref()
        .unwrap()
        .consensus_constants

So we don't need to store any of these

node/src/actors/chain_manager/handlers.rs
```diff
@@ -284,11 +285,21 @@ impl Handler<AddBlocks> for ChainManager {
...
             self.sm_state
         );
+
+        let consensus_constants = self
+            .chain_state
+            .chain_info
+            .as_ref()
+            .unwrap()
+            .consensus_constants
+            .clone();
```
tmpolaczyk: You probably can avoid this clone by using a reference:

    let consensus_constants = &self

node/src/actors/chain_manager/handlers.rs
```diff
@@ -542,8 +554,18 @@ impl Handler<PeersBeacons> for ChainManager {
 
                     let our_beacon = self.get_chain_beacon();
 
+                    let consensus_constants = self
+                        .chain_state
+                        .chain_info
+                        .as_ref()
+                        .unwrap()
+                        .consensus_constants
+                        .clone();
```
tmpolaczyk: Same, try to avoid clone

node/src/actors/chain_manager/mining.rs
```diff
@@ -88,21 +88,16 @@ impl ChainManager {
...
         let current_epoch = self.current_epoch.unwrap();
-        let mining_bf = self
+        let consensus_constants = self
             .chain_state
             .chain_info
             .as_ref()
             .unwrap()
             .consensus_constants
-            .mining_backup_factor;
+            .clone();
```
tmpolaczyk: Remove clone

PR #1126, opened 2020-03-25
feat(node): use all beacons in consensus, not only expected ones

tmpolaczyk:
Additionally, now a `LastBeacon` message is sent to all inbound peers as soon as they consolidate their connection, which greatly reduces the synchronization delay.

### DIFF
--- a/node/src/actors/session/actor.rs
+++ b/node/src/actors/session/actor.rs
@@ -62,24 +62,20 @@ impl Actor for Session {
                     session_type: self.session_type,
                 })
                 .into_actor(self)
-                .then(|res, act, ctx| {
-                    match res {
-                        Ok(Ok(_)) => {
-                            debug!(
-                                "Successfully registered session {:?} into SessionManager",
-                                act.remote_addr
-                            );
+                .then(|res, act, ctx| match res {
+                    Ok(Ok(_)) => {
+                        debug!(
+                            "Successfully registered session {:?} into SessionManager",
+                            act.remote_addr
+                        );
 
-                            actix::fut::ok(())
-                        }
-                        _ => {
-                            error!("Session register into Session Manager failed");
-                            // FIXME(#72): a full stop of the session is not correct (unregister should
-                            // be skipped)
-                            ctx.stop();
+                        actix::fut::ok(())
+                    }
+                    _ => {
+                        error!("Session register into Session Manager failed");
+                        ctx.stop();
 
-                            actix::fut::err(())
-                        }
+                        actix::fut::err(())
                     }
                 })
                 .and_then(|_, act, _ctx| {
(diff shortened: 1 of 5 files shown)


lrubiorod:
Add new tests

aesedepece:
At first thought, I'm concerned about possible security implications of this, we would need to verify there are none.

aesedepece:
There's a couple of additional changes that may be worth considering:
- I think we will found ourselves at the end of the epoch realizing that we are missing the beacons from every peer that we consolidated after the "hey it's time to exchange beacons" moment. Shall we send our beacon to outbound peers as soon as we consolidate them? (notwithstanding that it will also be sent once the next checkpoint is crossed)
- Another one I don't remember right now, lol.

aesedepece:
>     * I think we will found ourselves at the end of the epoch realizing that we are missing the beacons from every peer that we consolidated after the "hey it's time to exchange beacons" moment. Shall we send our beacon to outbound peers as soon as we consolidate them? (notwithstanding that it will also be sent once the next checkpoint is crossed)

Ok, this doesn't actually happen because `PeersBeacons` gets delayed until next beacons exchange round:

> Not sending PeersBeacons message because of lack of peers (still bootstrapping)
> Not sending PeersBeacons message because of lack of peers (still bootstrapping)
> Received LAST_BEACON: #20085: 4329c56f6e8a579c2e4869d7dafc1dd8c8a1d162be431385724f000ffe3c08db
> Sending PeersBeacons message

Of course the "still bootstrapping" message now means something different than before — it doesn't imply that `ChainManager` is now in `WaitingConsensus` state. It simply says that at some point during the last epoch, after beacons were exchanged, we swapped some of our peers.

This makes my suggestion not critical, but still I think it we can implement it so as to increase the chances that we arrive to successful consensus at the first try without having to delay sending the `PeersBeacons` message.

tmpolaczyk:
"still bootstrapping" is supposed to indicate that the number of outbound peers is not enough, and we are waiting for some more. If we reach the outbound_limit, the error message should be "not enough beacons"

aesedepece:
Have been testing it for a while and I can confirm this works as intended.

@tmpolaczyk what do you think about sending our beacon to outbound peers as soon as we consolidate them?

tmpolaczyk:
> @tmpolaczyk what do you think about sending our beacon to outbound peers as soon as we consolidate them?

Do you mean to our inbound peers? The beacons used for consensus are the ones from our outbound peers, and they send their beacon to their inbound peers.

aesedepece:
> Do you mean to our inbound peers? The beacons used for consensus are the ones from our outbound peers, and they send their beacon to their inbound peers.

Yes, I meant inbound, sorry.

aesedepece:
Cool, I'll try it now again.

tmpolaczyk:
Sending the beacon as soon as the peer is consolidated greatly reduces the synchronization delay, but we need to test it more to make sure it doesn't break with more peers or more blocks to sync.

aesedepece:
Works nice for me in the typical testing setup: 5 nodes, outbound limit 4, killing 1 doesn't break consensus for its peers as they get a replacement peer inside the same epoch.

Regarding synchronization delay, I must say that I saw some nodes print the SYNCED banner just 1 second after restarting them. Wild.

aesedepece:
If the tests pass, this has my explicit approval.

PR #1127, opened 2020-03-25
Wallet: initial block sync & downtime catch-up

aesedepece:
This implements initial block synchronization for the wallet and unifies block handling and transaction indexing so they are dealt with in the same way regardless of they being discovered through initial sync or real time block notifications.

BREAKING CHANGE: I accidentally broke the temporary wallet funding method, so I took the chance to remove it altogether.


aesedepece:
The latest changes leverage the existing workers architecture to run all the synchronization business logic asynchronously, so it doesn't block the main server thread.

wallet/src/actors/worker/methods.rs
```diff
@@ -283,4 +297,175 @@ impl Worker {
...
+
+                // Process each block
+                futures03::executor::block_on(self.handle_block(block, wallet_id, wallet.clone()))?;
+                latest_epoch = epoch;
+            }
+
+            // Persist the epoch of the last synced block. Doing this by batches instead of by block
+            // should be safe under the assumption that the same transaction cannot be processed
+            // twice by a single wallet.
+            wallet.update_last_sync(latest_epoch)?;
```
aesedepece: This assumption needs to be validated.

wallet/src/actors/worker/methods.rs
```diff
@@ -283,4 +297,175 @@ impl Worker {
...
+            slf.do_send(worker::IndexTxns(
+                wallet_id.to_owned(),
+                wallet.clone(),
+                block_txns.clone(),
+                model::BlockInfo {
+                    epoch: block_epoch,
+                    hash: block_hash.clone(),
+                },
+            ));
+        }
```
aesedepece: This should probably block. Otherwise we risk having a false positive in processing a block, and therefore we miss its contents.

aesedepece:
One significant change to the wallet covertly added in this pull request is that it now panics at the beginning if a Witnet node is not properly configured. It just made no sense that it continued without it only to give errors when it found there was no node in place. Basically because you can't configure the connection to the node in runtime. If it's not available upon starting up... why even bother?

wallet/src/actors/worker/methods.rs
tmpolaczyk: I think the `handle_block` method should save the hash of the last handled block, and then when handling a new block do a quick check that `block_beacon.hash_prev_block` is equal to that hash. This way we avoid any problems with missing blocks, duplicated blocks, forked blocks, and blocks being processed out of order.

wallet/src/actors/worker/methods.rs
```diff
@@ -283,4 +295,175 @@ impl Worker {
...
+            .into_iter()
+            .map(types::Transaction::from);
+        let vtt_txns = block
+            .txns
+            .value_transfer_txns
+            .into_iter()
+            .map(types::Transaction::from);
+        let block_txns = dr_txns.chain(vtt_txns).collect::<Vec<types::Transaction>>();
+
+        for slf in &self.own_address {
```
tmpolaczyk: Iterating over an option

```suggestion
        if let Some(slf) = self.own_address.as_ref() {
```

aesedepece: image

tmpolaczyk: Anyway, why is this an if? If `self.own_address` is `None`, this method will do nothing, so can't we just panic?

wallet/src/actors/worker/methods.rs
aesedepece: Great suggestion. It works like that now :ok_hand:

wallet/src/actors/worker/methods.rs
aesedepece: It blocks now.

wallet/src/actors/worker/methods.rs
aesedepece: Agree.

Anyhow, the two last commit remove the need for worker to know their own address altogether.

wallet/src/actors/app/methods.rs
```diff
@@ -460,59 +463,35 @@ impl App {
...
-                },
-            ));
-        }
 
-        log::trace!("notifying balances to sessions");
-        for (wallet, sink) in self.state.notifiable_wallets() {
-            self.params
-                .worker
-                .do_send(worker::NotifyBalance(wallet, sink));
+        for (_id, wallet) in self
```
tmpolaczyk: So we need to handle the block once for every wallet. What should happen if we create a new wallet after the initial block sync? Lose all the "old" UTXOs or sync again?

aesedepece: Every new wallet is initiated in such a way that its `last_sync` is the genesis block. As a consequence, regardless of when the wallet is created or what blocks we have processed in the past, it syncs from the beginning and the final state is just right :)

aesedepece: Needless to say that the fact that one wallet syncs at some point doesn't affect what was already processed for other wallets.

tmpolaczyk: Oh, nice! Do we want to store all the blocks somewhere to avoid asking the node more than once, or do we prefer to not store another copy of the blockchain in the wallet?

aesedepece: Yes, let's avoid having the chain. At the end of the day that's the whole point of it being a separate component than the node — it should save users the disk space burden of running a full node.

On a side note, asking for the whole chain may sound overkill, but take into account that creating new wallets is done sparingly. This argument will specially get stronger if at some point Sheikah enables accounting, i.e. having more than one wallet for different purposes (daily expenses, hodling, etc.) derived from a single master key. In theory it's already supported by the wallet!

aesedepece:
@Tommytrg do you want to try this so as to verify it fulfills everyone's expectations? :grimacing:

wallet/src/actors/worker/methods.rs
```diff
@@ -223,7 +241,12 @@ impl Worker {
         block: &model::BlockInfo,
         txns: &[types::Transaction],
     ) -> Result<()> {
+        log::debug!("trying to index txns from epoch {}", block.epoch);
         wallet.index_transactions(block, txns)?;
+        wallet.update_last_sync(CheckpointBeacon {
+            checkpoint: block.epoch,
+            hash_prev_block: Hash::from(block.hash.clone()),
```
tmpolaczyk: Maybe instead of implementing conversions between `Vec<u8>` and `Hash`, it would be better to just change the type of `block_info.hash_prev_block` to `Hash`.

```suggestion
            hash_prev_block: block.hash.clone(),
```

wallet/src/actors/worker/methods.rs
```diff
@@ -283,4 +306,195 @@ impl Worker {
...
+                self.handle_block(block.clone(), wallet.clone())?;
+                latest_beacon = CheckpointBeacon {
+                    checkpoint: epoch,
+                    hash_prev_block: Hash::from_str(&id)?,
+                };
+            }
+
+            // Keep asking for new batches of blocks until we get less than expected, which signals
+            // that there are no more blocks to process.
+            // The `.unwrap()` here is provably safe given that no value in an `usize` can overflow
```
tmpolaczyk: Which unwrap?

wallet/src/actors/worker/methods.rs
aesedepece: Great suggestion. I'll measure the impact and see if the moment is now.

wallet/src/actors/worker/methods.rs
aesedepece: Lol.

wallet/src/model.rs
```diff
@@ -56,6 +56,9 @@ pub struct ExtendedKeyedSignature {
     pub chaincode: String,
 }
 
+#[derive(Debug, Serialize)]
+pub struct Block {}
```
tmpolaczyk: Dead code?

wallet/src/actors/worker/methods.rs
aesedepece: Done. Thanks once again!

PR #1130, opened 2020-03-26
Refactor Tally validations

### LINKED ISSUE #695: Refactor tally validation to use `create_tally`

### LINKED ISSUE #1113: [Collaterals]: Tally Transactions

lrubiorod:
Close #695

### DIFF
--- a/data_structures/src/data_request.rs
+++ b/data_structures/src/data_request.rs
@@ -338,6 +338,25 @@ impl DataRequestPool {
     }
 }
 
+/// Return the change that should be returned to the creator of the data request if
+/// some witness fails to commit, fails to reveal, or reports a value out of consensus.
+/// If the change is 0, the change `ValueTransferOutput` should not be created
+pub fn calculate_tally_change(
+    commits_count: usize,
+    reveals_count: usize,
+    honests_count: usize,
+    dr_output: &DataRequestOutput,
+) -> u64 {
+    let commits_count = commits_count as u64;
+    let reveals_count = reveals_count as u64;
+    let honests_count = honests_count as u64;
+    let witnesses = u64::from(dr_output.witnesses);
+
+    dr_output.witness_reward * (witnesses - honests_count)
+        + dr_output.reveal_fee * (witnesses - reveals_count)
+        + dr_output.commit_fee * (witnesses - commits_count)
+}
+
 pub fn create_tally<RT, S: ::std::hash::BuildHasher>(
     dr_pointer: Hash,
     dr_output: &DataRequestOutput,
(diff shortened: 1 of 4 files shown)


validations/src/tests.rs
```diff
@@ -78,23 +54,21 @@ fn sign_t2<H: Hashable>(tx: &H) -> KeyedSignature {
...
+fn test_sign_tx() {
+    let tx = &RevealTransaction::default();
 
-    let secp = &Secp256k1::new();
-    let secret_key =
-        Secp256k1_SecretKey::from_slice(&[0x69; 32]).expect("32 bytes, within curve order");
-    let public_key = Secp256k1_PublicKey::from_secret_key(secp, &secret_key);
-    let public_key = PublicKey::from(public_key);
-    assert_eq!(public_key.pkh(), MY_PKH_3.parse().unwrap());
+    let ks = sign_tx([0xcd; 32], tx);
```
tmpolaczyk: I would prefer to use a constant instead of hardcoding the private key:

```suggestion
    let ks = sign_tx(PRIV_KEY_1, tx);
```

And maybe we can rename `MY_PKH` to `MY_PHK_1` to be more consistent

tmpolaczyk:
Missing tests for errors `TransactionError::DishonestReward`, `TransactionError::InvalidReward`, `TransactionError::InvalidTallyValue`

data_structures/src/data_request.rs
```diff
@@ -338,6 +338,22 @@ impl DataRequestPool {
     }
 }
 
+pub fn calculate_tally_change(
```
tmpolaczyk: Documentation

```suggestion
/// Return the change that should be returned to the creator of the data request if some witness fails to commit, fails to reveal, or reports a value out of consensus.
/// If the change is 0, the change `ValueTransferOutput` should not be created
pub fn calculate_tally_change(
```

validations/src/tests.rs
```diff
@@ -77,24 +53,24 @@ fn sign_t2<H: Hashable>(tx: &H) -> KeyedSignature {
         public_key,
     }
 }
+static PRIV_KEY_1: [u8; 32] = [0xcd; 32];
+static PRIV_KEY_2: [u8; 32] = [0x43; 32];
+static PRIV_KEY_3: [u8; 32] = [0x69; 32];
+static MY_PKH_1: &str = "wit18cfejmk3305y9kw5xqa59rwnpjzahr57us48vm";
+static MY_PKH_2: &str = "wit1z8mxkml4a50dyysqczsp7gj5pnvz3jsldras8t";
+static MY_PKH_3: &str = "wit164gu2l8p7suvc7zq5xvc27h63td75g6uspwpn5";
+#[test]
```
tmpolaczyk: New line here

```suggestion

#[test]
```

validations/src/validations.rs
```diff
@@ -918,117 +886,152 @@ pub fn validate_tally_transaction<'a>(
...
                 return Err(TransactionError::MultipleRewards { pkh: output.pkh }.into());
             }
             pkh_rewarded.insert(output.pkh);
         }
+        total_tally_value += output.value;
     }
 
-    Ok(())
+    let expected_dr_value = dr_state.data_request.checked_total_value()?;
+    let found_dr_value = dr_state.info.commits.len() as u64 * dr_state.data_request.commit_fee
```
tmpolaczyk: This could be simplified if we just sum the total value of the outputs, that value must be equal to the `expected_dr_value`.

PR #1136, opened 2020-03-31
fix(peering): add public address field to fix bucketing problems

### LINKED ISSUE #1055: New bucket bug
I have checked that there are not saving new addresses in new bucket that are different from config file

lrubiorod:
Fix #1055

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -150,6 +150,11 @@ pub struct Connections {
     /// peers should bind to
     pub server_addr: SocketAddr,
 
+    /// Public address
+    #[partial_struct(skip)]
+    #[partial_struct(serde(default))]
+    pub public_addr: Option<SocketAddr>,
+
     /// Maximum number of concurrent connections the server should
     /// accept
     pub inbound_limit: u16,
@@ -472,6 +477,7 @@ impl Connections {
                 .server_addr
                 .to_owned()
                 .unwrap_or_else(|| defaults.connections_server_addr()),
+            public_addr: config.public_addr,
             inbound_limit: config
                 .inbound_limit
                 .to_owned()
@@ -813,8 +819,10 @@ mod tests {
     #[test]
     fn test_connections_from_partial() {
         let addr: SocketAddr = "127.0.0.1:3000".parse().unwrap();
+        let public_addr: SocketAddr = "127.0.0.1:3003".parse().unwrap();
         let partial_config = PartialConnections {
             server_addr: Some(addr),
+            public_addr: Some(public_addr),
             inbound_limit: Some(3),
             outbound_limit: Some(4),
             known_peers: [addr].iter().cloned().collect(),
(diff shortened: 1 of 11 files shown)


node/src/actors/sessions_manager/actor.rs
```diff
@@ -25,7 +25,7 @@ impl Actor for SessionsManager {
 
                 // Set server address, connections limits and handshake timeout
                 act.sessions
-                    .set_server_address(config.connections.server_addr);
+                    .set_server_address(config.connections.public_addr);
```
tmpolaczyk: Maybe we should rename this method and `self.server_address` into `self.public_address`?

```suggestion
                    .set_public_address(config.connections.public_addr);
```

aesedepece:
It works!

node/src/actors/sessions_manager/actor.rs
aesedepece: Agreed.

node/src/actors/session/mod.rs
```diff
@@ -45,8 +45,8 @@ impl HandshakeFlags {
 
 /// Session representing a TCP connection
 pub struct Session {
-    /// Server socket address (local peer)
-    server_addr: SocketAddr,
+    /// Server public socket address (local peer)
```
tmpolaczyk: This comment is not clear enough

```suggestion
    /// Public address of the node (the one used by other peers to connect to ours)
```

config/src/config.rs
```diff
@@ -150,6 +150,9 @@ pub struct Connections {
     /// peers should bind to
     pub server_addr: SocketAddr,
 
+    /// Public address
+    pub public_addr: SocketAddr,
```
tmpolaczyk: Does it make sense to make this an `Option`? For people that don't have a public IP.

```suggestion
    pub public_addr: Option<SocketAddr>,
```

lrubiorod: They only should have to put the same IP that in server_address

aesedepece: > They only should have to put the same IP that in server_address

For it to work, it is true that leaving the default address `0.0.0.0:...` just does the job as you still have to put some IP in the message that is exchanged. 

I am not claiming though that that's ideal in the long term. This is just the option with the lowest impact (it doesn't cause a hard fork as the protobuf schema remains intact).

data_structures/src/builders.rs
```diff
@@ -66,7 +66,7 @@ impl Message {
                 version: PROTOCOL_VERSION,
                 timestamp: get_timestamp(),
                 capabilities: CAPABILITIES,
-                sender_address: to_address(sender_addr),
+                sender_address: sender_addr.map(to_address),
```
tmpolaczyk, line 69: Maybe we can serialize `None` as `SocketAddr::default()` to avoid that ugly `impl ProtobufConvert`

```suggestion
                sender_address: sender_addr.unwrap_or_default(),
```

PR #1140, opened 2020-04-01
feat(docker): enable auto-detection of oneself's public IP

aesedepece:
- Checks whether a non-default IP exists in the config file. If not:
  - Guesses own public IP using a third party.
  - Checks what's the listening port for the node server.
  - Tries to write the public IP + port into the config file.

### DIFF
--- a/docker/witnet-rust/runner.sh
+++ b/docker/witnet-rust/runner.sh
@@ -19,10 +19,48 @@ URL="https://github.com/witnet/witnet-rust/releases/download/$VERSION/witnet-$VE
 
 FILENAME="$VERSION.tar.gz"
 FOLDERNAME="."
+CONFIG_FILE_FROM_CMD=`echo "$@" | sed -E 's/(.*-c\s*)?(.*\.toml)?.*/\2/'`
+CONFIG_FILE=${CONFIG_FILE_FROM_CMD:-witnet.toml}
+DEFAULT_IP="0.0.0.0"
+DEFAULT_PORT="21337"
+DEFAULT_ADDR="$DEFAULT_IP:$DEFAULT_PORT"
+
+echo "Reading configuration from $CONFIG_FILE"
+
+function read_public_addr_from_config {
+    echo "Reading public_addr from config file";
+    PUBLIC_ADDR_FROM_CONFIG=`grep public_addr $CONFIG_FILE | cut -d "\"" -f2`;
+    LISTENING_PORT_FROM_CONFIG=`grep "server_addr" $CONFIG_FILE | head -1 | cut -d "\"" -f2 | cut -d ":" -f2`
+}
+
+function guess_public_addr {
+    echo "Trying to guess public_addr";
+    API_URL="http://bot.whatismyipaddress.com/";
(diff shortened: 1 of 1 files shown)


docker/witnet-rust/runner.sh
```diff
@@ -19,10 +19,48 @@ URL="https://github.com/witnet/witnet-rust/releases/download/$VERSION/witnet-$VE
 
 FILENAME="$VERSION.tar.gz"
 FOLDERNAME="."
+CONFIG_FILE_FROM_CMD=`echo "$@" | sed -E 's/(.*-c\s*)?(.*\.toml)?.*/\2/'`
+CONFIG_FILE=${CONFIG_FILE_FROM_CMD:-witnet.toml}
```
aesedepece, line 23: I have updated my commit so that the path for the config file is not hard-coded: it reads it now from the command line, i.e. it parses `-c /whatever/path/is/the/config.toml` :wink:

aesedepece:
@witnet/devs is there some Bash hacker in the room?

tmpolaczyk:
> @witnet/devs is there some Bash hacker in the room?

Haha, this is too much magic.

I was wondering why are we limiting this feature to docker users, it would also be useful to have a similar script when running the node from source, or from the distributed binary. I guess we can implement this functionality in the node: if `public_addr` is set to `0.0.0.0`, try to guess the public IP and use that instead.

aesedepece:
That's a great point indeed.

I've been assuming that Docker users just want to run the node without caring much about privacy or trust.

As an example, the current Docker image downloads the binary on the spot but doesn't verify its integrity against my public key, so there's no guarantee that what they run is the same as I signed. (BTW at some point we should set up some multisignature scheme or anything not centrallized in a single person).

In that same line, I assumed that querying a third party for IP discovery was a reasonable privacy compromise.

However, I'm also assuming that someone running the node from the native binary (or compiling from source code) will be likely more leaning towards privacy and security, and therefore would prefer the software not to use third party services by default.

I agree with you on how great would be having IP discovery in the node software itself. Specially if it could be an opt-in feature that only runs if the address is not set yet in the config file and is clearly documented so that no user can be surprised by the fact that it's pinging 3rd parties.

aesedepece:
That being said, another approach that I like even more would be making your peers report your public IP back to you upon handshaking so that you can guess it based of what they reported, using our typical consensus-building functions. In fact this should adapt relatively well to public IP address changes in runtime.

bertux:
> > @witnet/devs is there some Bash hacker in the room?
> 
> Haha, this is too much magic.
> 
> I was wondering why are we limiting this feature to docker users, it would also be useful to have a similar script when running the node from source, or from the distributed binary. I guess we can implement this functionality in the node: if `public_addr` is set to `0.0.0.0`, try to guess the public IP and use that instead.

that's why I will try to add it both in the build of versioned Docker containers after every release and also in the systemd integration I've merged into master few minutes ago.

aesedepece:
Merged as d556b0f6cd1338c86f85675df265c3da53839216.

PR #1142, opened 2020-04-02
chore: update eager's peer nodes

igorline:
Adding 3 more nodes I'm running and apparently my ISP changed my IP so had to change it to keep my raspberry node in the list


aesedepece:
Merged as 71ba668791f7e169f013073a9b90f6760dcfbc50. Thanks!

PR #1159, opened 2020-04-09
[wip] Utxo splitting

### LINKED ISSUE #1115: [Collaterals] Smart UTXO splitting / joining
As recommended by https://github.com/witnet/WIPs/pull/18

lrubiorod:


data_structures/src/chain.rs
```diff
@@ -894,6 +894,8 @@ pub struct ValueTransferOutput {
     /// timestamp. That is, they cannot be used as an input in any transaction of a
     /// subsequent block proposed for an epoch whose opening timestamp predates the time lock.
     pub time_lock: u64,
+    /// This value indicates how many copies of the value exists
+    pub multiplier: u32,
```
aesedepece, line 898:
```suggestion
    /// Indicates how many identical outputs this output actually represents.
    /// This feature enables a more compact representation of successive outputs that
    /// have the same value and spending conditions. 
    /// E.g. if `multiplier = 4`, this output will be demultiplexed into 4 identical
    /// outputs that will be inserted into the UTXO set separately, will have their
    /// own output pointers and will be independently spendable.
    pub multiplier: u32,
```

data_structures/src/chain.rs
```diff
@@ -2246,14 +2248,19 @@ fn update_utxo_outputs(
...
-            transaction_id: txn_hash,
-            output_index: u32::try_from(index).unwrap(),
-        };
+    let mut index = 0;
+    for output in outputs.iter() {
+        for i in 0..output.multiplier {
+            // Add the new outputs to the utxo_diff
+            let output_pointer = OutputPointer {
+                transaction_id: txn_hash,
+                output_index: u32::try_from(index + i).unwrap(),
```
aesedepece, line 2257: I may be missing something here, but why is this `try_from` needed under the assumption that `output.multiplier` is `u32` and therefore `i` and `index` have their type derived from it?

aesedepece:
I have a major concern about the lack of limitations to this feature.

## The problem

Someone could try to use a ludicrously high multiplier on a very small value output with views to making every node in the network run out of memory. In practice, the kernel would most likely step in and kill the node process because of memory abuse. Long story short: it could lead to generalized network disruption and the technical impossibility to ever synchronize a node.

As a practical case, under the assumption that an UTXO set entry takes _at least_ 48 bytes (32 for the transaction hash plus 8 for the value plus 8 for the timestamp), an attacker could use the following parameters to devour every node's memory:

- Total transaction value: `0.022369622 Wits`
- Only one output with `value = 1 nanoWit` and `multiplier = 22369622`
- Total UTXO set size increase: 1GiB

Scaling the attack is super easy. A transaction spending `1 wit` easily buys you 44.7GiB, and the seemingly innocent `500 wit` block reward can buy you the whopping amount of 21.8TiB. I wish I had that much RAM to spare! 

## Mitigation

It becomes apparent that this feature—even if highly beneficial in view of collateralized UTXOs—cannot be implemented without some sort of limitations.

We can mitigate this in many ways. Some of them, top of my mind:
1. Limit the value of `multiplier`. We can foresee that there will be no need for splitting a big output in more than a few dozens or hundred smaller outputs.
2. Limit how small a "multiplied" output can be. Basically requiring that the higher the multiplier, the higher the minimum value of the multiplied output needs to be.
3. Leave the market do their thing. Miners would calculate the __weight__ that the transaction would take in the UTXO set if demultiplexed, and de-prioritize them because the fee per weight unit is too low.

My personal opinion on those three options:
1. The simplest—it should just work. However, static limits are difficult to assess. I personally believe that the range of an `u8` could even suffice, but we would need to run the numbers and see what makes sense.
2. I think this theoretically allows the `multiplier` feature to scale to many different use cases, as there's no static limit but rather an enforced ratio with respect to `value`. Yet I have the hunch that again finding that ratio would require some fine-tuning.
3. This actually makes sense in the long term, but in the absence of a fee market it is just too fragile. As a matter of fact, without an actual bottom threshold of fee per weight unit, the network would be exploitable always that blocks are not full. 

By the way, even if the 3rd is a kamikaze option, it made me think that we should consider that the `multiplier` feature should always increase the transaction weight.

tmpolaczyk:
> Limit the value of multiplier. We can foresee that there will be no need for splitting a big output in more than a few dozens or hundred smaller outputs.

Then we can simply not implement this feature, as the worst case scenario will be wasting dozens or hundreds of 48-byte UTXOs.

aesedepece:
> > Limit the value of multiplier. We can foresee that there will be no need for splitting a big output in more than a few dozens or hundred smaller outputs.
> 
> Then we can simply not implement this feature, as the worst case scenario will be wasting dozens or hundreds of 48-byte UTXOs.

It is true that this could be proven not critical, but it is still a nice optimization with regards to:
- more compact data structures when on the wire
- potentially subsidizing multiplexed transactions by giving them a weight discount

data_structures/src/chain.rs
```diff
@@ -2293,6 +2317,27 @@ pub fn generate_unspent_outputs_pool(
...
+            let value = all_utxos.get(o).map(|vt| i128::from(vt.value)).unwrap_or(0);
+
+            if bigger_first {
+                -value
+            } else {
+                value
+            }
+        })
+        .cloned()
+        .collect()
```
aesedepece, line 2338: So functional :smirk:

tmpolaczyk:
I would prefer to have the multiplier field at the protobuf level, and once a transaction is deserialized it magically duplicates the value transfer outputs. This way we avoid adding extra complexity to the node.

aesedepece:
> I would prefer to have the multiplier field at the protobuf level, and once a transaction is deserialized it magically duplicates the value transfer outputs. This way we avoid adding extra complexity to the node.

I'm cool with that option too.

aesedepece:
Needs rebasing.

lrubiorod:
Continues in #1168

PR #1156, opened 2020-04-07
Merge T8 branch into master

tmpolaczyk:

PR #1160, opened 2020-04-09
Vrf as poe input

girazoki:


node/src/actors/chain_manager/handlers.rs
```diff
@@ -274,6 +278,21 @@ impl Handler<GetHighestCheckpointBeacon> for ChainManager {
...
+        if let Some(chain_info) = &self.chain_state.chain_info {
+            Ok(chain_info.highest_vrf_output)
+        } else {
+            log::error!("No ChainInfo loaded in ChainManager");
+
+            Err(ChainInfoError::ChainInfoNotFound.into())
+        }
+    }
+}
+
```
aesedepece, line 295: The handler for `GetHighestVrfInput` is a getter for `highest_vrf_output`? I think I understand the rationale, but still sounds confusing to me :rofl:

node/src/actors/chain_manager/mining.rs
```diff
@@ -126,11 +134,14 @@ impl ChainManager {
         }
         // The highest checkpoint beacon should contain the current epoch
         beacon.checkpoint = current_epoch;
+        vrf_input.checkpoint = current_epoch;
+
+        vrf_input.checkpoint = current_epoch;
```
aesedepece, line 139: Code repetition!

node/src/actors/chain_manager/handlers.rs
girazoki: Good catch, initially I called the variable highest_vrf_input and thats why the handler is named like that haha. Changing ti

data_structures/src/vrf.rs
```diff
@@ -140,6 +154,10 @@ impl BlockEligibilityClaim {
                 Hash::SHA256(sha256)
             })
     }
+    /// Create a VRF proof for a given message. Needed to generate the input for following eligibilities.
```
aesedepece, line 157: Looks to me like this comment could use an empty line on top.
Also I can't understand what "for following eligibilities" means.

node/src/actors/chain_manager/mining.rs
```diff
@@ -111,6 +112,13 @@ impl ChainManager {
             _ => return,
         };
 
+        // Check eligibility
+        // S(H(beacon))
+        let mut vrf_input = match self.handle(GetHighestVrfOutput, ctx) {
+            Ok(b) => b,
+            _ => return,
+        };
```
aesedepece, line 120: IMO `self.handle()` is code smell for all cases where the handler discards the context.

I'd move the logic from the handler into a `function get_highes_vrf_output` in `impl ChainManager` and then invoke that function from the handler and here.

This gives us better testability, as it allows unit testing the logic against the minimal set of data it is set to operate on (`struct ChainManager`), without intrinsics of the actor system getting in the middle (e.g. `ctx`).  

P.S. I know `AddTransaction` is also being used through `self.handle()`. I think we wasn't aware of these concerns at the time we wrote that. Feel free to leave it like it is or make the same change there too.

data_structures/src/chain.rs
```diff
@@ -53,6 +53,9 @@ pub struct ChainInfo {
 
     /// Checkpoint of the last block in the blockchain
     pub highest_block_checkpoint: CheckpointBeacon,
+
+    /// Checkpoint of the last vrf otuput in the blockchain
+    pub highest_vrf_output: CheckpointBeacon,
```
tmpolaczyk, line 58: This field would benefit from having a different type, so we never accidentally use a `CheckpointBeacon` (epoch + block hash) instead of (epoch + vrf hash). We would need to change it in `VrfMessage::block_mining` and `VrfMessage::data_request`, and all the methods that use them. Not sure how to call this new type though.

validations/src/validations.rs
```diff
@@ -1458,15 +1462,15 @@ pub fn validate_new_transaction(
         .map(|_| ()),
         Transaction::Commit(tx) => {
             // We need a checkpoint beacon with the current epoch and the hash of the previous block
-            let dr_beacon = CheckpointBeacon {
+            let _dr_beacon = CheckpointBeacon {
```
tmpolaczyk, line 1465: If this is unused you can just remove it

node/src/actors/chain_manager/mod.rs
```diff
@@ -403,6 +414,31 @@ impl ChainManager {
...
+                let vrf_ctx = match self.vrf_ctx.as_mut() {
+                    Some(x) => x,
+                    None => {
+                        error!("No VRF context available");
+                        return;
+                    }
+                };
+
+                // If synching from genesis, start consolidating as genesis hash.
+                // Else, take hash the proof to obtain the new vrf output.
```
aesedepece, line 427:
```suggestion
                // Decide the input message for the VRF of this block candidate:
                //   If the candidate builds right on top of the genesis block, use candidate's own checkpoint and the genesis block hash.
                //   Else, use use candidate's own checkpoint and the hash of the VRF proof from the block it builds on.
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -403,6 +414,31 @@ impl ChainManager {
...
+                        checkpoint: block_epoch,
+                        hash_prev_block: block_hash,
+                    },
+                    _ => {
+                        let proof_hash = block.block_header.proof.proof_to_hash(vrf_ctx).unwrap();
+                        CheckpointBeacon {
+                            checkpoint: block_epoch,
+                            hash_prev_block: proof_hash,
+                        }
+                    }
```
aesedepece, line 439: I have my reservations about usage of `CheckpointBeacon` for purposes other than acting as such.

I mean, it sounds counter-intuitive to me that `hash_prev_block` contains something that is not literally the hash of the previous block.

EDIT: this is a code only concern under the assumption that the `Hasher` that we are using doesn't take into account the name of the fields but only the types and order in which they were defined.

girazoki: Understood!

data_structures/src/chain.rs
```diff
@@ -53,6 +53,9 @@ pub struct ChainInfo {
 
     /// Checkpoint of the last block in the blockchain
     pub highest_block_checkpoint: CheckpointBeacon,
+
+    /// Checkpoint of the last vrf otuput in the blockchain
```
tmpolaczyk, line 57: This comment is a bit confusing

```suggestion
    /// Checkpoint and VRF hash of the highest block in the blockchain
```

data_structures/src/chain.rs
```diff
@@ -265,6 +268,19 @@ pub struct CheckpointBeacon {
...
+/// Checkpoint VRF structure
+#[derive(
+    Copy, Clone, Debug, Default, Eq, Hash, PartialEq, Serialize, Deserialize, ProtobufConvert,
+)]
+#[protobuf_convert(pb = "witnet::CheckpointVRF")]
+#[serde(rename_all = "camelCase")]
+pub struct CheckpointVRF {
+    /// The serial number for an epoch
+    pub checkpoint: Epoch,
+    /// The 256-bit hash of the previous block header
```
tmpolaczyk, line 280:
```suggestion
    /// The 256-bit hash of the VRF used in the previous block
```

aesedepece:
Needs rebasing after #1156 was merged.

data_structures/src/chain.rs
aesedepece: `block chain` indeed :stuck_out_tongue:

data_structures/src/vrf.rs
```diff
@@ -113,40 +127,45 @@ pub struct BlockEligibilityClaim {
...
-            .verify(vrf, &VrfMessage::block_mining(beacon))
+            .verify(vrf, &VrfMessage::block_mining(vrf_input))
             .map(|v| {
                 let mut sha256 = [0; 32];
                 sha256.copy_from_slice(&v);
 
                 Hash::SHA256(sha256)
             })
     }
+    /// Output the hash of the VRF proof.
```
aesedepece, line 157:
```suggestion

    /// Output the hash of the VRF proof.
```

node/src/actors/chain_manager/mining.rs
aesedepece: What do you think about this?

girazoki: I am on it, trying to understand it first hehe.

lrubiorod: If we are using form ChainManager (mining is inside of ChainManager). Why dont use the field 'chain_info.highest_block_checkpoint' instead of a handler?

aesedepece: Oops it is true that `GetHighestVrfOutput` is not being used from any other actor.

I'd drop the handler altogether but create the getter as proposed before just in case in the future we need to add the handler back or use that logic somewhere else.

girazoki: Should I change the way we get the beacon as well then?

girazoki: I mean from the state of the chain manager actor as  'chain_info.highest_vrf_checkpoint' and  'chain_info.highest_block_checkpoint'

node/src/actors/chain_manager/mod.rs
```diff
@@ -334,9 +337,17 @@ impl ChainManager {
                     .consensus_constants
                     .mining_backup_factor;
                 let mut signatures_to_verify = vec![];
+                let mut vrf_input = self
```
lrubiorod, line 340: due to always that load the vrf_input, you change the checkpoint. Should be only saved the vrf_proof_hash in chain_info?

node/src/actors/chain_manager/mining.rs
tmpolaczyk: > Should I change the way we get the beacon as well then?

Yes please

node/src/actors/chain_manager/mod.rs
girazoki: I dont know I think it does not hurt to have it, specially since it is updated every time a block is consolidated. What do you guys think? In any case we would have to construct a beacon-like structure for the validations..

tmpolaczyk: Whatever is simpler to implement.

lrubiorod: Oks, simpler is to not change things xDD

node/src/actors/chain_manager/mining.rs
```diff
@@ -875,13 +873,14 @@ mod tests {
 
         // Fields required to mine a block
         let block_beacon = CheckpointBeacon::default();
+        let vrf_input = CheckpointVRF::default();
```
lrubiorod, line 876: could you use a CheckpointVRF that has a different hash from CheckpointBeacon, only to ensure that the test is working properly and it is not using the CheckpointBeacon instead of CheckpointVRF

node/src/actors/chain_manager/handlers.rs
```diff
@@ -274,6 +278,21 @@ impl Handler<GetHighestCheckpointBeacon> for ChainManager {
     }
 }
 
+/// Handler for GetHighestVrfOutput message
+impl Handler<GetHighestVrfOutput> for ChainManager {
```
lrubiorod, line 282: Remove this handler in case it is not finally used

node/src/actors/messages.rs
```diff
@@ -55,6 +55,14 @@ impl Message for GetHighestCheckpointBeacon {
     type Result = Result<CheckpointBeacon, failure::Error>;
 }
 
+/// Message to obtain the highest block checkpoint managed by the `ChainManager`
```
lrubiorod, line 58: Also removed the handler's message

validations/src/validations.rs
```diff
@@ -1653,7 +1651,7 @@ impl fmt::Display for Wit {
 const INITIAL_BLOCK_REWARD: u64 = 500 * NANOWITS_PER_WIT;
 const HALVING_PERIOD: Epoch = 1_750_000;
 
-/// Calculate the block mining reward.
+/// Calculate the block mining r    eward.
```
lrubiorod, line 1654: typo?

validations/src/tests.rs
```diff
@@ -4419,12 +4440,10 @@ fn block_duplicated_commits() {
     };
     let current_epoch = 1000;
     let last_block_hash = LAST_BLOCK_HASH.parse().unwrap();
-    let block_beacon = CheckpointBeacon {
+    let vrf_input = CheckpointVRF {
         checkpoint: current_epoch,
-        hash_prev_block: last_block_hash,
+        hash_prev_vrf: last_block_hash,
```
lrubiorod, line 4445: should be it LAST_VRF_INPUT?

girazoki: indeed.

schemas/witnet/witnet.proto
```diff
@@ -256,6 +256,11 @@ message CheckpointBeacon {
     Hash hash_prev_block = 2;
 }
 
+message CheckpointVRF {
+    fixed32 checkpoint = 1;
```
aesedepece, line 260: Why not `uint32`?

schemas/witnet/witnet.proto
```diff
@@ -286,7 +291,7 @@ message VrfProof {
 }
 
 message DataRequestVrfMessage {
-    CheckpointBeacon beacon = 1;
+    CheckpointVRF vrf_input = 1;
     Hash dr_hash = 2;
 }
```
aesedepece, line 296: Maybe I'm missing something, but why are we putting `DataRequestVrfMessage` and `CheckpointVRF` into the schema if these structures are not being ever sent over the wire?

girazoki: Data Request VRF message was in the schema before this PR so I am assuming is somehow being sent right?

schemas/witnet/witnet.proto
girazoki: I just followed the same structure as `CheckpointBeacon`. Is there a reason to have a fixex32 in the beacon and not in the vrf?

schemas/witnet/witnet.proto
tmpolaczyk: We use the schema to define how to serialize this structure, because it needs to be serialized to bytes in order to create a VRF proof.

schemas/witnet/witnet.proto
tmpolaczyk: `unit32` uses variable length encoding, so smaller values take fewer bytes, but values greater than or equal to `2^28` use 5 bytes. `fixed32` always uses 4 bytes, and is slightly cheaper to serialize and deserialize.

schemas/witnet/witnet.proto
aesedepece: Got it, sorry for my blatant ignorance :sweat_smile:

schemas/witnet/witnet.proto
aesedepece: Everyone here knows I'm a big fan of variable length encoding, but I understand @tmpolaczyk's rationale and in this case I do believe the right option is `fixed32`.

Otherwise we would be doing "early chain optimization", i.e. making encoding more compact in the early days with a negative trade-off for the future, once the project has gone mainstream ant the protocol would actually get the most of smaller data structures. No thanks!

PR #1166, opened 2020-04-15
feat(wallet): provide additional wallet transaction details

mariocao:
This PR:

- defines types following the specification discussed in the issue https://github.com/witnet/sheikah/issues/960
- implements additional functionality to the `index_txns` logic (in `worker/methods`) in order to query the node for previous transactions that contain the amounts of the transaction `inputs`.
- extends the `_index_transaction` logic (in `wallet/mod`) in order to:
  - use the previously queried additional information
  - store a `BalanceMovement` key/value instead of multiple keys
  - adds two new fields (responsible for this PR): `miner_fee` and `transaction_data` (which contains not only the inputs as `OutputPointers` but their `value` and `pkh`
 - adds field called `tally` in the tally transaction data (`TallyData`)
 - fills out a field called `tally` in previously indexed data request transaction (in `DrData`)
- implements additional functionality to query the node for data request reports


wallet/src/actors/worker/methods.rs
```diff
@@ -313,53 +319,92 @@ impl Worker {
...
+                types::Transaction::Tally(tally) => Some(IndexTransactionQuery::DataRequestReport(
+                    tally.dr_pointer.to_string(),
+                )),
+                _ => None,
             })
             .collect();
 
-        let query_vtt_outputs_futures: Result<Vec<Vec<ValueTransferOutput>>> = txns_output_pointers
+        let query_results: Result<Vec<model::ExtendedTransaction>> = txns
             .iter()
```
tmpolaczyk, line 366: You should be able to remove all the `transaction.clone()` from below if you make this iterator take `txns` by value:

```suggestion
            .into_iter()
```

wallet/src/types.rs
```diff
@@ -172,3 +172,14 @@ pub struct ExtendedKeyedSignature {
 
 #[derive(Clone, Debug, Deserialize)]
 pub struct ChainEntry(pub u32, pub String);
+
+/// Format of the output of getTransaction
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct GetTransactionOutput {
```
tmpolaczyk, line 179: This is a very bad name because a transaction has inputs and outputs, and this struct does not return the transaction outputs. I'm fine with `GetTransactionResult` or `GetTransactionResponse`
```suggestion
pub struct GetTransactionResult {
```

mariocao: I agree :smile:

wallet/src/actors/worker/methods.rs
```diff
@@ -307,6 +319,202 @@ impl Worker {
...
+        let retrieve_responses = async { futures03::future::try_join_all(txn_futures).await };
+        let transactions: Vec<types::Transaction> =
+            futures03::executor::block_on(retrieve_responses)?;
+        log::info!(
+            "Received {} wallet transactions from node",
+            transactions.len()
+        );
+
+        let result: Result<Vec<types::VttOutput>> = transactions
+            .iter()
```
tmpolaczyk, line 417: Same, you can probably remove all the `.map(types::VttOutput::clone)`

```suggestion
            .into_iter()
```

wallet/src/repository/wallet/mod.rs
```diff
@@ -330,25 +321,68 @@ where
...
+        txns: &[types::Transaction],
+    ) -> Result<Vec<types::Transaction>> {
+        let state = self.state.read()?;
+
+        let mut filtered_txns = vec![];
+        for txn in txns {
+            // Inputs and outputs from different transaction types
+            let (inputs, outputs) = match txn {
+                types::Transaction::ValueTransfer(vt) => {
+                    (vt.body.inputs.clone(), vt.body.outputs.clone())
```
tmpolaczyk, line 336: Maybe you can remove this clones by using references?

```suggestion
                    (&vt.body.inputs, &vt.body.outputs)
```

wallet/src/actors/worker/methods.rs
mariocao: Unfortunately, it seems not possible to get rid of the `clone`s :disappointed:

wallet/src/repository/wallet/mod.rs
mariocao: Tried but seems not possible.

wallet/src/repository/wallet/mod.rs
```diff
@@ -785,6 +877,171 @@ fn convert_block_epoch_to_timestamp(epoch_constants: EpochConstants, epoch: Epoc
...
+                        })
+                        .map_err(|err| Error::RevealRadDecode(err.to_string()))
+                })
+                .collect::<Result<HashMap<PublicKeyHash, model::Reveal>>>()?;
+
+            // Set not in consensus reveals
+            for pkh in &tally.slashed_witnesses {
+                let liar = reveals.get(&pkh).cloned();
+                if let Some(reveal) = liar {
+                    reveals.insert(
```
tmpolaczyk, line 1022: It looks like you are trying to mutate. You can use `reveals.get_mut()` and then `reveal.value.in_consensus = false`.

tmpolaczyk: And in the case where someone is penalized but has no reveal, you can try to insert a `RadonTypes::RadonError(NoReveal)`.

mariocao: I like it!

mariocao:
@tmpolaczyk, do you mind reviewing/accepting the PR?

@gabaldon already tested it and she already has a PR connecting Sheikah with the transaction details. :clap:

wallet/src/repository/wallet/mod.rs
```diff
@@ -588,47 +627,96 @@ where
...
+                balance_movement,
             )?;
-            batch.put(keys::transaction_block(account, txn_id), block_info)?;
 
-            // Transaction indexes
-            batch.put(keys::transactions_index(txn_hash.as_ref()), txn_id)?;
-            batch.put(keys::transaction_hash(account, txn_id), txn_hash.as_ref())?;
+            // Try to update data request tally report if it was previously indexed
+            if let types::Transaction::Tally(tally) = &txn.transaction {
+                log::debug!("Updating data request transaction with report from tally transaction");
```
aesedepece, line 682: Given that this is `debug` level logging, let's take the opportunity to state which specific request and tally transactions were involved in this.

wallet/src/model.rs
```diff
@@ -80,50 +75,113 @@ where
...
+    pub inputs: Vec<Input>,
+    pub outputs: Vec<Output>,
 }
 
-pub struct UnsupportedTransactionType(pub String);
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct DrData {
+    pub inputs: Vec<Input>,
+    pub outputs: Vec<Output>,
+    // #[serde(skip_serializing_if = "Option::is_none")]
```
aesedepece, line 140: What about this comment? :thinking:

wallet/src/model.rs
```diff
@@ -80,50 +75,113 @@ where
...
+    pub tally: Option<TallyReport>,
+}
 
-impl TryFrom<&types::Transaction> for TransactionType {
-    type Error = UnsupportedTransactionType;
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct TallyData {
+    pub request_transaction_hash: String,
+    pub outputs: Vec<Output>,
+    // #[serde(skip_serializing_if = "Option::is_none")]
```
aesedepece, line 148: Same over here.

wallet/src/model.rs
```diff
@@ -80,50 +75,113 @@ where
...
+            MovementType::Negative => write!(f, "negative"),
+        }
+    }
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct Transaction {
+    pub hash: String,
     pub timestamp: i64,
+    // #[serde(skip_serializing_if = "Option::is_none")]
```
aesedepece, line 110: This too.

mariocao:
@aesedepece there you have your change requests

PR #1168, opened 2020-04-15
Split mint and utxo splitting

### LINKED ISSUE #1115: [Collaterals] Smart UTXO splitting / joining

lrubiorod:
Close #1115


node/src/actors/chain_manager/mod.rs
```diff
@@ -936,7 +942,7 @@ where
 fn update_reputation(
     rep_eng: &mut ReputationEngine,
     consensus_constants: &ConsensusConstants,
-    miner_pkh: PublicKeyHash,
+    miner_pkh: Option<PublicKeyHash>,
```
tmpolaczyk, line 945: This should not be an option, and it should be the pkh used in the VRF proof, not the one used in the mint.

node/src/actors/chain_manager/handlers.rs
```diff
@@ -868,6 +869,20 @@ impl Handler<GetBalance> for ChainManager {
     }
 }
 
+impl Handler<GetUtxosCount> for ChainManager {
```
tmpolaczyk, line 872: Wouldn't it be more useful to have a `GetUtxos` method?

schemas/witnet/witnet.proto
```diff
@@ -203,7 +203,7 @@ message TallyTransaction {
 message MintTransaction {
     // The epoch field is used to ensure that each MintTransaction has a unique hash
     fixed32 epoch = 1;
-    ValueTransferOutput output = 2;
+    repeated ValueTransferOutput outputs = 3;
```
tmpolaczyk, line 206:
```suggestion
    repeated ValueTransferOutput outputs = 2;
```

node/src/actors/chain_manager/mining.rs
```diff
@@ -757,17 +771,42 @@ fn build_block(
...
     // Include Mint Transaction by miner
-    let reward = block_reward(epoch) + transaction_fees;
+    let mut reward = block_reward(epoch) + transaction_fees;
 
     // Build Mint Transaction
-    let mint = MintTransaction::new(
-        epoch,
-        ValueTransferOutput {
+    let mint = if split_mint {
+        let mut vt_outputs = vec![];
```
tmpolaczyk, line 778: This logic should be in a separated method, so we can test it more easily.

validations/src/tests.rs
```diff
@@ -145,6 +145,50 @@ fn mint_invalid_epoch() {
...
+}
+
+#[test]
+fn mint_split_valid() {
+    let epoch = 0;
+    let reward = block_reward(epoch);
+    let total_fees = 100;
+    let output1 = ValueTransferOutput {
+        pkh: Default::default(),
+        value: total_fees,
```
tmpolaczyk, line 179: Add another test where one of the values is 0. Also, we should put a strict limit on the size of the `MintTransaction` to prevent miners from filling blocks with junk.

validations/src/validations.rs
```diff
@@ -265,6 +265,13 @@ pub fn validate_mint_transaction(
         }
         .into())
     } else {
+        let first_pkh = mint_tx.outputs.first().unwrap().pkh;
```
tmpolaczyk, line 268: You will like https://docs.rs/itertools/0.9.0/itertools/trait.Itertools.html#method.all_equal

src/cli/node/with_node.rs
```diff
@@ -184,6 +229,12 @@ pub enum Command {
         #[structopt(short = "n", long = "node")]
         node: Option<SocketAddr>,
     },
+    #[structopt(name = "getUtxoInfo", about = "Get the utxo's number of the node")]
```
tmpolaczyk, line 232:
```suggestion
    #[structopt(name = "getUtxoInfo", about = "Get the unspent transaction outputs of the node")]
```

config/src/config.rs
```diff
@@ -325,6 +325,8 @@ pub struct Mining {
     pub data_request_timeout: Duration,
     /// Genesis block path
     pub genesis_path: String,
+    /// Binary flag to create a mint with a split reward
+    pub split_mint: bool,
```
aesedepece, line 329: Why not putting the value of the outputs itself and defaulting to whatever is the block reward?

lrubiorod: I include documentation in witnet.toml explaining it

data_structures/src/transaction.rs
```diff
@@ -466,14 +466,53 @@ pub struct MintTransaction {
...
+        collateral_minimum: u64,
+        collateral_age: u32,
+        data_request_max_retrievals_per_epoch: u16,
+    ) -> Self {
+        let mut vt_outputs = vec![];
+        let mut utxo_counter =
+            get_utxo_info(own_utxos, unspent_outputs_pool, collateral_minimum, 0)
+                .bigger_than_min_counter
+                + 1;
+        let utxos_to_achieve = collateral_age
```
tmpolaczyk, line 493: This can just be an argument

data_structures/src/transaction.rs
```diff
@@ -466,14 +466,53 @@ pub struct MintTransaction {
...
+        reward: u64,
+        own_pkh: PublicKeyHash,
+        own_utxos: &HashMap<OutputPointer, u64>,
+        unspent_outputs_pool: &UnspentOutputsPool,
+        collateral_minimum: u64,
+        collateral_age: u32,
+        data_request_max_retrievals_per_epoch: u16,
+    ) -> Self {
+        let mut vt_outputs = vec![];
+        let mut utxo_counter =
```
tmpolaczyk, line 489: Actually, this should also be an argument calculated externally, so the final list of arguments should be:


    pub fn with_split_utxos(
        epoch: Epoch,
        reward: u64,
        own_pkh: PublicKeyHash,
        collateral_minimum: u64,
        utxos_to_achieve: usize,
    ) -> Self {

data_structures/src/chain.rs
```diff
@@ -2099,6 +2100,50 @@ impl UnspentOutputsPool {
...
 
+/// Information about our own UTXOs
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct UtxoInfo {
+    /// Vector of OutputPointer with their values
+    pub utxos: Vec<(OutputPointer, u64)>,
+    /// Counter of UTXOs bigger than the minimum collateral
+    pub bigger_than_min_counter: usize,
+    /// Counter of UTXOs bigger than the minimum collateral and older than collateral coinage
+    pub old_utxos_counter: usize,
```
tmpolaczyk, line 2111: Maybe it would be better to calculate this in the client, not here. So for example users can see which UTXOs can be used for collateral, not just how many. We can do that by adding `included_in_block_number` to `utxos`, but we also need the `collateral_minimum` value.

tmpolaczyk: If this is too much we can open a new issue for adding features to `utxosInfo` :P

data_structures/src/chain.rs
```diff
@@ -2099,6 +2101,137 @@ impl UnspentOutputsPool {
...
+
+    pub fn contains_key<Q: ?Sized>(&self, k: &Q) -> bool
+    where
+        OutputPointer: std::borrow::Borrow<Q>,
+        Q: std::hash::Hash + Eq,
+    {
+        self.map.contains_key(k)
+    }
+
+    pub fn insert(&mut self, k: OutputPointer, v: u64, bigger_than_min: bool) -> Option<u64> {
```
tmpolaczyk, line 2139: It would be a better API if the `minimum_collateral` is stored in self, using `fn new(minimum_collateral: u32)` constructor. Then insert and remove can be simplified:
```suggestion
    pub fn insert(&mut self, k: OutputPointer, v: u64) -> Option<u64> {
```

data_structures/src/chain.rs
```diff
@@ -2099,6 +2101,137 @@ impl UnspentOutputsPool {
...
+    pub fn is_empty(&self) -> bool {
+        self.map.is_empty()
+    }
+}
+
+/// Information about our own UTXOs
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct UtxoInfo {
+    /// Vector of OutputPointer with their values, time_locks and if it is ready for collateral
+    pub utxos: Vec<(OutputPointer, u64, u64, bool)>,
```
tmpolaczyk, line 2188: Can you create a struct for this field? This would improve the JSON-RPC API by giving the fileds a name:

```rust
struct UtxoMetadata {
    output_pointer: OutputPointer,
    value: u64,
    timelock: u64,
    ready_for_collateral: bool,   
}
```

```suggestion
    pub utxos: Vec<UtxoMetadata>,
```

validations/src/validations.rs
```diff
@@ -265,6 +266,18 @@ pub fn validate_mint_transaction(
...
+
+        if !mint_tx.outputs.iter().map(|vto| vto.pkh).all_equal() {
+            return Err(BlockError::DifferentPkhsInMint.into());
+        }
+
+        for output in mint_tx.outputs.iter() {
+            if outputs_len > 1 && output.value < collateral_minimum {
+                return Err(BlockError::TooSplitMint.into());
+            }
+        }
```
aesedepece, line 279: These are consensus-critical statements that change the protocol in a hard-fork way. They could use code comments

data_structures/src/chain.rs
```diff
@@ -2465,6 +2598,27 @@ pub fn generate_unspent_outputs_pool(
     unspent_outputs
 }
 
+/// Method to sort own_utxos by value
+pub fn sort_own_utxos(
```
tmpolaczyk, line 2602: This can be a method of `OwnUnspentOutputsPool` so in the future we can optimize the implementation by adding a sorted index or a cache.

data_structures/src/error.rs
```diff
@@ -268,6 +268,10 @@ pub enum BlockError {
         fees_value: u64,
         reward_value: u64,
     },
+    #[fail(display = "Different PublicKeyHashes found in the Mint transaction")]
```
tmpolaczyk, line 271: :laughing: 

```suggestion
    #[fail(display = "Different addresses found in the Mint transaction")]
```

data_structures/src/error.rs
```diff
@@ -268,6 +268,10 @@ pub enum BlockError {
...
+    DifferentPkhsInMint,
+    #[fail(display = "MintTransaction is split in too small sizes")]
```
tmpolaczyk, line 273: This error may be hard to understand, but no idea how to improve it.

```suggestion
    #[fail(display = "MintTransaction output is smaller than the collateral minimum")]
```

data_structures/src/transaction.rs
```diff
@@ -466,14 +466,44 @@ pub struct MintTransaction {
...
         }
     }
 
+    pub fn with_split_utxos(
+        epoch: Epoch,
+        reward: u64,
+        own_pkh: PublicKeyHash,
+        collateral_minimum: u64,
+        utxo_counter: usize,
+        utxos_to_achieve: usize,
```
tmpolaczyk, line 483: Can you refactor this into only one argument? By changing the calls to `utxos_to_achieve.saturating_sub(utxo_counter)`.

```suggestion
        utxos_to_achieve: usize,
```

node/src/actors/chain_manager/transaction_factory.rs
```diff
@@ -107,14 +113,20 @@ pub fn insert_change_output(
...
+    own_utxos: &mut OwnUnspentOutputsPool,
     own_pkh: PublicKeyHash,
     all_utxos: &UnspentOutputsPool,
     timestamp: u64,
     tx_pending_timeout: u64,
+    // Sorted utxo by value
+    // None: no sorted
+    // Some(true): bigger value first
+    // Some(false): lower value first
+    sorted_bigger: Option<bool>,
```
tmpolaczyk, line 129: This should be an enum:

```rust
enum UtxoSelectionStrategy {
    None, // or Random
    BigFirst,
    SmallFirst,
}
```

```suggestion
    utxo_selection_strategy: UtxoSelectionStrategy,
```

data_structures/src/chain.rs
```diff
@@ -2099,6 +2101,137 @@ impl UnspentOutputsPool {
...
+    where
+        OutputPointer: std::borrow::Borrow<Q>,
+        Q: std::hash::Hash + Eq,
+    {
+        self.map.contains_key(k)
+    }
+
+    pub fn insert(&mut self, k: OutputPointer, v: u64, bigger_than_min: bool) -> Option<u64> {
+        let old_ts = self.map.insert(k, v);
+        if bigger_than_min {
```
tmpolaczyk, line 2141: Here you can also add a check like the one in `remove`

```suggestion
        if old_ts.is_none() && bigger_than_min {
```

data_structures/src/chain.rs
aesedepece: I support that idea. The node itself should always give raw data over the JSONRPC interface. Then, we can process it and make it more meaningful to end users in the CLI. This also has the benefit of offloading the node server as any additional computation is done on the client side.

validations/src/validations.rs
```diff
@@ -1474,13 +1487,18 @@ pub fn validate_block_transactions(
...
+            block_beacon.checkpoint,
+            collateral_minimum,
+        )?;
 
         // Insert mint in utxo
         update_utxo_diff(
             &mut utxo_diff,
             vec![],
-            vec![&block.txns.mint.output],
+            block.txns.mint.outputs.iter().collect(),
```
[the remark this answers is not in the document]
lrubiorod: It is not really a clone, it would change a Vec<ValueTransaferOutput> to Vec<&ValueTransferOutput>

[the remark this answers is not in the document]
aesedepece: oh cool!

lrubiorod:
> What do you think about a `--dry-run` flag that gives you a preview of what the transaction resulting of calling `join` or `split` will be? I feel dizzy when invoking this methods right now see_no_evil
> 
> IMO, `send` and other methods could also use that flag indeed.

Done!!

data_structures/src/chain.rs
lrubiorod: Oks. Now I return utxo, value, time_lock and if it is ready to collaterize

src/cli/node/json_rpc_client.rs
```diff
@@ -115,6 +115,52 @@ pub fn get_pkh(addr: SocketAddr) -> Result<(), failure::Error> {
...
+    let utxos_len = utxo_info.utxos.len();
+    let mut table = Table::new();
+    table.set_format(*prettytable::format::consts::FORMAT_NO_BORDER_LINE_SEPARATOR);
+    table.set_titles(row![
+        "OutputPointer",
+        "Value (in wits)",
+        "Time lock",
+        "Ready for collateral"
+    ]);
+    for utxo_metadata in utxo_info.utxos.into_iter().sorted_by_key(|um| um.value) {
```
tmpolaczyk, line 134: Here you can sort by value and then by output pointer, so the list is sorted when there are many UTXOs with the same value:

```suggestion
    for utxo_metadata in utxo_info.utxos.into_iter().sorted_by_key(|um| (um.value, um.output_pointer)) {
```

But you need to implement `Ord` for `OutputPointer`, which should be something like:

```rust
impl Ord for OutputPointer {
    fn cmp(&self, other: &Self) -> Ordering {
        (self.transaction_id, self.output_index).cmp(&(other.transaction_id, &other.output_index))
    }
}
```

node/src/actors/chain_manager/mining.rs
```diff
@@ -758,16 +771,29 @@ fn build_block(
 
     // Include Mint Transaction by miner
     let reward = block_reward(epoch) + transaction_fees;
+    let utxos_to_achieve =
```
tmpolaczyk, line 774: Can you explain a bit why we need this?

data_structures/src/transaction.rs
```diff
@@ -466,14 +466,43 @@ pub struct MintTransaction {
...
+    pub fn new(epoch: Epoch, outputs: Vec<ValueTransferOutput>) -> Self {
         MintTransaction {
             epoch,
-            output,
+            outputs,
             hash: MemoHash::new(),
         }
     }
 
+    pub fn with_split_utxos(
```
tmpolaczyk, line 477:
```suggestion
    // Try to create a MintTransaction with number of outputs equal to `utxos_required`, where every one of those outputs must be equal to or greater than `collateral_minimum`. If the reward is too small, the number of outputs may be smaller than `utxos_required`. The output value will only be smaller than `collateral_minimum` if the `reward` is smaller than `collateral_minimum`, in which case this function will create a MintTransaction with exactly one output
    pub fn with_split_utxos(
```

node/src/actors/messages.rs
```diff
@@ -169,6 +170,8 @@ pub struct BuildVtt {
     pub vto: Vec<ValueTransferOutput>,
     /// Fee
     pub fee: u64,
+    /// Strategy to sort the unspent outputs pool
+    pub utxo_strategy: UtxoSelectionStrategy,
```
tmpolaczyk, line 174: We can add a serde attribute to deserialize as the default if the value is missing, this way we don't break compatibility, if someone was using the JSON-RPC send method it will keep working the same way as before.

I think you will have to implement Default as UtxoSelectionStrategy::Random.

```suggestion
    #[serde(default)]
    pub utxo_strategy: UtxoSelectionStrategy,
```

src/cli/node/with_node.rs
```diff
@@ -231,6 +291,63 @@ pub enum Command {
...
+        /// Utxo's size
+        #[structopt(long = "size")]
+        size: u64,
+        /// Fee
+        #[structopt(long = "fee")]
+        fee: u64,
+        /// Time lock
+        #[structopt(long = "time-lock")]
+        time_lock: Option<u64>,
+        /// Dry-run flag
```
tmpolaczyk, line 321:
```suggestion
        /// Print the request that would be sent to the node and exit without doing anything
```

PR #1169, opened 2020-04-15
Make wallet server report node and own status to clients

### LINKED ISSUE #1152: Make wallet inform / notify clients about wallet syncing status and overall node synchronization
The wallet should expose a JSON-RPC method called `get_status` providing the following information about the status of the node and the wallet synchronization:

- node connection
- node height
- wallet height

aesedepece:
This is more or less complete, it is only missing the network label, which the node doesn't provide. I'll do some magic tricks to get that.

Fix #1152


aesedepece:
Here's how it's looking so far from Sheikah:
image

aesedepece:
image

aesedepece:
image

wallet/src/actors/app/methods.rs
```diff
@@ -22,12 +23,12 @@ impl App {
     /// Return a new subscription id for a session.
     pub fn next_subscription_id(
         &mut self,
-        session_id: types::SessionId,
+        session_id: &types::SessionId,
     ) -> Result<types::SubscriptionId> {
-        if self.state.is_session_active(&session_id) {
+        if self.state.is_session_active(session_id) {
             // We are re-using the session id as the subscription id, this is because using a number
             // can let any client call the unsubscribe method for any other session.
```
tmpolaczyk, line 30: If we ever want to support more than one subscription id per session, we can just use `format!("{}-{}", session_id, subscription_id)` as the subscription id.

wallet/src/actors/app/methods.rs
```diff
@@ -445,31 +446,40 @@ impl App {
     }
 
     /// Handle new block notifications received from the node.
-    pub fn handle_block_notification(&self, value: types::Json) -> Result<()> {
-        log::trace!("received block notification");
+    pub fn handle_block_notification(&mut self, value: types::Json) -> Result<()> {
         let block = serde_json::from_value::<types::ChainBlock>(value).map_err(node_error)?;
 
-        for (_id, wallet) in self
+        // This iterator is collected early so as to free the immutable reference to `self`.
```
tmpolaczyk, line 452: This is because `self.state.get_sink` takes a mutable reference to self. No idea about what's the current architecture, but if every wallet has a sink, why not store it in the wallet instead of in self.state?

wallet/src/actors/app/methods.rs
aesedepece: The silly thing here is that a session already means one subscriber through one WS connection. I don't think we need more than one sink.

Although who knows, wasn't Bill Gates who said that 640KB was more memory than anyone will ever need? :rofl:

wallet/src/actors/app/methods.rs
aesedepece: It was inside `self.state.wallets` before, which was even crazier, lol.

Whoever wrote all that part of the wallet left no hints at all at why we got `Wallet`, `SessionWallet`, `State`, `Params` and a myriad more data structures that resemble each other without offering any additional convenience :shrug:

wallet/src/types.rs
```diff
@@ -172,3 +203,39 @@ pub struct ExtendedKeyedSignature {
...
+
+/// Friendly events that can be sent to subscribed clients to let them now about significant
+/// activity related to their wallets.
+#[derive(Debug, Serialize)]
+pub enum Event {
+    /// The basic information of a new block that has already been processed
+    Block(model::BlockInfo),
+    // TODO: turn this into `Movement(Movement)`
+    ///// A new movement (transaction) affecting balance.
+    //Movement,
```
aesedepece, line 234: This is here for when this PR and #1166 meet each other and have baby PRs.

wallet/src/params.rs
```diff
@@ -43,6 +44,54 @@ impl Default for Params {
...
+
+    /*
+    /// Update the `network` field in runtime, as it is unknown at boot.
+    /// This is a best-effort method. It will silently do nothing if the write lock on `network`.
+    pub fn update_network(&self, new_network: String) {
+        let lock = (*self.network).write();
+        if let Ok(mut network) = lock {
+            *network = Some(new_network)
+        }
+    }*/
```
aesedepece, line 96: As explained in the PR comment, I'm still undecided on what to do with this.

We could try to guess the network from a few different places, e.g. last bytes of the bootstrap hash.

Of course we could ask the node through JSON-RPC. But the bottom line is that not even the node is aware of what's the "identifier" of the network it is operating on. 

So maybe I'd go with a combination of those two things: asking the node and let the node guess from the bootstrap hash.

tmpolaczyk: If we want to show the network as "testnet-7", "testnet-8", "mainnet", or even some custom names, then we can add a `network_name` field to `witnet.toml`, and add that to the status JSON-RPC method of the node. If we only want testnet/mainnet then I think there is a variable in the wallet like `testnet = true`, and the wallet will show an error when connecting to a node running on mainnet because the addresses have a different prefix (twit vs wit).

aesedepece: Let's do the latter for the time being.

aesedepece:
I'm rebasing this now, making sure everything works and trying to send movement event notifications.

PR #1143, opened 2020-04-02
Update of operator-5 for testnet 7.3

bertux:
stop using local config file

PR #1173, opened 2020-04-16
fix(docker-compose): do not require customized config file

bertux:
to do as what has been done for the 5 nodes integration of Docker-Compose at https://github.com/witnet/witnet-rust/pull/1143

### DIFF
--- a/docker/compose/bertux-operator-stable/docker-compose.yaml
+++ b/docker/compose/bertux-operator-stable/docker-compose.yaml
@@ -4,10 +4,9 @@ services:
   node:
     image: witnet/witnet-rust
     restart: unless-stopped
-    command: ["latest", "--config", "/.witnet.toml", "--debug", "node", "server"]
+    command: ["latest", "node", "server"]
     network_mode: host
     ports:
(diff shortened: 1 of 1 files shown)


aesedepece:
I assume that the idea is that if an operator wants to use a custom config file, they will simply modify the command in `docker-compose.yaml` . Am I right?

bertux:
> I assume that the idea is that if an operator wants to use a custom config file, they will simply modify the command in `docker-compose.yaml` . Am I right?

yes you are totally right, it's to make the default version working "out of the box" without having first to document how to use the configuration file. Then it can be documented in a next version of the ongoing documentation of docker-compose I've already proposed.
And we can still make another directory for the version of docker-compose.yml which uses a customized config file where the public IP address will be added.
Because it's also connected to https://github.com/witnet/witnet-rust/pull/1172 which doesn't include https://github.com/witnet/witnet-rust/pull/1140 about detecting automatically the public IP. Public IP teller can become one of our community tools instead of relying on a third-party service unrelated to Witnet.

PR #1198, opened 2020-04-20
feat(docker-compose): add getVersion wrapper

bertux:


aesedepece:
I may be wrong, but I think that the correct command for getting the version is:
```console
witnet node --version
```

bertux:
I've copied what I was using before on my node but you are right it doesn't work anymore and now it's needed to use --version so I'm correcting it, thank you to have seen it.

aesedepece:
@bertux This is one trick I normally do for running the same command on all nodes:
```console
docker ps | grep witnet-rust | cut -d' ' -f1 | parallel docker exec {} ./witnet node --version
```

Needless to say that it has `parallel` as a dependency :sweat_smile:

bertux:
> @bertux This is one trick I normally do for running the same command on all nodes:
> 
> ```
> docker ps | grep witnet-rust | cut -d' ' -f1 | parallel docker exec {} ./witnet node --version
> ```
> 
> Needless to say that it has `parallel` as a dependency 😅

Thank you for this trick, I will add it to the script, and it would be even better to be able to use Docker tags to select the docker containers running witnet-rust

Issue #1051, opened 2020-02-27
All `--pkh` arguments in the CLI should be renamed to `--address`


aesedepece:
## Description

So far, all methods optionally accepting addresses as argument expect those to be preceded by the `--pkh` flag.

`PKH` stands for "Public Key Hash". While it is true that addresses in Witnet are derived from the hash of the public key, there's no 1:1 correspondence between the concepts of "PKH" and "address". Namely, an address is the Bech32 serialized form of the first 20 bytes of a PKH.

Moreover, most smart contract developers and the broader Ethereum community is most likely not familiar at all with the concept of a PKH (even if Ethereum addresses are generated in the same way as Witnet addresses although serialized differently).

For all of this, I see no reason why we should keep miscalling addresses.

## Proposal

- Renaming all `--pkh` arguments as `--address`. 

## Consequences

- This is not a breaking change at the consensus / protocol level.
- This is an API breaking change if implemented at JSONRPC level.
- This is a minor usability challenge if only implemented at CLI level.
- This change should be reflected in the documentation.

tmpolaczyk:
We could add `--address` as an alias to `--pkh`, so both syntaxes work. If we want to deprecate `--pkh`, then it is a breaking change.

aesedepece:
I have updated the "consequences" section to better describe the severity of the changes. As you can see, they are different depending on whether we only change the CLI command or also the JSONRPC command.

I'm totally in for the alias. Then we can discuss if we want to keep the `--pkh` or not. For reasons of maintainability and preventing more severe breaking changes in the future, I'd stick to the smallest API possible, so I'd remove it.

bertux:
implementing it at https://github.com/bertux/witnet-rust/tree/cli-pkh-rename

bertux:
GetPkh name needs aliasing and updated about description
GetBalance GetReputation Send needs aliasing of their long parameter --pkh

PR #1200, opened 2020-04-20
feat(cli): rename pkh option into address

bertux:
first draft for https://github.com/witnet/witnet-rust/issues/1051
still to need to test how to add the aliasing because I've seen for the command but not yet for the options

### DIFF
--- a/src/cli/node/with_node.rs
+++ b/src/cli/node/with_node.rs
@@ -179,7 +179,7 @@ pub enum Command {
         #[structopt(short = "n", long = "node")]
         node: Option<SocketAddr>,
         /// Public key hash for which to get balance. If omitted, defaults to the node pkh.
-        #[structopt(long = "pkh")]
+        #[structopt(long = "address")]
         pkh: Option<String>,
     },
     #[structopt(name = "getPkh", about = "Get the public key hash of the node")]
@@ -197,7 +197,7 @@ pub enum Command {
         #[structopt(short = "n", long = "node")]
         node: Option<SocketAddr>,
         /// Public key hash for which to get reputation. If omitted, defaults to the node pkh.
-        #[structopt(long = "pkh")]
+        #[structopt(long = "address")]
         pkh: Option<String>,
         /// Print all the reputation?
         #[structopt(long = "all", conflicts_with = "pkh")]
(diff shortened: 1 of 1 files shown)


aesedepece:
So far so good!

bertux:
I've compiled and tested on my node and it works for both --address and --pkh
To stay coherent I have also renamed getPkh into getAddress but I've added getPkh as alias so both work also on my node server.
I've not changed the RPC to avoid breaking change but it helped me to understand better how it is called by command line.

src/cli/node/with_node.rs
```diff
@@ -178,12 +178,16 @@ pub enum Command {
         /// Socket address of the Witnet node to query.
         #[structopt(short = "n", long = "node")]
         node: Option<SocketAddr>,
-        /// Public key hash for which to get balance. If omitted, defaults to the node pkh.
-        #[structopt(long = "pkh")]
-        pkh: Option<String>,
+        /// Public address for which to get balance. If omitted, defaults to the node address.
```
aesedepece, line 181:
```suggestion
        /// Address for which to get balance. If omitted, defaults to the node address.
```

src/cli/node/with_node.rs
```diff
@@ -196,11 +200,11 @@ pub enum Command {
         /// Socket address of the Witnet node to query.
         #[structopt(short = "n", long = "node")]
         node: Option<SocketAddr>,
-        /// Public key hash for which to get reputation. If omitted, defaults to the node pkh.
-        #[structopt(long = "pkh")]
-        pkh: Option<String>,
+        /// Public address for which to get reputation. If omitted, defaults to the node address.
```
aesedepece, line 203:
```suggestion
        /// Address for which to get reputation. If omitted, defaults to the node address.
```

src/cli/node/with_node.rs
```diff
@@ -219,9 +223,9 @@ pub enum Command {
         /// Socket address of the Witnet node to query.
         #[structopt(short = "n", long = "node")]
         node: Option<SocketAddr>,
-        /// Public key hash of the destination
-        #[structopt(long = "pkh")]
-        pkh: String,
+        /// Public address of the destination
```
aesedepece, line 226:
```suggestion
        /// Address of the destination
```

PR #1204, opened 2020-04-21
fix(jsonrpc): remove subscriptions on session close

tmpolaczyk:
This fixes an error found in one of the bootstrap nodes, which stopped consolidating blocks when it got to 925 subscriptions.


tmpolaczyk:
You can try this by running a node, and then on a different terminal:

    cargo run -- node raw

And paste

    {"jsonrpc":"2.0","id":"1","method":"witnet_subscribe","params":["newBlocks"]}

To add a subscription. You can paste that multiple times to add multiple subscriptions. When you close that terminal, all the subscriptions will be closed, there is a log line that prints the number of active subscriptions.

PR #1207, opened 2020-04-22
feat(wallet): add address additional information

mariocao:
In this PR:
 - Extend address information with a new field called `info`.
 - Fix duplicated field contents `pkh` and `address` (currently `pkh` and `address` fields are the same).

These are the data structure returned by the wallet server `get_addresses`:

```rust
#[derive(Debug, Serialize, PartialEq)]
pub struct Address {
    pub address: String,
    pub index: u32,
    pub keychain: u32,
    pub account: u32,
    pub path: String,
    pub info: AddressInfo,
    pub pkh: types::PublicKeyHash,
}

#[derive(Debug, Serialize, Deserialize, PartialEq)]
pub struct AddressInfo {
    pub label: Option<String>,
    pub received_payments: Vec<String>,
    pub received_amount: u64,
    pub first_payment_date: Option<i64>,
    pub last_payment_date: Option<i64>,
}
```

where the `received_payments` is a list of output_pointers of the form `<txn_id>:<index>`.

### DIFF
--- a/wallet/src/model.rs
+++ b/wallet/src/model.rs
@@ -30,7 +30,7 @@ pub struct Address {
     pub keychain: u32,
     pub account: u32,
     pub path: String,
-    pub label: Option<String>,
+    pub info: AddressInfo,
     pub pkh: types::PublicKeyHash,
 }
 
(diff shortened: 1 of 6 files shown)


wallet/src/model.rs
```diff
@@ -40,6 +40,15 @@ pub struct Addresses {
     pub total: u32,
 }
 
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+pub struct AddressInfo {
+    pub label: Option<String>,
```
tmpolaczyk, line 45: Does it make sense to add the serde skip attribute to the optional fields?

    #[serde(skip_serializing_if = "Option::is_none")]

mariocao: We checked that the `skip_serializing_if` is not working properly with the `bincode` which we use for the DB. :disappointed:  Hopefully, we'll change it in the near future for `protobuf` or similar.

PR #1212, opened 2020-04-23
fix(node): send PeersBeacons even if peers < outbound limit

tmpolaczyk:
This fixes a bug where the node would move to WaitingConsensus as soon
as one of its peers closed the connection.

Now the SessionsManager will always send the list of beacons to the
ChainManager. If there is consensus after counting the missing peers as
"NO BEACON", the node will start synchronizing even if it's still
missing some peers.

For example, if a node has `outbound_limit = 5` and `consensus_c = 70`, it means that it's waiting for 5 peers, and at least 70% of them must agree (in this case 4 peers out of 5). So before this PR the node would wait for 5 peers, and then calculate the consensus. After this PR, the node calculates the consensus once every epoch, and if at least 4 peers reported the same beacon, it will start synchronizing even if it is still missing some peers.

---------------

If the node has less peers than "consensus_c %", so in the previous example 3 peers out of 5, it will not drop any peers.

This PR also reverts one change introduced in commit 15e9313c06e933adfc438c7fa29b642aa98f8369 (PR  https://github.com/witnet/witnet-rust/pull/1126). We found a case where the node wrongly assumes that it got all the beacons for this epoch, when in reality it just got the beacons from last epoch when connecting to the peer:

image

So now we do not send the beacon when establishing a new connection. That makes synchronization up to one epoch slower, and now the nodes cannot sync before epoch 0, unless we add a hack to send beacons periodically before epoch 0.

### DIFF
--- a/node/src/actors/chain_manager/handlers.rs
+++ b/node/src/actors/chain_manager/handlers.rs
@@ -543,24 +543,45 @@ impl Handler<PeersBeacons> for ChainManager {
         // Calculate the consensus, or None if there is no consensus
         let consensus_threshold = self.consensus_c as usize;
         let consensus = peers_beacons.consensus(consensus_threshold);
-
         let pb = peers_beacons.pb;
+        let outbound_limit = peers_beacons.outbound_limit;
+        let pb_len = pb.len();
+        let peers_needed_for_consensus = outbound_limit
+            .map(|x| {
+                // ceil(x * consensus_threshold / 100)
+                (usize::from(x) * consensus_threshold + 99) / 100
+            })
+            .unwrap_or(1);
+        let peers_to_unregister = if let Some(consensus_beacon) = consensus {
+            // Consensus: unregister peers which have a different beacon
+            pb.into_iter()
+                .filter_map(|(p, b)| {
+                    if b != Some(consensus_beacon) {
+                        Some(p)
+                    } else {
+                        None
+                    }
+                })
+                .collect()
+        } else if pb_len < peers_needed_for_consensus {
+            // Not enough outbound peers, do not unregister any peers
+            log::debug!(
+                "Got {} peers but need at least {} to calculate the consensus",
+                pb_len,
+                peers_needed_for_consensus
+            );
+            vec![]
+        } else {
+            // No consensus: unregister all peers
+            log::warn!("No consensus: unregister all peers");
+            pb.into_iter().map(|(p, _b)| p).collect()
+        };
+
         match self.sm_state {
             StateMachine::WaitingConsensus => {
                 // As soon as there is consensus, we set the target beacon to the consensus
                 // and set the state to Synchronizing
                 if let Some(consensus_beacon) = consensus {
-                    // Consensus: unregister peers which have a different beacon
-                    let peers_out_of_consensus = pb
-                        .into_iter()
-                        .filter_map(|(p, b)| {
-                            if b != Some(consensus_beacon) {
-                                Some(p)
-                            } else {
-                                None
-                            }
-                        })
-                        .collect();
                     self.target_beacon = Some(consensus_beacon);
 
                     let our_beacon = self.get_chain_beacon();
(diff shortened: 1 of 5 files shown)


node/src/actors/chain_manager/handlers.rs
```diff
@@ -511,12 +524,16 @@ impl Handler<PeersBeacons> for ChainManager {
...
-            .and_then(|x| *x);
+        let consensus = mode_consensus(
+            pb.iter()
+                .map(|(_p, b)| b)
+                .chain(std::iter::repeat(&None).take(num_missing_peers)),
+            consensus_threshold,
+        )
+        // Flatten result:
+        // None (consensus % below threshold) should be treated the same way as
+        // Some(None) (most of the peers did not send a beacon)
```
aesedepece, line 535: What's this commenting on?

node/src/actors/sessions_manager/handlers.rs
```diff
@@ -320,11 +320,18 @@ impl Handler<EpochNotification<()>> for SessionsManager {
...
-                act.try_send_peers_beacons(ctx)
-            {
-                // Send it even if it is incomplete, and unregister the peers which have not sent a beacon
-                act.send_peers_beacons(ctx);
+            match act.try_send_peers_beacons(ctx) {
+                Ok(_) => {}
+                Err(NotSendingPeersBeaconsBecause::AlreadySent) => {}
+                Err(NotSendingPeersBeaconsBecause::BootstrapNeeded) => {
+                    // If the number of peers is less than the outbound limit
+                    // and unregister the peers which have not sent a beacon
```
aesedepece, line 328: Weird sentence hehehe :sweat_smile:

node/src/actors/chain_manager/handlers.rs
tmpolaczyk: The and_then below, which maps Some(None) to None

node/src/actors/chain_manager/handlers.rs
```diff
@@ -511,12 +524,16 @@ impl Handler<PeersBeacons> for ChainManager {
...
-        // In case of tie returns None
-        let consensus = mode_consensus(pb.iter().map(|(_p, b)| b), consensus_threshold)
-            // Flatten result:
-            // None (no consensus) should be treated the same way as
-            // Some(None) (the consensus is that there is no consensus)
-            .and_then(|x| *x);
+        let consensus = mode_consensus(
+            pb.iter()
+                .map(|(_p, b)| b)
+                .chain(std::iter::repeat(&None).take(num_missing_peers)),
```
aesedepece, line 530: Lol to this line.

tmpolaczyk: Hacks everywhere

node/src/actors/chain_manager/handlers.rs
aesedepece: My bad!

lrubiorod:
review what happens with yout outbounds when there are too much outbounds missing

lrubiorod:
image

When a node start later, it looks like the PeersBeacons message arrives late and, unless it has the real beacons, only check some beacons from one epoch before. It produces that being well synchronized it thinks that is not, and restart again the synchronization

tmpolaczyk:
@lrubiorod this happens because we didn't clean the old beacons. We normally clean the old beacons after sending the PeersBeacons message, but in this case that's the first time we send the PeersBeacons message so the node didn't do that. The beacons that were stored are the ones that we receive when connecting to a new outbound peer. In an attempt to quickly fix that bug I just reverted commit 15e9313c06e933adfc438c7fa29b642aa98f8369

Note that while the node can keep syncing, this bug can happen every time we drop a peer and connect to a new one, so we can't just not fix it...

tmpolaczyk:
> review what happens with yout outbounds when there are too much outbounds missing

So in the current implementation if there is only 1 peer out of 4, the consensus is `None` and all the peers are dropped. I will change it so that we will not drop any peers if the number of peers divided by the outbound limit is below "consensus_c %".

PR #1213, opened 2020-04-24
fix(clippy): apply new clippy suggestions after rust update

lrubiorod:


p2p/src/sessions/mod.rs
```diff
@@ -249,11 +249,7 @@ where
     ) -> Result<(), failure::Error> {
         // If this is an outbound consolidated session, try to remove it from the consensus list
         if let (SessionType::Outbound, SessionStatus::Consolidated) = (session_type, status) {
-            match self.unconsensus_session(address) {
-                // Explicitly ignore the result because we have no guarantees that this session was
```
tmpolaczyk, line 253: Can you keep this comment?

PR #1214, opened 2020-04-24
fix(wallet): index genesis block in initial sync

mariocao:
The wallet was not indexing the genesis block at the initial wallet synchronization, because of the `since_beacon` is pointing to epoch 0 and the sync logic start at `checkpoint + 1`.

The solution was to handle this specific case before the `loop` that iterates synchronization batches.

### DIFF
--- a/wallet/src/actors/worker/methods.rs
+++ b/wallet/src/actors/worker/methods.rs
@@ -566,6 +566,22 @@ impl Worker {
         let mut since_beacon = since_beacon;
         let first_beacon = since_beacon;
 
+        // If first initial synchronization (checkpoint == 0), genesis block should be indexed
+        if latest_beacon.checkpoint == 0 {
+            let gen_fut = self.get_block_chain(0, 1);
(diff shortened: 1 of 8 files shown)


wallet/src/actors/worker/methods.rs
```diff
@@ -566,6 +566,22 @@ impl Worker {
         let mut since_beacon = since_beacon;
         let first_beacon = since_beacon;
 
+        // If first initial synchronization (checkpoint == 0), genesis block should be indexed
+        if latest_beacon.checkpoint == 0 {
+            let gen_fut = self.get_block_chain(0, 1);
+            let gen_res: Vec<ChainEntry> = futures03::executor::block_on(gen_fut)?;
+            let gen_entry = gen_res
+                .get(0)
+                .expect("A Witnet chain should always have at least one genesis block");
```
aesedepece, line 575:
```suggestion
                .expect("A Witnet chain should always have a genesis block");
```

PR #1215, opened 2020-04-27
feat(cli): add long flag to getUtxoInfo

lrubiorod:
Dont show all the utxo information (because it could be very noisy if you have so many utxos) if you don specific require with "--long" flag


src/cli/node/with_node.rs
```diff
@@ -250,6 +251,9 @@ pub enum Command {
         /// Show all the information about utxos
         #[structopt(long = "long")]
         long: bool,
+        /// Public key hash for which to get reputation. If omitted, defaults to the node pkh.
```
tmpolaczyk, line 254:
```suggestion
        /// Public key hash for which to get UTXO information. If omitted, defaults to the node pkh.
```

PR #1221, opened 2020-04-28
Fix docker image so it correctly sets the public address unless disabled

aesedepece:
This fixes the `witnet/witnet-rust` docker image so it correctly sets the public address when starting unless explicitly disabled using the `PUBLIC_ADDR_DISCOVERY=false` environment flag:

```console
docker run -d \
    --volume ~/.witnet:/.witnet \
    --name witnet_node \
    --publish 21337:21337 \
    -e PUBLIC_ADDR_DISCOVERY=false \
    witnet/witnet-rust
```

### DIFF
--- a/docker/witnet-rust/Dockerfile
+++ b/docker/witnet-rust/Dockerfile
@@ -1,4 +1,4 @@
-FROM --platform=$TARGETPLATFORM ubuntu:disco
+FROM --platform=$TARGETPLATFORM ubuntu:focal
 
 # Install basic environment dependencies
 RUN apt-get update && apt-get install -y --no-install-recommends \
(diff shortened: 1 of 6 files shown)


bertux:
I think you have forgotten to use chmod +x *.sh instead of chmod +x runner.sh because now there are other scripts.

docker/witnet-rust/Dockerfile
```diff
@@ -22,11 +22,11 @@ EXPOSE 21338
 EXPOSE 11212
 
 # Run the install script
-RUN ["chmod", "+x", "./runner.sh"]
-RUN ["./runner.sh"]
+RUN ["chmod", "+x", "./*.sh"]
```
aesedepece, line 25: For some reason this doesn't work.

bertux:
thank you @aesedepece for my first review request here, I will check that on my node server to make it work.

aesedepece:
Merged as b4291040d408c4d1670ee2c8efb94a7f99a48bd2.

PR #1224, opened 2020-04-29
feat(wallet): notify initial wallet status as soon as possible

aesedepece:


wallet/src/actors/app/state.rs
```diff
@@ -56,19 +56,28 @@ impl State {
         }
     }
 
+    /// Get all wallets for a session
+    pub fn wallets(
```
mariocao, line 60:
```suggestion
    pub fn session_wallets(
```

Maybe something more descriptive?

wallet/src/actors/app/state.rs
```diff
@@ -56,19 +56,28 @@ impl State {
         }
     }
 
+    /// Get all wallets for a session
+    pub fn get_wallets_by_session(
+        &self,
+        session_id: &types::SessionId,
+    ) -> Result<HashMap<String, types::SessionWallet>> {
```
tmpolaczyk, line 63: Cloning an entire hashmap may be too much, just return a reference and clone in the caller if needed.

```suggestion
    pub fn get_wallets_by_session<'a>(
        &'a self,
        session_id: &types::SessionId,
    ) -> Result<&'a HashMap<String, types::SessionWallet>> {
```

aesedepece: Great idea. I didn't do it initially because the biggest part of the HashMap is `Arc`ed, so it felt permissible at the time, but I agree we'll be better off returning a reference. Thanks!

aesedepece: Done :wink:

PR #1225, opened 2020-04-29
Avoid show redundant block and transaction messages in debug log level

lrubiorod:
Created a hashset in SessionsManager to avoid showing redundant messages received from different sessions

It also implemented a little fix about new blocks handling, because it should be only handle as a new Candidate, but it was handle as a Candidate and as a requested block

And a little refactor in code to use log::debug!() call instead of only debug!() call

### DIFF
--- a/node/src/actors/session/mod.rs
+++ b/node/src/actors/session/mod.rs
@@ -1,22 +1,19 @@
 use std::{collections::HashMap, net::SocketAddr, time::Duration};
 
-use actix::io::FramedWrite;
+use actix::{io::FramedWrite, SystemService};
 
 use ansi_term::Color::Green;
-
-use log::{debug, error, trace};
 
 use tokio::{io::WriteHalf, net::TcpStream};
 
 use witnet_data_structures::{
-    chain::{Block, Hash},
+    chain::{Block, Epoch, Hash},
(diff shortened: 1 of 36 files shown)


node/src/actors/session/mod.rs
```diff
@@ -129,34 +126,63 @@ impl Session {
...
-        trace!("\t{:?}", msg);
+        match msg.kind {
+            Command::Transaction(_) | Command::Block(_) => {
+                let log_data = format!(
+                    "{} Received {} message ({} bytes)",
+                    Green.bold().paint("[<]"),
+                    Green.bold().paint(msg.kind.to_string()),
+                    bytes.len(),
+                );
+                SessionsManager::from_registry().do_send(LogMessage { log_data })
```
tmpolaczyk, line 173: Maybe we can also save `self.remote_addr`, even if we do not print it, so that we don't lose that information. This would mean changing the `HashSet<String>` into a `HashMap<String, Vec<SocketAddr>>`.

lrubiorod: This PR is only to change the way that we show the messages, but it doesnt affect to the witnet logic, so... why we would need to save that information?

PR #1227, opened 2020-04-30
BLS signatures

### LINKED ISSUE #1228: Generate BLS key upon start
Generate a BLS key upon start of the node. Currently there is no need to store the BLS key if the node is stopped. We can re-generate a BLS key every time we start the node, as this would only update every other nodes mapping.

### LINKED ISSUE #1230: Update Signature Manager API
The signature manager API should be updated to be able to generate signatures for both BLS and ECDSA. Currently only the latter is supported. This can be done through extending the API or by adding a flag to the existing methods indicating the algorithm to use.

tmpolaczyk:
Close #1230
Close #1228

Basic support for BLS signatures in the node: data structures and signature manager methods. The BLS key is imported from the storage, or generated if it does not exist, but there is no way to manually import or change it.

See also: https://github.com/witnet/bls-signatures-rs/pull/7


data_structures/src/chain.rs
```diff
@@ -37,6 +37,9 @@ use crate::{
     vrf::{BlockEligibilityClaim, DataRequestEligibilityClaim},
 };
 use bech32::{FromBase32, ToBase32};
+use bls_signatures_rs::bn256;
+use bls_signatures_rs::bn256::Bn256;
+use bls_signatures_rs::MultiSignature;
```
lrubiorod, line 42: fmt

data_structures/src/chain.rs
```diff
@@ -1042,15 +1036,72 @@ pub struct ExtendedSecretKey {
...
 }
 
-// FIXME: ExtendedSecretKey shouldn't implement Default
-impl Default for ExtendedSecretKey {
-    fn default() -> Self {
+/// BLS data structures
+
+#[derive(Debug, Eq, PartialEq, Clone, Serialize, Deserialize, ProtobufConvert)]
+#[protobuf_convert(pb = "witnet::BlsPublicKey")]
+pub struct BlsPublicKey {
```
girazoki, line 1043: Can we name it somehow with the curve we are referring to?

tmpolaczyk: Something like `Bn256PublicKey`?

node/src/signature_mngr.rs
```diff
@@ -51,6 +53,14 @@ pub fn sign_data(data: [u8; 32]) -> impl Future<Item = KeyedSignature, Error = f
     addr.send(Sign(data.to_vec())).flatten()
 }
 
+/// Sign a piece of (Hashable) data with the stored key.
+///
+/// This might fail if the manager has not been initialized with a key
+pub fn bls_sign(message: Vec<u8>) -> impl Future<Item = BlsKeyedSignature, Error = failure::Error> {
+    let addr = SignatureManagerAdapter::from_registry();
+    addr.send(BlsSign(message)).flatten()
+}
```
aesedepece, line 62: Why `Vec<u8>`? It makes somehow hard to guess whether I should hash the bytes in advance or not.

Maybe we should make it uniform with how we deal with ECDSA: having separate `fn([u8; 32])` and `fn<T>(&T) where T: Hashable` functions.

tmpolaczyk: BLS signatures are implemented by hashing the message concatenated with a counter, like `H(message || counter)`. So if we use the hash instead of the message as the input, it would be hashing twice: `H(H(message) || counter)`. ECDSA works differently and can accept the message hash as input. So we can make the API uniform with ECDSA, but we will be hashing twice.

data_structures/src/chain.rs
girazoki: works

data_structures/Cargo.toml
```diff
@@ -7,11 +7,12 @@ version = "0.3.2"
 workspace = ".."
 
 [dependencies]
+bls-signatures-rs = { git = "https://github.com/witnet/bls-signatures-rs", rev = "beab4942fc0fc91ff57b1d579d6dbfa1e4c590c3" }
```
mariocao, line 10: Maybe we could publish it to crates.io :)

node/src/signature_mngr.rs
```diff
@@ -67,6 +78,14 @@ pub fn public_key() -> impl Future<Item = PublicKey, Error = failure::Error> {
     addr.send(GetPublicKey).flatten()
 }
 
+/// Get the public key.
```
mariocao, line 81:
```suggestion
/// Get the BLS public key.
```

node/src/signature_mngr.rs
```diff
@@ -151,6 +189,36 @@ fn create_master_key() -> Box<dyn Future<Item = ExtendedSK, Error = failure::Err
     }
 }
 
+fn create_bls_keys() -> (Bn256SecretKey, Bn256PublicKey) {
+    let mut rng = thread_rng();
+    // Loop: if something fails, generate a new secret key and try again
```
aesedepece, line 194: Because generating keys is free.

PR #1234, opened 2020-05-04
log(mining): show message when your block is mined in chain or you ga…

lrubiorod:
…in or lose reputation

### DIFF
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -2393,6 +2393,25 @@ pub fn get_utxo_info(
 
 pub type Blockchain = BTreeMap<Epoch, Hash>;
 
+/// Node stats
+#[derive(Debug, Default, Clone, Serialize, Deserialize, PartialEq)]
+pub struct NodeStats {
+    /// Number of proposed blocks
+    pub block_proposed_count: u32,
+    /// Number of blocks included in the block chain
+    pub block_mined_count: u32,
+    /// Number of times we were eligible to participate in a Data Request
+    pub dr_eligibility_count: u32,
+    /// Number of proposed commits
+    pub commits_proposed_count: u32,
+    /// Number of commits included in a data request
+    pub commits_count: u32,
+    /// Last block proposed
+    pub last_block_proposed: Hash,
+    /// Number of slashed commits
+    pub out_of_consensus_count: u32,
+}
+
 /// Blockchain state (valid at a certain epoch)
 #[derive(Debug, Default, Clone, Serialize, Deserialize, PartialEq)]
 pub struct ChainState {
@@ -2409,6 +2428,8 @@ pub struct ChainState {
     pub own_utxos: OwnUnspentOutputsPool,
     /// Reputation engine
     pub reputation_engine: Option<ReputationEngine>,
+    /// Node mining stats
+    pub node_stats: NodeStats,
     /// Alternative public key mapping
     pub alt_keys: AltKeys,
 }
--- a/node/src/actors/chain_manager/mod.rs
+++ b/node/src/actors/chain_manager/mod.rs
@@ -62,8 +62,8 @@ use witnet_data_structures::{
     chain::{
         penalize_factor, reputation_issuance, Alpha, AltKeys, Block, Bn256PublicKey, ChainState,
         CheckpointBeacon, CheckpointVRF, ConsensusConstants, DataRequestReport, Epoch,
-        EpochConstants, Hash, Hashable, InventoryItem, OwnUnspentOutputsPool, PublicKeyHash,
-        Reputation, ReputationEngine, TransactionsPool, UnspentOutputsPool,
+        EpochConstants, Hash, Hashable, InventoryItem, NodeStats, OwnUnspentOutputsPool,
+        PublicKeyHash, Reputation, ReputationEngine, TransactionsPool, UnspentOutputsPool,
     },
     data_request::DataRequestPool,
     radon_report::{RadonReport, ReportContext},
@@ -466,6 +466,7 @@ impl ChainManager {
                     self.own_pkh,
                     &mut self.chain_state.own_utxos,
                     epoch_constants,
+                    &mut self.chain_state.node_stats,
                 );
 
                 let miner_pkh = block.block_header.proof.proof.pkh();
@@ -480,6 +481,7 @@ impl ChainManager {
                         rep_info,
                         log_level,
                         block_epoch,
+                        self.own_pkh.unwrap_or_default(),
                     );
                 }
 
@@ -527,6 +529,11 @@ impl ChainManager {
                             .chain_state
                             .data_request_pool
                             .update_data_request_stages();
+
+                        if block_hash == self.chain_state.node_stats.last_block_proposed {
+                            self.chain_state.node_stats.block_mined_count += 1;
+                            log::info!("Congratulations! Your block was consolidated into the block chain by an apparent majority of peers");
+                        }
 
                         show_info_dr(&self.chain_state.data_request_pool, &block);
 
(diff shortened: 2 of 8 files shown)


node/src/actors/chain_manager/mod.rs
```diff
@@ -987,6 +995,13 @@ fn update_reputation(
     // Penalize liars and accumulate the reputation
     // The penalization depends on the number of lies from the last epoch
     let liars_and_penalize_function = liars.iter().map(|(pkh, num_lies)| {
+        if own_pkh == *pkh {
+            log::warn!(
```
tmpolaczyk, line 999:
```suggestion
            log::info!(
```

I mean, losing reputation is bad but it should be logged at the same level as gaining reputation.

node/src/actors/chain_manager/mod.rs
```diff
@@ -1115,6 +1138,13 @@ fn show_info_dr(data_request_pool: &DataRequestPool, block: &Block) {
             )
         });
 
+    if Some(block_hash) == mined_block {
+        log::info!(
+            "Congratulations! Your block {} was consolidated in chain",
+            block_hash
```
tmpolaczyk, line 1144: The block hash is redundant because it will be printed in the log immediately below
```suggestion
```

tmpolaczyk: What do you think about printing the number of mined blocks so far? Something like "your node has mined {} blocks so far".

aesedepece: Will that be like the 40th entry in `struct ChainManager`? :stuck_out_tongue:

I think it's cool as long as it's in a `struct Stats` or similar where we can eventually throw any further metrics that we want to track in the future.

bertux: Having a struct Strats would help for the Prometheus instrumentation of metrics I'm studying.

aesedepece: @witnet/devs yet this will be a problem with bincode because we can't add new fields.

node/src/actors/chain_manager/mod.rs
```diff
@@ -987,6 +995,13 @@ fn update_reputation(
     // Penalize liars and accumulate the reputation
     // The penalization depends on the number of lies from the last epoch
     let liars_and_penalize_function = liars.iter().map(|(pkh, num_lies)| {
+        if own_pkh == *pkh {
+            log::warn!(
+                "Your reputation has been slashed. Your lies number was {}",
```
aesedepece, line 1000:
```suggestion
                "Your reputation score has been slashed because you witnessed {} times out of consensus",
```

aesedepece: Can we know how much did we lost?

node/src/actors/chain_manager/mod.rs
```diff
@@ -1115,6 +1138,13 @@ fn show_info_dr(data_request_pool: &DataRequestPool, block: &Block) {
             )
         });
 
+    if Some(block_hash) == mined_block {
+        log::info!(
+            "Congratulations! Your block {} was consolidated in chain",
```
aesedepece, line 1143:
```suggestion
            "Congratulations! Your block {} was consolidated into the block chain by an apparent majority of peers",
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -167,6 +167,8 @@ pub struct ChainManager {
     tx_pending_timeout: u64,
     /// Magic number from ConsensusConstants
     magic: u16,
+    /// Last block mined
+    mined_block: Option<Hash>,
```
aesedepece, line 171:
```suggestion
    /// Hash of the last block mined by this node since last start
    last_mined_block: Option<Hash>,
```

node/src/actors/chain_manager/mod.rs
mariocao: I agree, also to be consistent with the case when reputation is gained.

node/src/actors/chain_manager/mod.rs
lrubiorod: done!!

src/cli/node/json_rpc_client.rs
```diff
@@ -817,6 +817,32 @@ pub fn get_known_peers(addr: SocketAddr) -> Result<(), failure::Error> {
...
+    let node_stats: NodeStats = parse_response(&response)?;
+
+    println!(
+        "Block mining stats:\n\
+     - Proposed blocks: {}\n\
+     - Blocks included in the block chain: {}\n\
+    Data Request mining stats:\n\
+     - Data Requests with eligibility: {}\n\
+     - Commits proposed: {}\n\
+     - Commits included in a Data Request: {}\n\
```
aesedepece, line 833:
```suggestion
     - Proposed commits: {}\n\
     - Accepted commits: {}\n\
```

src/cli/node/json_rpc_client.rs
```diff
@@ -817,6 +817,32 @@ pub fn get_known_peers(addr: SocketAddr) -> Result<(), failure::Error> {
...
+     - Slashed commits:{}",
```
aesedepece, line 834:
```suggestion
     - Slashed commits: {}",
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -991,9 +1014,22 @@ fn update_reputation(
...
     // The penalization depends on the number of lies from the last epoch
     let liars_and_penalize_function = liars.iter().map(|(pkh, num_lies)| {
+        if own_pkh == *pkh {
+            let after_slashed_rep = f64::from(own_rep.0)
+                * consensus_constants
+                    .reputation_penalization_factor
+                    .powf(f64::from(*num_lies));
+            let slasehd_rep = own_rep.0 - (after_slashed_rep as u32);
+            log::info!(
+                "Your reputation score has slashed by {} points",
```
aesedepece, line 1028:
```suggestion
                "Your reputation score has been slashed by {} points",
```

aesedepece:
Can't we log expiration too?

lrubiorod:
> Can't we log expiration too?

It is more difficult because reputation expires in batches

node/src/actors/chain_manager/mod.rs
```diff
@@ -991,9 +1014,22 @@ fn update_reputation(
...
+    let own_rep = rep_eng.trs().get(&own_pkh);
     // Penalize liars and accumulate the reputation
     // The penalization depends on the number of lies from the last epoch
     let liars_and_penalize_function = liars.iter().map(|(pkh, num_lies)| {
+        if own_pkh == *pkh {
+            let after_slashed_rep = f64::from(own_rep.0)
+                * consensus_constants
+                    .reputation_penalization_factor
+                    .powf(f64::from(*num_lies));
+            let slasehd_rep = own_rep.0 - (after_slashed_rep as u32);
```
tmpolaczyk, line 1026:
```suggestion
            let slashed_rep = own_rep.0 - (after_slashed_rep as u32);
```

aesedepece: image

data_structures/src/chain.rs
```diff
@@ -2393,6 +2393,25 @@ pub fn get_utxo_info(
...
 
+/// Node stats
+#[derive(Debug, Default, Clone, Serialize, Deserialize, PartialEq)]
+pub struct NodeStats {
+    /// Number of proposed blocks
+    pub block_proposed_count: u32,
+    /// Number of blocks included in the block chain
+    pub block_mined_count: u32,
+    /// Data Requests with elegibility
+    pub dr_with_eligibility_count: u32,
```
tmpolaczyk, line 2404: This is misleading the name suggests that this is the number of different data requests that we could have participated in, but if we propose 2 commits in 2 different rounds in the same data requests it counts as 2 data requests. So it should be renamed to something like "number of times we were eligible to participate in a data request".

Issue #1236, opened 2020-05-06
Superblock signing and broadcasting


lrubiorod:
Sign the superblock with BLS and broadcasting it

PR #1237, opened 2020-05-06
Create and mine superblock

### LINKED ISSUE #1182: Superblock production
The goal is to generate a signature using the library in https://github.com/witnet/bls-signatures-rs and the scope defined in #1176.

This issue should include:

- Adding a new consensus constant indicating the super block proposal period.
- Everytime epoch % block proposal period = 0 is reached, the **ARS members** should construct a superblock taking the last *period* DR and tally roots, computing a merkle tree to construct superblock DR and tally roots. With those roots, the superblock index and the previous superblock hash a superblock is constructed.

lrubiorod:
Close #1182

### DIFF
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -544,6 +544,34 @@ pub struct BlockMerkleRoots {
     pub tally_hash_merkle_root: Hash,
 }
 
+/// `SuperBlock` abridges the tally and data request information that happened during a
+/// `superblock_period` number of Witnet epochs as well the ARS members merkle root
+/// as of the last block in that period.
+/// This is needed to ensure that the security and trustlessness properties of Witnet will
+/// be relayed to bridges with other block chains.
+#[derive(Debug, Eq, PartialEq, Clone, Serialize, Deserialize, ProtobufConvert, Default)]
+#[protobuf_convert(pb = "witnet::SuperBlock")]
+pub struct SuperBlock {
+    /// Merkle root of the Active Reputation Set members included into the previous SuperBlock
+    pub ars_root: Hash,
+    /// Merkle root of the data requests in the blocks created since the last SuperBlock
+    pub data_request_root: Hash,
+    /// Superblock index,
+    pub index: u32,
+    /// Hash of the block that this SuperBlock is attesting as the latest block in the block chain,
+    pub last_block: Hash,
+    /// Hash of the block that the previous SuperBlock used for its own `last_block` field,
+    pub last_block_in_previous_superblock: Hash,
+    /// Merkle root of the tallies in the blocks created since the last SuperBlock
+    pub tally_root: Hash,
+}
+
+impl Hashable for SuperBlock {
+    fn hash(&self) -> Hash {
+        calculate_sha256(&self.to_pb_bytes().unwrap()).into()
+    }
+}
+
 /// Digital signatures structure (based on supported cryptosystems)
 #[derive(Debug, Eq, PartialEq, Clone, Serialize, Deserialize, ProtobufConvert)]
 #[protobuf_convert(pb = "witnet::Signature")]
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -153,6 +153,9 @@ pub struct ConsensusConstants {
 
     /// Minimum input age of an UTXO for being a valid collateral
     pub collateral_age: u32,
+
+    /// Build a superblock every `superblock_period` epochs
+    pub superblock_period: u16,
 }
 
 impl ConsensusConstants {
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -830,6 +830,12 @@ impl fmt::Display for PublicKeyHash {
     }
 }
 
+impl Hashable for PublicKeyHash {
+    fn hash(&self) -> Hash {
+        calculate_sha256(&self.to_pb_bytes().unwrap()).into()
+    }
+}
+
 /// Error when parsing hash from string
 #[derive(Debug, Fail)]
 pub enum PublicKeyHashParseError {
--- a/node/src/actors/chain_manager/mining.rs
+++ b/node/src/actors/chain_manager/mining.rs
@@ -1,3 +1,9 @@
+use actix::{
+    ActorFuture, AsyncContext, Context, ContextFutureSpawner, Handler, SystemService, WrapFuture,
+};
+use ansi_term::Color::{White, Yellow};
+use futures::future::{join_all, Future};
+use itertools::Itertools;
 use std::{
     cmp::Ordering,
     collections::HashSet,
@@ -8,17 +14,11 @@ use std::{
     },
 };
 
-use actix::{
-    ActorFuture, AsyncContext, Context, ContextFutureSpawner, Handler, SystemService, WrapFuture,
-};
-use ansi_term::Color::{White, Yellow};
-use futures::future::{join_all, Future};
-
 use witnet_data_structures::{
     chain::{
         Block, BlockHeader, BlockMerkleRoots, BlockTransactions, Bn256PublicKey, CheckpointBeacon,
-        CheckpointVRF, EpochConstants, Hashable, OwnUnspentOutputsPool, PublicKeyHash,
-        ReputationEngine, TransactionsPool, UnspentOutputsPool, ValueTransferOutput,
+        CheckpointVRF, EpochConstants, Hash, Hashable, OwnUnspentOutputsPool, PublicKeyHash,
+        ReputationEngine, SuperBlock, TransactionsPool, UnspentOutputsPool, ValueTransferOutput,
     },
     data_request::{calculate_witness_reward, create_tally, DataRequestPool},
     error::TransactionError,
@@ -33,7 +33,7 @@ use witnet_rad::{error::RadError, types::serial_iter_decode};
 use witnet_util::timestamp::get_timestamp;
 use witnet_validations::validations::{
     block_reward, calculate_randpoe_threshold, calculate_reppoe_threshold, dr_transaction_fee,
-    merkle_tree_root, update_utxo_diff, vt_transaction_fee, UtxoDiff,
+    hash_merkle_tree_root, merkle_tree_root, update_utxo_diff, vt_transaction_fee, UtxoDiff,
 };
 
 use crate::{
@@ -42,7 +42,10 @@ use crate::{
             transaction_factory::{build_commit_collateral, sign_transaction},
             ChainManager, StateMachine,
         },
-        messages::{AddCandidates, AddCommitReveal, ResolveRA, RunTally},
+        inventory_manager::InventoryManager,
+        messages::{
+            AddCandidates, AddCommitReveal, GetBlocksEpochRange, GetItemBlock, ResolveRA, RunTally,
+        },
         rad_manager::RadManager,
     },
     signature_mngr,
@@ -93,7 +96,7 @@ impl ChainManager {
         let current_epoch = self.current_epoch.unwrap();
 
         let chain_info = self.chain_state.chain_info.as_ref().unwrap();
-
+        let genesis_hash = chain_info.consensus_constants.genesis_hash;
         let max_block_weight = chain_info.consensus_constants.max_block_weight;
 
         let mining_bf = chain_info.consensus_constants.mining_backup_factor;
@@ -125,6 +128,26 @@ impl ChainManager {
 
         let own_pkh = self.own_pkh.unwrap_or_default();
         let is_ars_member = rep_engine.is_ars_member(&own_pkh);
+
+        let superblock_period = u32::from(chain_info.consensus_constants.superblock_period);
+        if rep_engine.is_ars_member(&own_pkh) && current_epoch % superblock_period == 0 {
+            // FIXME(#1236): ARS Members have to use the BLS address when forwarding
+            // TODO: ARS Member addresses have to be selected from the last SuperBlock
+            let ars_members: Vec<PublicKeyHash> = rep_engine
+                .ars()
+                .active_identities()
+                .cloned()
+                .sorted()
+                .collect();
+
+            self.superblock_forwarding(
+                ctx,
+                current_epoch,
+                superblock_period,
+                ars_members,
+                genesis_hash,
+            );
+        }
 
         // Create a VRF proof and if eligible build block
         signature_mngr::vrf_prove(VrfMessage::block_mining(vrf_input))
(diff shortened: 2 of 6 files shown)


data_structures/src/chain.rs
```diff
@@ -537,6 +540,30 @@ pub struct BlockMerkleRoots {
...
+    pub tally_root: Hash,
+    /// Merkle root of the current Active Reputation Set members
+    pub ars_root: Hash,
+    /// Superblock index,
+    pub index: u32,
+    /// Superblock prev hash,
+    pub prev_block_hash_id: Hash,
+    /// Superblock current hash,
+    pub current_block_hash_id: Hash,
+}
```
aesedepece, line 559: Please order keys alphabetically.

data_structures/src/chain.rs
```diff
@@ -537,6 +540,30 @@ pub struct BlockMerkleRoots {
...
+    /// Merkle root of the last superblock period blocks tally merkles root
+    pub tally_root: Hash,
+    /// Merkle root of the current Active Reputation Set members
+    pub ars_root: Hash,
+    /// Superblock index,
+    pub index: u32,
+    /// Superblock prev hash,
+    pub prev_block_hash_id: Hash,
+    /// Superblock current hash,
+    pub current_block_hash_id: Hash,
```
aesedepece, line 558:
```suggestion
    /// Hash of the block that the previous superblock used for its own `last_block` field,
    pub last_block_in_previous_superblock: Hash,
    /// Hash of the block that this superblock is attesting as the latest block in the block chain,
    pub last_block: Hash,
```

node/src/actors/chain_manager/mining.rs
```diff
@@ -125,6 +128,25 @@ impl ChainManager {
 
         let own_pkh = self.own_pkh.unwrap_or_default();
 
+        let superblock_period = u32::from(chain_info.consensus_constants.superblock_period);
+        if rep_engine.ars().contains(&own_pkh) && current_epoch % superblock_period == 0 {
+            // FIXME(#1236): ARS Members has to use the BLS address when forwarding
```
aesedepece, line 133:
```suggestion
            // FIXME(#1236): ARS Members have to use the BLS address when forwarding
```

node/src/actors/chain_manager/mining.rs
```diff
@@ -125,6 +128,25 @@ impl ChainManager {
 
         let own_pkh = self.own_pkh.unwrap_or_default();
 
+        let superblock_period = u32::from(chain_info.consensus_constants.superblock_period);
+        if rep_engine.ars().contains(&own_pkh) && current_epoch % superblock_period == 0 {
```
aesedepece, line 132:
```suggestion
        if rep_engine.is_ars_member(&own_pkh) && current_epoch % superblock_period == 0 {
```

data_structures/src/chain.rs
```diff
@@ -795,6 +822,12 @@ impl fmt::Display for PublicKeyHash {
     }
 }
 
+impl Hashable for PublicKeyHash {
+    fn hash(&self) -> Hash {
+        calculate_sha256(&self.to_pb_bytes().unwrap()).into()
+    }
+}
+
```
mariocao, line 830: As discussed with @aesedepece , maybe we could use a blanket implementation in order to avoid code duplication:

```rust
impl<T> Hashable for T
    where T: ProtobufConvert
{
    fn hash(&self) -> Hash {
        calculate_sha256(&self.to_pb_bytes().unwrap()).into()
    }
}
```

data_structures/src/chain.rs
```diff
@@ -537,6 +540,30 @@ pub struct BlockMerkleRoots {
     pub tally_hash_merkle_root: Hash,
 }
 
+/// Block merkle tree roots
+#[derive(Debug, Eq, PartialEq, Clone, Serialize, Deserialize, ProtobufConvert, Default)]
+#[protobuf_convert(pb = "witnet::SuperBlock")]
+pub struct SuperBlock {
+    /// Merkle root of the last superblock period blocks data request merkles root
```
clbartoli, line 547: The description is not super clear to me, maybe something like
"Merkle root of the data requests in the blocks created since the last Superblock" ?

data_structures/src/chain.rs
lrubiorod: image

lrubiorod: I think it should be do it in a specific issue to fix that, dont you think that?

node/src/actors/chain_manager/mining.rs
```diff
@@ -125,6 +128,25 @@ impl ChainManager {
...
+            // FIXME(#1236): ARS Members have to use the BLS address when forwarding
+            let ars_members: Vec<PublicKeyHash> = rep_engine
```
tmpolaczyk, line 134: Shouldn't this be the ARS members from 2 epochs ago?

lrubiorod: It will be modified after Reputation Merkelitation improvement

lrubiorod: I added a TODO

data_structures/src/chain.rs
```diff
@@ -538,6 +541,30 @@ pub struct BlockMerkleRoots {
     pub tally_hash_merkle_root: Hash,
 }
 
+/// Block merkle tree roots
+#[derive(Debug, Eq, PartialEq, Clone, Serialize, Deserialize, ProtobufConvert, Default)]
+#[protobuf_convert(pb = "witnet::SuperBlock")]
+pub struct SuperBlock {
+    /// Merkle root of the current Active Reputation Set members
```
tmpolaczyk, line 548: So here "current" means "after consolidating the block with hash `last_block`"?

data_structures/src/chain.rs
```diff
@@ -538,6 +541,30 @@ pub struct BlockMerkleRoots {
     pub tally_hash_merkle_root: Hash,
 }
 
+/// Block merkle tree roots
```
tmpolaczyk, line 544: Explain what it is, why it is needed, and maybe how it's created.

node/src/actors/chain_manager/mining.rs
```diff
@@ -836,6 +953,34 @@ fn build_block(
     (block_header, txns)
 }
 
+fn build_superblock(
```
tmpolaczyk, line 956: Should this be in `impl SuperBlock`

node/src/actors/chain_manager/mining.rs
```diff
@@ -836,6 +953,34 @@ fn build_block(
...
+    block_headers: &[BlockHeader],
+    ars_identities: &[PublicKeyHash],
```
tmpolaczyk, line 958: Maybe add a comment explaining that these are sorted by pkh.

```suggestion
    sorted_ars_identities: &[PublicKeyHash],
```

node/src/actors/chain_manager/mining.rs
```diff
@@ -836,6 +953,34 @@ fn build_block(
...
+    index: u32,
+    previous_hash: Hash,
```
tmpolaczyk, line 960: I like the new name, and a comment or an example would be nice

```suggestion
    last_block_in_previous_superblock: Hash,
```

schemas/witnet/witnet.proto
```diff
@@ -311,3 +312,12 @@ message DataRequestVrfMessage {
 message DataRequestEligibilityClaim {
     VrfProof proof = 1;
 }
+
+message SuperBlock {
+    Hash ars_root = 1;
+    Hash data_request_root = 2;
+    uint32  index = 3;
```
tmpolaczyk, line 319:
```suggestion
    uint32 index = 3;
```

data_structures/src/chain.rs
```diff
@@ -153,6 +153,9 @@ pub struct ConsensusConstants {
 
     /// Minimum input age of an UTXO for being a valid collateral
     pub collateral_age: u32,
+
+    /// Superblock period
```
tmpolaczyk, line 157:
```suggestion
    /// Build a superblock every `superblock_period` epochs
```

data_structures/src/chain.rs
aesedepece: True, it would now be something like the Active Reputation Set that got anchored into the previous superblock

node/src/actors/chain_manager/mining.rs
lrubiorod: now are sorted by pkh, but in the future it will be sorted by bls signatures, so... maybe only with the variable name could be enough

node/src/actors/chain_manager/mining.rs
lrubiorod: now it is similar to build block, a private function in mining. Moreover, moving that function to data_structures creates problems with cyclic dependencies that maybe could solved in another issue

data_structures/src/chain.rs
```diff
@@ -541,6 +544,33 @@ pub struct BlockMerkleRoots {
     pub tally_hash_merkle_root: Hash,
 }
 
+/// SuperBlock is a struct that resume the tally and the data request information that happened
+/// during a superblock_period and the ARS member merkle root. That struct is required for ensure
+/// that the security and trustless properties of Witnet will be relayed to the bridge with other
+/// block chains
```
aesedepece, line 550:
```suggestion
/// `SuperBlock` abridges the tally and data request information that happened during a
/// `superblock_period` number of Witnet epochs as well the ARS members merkle root
/// as of the last block in that period.
/// This is needed to ensure that the security and trustlessness properties of Witnet will
/// be relayed to bridges with other block chains.
```

node/src/actors/chain_manager/mining.rs
```diff
@@ -1276,6 +1423,127 @@ mod tests {
         assert!(verify(secp, &public_key2, &data, &signature2).is_ok());
     }
 
+    #[test]
+    fn test_superblock_creation_no_blocks() {
+        let default_hash = Hash::default();
+        let superblock = build_superblock(&[], &[], 0, default_hash);
+        assert!(superblock.is_none(), true);
```
tmpolaczyk, line 1430:
```suggestion
        assert_eq!(superblock, None);
```

data_structures/src/chain.rs
mariocao: no prob, it was just a suggestion! :)

node/src/actors/chain_manager/mining.rs
```diff
@@ -126,6 +129,26 @@ impl ChainManager {
         let own_pkh = self.own_pkh.unwrap_or_default();
         let is_ars_member = rep_engine.is_ars_member(&own_pkh);
 
+        let superblock_period = u32::from(chain_info.consensus_constants.superblock_period);
+        if rep_engine.is_ars_member(&own_pkh) && current_epoch % superblock_period == 0 {
+            // FIXME(#1236): ARS Members have to use the BLS address when forwarding
+            // FIXME: ARS Member addresses have to be selected from the last SuperBlock
```
aesedepece, line 135: Actually it is from the last block in the superblock, isn't it?

node/src/actors/chain_manager/mining.rs
```diff
@@ -538,6 +561,100 @@ impl ChainManager {
...
+            GetBlocksEpochRange::new_with_limit(init_epoch..=final_epoch, 0),
+            ctx,
+        ))
+        .and_then(move |res| match res {
+            Ok(v) => {
+                let block_hashes: Vec<Hash> = v.into_iter().map(|(_epoch, hash)| hash).collect();
+                futures::future::ok(block_hashes)
+            }
+            Err(e) => {
+                log::error!("{}", e);
```
tmpolaczyk, line 591: Add some context to this errors?
```suggestion
                log::error!("Error in GetBlocksEpochRange: {}", e);
```

node/src/actors/chain_manager/mining.rs
```diff
@@ -126,6 +129,29 @@ impl ChainManager {
         let own_pkh = self.own_pkh.unwrap_or_default();
         let is_ars_member = rep_engine.is_ars_member(&own_pkh);
 
+        let superblock_period = u32::from(chain_info.consensus_constants.superblock_period);
+        // FIXME: Only ARS members from the last block of the previous SuperBlock can create a SuperBlock
+        if rep_engine.is_ars_member(&own_pkh) && current_epoch % superblock_period == 0 {
+            // FIXME(#1236): ARS Members have to include the BLS signature instead of PublicKeyHash
+            // FIXME: After Reputation Merkelitation, only the ARS Members from the previous block
+            // consolidated will be added, to avoid include addresses that could be wrong later by
```
mariocao, line 137:
```suggestion
            // FIXME: After Reputation Merkelitation, only the ARS Members from the previous consolidated
            // block will be added, to avoid include addresses that could be wrong later by
```

node/src/actors/chain_manager/mining.rs
```diff
@@ -863,6 +983,36 @@ fn build_block(
     (block_header, txns)
 }
 
+/// Function that return an Option of a SuperBlock or None in case if there are not blocks
+/// to produce a SuperBlock
```
aesedepece, line 987: This doc comment is telling me no more than the function signature. And in fact, `None` is still an `Option` of a `SuperBlock` :stuck_out_tongue:  

```suggestion
/// Produces a `SuperBlock` that includes the blocks in `block_headers` if there is at least one of them.
```

node/src/actors/chain_manager/mining.rs
```diff
@@ -863,6 +983,36 @@ fn build_block(
...
+    let merkle_drs: Vec<Hash> = block_headers
+        .iter()
+        .map(|b| b.merkle_roots.dr_hash_merkle_root)
+        .collect();
+    let merkle_tallies: Vec<Hash> = block_headers
+        .iter()
+        .map(|b| b.merkle_roots.tally_hash_merkle_root)
+        .collect();
+
+    let pkh_hashes: Vec<Hash> = sorted_ars_identities.iter().map(|pkh| pkh.hash()).collect();
```
aesedepece, line 1004: Hashes of the public key hashes? :open_mouth:

lrubiorod: currently our pkh is not a really hash, it is only the first part so... hashing the hashes :P 
The other solution should be to have a real PublicKeyHash, why do we reduce them before?

lrubiorod: Anyway, now we are creating a merkleroot with our pkh, but I remeber that it would be with the BLS addresses, that it will be done in the issue #1236

aesedepece: It's OK for the time being.

And yes, we should stop trimming PKHs asap.

PR #1238, opened 2020-05-07
log(reputation): fix reputation logging in Live state

lrubiorod:


node/src/actors/chain_manager/mod.rs
```diff
@@ -445,7 +445,7 @@ impl ChainManager {
 
                 // Print reputation logs on debug level on synced state,
```
mariocao, line 446:
```suggestion
                // Print reputation logs on debug level on Live state,
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -1058,27 +1058,28 @@ fn update_reputation(
     {
         log::error!("Error updating reputation in consolidation: {}", e);
     }
+    if log_level != log::Level::Trace {
```
tmpolaczyk, line 1061: We can also skip this if the current log level is INFO or ERROR, so try doing:
```suggestion
    if log_level >= log::max_level() {
```

tmpolaczyk: Even better:
```suggestion
    if log_enabled!(log_level) {
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -1058,27 +1058,28 @@ fn update_reputation(
     {
         log::error!("Error updating reputation in consolidation: {}", e);
     }
+    if log::log_enabled!(log_level) {
```
aesedepece, line 1061: Maybe this:
```suggestion

    if log::log_enabled!(log_level) {
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -1069,26 +1069,28 @@ fn update_reputation(
...
-            ' '
-        };
-        log::log!(log_level, "    [{}] {}: {}", active, pkh, rep.0);
+        log::log!(log_level, "Total Reputation: {{");
+        for (pkh, rep) in rep_eng
+            .trs()
+            .identities()
+            .sorted_by_key(|&(_, &r)| std::cmp::Reverse(r))
+        {
+            let active = if rep_eng.ars().contains(pkh) {
```
aesedepece, line 1085:
```suggestion
            let active = if rep_eng.is_ars_member(pkh) {
```

PR #1239, opened 2020-05-07
fix(wallet): fix get addresses offset and limit

mariocao:


wallet/src/actors/app/handlers/get_addresses.rs
```diff
@@ -27,11 +25,9 @@ impl Handler<GetAddressesRequest> for app::App {
         let offset = msg
             .offset
             .unwrap_or_else(|| constants::DEFAULT_PAGINATION_OFFSET);
-        let limit = cmp::min(
-            msg.offset
-                .unwrap_or_else(|| constants::DEFAULT_PAGINATION_LIMIT),
-            constants::MAX_PAGINATION_LIMIT,
```
tmpolaczyk, line 33: Why remove the limit?

wallet/src/actors/app/handlers/get_addresses.rs
```diff
@@ -28,7 +28,7 @@ impl Handler<GetAddressesRequest> for app::App {
             .offset
             .unwrap_or_else(|| constants::DEFAULT_PAGINATION_OFFSET);
         let limit = cmp::min(
-            msg.offset
+            msg.limit
                 .unwrap_or_else(|| constants::DEFAULT_PAGINATION_LIMIT),
             constants::MAX_PAGINATION_LIMIT,
```
tmpolaczyk, line 33: Why keep the limit? :smiley:

mariocao: To use a similar logic as in `get_transactions` :laughing:

mariocao: I removed it along with the one in `get_transactions`

PR #1242, opened 2020-05-12
feat(data_structures): insert block number from input when pkh would …

lrubiorod:
…be the same than output

This PR allows that new ValueTransferOutput keeps the same block inclusion number than their inputs in case that they have the same pkh

Modify pending timeout for commits due to CommitTransactions are only valid in the same Epoch, and you dont need to block that utxos more than one epoch

Remove smart mint splitting and changed by allowing to include a new address (different from node's address) in the MintTransaction, splitting the reward between node and that address

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -38,15 +38,14 @@
 //! // Default config for mainnet
 //! // Config::from_partial(&PartialConfig::default_mainnet());
 //! ```
-use std::collections::HashSet;
-use std::net::SocketAddr;
-use std::path::PathBuf;
-use std::time::Duration;
+use std::{collections::HashSet, net::SocketAddr, path::PathBuf, time::Duration};
 
 use serde::{Deserialize, Deserializer, Serialize};
 
-use crate::defaults::{Defaults, Testnet};
-use crate::dirs;
+use crate::{
+    defaults::{Defaults, Testnet},
+    dirs,
+};
 use partial_struct::PartialStruct;
 use witnet_crypto::hash::HashFunction;
 use witnet_data_structures::chain::{ConsensusConstants, Environment, PartialConsensusConstants};
@@ -326,8 +325,12 @@ pub struct Mining {
     pub data_request_timeout: Duration,
     /// Genesis block path
     pub genesis_path: String,
-    /// Binary flag to create a mint with a split reward
-    pub split_mint: bool,
+    /// Percentage to redistribute mint reward in another address
+    pub mint_external_percentage: u8,
+    /// Address where redistribute mint reward
+    #[partial_struct(skip)]
+    #[partial_struct(serde(default))]
+    pub mint_external_address: Option<String>,
 }
 
 /// NTP-related configuration
@@ -598,10 +601,11 @@ impl Mining {
                 .genesis_path
                 .clone()
                 .unwrap_or_else(|| defaults.mining_genesis_path()),
-            split_mint: config
-                .split_mint
+            mint_external_percentage: config
+                .mint_external_percentage
                 .to_owned()
-                .unwrap_or_else(|| defaults.mining_split_mint()),
+                .unwrap_or_else(|| defaults.mining_mint_external_percentage()),
+            mint_external_address: config.mint_external_address.clone(),
         }
     }
 }
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -2104,15 +2104,11 @@ pub enum DataRequestStage {
 /// Unspent Outputs Pool
 #[derive(Clone, Debug, Default, Eq, PartialEq, Serialize, Deserialize)]
 pub struct UnspentOutputsPool {
-    /// Map of output pointer to value transfer output
-    map: HashMap<OutputPointer, ValueTransferOutput>,
-    /// Map of transaction hash to a tuple of:
+    /// Map of output pointer to a tuple of:
+    /// * Value transfer output
     /// * The number of the block that included the transaction
     ///   (how many blocks were consolidated before this one).
-    /// * A reference count, used to keep this map clear after removing transactions.
-    ///   This reference count is the number of output pointers that point to this
-    ///   transaction hash, when it reaches 0 the entry should be removed.
-    transaction_block_number: HashMap<Hash, (u32, u32)>,
+    map: HashMap<OutputPointer, (ValueTransferOutput, u32)>,
 }
 
 impl UnspentOutputsPool {
@@ -2121,7 +2117,7 @@ impl UnspentOutputsPool {
         OutputPointer: std::borrow::Borrow<Q>,
         Q: std::hash::Hash + Eq,
     {
-        self.map.get(k)
+        self.map.get(k).map(|(vt, _n)| vt)
     }
 
     pub fn contains_key<Q: ?Sized>(&self, k: &Q) -> bool
@@ -2137,64 +2133,31 @@ impl UnspentOutputsPool {
         k: OutputPointer,
         v: ValueTransferOutput,
         block_number: u32,
-    ) -> Option<ValueTransferOutput> {
-        let transaction_id = k.transaction_id;
-        let old_vto = self.map.insert(k, v);
-        if old_vto.is_none() {
-            // Store block number for this transaction hash
-            // If the transaction hash already exists, do not update the block number
-            let (_block_number, refcount) = self
-                .transaction_block_number
-                .entry(transaction_id)
-                .or_insert((block_number, 0));
-            // Increase the refcount if this is a new output pointer
-            // Entries should be removed from the map when the refcount reaches 0 again
-            *refcount += 1;
-        }
-
-        old_vto
+    ) -> Option<(ValueTransferOutput, u32)> {
+        self.map.insert(k, (v, block_number))
     }
 
-    pub fn remove(&mut self, k: &OutputPointer) -> Option<ValueTransferOutput> {
-        let vto = self.map.remove(k);
-
-        if vto.is_some() {
-            // Decrease refcount if this is a new output pointer
-            let (_epoch, refcount) = self
-                .transaction_block_number
-                .get_mut(&k.transaction_id)
-                .unwrap();
-            *refcount -= 1;
-            // Entries should be removed from the map when the refcount reaches 0
-            if *refcount == 0 {
-                self.transaction_block_number.remove(&k.transaction_id);
-            }
-        }
-
-        vto
+    pub fn remove(&mut self, k: &OutputPointer) -> Option<(ValueTransferOutput, u32)> {
+        self.map.remove(k)
     }
 
     pub fn drain(
         &mut self,
-    ) -> std::collections::hash_map::Drain<OutputPointer, ValueTransferOutput> {
+    ) -> std::collections::hash_map::Drain<OutputPointer, (ValueTransferOutput, u32)> {
         self.map.drain()
     }
 
-    pub fn iter(&self) -> std::collections::hash_map::Iter<OutputPointer, ValueTransferOutput> {
+    pub fn iter(
+        &self,
+    ) -> std::collections::hash_map::Iter<OutputPointer, (ValueTransferOutput, u32)> {
         self.map.iter()
     }
 
     /// Returns the number of the block that included the transaction referenced
     /// by this OutputPointer. The difference between that number and the
-    /// current number of consolidated blocks is the "coin age".
+    /// current number of consolidated blocks is the "collateral age".
     pub fn included_in_block_number(&self, k: &OutputPointer) -> Option<Epoch> {
-        if self.map.contains_key(k) {
-            self.transaction_block_number
-                .get(&k.transaction_id)
-                .map(|(epoch, _refcount)| *epoch)
-        } else {
-            None
-        }
+        self.map.get(k).map(|(_vt, n)| *n)
     }
 }
 
@@ -2385,7 +2348,7 @@ pub fn get_utxo_info(
     let utxos = if let Some(pkh) = pkh {
         all_utxos
             .iter()
-            .filter_map(|(o, vto)| {
+            .filter_map(|(o, (vto, _))| {
                 if vto.pkh == pkh {
                     Some(create_utxo_metadata(
                         vto,
@@ -3740,8 +3703,8 @@ mod tests {
 
     #[test]
     fn utxo_set_insert_same_transaction_different_epoch() {
-        // Inserting the same transaction twice with different block number keeps
-        // the old block number but the new transaction
+        // Inserting the same transaction twice with different indexes means a different UTXO
+        // so, each UTXO keeps their own block number
         let mut p = UnspentOutputsPool::default();
         let v = || ValueTransferOutput::default();
 
@@ -3757,7 +3720,7 @@ mod tests {
                 .unwrap();
 
         p.insert(k1.clone(), v(), 1);
-        assert_eq!(p.included_in_block_number(&k1), Some(0));
+        assert_eq!(p.included_in_block_number(&k1), Some(1));
     }
 
     #[test]
--- a/data_structures/src/transaction.rs
+++ b/data_structures/src/transaction.rs
@@ -479,49 +479,53 @@ impl MintTransaction {
         }
     }
 
-    /// Try to create a MintTransaction with number of outputs equal to `utxos_required`,
-    /// where every one of those outputs must be equal to or greater than `collateral_minimum`.
-    /// If the reward is too small, the number of outputs may be smaller than `utxos_required`.
-    /// The output value will only be smaller than `collateral_minimum` if the `reward` is smaller
-    /// than `collateral_minimum`, in which case this function will create a MintTransaction with
-    /// exactly one output
-    pub fn with_split_utxos(
+    /// This method creates a MintTransaction with a reward split between the node and an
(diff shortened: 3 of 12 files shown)


config/src/config.rs
```diff
@@ -324,8 +323,12 @@ pub struct Mining {
     pub data_request_timeout: Duration,
     /// Genesis block path
     pub genesis_path: String,
-    /// Binary flag to create a mint with a split reward
-    pub split_mint: bool,
+    /// Percentage to redistribute mint reward in another address
```
tmpolaczyk, line 326: I would prefer to not implement this feature yet, open a different PR for that

aesedepece: +1

data_structures/src/chain.rs
```diff
@@ -2137,64 +2133,31 @@ impl UnspentOutputsPool {
...
+    pub fn iter(
+        &self,
+    ) -> std::collections::hash_map::Iter<OutputPointer, (ValueTransferOutput, u32)> {
         self.map.iter()
     }
 
     /// Returns the number of the block that included the transaction referenced
     /// by this OutputPointer. The difference between that number and the
-    /// current number of consolidated blocks is the "coin age".
+    /// current number of consolidated blocks is the "collateral age".
```
aesedepece, line 2158: THANKS FOR THIS :pray:

node/src/actors/chain_manager/mining.rs
```diff
@@ -443,13 +440,19 @@ impl ChainManager {
...
                         None => {
                             log::error!("ChainInfo is None");
                             return actix::fut::err(());
                         },
                     };
+
+                    // The timestamp included when using collateral is increased in (timeout - checkpoint_period)
+                    // to ensure that if your commit has not been accepted you can use your utxo in the next epoch
+                    // or at least in two epochs
+                    let ts_for_collateral = timestamp + tx_pending_timeout - u64::from(checkpoint_period) ;
```
aesedepece, line 454: Is there any way to test this?

witnet.toml
```diff
@@ -68,11 +68,11 @@ enabled = true
...
-# Enables or disables automatically splitting block mining rewards.
-# If set to `true`, the rewards will be split into multiple outputs of the same value as `collateral_minimum`
-# until the node's set of outputs available for collateralization outnumbers `data_request_max_retrievals_per_epoch`
-# multiplied by `collateral_coinage`, i.e. makes sure it always has spare outputs to collateralize.
-split_mint = true
+# 'mint_pkh' and 'mint_percentage' fields allows to the node operator splitting the mint reward
+# between the node address and another address, 'mint_percentage' indicates how much percentage
+# is provided to the 'mint_pkh'
+#mint_pkh = "twit1jqgf4rxjrgas3kdhj3t4cr3mg3n33m8zw0aulr"
+#mint_percentage = 50
```
aesedepece, line 75: I think `mint_percentage` should be explicit about the fact that we are choosing the share for the external address and not the address of the node.
```suggestion
#mint_external_percentage = 50
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -169,6 +167,10 @@ pub struct ChainManager {
     tx_pending_timeout: u64,
     /// Magic number from ConsensusConstants
     magic: u16,
+    /// External mint address
+    external_address: Option<PublicKeyHash>,
+    /// Mint Percentage to share with the external address
+    external_percentage: u8,
```
aesedepece, line 173: Are we locally ensuring that `external_percentage > 0` and `external_percentage <= 100`?

That's totally something that is worth being validated at start time, otherwise any blocks produced would be deemed invalid by other peers.

node/src/actors/chain_manager/mining.rs
```diff
@@ -916,37 +913,31 @@ fn build_block(
...
-            vec![ValueTransferOutput {
-                pkh: own_pkh,
-                value: reward,
+    let mut vts = vec![];
+    if let Some(external_pkh) = external_address {
+        // In case of a specified pkh, the reward will be redistributed between the node's pkh
+        // and the specified one using the external_percentage value
+        let external_reward = (reward * u64::from(external_percentage)) / 100;
+        if external_reward > 0 {
+            reward -= external_reward;
```
aesedepece, line 925: Without ensuring that `external_percentage <= 100`, this can easily underflow or lead to `reward = 0`.

config/src/config.rs
lrubiorod: We discussed before that it could be reviewed together

node/src/actors/chain_manager/mining.rs
```diff
@@ -916,37 +913,31 @@ fn build_block(
...
+                pkh: external_pkh,
+                value: external_reward,
                 time_lock: 0,
-            }],
-        )
-    };
+            })
+        }
+    }
+    if reward > 0 {
```
aesedepece, line 933:
```suggestion

   // If `external_percentage` is `100`, the external address will receive the entire
   // reward, and the output assigning tokens to the node is not needed.
    if reward > 0 {
```

node/src/actors/chain_manager/mining.rs
```diff
@@ -916,37 +913,31 @@ fn build_block(
...
+            vts.push(ValueTransferOutput {
+                pkh: external_pkh,
+                value: external_reward,
                 time_lock: 0,
-            }],
-        )
-    };
+            })
+        }
+    }
```
aesedepece, line 932: This is a sort of weird positioning for this block, because unless I'm misreading, this will cause the classical self-address mint output to come second instead of first.

config/src/config.rs
aesedepece: ...as long as it's in different commits :rofl:

node/src/actors/chain_manager/mining.rs
lrubiorod: The problem is that 'try_mine_data_request' is a ChainManager's impl function, so now we dont do tests between actors. So the way to test it is... creating a local testnet and check it :P

node/src/actors/chain_manager/mod.rs
lrubiorod: if it is 0, nothing happens, but yes, it is needed to limit the external_percentage to 100

node/src/actors/chain_manager/mining.rs
lrubiorod: yes, it was for simplify the conditions, but could be changed

data_structures/src/chain.rs
```diff
@@ -3740,8 +3703,8 @@ mod tests {
 
     #[test]
     fn utxo_set_insert_same_transaction_different_epoch() {
-        // Inserting the same transaction twice with different block number keeps
-        // the old block number but the new transaction
+        // Inserting the same transaction twice with different index means a different UTXO
+        // so, each UTXO keeps their own block number
```
aesedepece, line 3707:
```suggestion
        // Inserting the same transaction twice with different indexes means a different UTXO
        // so, each UTXO keeps their own block number
```

node/src/actors/chain_manager/mining.rs
lrubiorod: Im trying and it is not working properly... u_u

lrubiorod: done!! At the end it was only change the timeout that we want, not the timestamp. And it could be tested

node/src/actors/chain_manager/actor.rs
```diff
@@ -79,8 +78,14 @@ impl ChainManager {
...
                 act.mining_enabled = config.mining.enabled;
 
-                // Flag to indicate if we want to split the mint transaction
-                act.split_mint = config.mining.split_mint;
+                // External address mint sharing parameters
+                act.external_address = if let Some(s) = &config.mining.mint_pkh {
+                    PublicKeyHash::from_str(s.as_str()).ok()
+                } else {
+                    None
+                };
```
aesedepece, line 86: I'd prefer not to destructure the `Option` if the whole expression is returning a new `Option`, we make the most of `and_then` for that:
```suggestion
                act.external_address = &config
                    .mining.mint_pkh.and_then(|pkh|
                        PublicKeyHash::from_str(pkh.as_str()).ok()
                    };
```

witnet.toml
```diff
@@ -68,11 +68,11 @@ enabled = true
...
 genesis_path = "genesis_block.json"
-# Enables or disables automatically splitting block mining rewards.
-# If set to `true`, the rewards will be split into multiple outputs of the same value as `collateral_minimum`
-# until the node's set of outputs available for collateralization outnumbers `data_request_max_retrievals_per_epoch`
-# multiplied by `collateral_coinage`, i.e. makes sure it always has spare outputs to collateralize.
-split_mint = true
+# 'mint_pkh' and 'mint_external_percentage' fields allows to the node operator splitting the mint reward
+# between the node address and another address, 'mint_external_percentage' indicates how much percentage
+# is provided to the 'mint_pkh'
+#mint_pkh = "twit1jqgf4rxjrgas3kdhj3t4cr3mg3n33m8zw0aulr"
```
aesedepece, line 74: - We can name this as external as well so that both settings are uniform and the user understand that they belong to the same feature.
- We should avoid using the "PKH" word in anything facing to the user and instead call them "addresses".
```suggestion
#mint_external_address = "twit1jqgf4rxjrgas3kdhj3t4cr3mg3n33m8zw0aulr"
```

node/src/actors/chain_manager/mining.rs
```diff
@@ -459,7 +456,10 @@ impl ChainManager {
                         own_pkh,
                         &act.chain_state.unspent_outputs_pool,
                         timestamp,
-                        tx_pending_timeout,
+                        // The timeout included when using collateral is only one epoch to ensure
+                        // that if your commit has not been accepted you can use your utxo in
+                        // the next epoch or at least in two epochs
+                        u64::from(checkpoint_period),
```
aesedepece, line 462: Makes sense.

witnet.toml
```diff
@@ -68,11 +68,11 @@ enabled = true
...
 data_request_timeout_milliseconds = 2000
 genesis_path = "genesis_block.json"
-# Enables or disables automatically splitting block mining rewards.
-# If set to `true`, the rewards will be split into multiple outputs of the same value as `collateral_minimum`
-# until the node's set of outputs available for collateralization outnumbers `data_request_max_retrievals_per_epoch`
-# multiplied by `collateral_coinage`, i.e. makes sure it always has spare outputs to collateralize.
-split_mint = true
+# 'mint_external_address' and 'mint_external_percentage' fields allows to the node operator splitting the mint reward
+# between the node address and another address, 'mint_external_percentage' indicates how much percentage
+# is provided to the 'mint_external_address'
```
aesedepece, line 73: For some reason this suggestion wasn't submitted, sorry.
```suggestion
# `mint_external_address` and `mint_external_percentage` enable splitting the mint reward between the node's own address and an "external" address, e.g. a the address of a wallet.
# `mint_external_percentage` indicates the percentage of the block rewards that will be assigned to `mint_external_address`.
```

node/src/actors/chain_manager/mining.rs
```diff
@@ -916,37 +910,39 @@ fn build_block(
...
-            }],
-        )
-    };
+    let mut vts = vec![];
+    let mut external_reward = 0;
+    let mut external_pkh = PublicKeyHash::default();
+    if let Some(pkh) = external_address {
+        // In case of a specified pkh, the reward will be redistributed between the node's pkh
+        // and the specified one using the external_percentage value
+        external_reward = (reward * u64::from(external_percentage)) / 100;
```
tmpolaczyk, line 922: In practice `reward * u64::from(external_percentage)` should never overflow, but can we replace it with `saturating_mul` just in case?

node/src/actors/chain_manager/mining.rs
```diff
@@ -916,37 +910,39 @@ fn build_block(
...
-        MintTransaction::new(
-            epoch,
-            vec![ValueTransferOutput {
-                pkh: own_pkh,
-                value: reward,
-                time_lock: 0,
-            }],
-        )
-    };
+    let mut vts = vec![];
```
tmpolaczyk, line 916: Can we extract the mint creation logic into a separate function? Something that takes `mint_reward, external_percentage, external_pkh` and returns a `MintTransaction`. This way it should be easier to test.

validations/src/validations.rs
```diff
@@ -1288,10 +1297,28 @@ pub fn update_utxo_diff(
     outputs: Vec<&ValueTransferOutput>,
     tx_hash: Hash,
 ) {
+    let mut input_pkh = if inputs.is_empty() {
+        None
+    } else {
+        let o = inputs.first().unwrap().output_pointer();
```
tmpolaczyk, line 1303: Instead of using unwrap, you can use pattern matching. So instead of `if inputs.is_empty()`, do `if let Some(first) = inputs.first()`

data_structures/src/transaction.rs
```diff
@@ -479,47 +479,50 @@ impl MintTransaction {
...
-        vt_outputs.push(ValueTransferOutput {
-            pkh: own_pkh,
-            value: reward,
-            time_lock: 0,
-        });
-
         MintTransaction::new(epoch, vt_outputs)
     }
 
     pub fn len(&self) -> usize {
```
tmpolaczyk, line 524: This method actually returns the number of transactions, so it it correct to return only 0 or 1. I think that's too misleading, maybe we should remove this method and just inline `if self.is_empty() { 0 } else { 1 }` where it is used.

lrubiorod: or directly put a 1, because a no mint is an error

data_structures/src/transaction.rs
```diff
@@ -479,47 +479,50 @@ impl MintTransaction {
...
     }
 
-    /// Try to create a MintTransaction with number of outputs equal to `utxos_required`,
-    /// where every one of those outputs must be equal to or greater than `collateral_minimum`.
-    /// If the reward is too small, the number of outputs may be smaller than `utxos_required`.
-    /// The output value will only be smaller than `collateral_minimum` if the `reward` is smaller
-    /// than `collateral_minimum`, in which case this function will create a MintTransaction with
-    /// exactly one output
-    pub fn with_split_utxos(
+    /// This method creates a MintTransaction with a reward split in the node and a external address
```
tmpolaczyk, line 482:
```suggestion
    /// This method creates a MintTransaction with a reward split between the node and a external address.
    /// external_percentage must be lower than or equal to 100.
    /// If external_address is None all the reward goes to own_pkh.
```

data_structures/src/transaction.rs
tmpolaczyk: Probably

data_structures/src/transaction.rs
```diff
@@ -479,49 +479,53 @@ impl MintTransaction {
...
             vt_outputs.push(ValueTransferOutput {
                 pkh: own_pkh,
-                value: collateral_minimum,
+                value: reward,
+                time_lock: 0,
+            });
+        }
+        // If `external_percentage` is `0` or there is not any address specified as 'external_address',
+        // the node address will receive the entire reward, and the output assigning tokens to
+        // the external address is not needed.
```
aesedepece, line 514:
```suggestion
        // If `external_percentage` is `0` or there is no address specified as 'external_address',
        // the node address will receive the entire reward, and the output assigning tokens to
        // the external address is not needed.
```

validations/src/validations.rs
```diff
@@ -1288,10 +1297,27 @@ pub fn update_utxo_diff(
     outputs: Vec<&ValueTransferOutput>,
     tx_hash: Hash,
 ) {
+    let mut input_pkh = if let Some(first) = inputs.first() {
+        utxo_diff.get(first.output_pointer()).map(|vt| vt.pkh)
+    } else {
+        None
+    };
```
aesedepece, line 1304: You know how much I love one-liners:
```suggestion
    let mut input_pkh = inputs.first().and_then(|first| utxo_diff.get(first.output_pointer()).map(|vt| vt.pkh));
```

validations/src/validations.rs
```diff
@@ -1288,10 +1297,27 @@ pub fn update_utxo_diff(
...
+
+    let mut block_number_input = 0;
     for input in inputs {
         // Obtain the OuputPointer of each input and remove it from the utxo_diff
         let output_pointer = input.output_pointer();
 
+        // Returns the pkh input in case that all pkh are the same
```
aesedepece, line 1311:
```suggestion
        // Returns the input PKH in case that all PKHs are the same
```

validations/src/validations.rs
```diff
@@ -261,28 +260,29 @@ pub fn validate_mint_transaction(
...
-        .into())
-    } else {
-        let outputs_len = mint_tx.outputs.len();
+        .into());
+    }
 
-        if !mint_tx.outputs.iter().map(|vto| vto.pkh).all_equal() {
-            return Err(BlockError::MultiplePkhsInMint.into());
-        }
+    if mint_tx.outputs.len() > 2 {
```
tmpolaczyk, line 271: I think this is not tested? That creating a mint with 2 outputs is fine but a mint with 3 outputs fails.

lrubiorod: line 149 of validations/src/tests.rs

tmpolaczyk: There is a misleading comment there:

    // Error: Mint outputs smaller than collateral minimum

PR #1243, opened 2020-05-12
Superblock broadcasting

### LINKED ISSUE #1233: Superblock forwarding
Nodes should forward superblock votes as long as they belong to the ARS and have not being forwarded yet. Therefore, each node should store the information of the votes already forwarded per superblock. When a new superblock epoch is reached, this information can be deleted.

### LINKED ISSUE #1236: Superblock signing and broadcasting

tmpolaczyk:
Close #1233
Close #1236 

This PR has a long list of unresolved questions:

### Superblock votes

```
message SuperBlockVote {
    Hash superblock_hash = 1;
    Bn256Signature bn256_signature = 2;
    KeyedSignature secp256k1_signature = 3;
    fixed32 superblock_index = 4;
}
```

We sign (superblock_index || block_hash) using bn256, and then we sign (superblock_index || block_hash || bls_signature) using secp256k1. The superblock_index is serialized in big endian. The bn256 public key is not needed because it can be obtained using the secp256k1 public key and the `AltKeys` mapping, so it is not included.

The `superblock_hash` field is also technically not needed because we can just try to verify the signature using our local `superblock_hash`, and if it fails it means it's a vote for a different superblock. But it is included for debugging purposes, we can remove it later after making sure it's not needed.

The `superblock_index` field is only used to make the forwarding easier to implement, we could remove it and the worst that will happen is that some invalid superblock votes will be forwarded when they shouldn't.

Normally when signing messages we define them in schema.proto, but this time since there are two different signatures in the same structure, I decided to just create some simple helper methods: `secp256k1_signature_message` and `bn256_signature_message`. They return the data that needs to be signed. An alternative strategy would be to use protocol buffers: serialize the incomplete SuperBlockVote, sign that with bn256, add the bn256 signature, serialize again, sign that with secp256k1, and add the secp256k1 signature. Same with verification, we would need to clone the vote, remove some fields, and serialize that. The good news is that thanks to the usage of helper methods we could easily change it.

### Superblock creation

Before this PR, only ARS members created superblocks. Now, everyone creates superblocks but only ARS members broadcast them. The reason is that in order to be able to be used by the bridge, the node needs to store the superblock votes. Also, the node needs to know the list of ARS members in order to validate the superblock votes, although a node could just forward all the received superblocks without validating ARS membership. We can discuss the different alternatives when we implement an option to allow node operators to disable superblock creation.

### Superblock forwarding

The node will forward all superblock votes that have a valid secp256k1 signature, are from the current superblock epoch, and are signed by members of the ARS. And also all the superblock votes that have a valid secp256k1 signature and are from the previous or the next superblock epoch, without checking ARS membership. This should prevent race conditions.

### DIFF
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -575,8 +575,41 @@ impl Hashable for SuperBlock {
 #[derive(Debug, Eq, PartialEq, Clone, ProtobufConvert)]
 #[protobuf_convert(pb = "witnet::SuperBlockVote")]
 pub struct SuperBlockVote {
-    pub signature: Bn256KeyedSignature,
     pub superblock_hash: Hash,
+    pub bn256_signature: Bn256Signature,
+    pub secp256k1_signature: KeyedSignature,
+}
+
+impl SuperBlockVote {
+    /// Create a new vote for the `superblock_hash`, but do not sign it.
+    /// The vote needs to be signed with a BN256 key, and that signature must
+    /// later be signed with a secp256k1 key
+    pub fn new_unsigned(superblock_hash: Hash) -> Self {
+        Self {
+            superblock_hash,
+            bn256_signature: Bn256Signature { signature: vec![] },
+            secp256k1_signature: Default::default(),
+        }
+    }
+    pub fn set_bn256_signature(&mut self, bn256_signature: Bn256Signature) {
+        self.bn256_signature = bn256_signature;
+    }
+    pub fn set_secp256k1_signature(&mut self, secp256k1_signature: KeyedSignature) {
+        self.secp256k1_signature = secp256k1_signature;
+    }
+    /// The message to be signed with the bn256 key is the hash of the superblock as bytes:
+    pub fn bn256_signature_message(&self) -> Vec<u8> {
+        self.superblock_hash.as_ref().to_vec()
+    }
+    /// The message to be signed with the secp256k1 key is the concatenation
+    /// of the hash of the superblock and the BN256 signature as bytes:
+    pub fn secp256k1_signature_message(&self) -> Vec<u8> {
+        [
+            self.superblock_hash.as_ref(),
+            &self.bn256_signature.signature,
+        ]
+        .concat()
+    }
 }
 
 /// Digital signatures structure (based on supported cryptosystems)
@@ -1140,6 +1173,12 @@ impl Bn256SecretKey {
         let signature = Bn256.sign(&self.bytes, &message)?;
 
         Ok(Bn256Signature { signature })
+    }
+}
+
+impl Hashable for Bn256Signature {
+    fn hash(&self) -> Hash {
+        calculate_sha256(&self.signature).into()
     }
 }
 
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -3045,6 +3045,46 @@ mod tests {
     }
 
     #[test]
+    fn test_superblock_hashable_trait() {
+        let superblock = SuperBlock {
+            ars_root: Hash::SHA256([1; 32]),
+            data_request_root: Hash::SHA256([2; 32]),
+            index: 1,
+            last_block: Hash::SHA256([3; 32]),
+            last_block_in_previous_superblock: Hash::SHA256([4; 32]),
+            tally_root: Hash::SHA256([5; 32]),
+        };
+        let expected = "4ee0395751a5b8d94217ba71623414721ab8dc8c1634a5c79769d5196a1b3993";
+        assert_eq!(superblock.hash().to_string(), expected);
+    }
+
+    #[test]
+    fn test_superblock_vote_bn256_signature_message() {
+        // If this test fails, the bridge will also fail
+        let superblock_hash = "4ee0395751a5b8d94217ba71623414721ab8dc8c1634a5c79769d5196a1b3993"
+            .parse()
+            .unwrap();
+        let superblock_vote = SuperBlockVote::new_unsigned(superblock_hash, 1);
+        let expected_bls =
+            hex::decode("000000014ee0395751a5b8d94217ba71623414721ab8dc8c1634a5c79769d5196a1b3993")
+                .unwrap();
+        assert_eq!(superblock_vote.bn256_signature_message(), expected_bls);
+    }
+
+    #[test]
+    fn test_superblock_vote_secp256k1_signature_message() {
+        let superblock_hash = "4ee0395751a5b8d94217ba71623414721ab8dc8c1634a5c79769d5196a1b3993"
+            .parse()
+            .unwrap();
+        let mut superblock_vote = SuperBlockVote::new_unsigned(superblock_hash, 1);
+        superblock_vote.bn256_signature.signature =
+            hex::decode("03100709d625d82c4eedf5f330d538b0cfa0dd68a5c505b6896902ed935b5cce0e")
+                .unwrap();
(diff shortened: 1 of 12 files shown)


data_structures/src/chain.rs
```diff
@@ -572,8 +572,61 @@ impl Hashable for SuperBlock {
     }
 }
 
+#[derive(Debug, Eq, PartialEq, Clone, Hash, ProtobufConvert)]
+#[protobuf_convert(pb = "witnet::SuperBlockVote")]
+pub struct SuperBlockVote {
```
lrubiorod, line 577: should we add the superblock index? That could be useful to compare new superblocks with the old ones

validations/src/validations.rs
```diff
@@ -6,6 +6,7 @@ use std::{
 };
 
 use itertools::Itertools;
+use witnet_crypto::hash::calculate_sha256;
```
lrubiorod, line 9: fmt

data_structures/src/chain.rs
tmpolaczyk: We can try that.

node/src/actors/chain_manager/mining.rs
```diff
@@ -129,8 +130,9 @@ impl ChainManager {
...
 
         let superblock_period = u32::from(chain_info.consensus_constants.superblock_period);
-        // FIXME: Only ARS members from the last block of the previous SuperBlock can create a SuperBlock
-        if rep_engine.is_ars_member(&own_pkh) && current_epoch % superblock_period == 0 {
+        // Everyone creates superblocks, but only ARS members broadcast it
+        // TODO: add config flag to disable superblock creation?
+        if current_epoch % superblock_period == 0 {
             // FIXME(#1236): ARS Members have to include the BLS signature instead of PublicKeyHash
             // FIXME: After Reputation Merkelitation, only the ARS Members from the previous consolidated
             // block will be added, to avoid include addresses that could be wrong later by
```
lrubiorod, line 138:
```suggestion
            // Superblock will be added, to avoid include addresses that could be wrong later by
```

lrubiorod: I am not sure how the ARS members is saved. Because we need the last ARS Members, that are the only nodes allowed to bootsrap the SuperBlockVotes, but you also need the ARS member from the previous consolidated block that will be the future nodes allowed to send the SuperBlockVotes

tmpolaczyk: I was thinking that when calling `build_superblock` number 10 we will pass the list of ARS identities from epoch 98 (or 99?). That list can be saved and used to validate all the superblock votes with index = 10. In order to validate superblocks with index = 11 we must wait to create the superblock number 11 locally. And in order to validate superblocks with index = 9, we could store a copy of the old list of members, but this PR just forwards it (after validating the secp256k1 signature). We should assess whether to save old superblock votes is useful, then we can implement validation of old votes.

tmpolaczyk: Okay, only ARS members from epoch 88 can create the superblock number 10. So we need to store two copies: ARS members from this superblock and ARS members from the previous superblock.

node/src/actors/chain_manager/superblock.rs
```diff
@@ -17,13 +17,15 @@ pub enum AddSuperBlockVote {
 /// State related to superblocks
 #[derive(Debug, Default)]
 pub struct SuperBlockState {
-    // Set of ARS identities that can currently send superblock votes
-    ars_identities: HashSet<PublicKeyHash>,
+    // Set of ARS identities that will be able to send superblock votes in the next superblock epoch
+    current_ars_identities: Option<HashSet<PublicKeyHash>>,
```
lrubiorod, line 21: why an option?

node/src/actors/chain_manager/superblock.rs
```diff
@@ -106,9 +108,19 @@ impl SuperBlockState {
         superblock_index: u32,
         last_block_in_previous_superblock: Hash,
     ) -> Option<SuperBlock> {
+        if let Some(prev_index) = self.current_superblock_index {
+            // TODO: remove this check
```
lrubiorod, line 112: remove this check :P

node/src/actors/chain_manager/superblock.rs
tmpolaczyk: Because we don't know the list of ARS identities when starting the node, we must wait for the first superblock mining call. This was incorrect before.

node/src/actors/chain_manager/superblock.rs
```diff
@@ -0,0 +1,331 @@
...
+            }
+            Some(superblock) => {
+                let superblock_hash = superblock.hash();
+                self.current_superblock_hash = Some(superblock_hash);
+
+                // Save ARS identities:
+                // previous = current
+                // current = sorted_ars_identities
+                {
+                    std::mem::swap(
```
lrubiorod, line 137: is there a test that check this hashset swap?

tmpolaczyk: Working on it

data_structures/src/types.rs
```diff
@@ -69,6 +72,11 @@ impl fmt::Display for Command {
                 }
                 write!(f, ": {}", tx.hash())
             }
+            Command::SuperBlockVote(sbv) => write!(
+                f,
+                "SUPERBLOCK_VOTE #{}: {}",
```
lrubiorod, line 77: Should we include the address that sign the superblock in the display?

tmpolaczyk: That may be too verbose.

There is already a log line in chain_manager/mod.rs add_superblock_vote:

                "Received superblock vote from identity {}: #{} {}",

lrubiorod:
the method log_received_messages in session/mod.rs line 167 is used to avoid showing redundant logs whent blocks or transactions are sending between nodes. Should we include the SuperblockVotes messages too?

data_structures/src/types.rs
lrubiorod: the idea is to skip the repeated messages, as I put it in another comment, but we need a way to distinguish when is the same supervote with the same signature, or when the signature is different

node/src/actors/chain_manager/superblock.rs
tmpolaczyk: Done, it's superblock_state_ars_identities

lrubiorod:
https://github.com/witnet/witnet-rust/pull/1243#discussion_r428690288
Remember

data_structures/src/types.rs
tmpolaczyk: Done.

node/src/actors/session/handlers.rs
```diff
@@ -705,3 +727,19 @@ fn send_last_beacon(session: &mut Session, beacon: CheckpointBeacon) {
     // Send LastBeacon msg
     session.send_message(beacon_msg);
 }
+
+fn send_superblock_vote(session: &mut Session, superblock_vote: SuperBlockVote) {
+    let superblock_vote_msg =
+        WitnetMessage::build_superblock_vote(session.magic_number, superblock_vote);
+    // Send SuperBlockVote msg
+    session.send_message(superblock_vote_msg);
+}
```
aesedepece, line 736: Undocumented function.

tmpolaczyk: Done

PR #1248, opened 2020-05-18
feat(node): integrate Sentry for automated bug reporting / telemetry

### LINKED ISSUE #1177: Add support for Sentry so users can voluntarily provide debug traces
Integrating Sentry would give us the opportunity to collect debug traces from users while on Testnet stage.

aesedepece:
This feature is disabled by default, but currently enabled by the `witnet.toml` configuration file.

Fix #1177

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,7 @@ itertools = "0.8.2"
 lazy_static = "1.4.0"
 log = "0.4.8"
 prettytable-rs = { version = "0.8.0", default-features = false }
-sentry = { version = "0.18.1", features = ["with_env_logger"] }
+sentry = { version = "0.18.1", features = ["with_env_logger"], optional = true }
 serde_json = "1.0.47"
 structopt = "0.3.9"
 terminal_size = "0.1.10"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ itertools = "0.8.2"
 lazy_static = "1.4.0"
 log = "0.4.8"
 prettytable-rs = { version = "0.8.0", default-features = false }
+sentry = { version = "0.18.1", features = ["with_env_logger"] }
 serde_json = "1.0.47"
 structopt = "0.3.9"
 terminal_size = "0.1.10"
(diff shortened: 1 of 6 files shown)


src/cli/mod.rs
```diff
@@ -75,7 +78,28 @@ fn init_logger(opts: LogOptions) {
...
         .filter_level(log::LevelFilter::Info)
         .filter_module("witnet", opts.level)
-        .init();
+        .filter_module("witnet_node", opts.level);
+
+    // Initialize Sentry (automated bug reporting) if explicitly enabled in configuration
+    if opts.sentry_telemetry {
+        // Initialize client
+        let guard = sentry::init(
+            "https://def0c5d0fb354ef9ad6dddb576a21624@o394464.ingest.sentry.io/5244595",
```
tmpolaczyk, line 87: Would it make sense to be able to change this URL in witnet.toml?

Cargo.toml
```diff
@@ -34,6 +34,7 @@ itertools = "0.8.2"
 lazy_static = "1.4.0"
 log = "0.4.8"
 prettytable-rs = { version = "0.8.0", default-features = false }
+sentry = { version = "0.18.1", features = ["with_env_logger"] }
```
tmpolaczyk, line 37: Can we make this dependency optional so that people can still compile the node without any telemetry? Although I don't know what's the command to disable some feature when compiling.

```suggestion
sentry = { version = "0.18.1", features = ["with_env_logger"], optional = true }
```

witnet.toml
```diff
@@ -75,7 +75,10 @@ genesis_path = "genesis_block.json"
 split_mint = true
 
 [log]
+# Logging level, i.e. from more verbose to quieter: "trace" > "debug" > "info" > "warn" > "error" > "nano"
```
lrubiorod, line 78: Nanoooo!! :P

src/cli/mod.rs
aesedepece: Let's postpone the decision until we have validated if Sentry is useful and how long it stays here.

PR #1249, opened 2020-05-19
Remove 'backup_witnesses' and 'extra_rounds' fields from `DataRequestOutput`

### LINKED ISSUE #1247: Implement new static backup factor / extra rounds

lrubiorod:
Close #1247


data_structures/src/chain.rs
```diff
@@ -2088,6 +2088,14 @@ impl DataRequestState {
         };
         self.info.current_stage = Some(self.stage);
     }
+
+    /// Function to calculate the backup witnesses required
+    pub fn backup_witnesses(&self) -> u16 {
```
tmpolaczyk, line 2093: This function should be better documented, maybe move it somewhere else as a function of `f(witnesses, commit_round) -> backup_witnesses` so it can be better tested.

src/cli/node/json_rpc_client.rs
```diff
@@ -563,18 +563,6 @@ impl fmt::Display for DataRequestTransactionInfo {
                     num_reveals,
                 )?;
             }
-            writeln!(
-                f,
-                "Commit rounds: {}/{}",
```
tmpolaczyk, line 568: Instead of removing this you can replace `{}/{}` with `{}`, and print `current_commit_round` but no `extra_commit_rounds`.

data_structures/src/data_request.rs
```diff
@@ -25,9 +25,18 @@ pub struct DataRequestPool {
     pub data_request_pool: HashMap<Hash, DataRequestState>,
     /// List of data requests that should be persisted into storage
     pub to_be_stored: Vec<DataRequestReport>,
+    /// Extra rounds for commitments and reveals
+    pub extra_rounds: u16,
 }
 
 impl DataRequestPool {
+    /// Create a new 'DataRequestPool' initialized with a extra rounds number
```
aesedepece, line 33:
```suggestion
    /// Create a new 'DataRequestPool' initialized with a number of extra rounds
```

src/cli/node/json_rpc_client.rs
```diff
@@ -565,15 +565,8 @@ impl fmt::Display for DataRequestTransactionInfo {
...
             writeln!(
                 f,
-                "Commit rounds: {}/{}",
+                "Commit rounds: {}",
                 data_request_state.current_commit_round,
-                1 + self.data_request_output.extra_commit_rounds
-            )?;
-            writeln!(
-                f,
-                "Reveal rounds: {}/{}",
```
tmpolaczyk, line 574: Same as above

schemas/witnet/witnet.proto
```diff
@@ -299,6 +296,7 @@ message ConsensusConstants {
     uint64 collateral_minimum = 13;
     uint32 collateral_age = 14;
     uint32 superblock_period = 15;
+    uint32 extra_rounds = 16;
```
tmpolaczyk, line 299: Would it be too bad to have two separate keys, `extra_commit_rounds` and `extra_reveal_rounds`?

lrubiorod: we dont find a strong reason to have different numbers for that, so thats the reason why we decide to use only one field

tmpolaczyk: Similarly, I don't see any reason to use only one field when we could use two fields, and try different combinations to find out what works better.

lrubiorod: Oks, we need more opinions to broke the tie :P. What do you think @aesedepece, @mariocao?

data_structures/src/data_request.rs
```diff
@@ -25,9 +25,18 @@ pub struct DataRequestPool {
...
 
 impl DataRequestPool {
+    /// Create a new 'DataRequestPool' initialized with a number of extra rounds
+    pub fn new(extra_rounds: u16) -> Self {
+        let mut data_request_pool = DataRequestPool::default();
+        data_request_pool.extra_rounds = extra_rounds;
+
+        data_request_pool
+    }
     /// Get all available data requests output pointers for an epoch
```
aesedepece, line 40:
```suggestion
    }

    /// Get all available data requests output pointers for an epoch
```

schemas/witnet/witnet.proto
aesedepece: __Tomayto, tomahto.__

data_structures/src/data_request.rs
```diff
@@ -25,9 +25,19 @@ pub struct DataRequestPool {
...
 }
 
 impl DataRequestPool {
+    /// Create a new 'DataRequestPool' initialized with a number of extra rounds
+    pub fn new(extra_rounds: u16) -> Self {
+        let mut data_request_pool = DataRequestPool::default();
+        data_request_pool.extra_rounds = extra_rounds;
+
+        data_request_pool
+    }
```
aesedepece, line 39: Expressions FTW!!!! :sunglasses:  
```suggestion
    pub fn new(extra_rounds: u16) -> Self {
        DataRequestPool {
            extra_rounds,
            ..Default::default()
        }
    }
```

aesedepece: Take this only as a suggestion. For what it's worth, it has my approval either way.

PR #1255, opened 2020-05-20
fix(wallet): index mint, commit and tally txns

### LINKED ISSUE #1250: Index mint transactions in wallet
PR #1242 allows setting a custom address in the mint transaction, but the wallet does not index mint transactions, so that feature currently cannot be used with sheikah

mariocao:
This PR closes #1250 but also indexes commit and tally transactions as their is not specific validation rule that forbids creating commit and tally transactions with other addresses.


aesedepece:
Merged as 1b174eab0bee1be612bba0eb5abc682eb3ab018c.

PR #1259, opened 2020-05-21
feat(validations): include output validations for commitments

lrubiorod:


validations/src/validations.rs
```diff
@@ -743,8 +743,12 @@ pub fn validate_commit_transaction(
     }
 
     let dr_output = &dr_state.data_request;
+    let pkh = co_tx.body.proof.proof.pkh();
 
     // A value transfer output cannot have zero value
```
mariocao, line 748: And may we should add as comment that it should also have the same PKH as output.

validations/src/validations.rs
```diff
@@ -743,8 +743,12 @@ pub fn validate_commit_transaction(
     }
 
     let dr_output = &dr_state.data_request;
+    let pkh = co_tx.body.proof.proof.pkh();
```
mariocao, line 746:
```suggestion
    let proof_pkh = co_tx.body.proof.proof.pkh();
```

validations/src/validations.rs
```diff
@@ -784,12 +795,11 @@ pub fn validate_commit_transaction(
     let commit_signature =
         validate_commit_reveal_signature(co_tx.hash(), &co_tx.signatures, signatures_to_verify)?;
 
-    let pkh = commit_signature.public_key.pkh();
-    let pkh2 = co_tx.body.proof.proof.pkh();
+    let pkh2 = commit_signature.public_key.pkh();
```
mariocao, line 798:
```suggestion
    let sign_pkh = commit_signature.public_key.pkh();
```

validations/src/validations.rs
```diff
@@ -744,7 +744,12 @@ pub fn validate_commit_transaction(
...
     let dr_output = &dr_state.data_request;
 
-    // A value transfer output cannot have zero value
+    // Commitment's output is only for change propose, so it only has to be one output and the
+    // address has to be the same than the address which creates the commitment
+    let proof_pkh = co_tx.body.proof.proof.pkh();
+    if co_tx.body.outputs.len() > 1 {
+        return Err(TransactionError::SeveralCommitOutputs.into());
+    }
     for (idx, output) in co_tx.body.outputs.iter().enumerate() {
```
tmpolaczyk, line 753: This for can be converted into an if-let, because there can only be 0 or 1 outputs. And idx is always 0.

```suggestion
    if let Some(output) = &co_tx.body.outputs {
```

tmpolaczyk:
Can you add a comment to `CommitTransactionBody` struct?

    // Change from collateral. The output pkh must be the same as the inputs, and there can only be one output
    pub outputs: Vec<ValueTransferOutput>,

PR #1262, opened 2020-05-26
Fix node stuck on Synchronizing

tmpolaczyk:
The purpose of this PR is to fix a strange issue where the node can get stuck in `Synchronizing` state forever. It is unknown if it actually fixes that issue though, as it is hard to reproduce.

Also, since this removes a check between the current epoch and the block epoch, now all the blocks that are not "expected" are assumed to be candidates and are sent to the ChainManager. Therefore we can see some new warnings:

    [2020-05-26T14:53:03Z WARN  witnet_node::actors::chain_manager] Block candidate's epoch differs from current epoch (9989 != 9988)

Update: we believe that are are 3 different issues that can get the node stuck in `Synchronizing`:

* If the node has 0 outbound consolidated peers when sending the `Anycast<LastBeacon>` message, it will be waiting for an `AddBlocks` message that will never arrive. This is now fixed by checking the result of the `Anycast` message, and moving back to `WaitingConsensus` on the error case.
* If the node tries to send an `Anycast<LastBeacon>` message to a `Session` that has just closed the connection (or a `Session` that was unregistered for being out of consensus), the `Anycast` will fail with error `Mailbox has closed`. This is fixed the same way as the previous point.
* If a `Session` actor sends a message to `SessionsManager` using `.wait(ctx)` while the `SessionsManager` actor is sending a message to this `Session` also using `.wait(ctx)`, both actors can enter into a deadlock, each one waiting for the message response from the other. It is not clear if this is exactly what happens, but removing the `.wait` from the `Anycast` handler should help with this issue.

To make sure that the node cannot get stuck in `Synchronizing` state anymore, we now have an additional timeout in `ChainManager` that will move the node back to `WaitingConsensus` state after N epochs in `Synchronizing` state, if the node does not receive any blocks.

### DIFF
--- a/node/src/actors/chain_manager/handlers.rs
+++ b/node/src/actors/chain_manager/handlers.rs
@@ -443,6 +443,11 @@ impl Handler<AddBlocks> for ChainManager {
             }
             StateMachine::AlmostSynced | StateMachine::Synced => {}
         };
+
+        // If we are not synchronizing, forget about when we started synchronizing
+        if self.sm_state != StateMachine::Synchronizing {
+            self.sync_waiting_for_add_blocks_since = None;
+        }
     }
 }
 
@@ -635,7 +640,7 @@ impl Handler<PeersBeacons> for ChainManager {
             pb.into_iter().map(|(p, _b)| p).collect()
         };
 
-        match self.sm_state {
+        let peers_to_unregister = match self.sm_state {
             StateMachine::WaitingConsensus => {
                 // As soon as there is consensus, we set the target beacon to the consensus
                 // and set the state to Synchronizing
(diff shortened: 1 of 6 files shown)


node/src/actors/chain_manager/handlers.rs
```diff
@@ -774,7 +767,22 @@ impl Handler<PeersBeacons> for ChainManager {
                     }
                 }
             }
+        };
+
+        if self.sm_state == StateMachine::Synchronizing {
+            if let Some(sync_start_epoch) = self.sync_waiting_for_add_blocks_since {
+                let current_epoch = self.current_epoch.unwrap();
+                let how_many_epochs_are_we_willing_to_wait_for_one_block_batch = 10;
```
tmpolaczyk, line 775: This is hardcoded to 10 epochs because it's a fix for when the node gets stuck, and that should never happen. We can use sentry to verify that it never happens, or decide a better timeout if it does happen to someone.

lrubiorod:
It would be easy to handle that: Missed epoch notification 26950. Moving to WaitingConsensus state? Or better in another issue?

tmpolaczyk:
> It would be easy to handle that: Missed epoch notification 26950. Moving to WaitingConsensus state? Or better in another issue?

When does that happen? I think it will probably disappear if we implement #1279 and decrease the batch size.

lrubiorod:
It is synced and looks good, so... approve!?? I dont know if we have to test synchronization process several times until be completely safu but...

PR #1263, opened 2020-05-26
feat(data_structures): make `MemoHash` mutably safe across threads

aesedepece:
This should allow us also to share blocks and transactions across threads without cloning :wink:


tmpolaczyk:
Can you add the test?

aesedepece:
> Can you add the test?

Sure, done, thanks!

PR #1264, opened 2020-05-26
fix(commits): skip commits with same inputs

lrubiorod:


tmpolaczyk:
Some tests would be nice

data_structures/src/chain.rs
```diff
@@ -1520,14 +1521,38 @@ impl TransactionsPool {
...
+                                .filter_map(|(_h, c)| {
+                                    co_hash_index.remove(&c).and_then(|commit_tx| {
+                                        let mut valid = true;
+                                        let mut hs_current_inputs = HashSet::new();
+                                        for input in commit_tx.body.collateral.iter() {
+                                            // This validations is to avoid commits with the same Input
+                                            if hs_inputs.contains(input)
+                                                || hs_current_inputs.contains(input)
+                                            {
+                                                valid = false;
```
tmpolaczyk, line 1535: We can stop processing the inputs after the first invalid one, right?

```suggestion
                                                valid = false;
                                                break;
```

data_structures/src/chain.rs
```diff
@@ -3344,6 +3370,119 @@ mod tests {
...
+        assert_eq!(transactions_pool.co_hash_index.len(), 2);
+        transactions_pool.insert(t3);
+        assert_eq!(transactions_pool.co_transactions.len(), 3);
+        assert_eq!(transactions_pool.co_hash_index.len(), 3);
+
+        let (commits_vec, commit_fees) = transactions_pool.remove_commits(&dr_pool);
+        assert_eq!(commits_vec.len(), 2);
+        // If the commit transactions with the same input would be returned,
+        // the commit fees exceed the 601 wits
+        assert!(commit_fees <= 601);
```
tmpolaczyk, line 3481: A strange way to check that... I would prefer something like
```suggestion
        // t2 does not conflict with anything so it must be present
        assert!(commits_vec.contains(&t2));
        // One of t1 or t3 must be present, but not both
        assert!(commits_vec.contains(&t1) ^ commits_vec.contains(&t3));
```

tmpolaczyk:
I guess using the same input in different commits for the same data request cannot happen because we validate that the commit pkh is the same as the input pkh when receiving the commit transaction, and also the commits are stored by pkh so there cannot be 2 commits with the same pkh for 1 data request. Correct?

data_structures/src/chain.rs
```diff
@@ -1520,14 +1521,39 @@ impl TransactionsPool {
...
-
-                            total_fee += dr_output.commit_fee * n_commits as u64;
+                            let filtered_commits: Vec<CommitTransaction> = commits
+                                .drain()
+                                .filter_map(|(_h, c)| {
+                                    co_hash_index.remove(&c).and_then(|commit_tx| {
+                                        let mut valid = true;
+                                        let mut hs_current_inputs = HashSet::new();
+                                        for input in commit_tx.body.collateral.iter() {
+                                            // This validations is to avoid commits with the same Input
```
aesedepece, line 1531:
```suggestion
                                            // Check that no Input from this or any other commitment is being spent twice
```

lrubiorod:
>  validate that the commit pkh is the same as the input pkh when receiving the commit transaction, and also the commits are stored by pkh so there cannot be 2 commits with the same pkh for 1 data

Right, this error only happen when there are two commit related with different data requests that use the same Input

data_structures/src/chain.rs
```diff
@@ -1508,6 +1508,7 @@ impl TransactionsPool {
     /// by the data request, and the value of all the fees obtained with those commits
     pub fn remove_commits(&mut self, dr_pool: &DataRequestPool) -> (Vec<CommitTransaction>, u64) {
         let mut total_fee = 0;
+        let mut hs_inputs = HashSet::new();
```
aesedepece, line 1511:
```suggestion
        let mut spent_inputs = HashSet::new();
```

data_structures/src/chain.rs
```diff
@@ -1520,14 +1521,39 @@ impl TransactionsPool {
...
-                                    .take(n_commits),
-                            );
-
-                            total_fee += dr_output.commit_fee * n_commits as u64;
+                            let filtered_commits: Vec<CommitTransaction> = commits
+                                .drain()
+                                .filter_map(|(_h, c)| {
+                                    co_hash_index.remove(&c).and_then(|commit_tx| {
+                                        let mut valid = true;
+                                        let mut hs_current_inputs = HashSet::new();
```
aesedepece, line 1529:
```suggestion
                                        let mut current_spent_inputs = HashSet::new();
```

data_structures/src/chain.rs
```diff
@@ -1520,14 +1521,39 @@ impl TransactionsPool {
...
+                                            hs_inputs.extend(hs_current_inputs);
+                                            Some(commit_tx)
+                                        } else {
+                                            None
+                                        }
+                                    })
+                                })
+                                .take(n_commits)
+                                .collect();
+
```
aesedepece, line 1552:
```suggestion

                            // Check once again that after filtering invalid commitments, we still have as many as requested
```

data_structures/src/chain.rs
```diff
@@ -1520,14 +1521,39 @@ impl TransactionsPool {
...
+                                            if hs_inputs.contains(input)
+                                                || hs_current_inputs.contains(input)
+                                            {
+                                                valid = false;
+                                                break;
+                                            } else {
+                                                hs_current_inputs.insert(input.clone());
+                                            }
+                                        }
+
```
aesedepece, line 1541:
```suggestion

                                        // Only mark the inputs of this commitment as used if all of them are valid and they will be returned by this function
```

PR #1265, opened 2020-05-27
Validate candidates once

tmpolaczyk:
Prior to this PR, the node did not store invalid candidates anywhere. So valid candidates were only validate once, but invalid candidates were validated every time they were received. This PR will validate all received block candidates once (per epoch). So receiving the same candidate twice will only trigger two validations if we receive it in different epochs. Which may make sense because the validity of a candidate depends on the current epoch.

Before this PR, the candidate validation consisted of comparing the block epoch with the current epoch and verifying the VRF hash. This PR adds an additional verification in between, check that the candidate builds on top of our current top of the blockchain. This changes an error that happens during synchronization: `The block has an invalid PoE`. Now that error is
```
Tried to validate candidate 1a626639804d5e676d5c9b5f332abc20849d86d01238d86977514c6e937af519: Ignoring block because previous hash ("675ddac30b85bec25b4f71f1dc1b7995fc136b0ca623c523fc171d77d2ad53a9") is different from our top block hash ("34652a0402b8a7cb89fa432ca13689f6bd4444ffd785f66d1b444a8a33666a54")

```
This allows us to skip the VRF verification when we know for sure that it will fail.

I tested this PR on the current testnet and the node synced successfully, but we should make sure that there are no strange issues when consolidating the last block.

### DIFF
--- a/node/src/actors/chain_manager/handlers.rs
+++ b/node/src/actors/chain_manager/handlers.rs
@@ -89,6 +89,7 @@ impl Handler<EpochNotification<EveryEpochPayload>> for ChainManager {
             self.sm_state = StateMachine::WaitingConsensus;
             // Clear candidates
             self.candidates.clear();
+            self.seen_candidates.clear();
         }
 
         if let Some(last_checked_epoch) = last_checked_epoch {
@@ -252,6 +253,7 @@ impl Handler<EpochNotification<EveryEpochPayload>> for ChainManager {
 
                         // Clear candidates
                         self.candidates.clear();
+                        self.seen_candidates.clear();
                     }
 
                     _ => {
@@ -659,6 +661,7 @@ impl Handler<PeersBeacons> for ChainManager {
                         let candidate = self.candidates.remove(&consensus_block_hash);
                         // Clear candidates, as they are only valid for one epoch
                         self.candidates.clear();
+                        self.seen_candidates.clear();
                         // TODO: Be functional my friend
                         if let Some((consensus_block, _consensus_block_vrf_hash)) = candidate {
                             match self.process_requested_block(ctx, &consensus_block) {
--- a/node/src/actors/chain_manager/handlers.rs
+++ b/node/src/actors/chain_manager/handlers.rs
@@ -201,7 +201,11 @@ impl Handler<EpochNotification<EveryEpochPayload>> for ChainManager {
                                         utxo_diff,
                                     ))
                                 }
-                                Err(e) => log::warn!("{}", e),
+                                Err(e) => log::warn!(
+                                    "Error when processing a block candidate {}: {}",
+                                    block_candidate.hash(),
+                                    e
+                                ),
                             }
                         }
 
@@ -800,7 +804,7 @@ impl Handler<BuildVtt> for ChainManager {
             msg.utxo_strategy,
         ) {
             Err(e) => {
-                log::error!("{}", e);
+                log::error!("Error when building value transfer transaction: {}", e);
                 Box::new(actix::fut::err(e.into()))
             }
             Ok(vtt) => {
@@ -816,7 +820,7 @@ impl Handler<BuildVtt> for ChainManager {
                             actix::fut::ok(tx_hash)
                         }
                         Err(e) => {
-                            log::error!("{}", e);
+                            log::error!("Failed to sign value transfer transaction: {}", e);
 
                             actix::fut::err(e)
                         }
@@ -849,7 +853,7 @@ impl Handler<BuildDrt> for ChainManager {
             self.tx_pending_timeout,
         ) {
             Err(e) => {
-                log::error!("{}", e);
+                log::error!("Error when building data request transaction: {}", e);
                 Box::new(actix::fut::err(e.into()))
             }
             Ok(drt) => {
@@ -866,7 +870,7 @@ impl Handler<BuildDrt> for ChainManager {
                             actix::fut::ok(tx_hash)
                         }
                         Err(e) => {
-                            log::error!("{}", e);
+                            log::error!("Failed to sign data request transaction: {}", e);
 
                             actix::fut::err(e)
                         }
--- a/node/src/actors/chain_manager/mod.rs
+++ b/node/src/actors/chain_manager/mod.rs
@@ -332,8 +332,15 @@ impl ChainManager {
     }
     #[allow(clippy::map_entry)]
     fn process_candidate(&mut self, block: Block) {
-        if let (Some(current_epoch), Some(rep_engine), Some(vrf_ctx), Some(secp_ctx)) = (
+        if let (
+            Some(current_epoch),
+            Some(chain_info),
+            Some(rep_engine),
+            Some(vrf_ctx),
+            Some(secp_ctx),
+        ) = (
             self.current_epoch,
+            self.chain_state.chain_info.as_ref(),
             self.chain_state.reputation_engine.as_ref(),
             self.vrf_ctx.as_mut(),
             self.secp.as_ref(),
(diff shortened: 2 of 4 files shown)


node/src/actors/chain_manager/handlers.rs
```diff
@@ -200,7 +201,11 @@ impl Handler<EpochNotification<EveryEpochPayload>> for ChainManager {
                                         utxo_diff,
                                     ))
                                 }
-                                Err(e) => log::warn!("{}", e),
+                                Err(e) => log::warn!(
+                                    "Process candidate {}: {}",
+                                    block_candidate.hash(),
+                                    e
+                                ),
```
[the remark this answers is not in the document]
tmpolaczyk: We received a block that was thought to be a block candidate, but a basic validation determined that it is not possible to append this block on top of our block chain tip. This mostly happens while synchronizing. Any suggestions on how to explain this in the log?

I thought about disabling block candidate validation while synchronizing but that makes it impossible to go to synced state.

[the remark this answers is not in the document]
mariocao: Maybe something like we do in the wallet:

```rust
log::warn!("Tried to process a candidate #{} that does not build directly on top of our tip of the chain #{}. (<REASON HERE>)",
                    block_beacon.checkpoint,
                    current_last_sync.checkpoint,
                );
```

node/src/actors/chain_manager/handlers.rs
```diff
@@ -797,7 +804,7 @@ impl Handler<BuildVtt> for ChainManager {
             msg.utxo_strategy,
         ) {
             Err(e) => {
-                log::error!("{}", e);
+                log::error!("BuildVtt: {}", e);
```
mariocao, line 807: Maybe something more descriptive

node/src/actors/chain_manager/mod.rs
```diff
@@ -374,10 +374,12 @@ impl ChainManager {
                             self.broadcast_item(InventoryItem::Block(block));
                             log::debug!("Send Candidate");
                         }
-                        Err(e) => log::warn!("{}", e),
+                        Err(e) => log::warn!("Validate candidate {}: {}", hash_block, e),
                     },
-                    Err(e) => log::warn!("{}", e),
+                    Err(e) => log::warn!("Validate candidate {}: {}", hash_block, e),
```
mariocao, line 379: What's the difference between this error and the previous one?

node/src/actors/chain_manager/handlers.rs
tmpolaczyk: For example?

node/src/actors/chain_manager/mod.rs
tmpolaczyk: This inner one validates the block epoch and the block hash_prev_block, and the second one validates the VRF proof. They are two steps of the same validation, so I will try to merge the two logs into one.

node/src/actors/chain_manager/handlers.rs
aesedepece: :shrug:

node/src/actors/chain_manager/mod.rs
```diff
@@ -744,7 +749,7 @@ impl ChainManager {
                     actix::fut::ok(())
                 }
                 Err(e) => {
-                    log::warn!("{}", e);
+                    log::warn!("Validate transacion {}: {}", msg.transaction.hash(), e);
```
aesedepece, line 752: I wouldn't be able to wrap my head around a warning saying "validate transaction". Is it telling me to go and validate it by hand? :stuck_out_tongue: 

Maybe something like this?
```suggestion
                    log::warn!("Error when validating transaction {}: {}", msg.transaction.hash(), e);
```

node/src/actors/chain_manager/handlers.rs
aesedepece, line 807:
```suggestion
                log::error!("Error when building value transfer transaction: {}", e);
```

node/src/actors/chain_manager/handlers.rs
```diff
@@ -813,7 +820,7 @@ impl Handler<BuildVtt> for ChainManager {
                             actix::fut::ok(tx_hash)
                         }
                         Err(e) => {
-                            log::error!("{}", e);
+                            log::error!("BuildVtt: failed to sign transaction: {}", e);
```
aesedepece, line 823:
```suggestion
                            log::error!("Failed to sign value transfer transaction: {}", e);
```

node/src/actors/chain_manager/handlers.rs
```diff
@@ -846,7 +853,7 @@ impl Handler<BuildDrt> for ChainManager {
             self.tx_pending_timeout,
         ) {
             Err(e) => {
-                log::error!("{}", e);
+                log::error!("BuildDrt: {}", e);
```
aesedepece, line 856:
```suggestion
                log::error!("Error when building data request transaction: {}", e);
```

node/src/actors/chain_manager/handlers.rs
```diff
@@ -863,7 +870,7 @@ impl Handler<BuildDrt> for ChainManager {
                             actix::fut::ok(tx_hash)
                         }
                         Err(e) => {
-                            log::error!("{}", e);
+                            log::error!("Failed to sign transaction: {}", e);
```
aesedepece, line 873:
```suggestion
                            log::error!("Failed to sign data request transaction: {}", e);
```

node/src/actors/chain_manager/handlers.rs
```diff
@@ -200,7 +201,11 @@ impl Handler<EpochNotification<EveryEpochPayload>> for ChainManager {
                                         utxo_diff,
                                     ))
                                 }
-                                Err(e) => log::warn!("{}", e),
+                                Err(e) => log::warn!(
+                                    "Tried to process a candidate {}: {}",
```
aesedepece, line 205:
```suggestion
                                    "Error when processing a block candidate {}: {}",
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -330,54 +332,57 @@ impl ChainManager {
...
-                        Err(e) => log::warn!("{}", e),
-                    },
-                    Err(e) => log::warn!("{}", e),
+                    verify_signatures(signatures_to_verify, vrf_ctx, secp_ctx)
+                }) {
+                    Ok(vrf_hash) => {
+                        self.candidates
+                            .insert(hash_block, (block.clone(), vrf_hash[0]));
+                        self.broadcast_item(InventoryItem::Block(block));
+                        log::debug!("Send Candidate");
```
aesedepece, line 378: Has this log line any use?

tmpolaczyk: Let's assume it doesn't

node/src/actors/chain_manager/mod.rs
```diff
@@ -330,54 +332,56 @@ impl ChainManager {
...
-                    Err(e) => log::warn!("{}", e),
+                    verify_signatures(signatures_to_verify, vrf_ctx, secp_ctx)
+                }) {
+                    Ok(vrf_hash) => {
+                        self.candidates
+                            .insert(hash_block, (block.clone(), vrf_hash[0]));
+                        self.broadcast_item(InventoryItem::Block(block));
+                    }
+                    Err(e) => {
+                        log::warn!("Tried to validate candidate {}: {}", hash_block, e);
```
aesedepece, line 380:
```suggestion
                        log::warn!("Error when trying to validate block candidate {}: {}", hash_block, e);
```

PR #1266, opened 2020-05-27
fix(wallet): avoid de-synchronization if node is closed

mariocao:
The node design's currently allows reorganizations of 1 block.

Problem:
 - When the node is closed and re-opened it does not have the last block
 - Specially, when running a local node, the last block is just lost and the wallet stays in an inconsistent state.

This PR improves the synchronization by only indexing a block once the next one arrives. :)

Blocked by PR #1263


wallet/src/actors/worker/methods.rs
```diff
@@ -746,24 +747,82 @@ impl Worker {
...
+                block_previous_beacon: block_beacon.hash_prev_block,
+                local_chain_tip: current_last_sync.hash_prev_block,
+            }));
+        }
 
+        // Index the previous stored block in `last_block`
+        let new_last_sync = self.index_block(current_last_block, &wallet, sink)?;
+
+        // Update wallet state with the last synchronization information:
+        // - last_sync: last epoch indexed and last indexed block hash
```
aesedepece, line 807:
```suggestion
        // - last_sync: last indexed epoch and block hash
```

PR #1267, opened 2020-05-28
Stop slashing commitments with "forced errors"

### LINKED ISSUE #1261: Discussion RadonErrors slashing and implement a solution
**Problem:** Currently we detected that some "honest" nodes (witnet-rust releases provided through this repository) are being slashed by returning RadonError::RetrieveTimeOutError, that could be produced because some APIs can not provide a proper solution for all the requests done by the witnesses. That means, that witnesses that send this kind of error could be slashed (lose reputation and collateral).

**Proposed solutions:**
 - _No slash:_ In this case, nodes are incentivized with reputation and reward in case of Error's consensus, but they will not be slashed in case of being out of consensus. **Cons:** nodes could only commit errors due to they never will be slashed, and they could win reputation in some cases. One response to this could be that the incentives for having normal behavior are more proficient that to do this.
- _No slash, any reward:_ They will not be slashed but also not be rewarded when they commit an Error. **Cons:** Nodes are not incentivized to commit RadonErrors
- _No slash, no reputation reward:_ They will not be slashed but also not be rewarded with reputation in case of Error's consensus, but in that case they will rewarded with the wits are specified in the Data Requests. **Cons:** are the wits reward enough as an incentive for the node operators to commit an error?

**Other related cuestions:**
- Currently, utxos created in TallyTransactions that come from the collateral has to wait a collateral age period. That produce a high quantity of utxos in the net (dust). Removing that wait, we reduce the number of dust, and also the high consume for having a big utxo set in memory, but combined with one of the solution proposed before, we could increase the easiness to commit an attack with several identities.

- Some nodes are showing a high percentage of Error commitment, if there will be a random response from the source, all nodes will have more or less the same errors percentage. So it would be great detect what factor increase your error commitments number

lrubiorod:
This PR adds the next logic:
- If you commit a RadonError you can not be slashed
- If you commit a RadonError and you are in consensus, you will earn the wits reward
- If you commit more RadonErrors than normal values (truths) in an epoch, you will not earn reputation.

image
image

Close #1261

### DIFF
--- a/data_structures/src/data_request.rs
+++ b/data_structures/src/data_request.rs
@@ -370,31 +370,45 @@ pub fn calculate_tally_change(
 pub fn calculate_witness_reward(
     commits_count: usize,
     reveals_count: usize,
-    honests_count: usize,
-    dr_output: &DataRequestOutput,
-    collateral_minimum: u64,
+    // Number of values that are out of consensus plus non-revealers
+    liars_count: usize,
+    // To calculate the reward, we consider errors_count as the number of errors that are
+    // out of consensus, it means, that they do not deserve any reward
+    errors_count: usize,
+    reward: u64,
+    collateral: u64,
 ) -> (u64, u64) {
-    let collateral = if dr_output.collateral == 0 {
-        collateral_minimum
-    } else {
-        dr_output.collateral
-    };
     if commits_count == 0 {
         (0, 0)
     } else if reveals_count == 0 {
         (collateral, 0)
     } else {
-        let liars_count = reveals_count - honests_count;
-        let slashed_count = (commits_count - reveals_count + liars_count) as u64;
-        let honests_count = honests_count as u64;
-        let slashed_collateral_reward = collateral * slashed_count / honests_count;
-        let slashed_collateral_remainder = (collateral * slashed_count) % honests_count;
+        let honests_count = (commits_count - liars_count - errors_count) as u64;
+        let liars_count = liars_count as u64;
+        let slashed_collateral_reward = collateral * liars_count / honests_count;
+        let slashed_collateral_remainder = (collateral * liars_count) % honests_count;
 
         (
-            dr_output.witness_reward + collateral + slashed_collateral_reward,
+            reward + collateral + slashed_collateral_reward,
             slashed_collateral_remainder,
         )
     }
+}
+/// Count how many commitments will be considered "errors" and how many will
+/// be considered "lies". This tells apart the case in which the committed value was
+/// an error from any other type of out-of-consensus situation, like non-reveals.
+fn calculate_errors_and_liars_count(errors: &[bool], liars: &[bool]) -> (usize, usize) {
+    liars
+        .iter()
+        .zip(errors.iter())
+        .fold((0, 0), |(l_count, e_count), x| match x {
+            // If it is out of consensus and it is an error we consider as a error
+            (true, true) => (l_count, e_count + 1),
+            // If it is out of consensus and it is not an error we consider as a lie
+            (true, false) => (l_count + 1, e_count),
+            // Rest of cases is an honest node
+            _ => (l_count, e_count),
+        })
 }
 
 pub fn create_tally<RT, S: ::std::hash::BuildHasher>(
@@ -412,69 +426,88 @@ where
     if let Stage::Tally(tally_metadata) = &report.metadata {
         let commits_count = committers.len();
         let reveals_count = revealers.len();
-        let mut slashed_witnesses = committers;
+        let mut out_of_consensus = committers;
+        let mut error_committers = vec![];
 
         let liars = &tally_metadata.liars;
+        let errors = &tally_metadata.errors;
 
-        if reveals_count != liars.len() {
-            return Err(TransactionError::MismatchingLiarsNumber {
-                reveals_n: reveals_count,
-                inputs_n: liars.len(),
-            }
-            .into());
-        }
+        assert_eq!(reveals_count, liars.len(), "Length of liars vector collected from tally ({}) does not match actual count of reveals ({})", reveals_count, liars.len());
+        assert_eq!(reveals_count, errors.len(), "Length of errors vector collected from tally ({}) does not match actual count of reveals ({})", reveals_count, errors.len());
 
-        let liars_count = liars.iter().fold(0, |count, liar| match liar {
-            true => count + 1,
-            false => count,
-        });
-        let honests_count = reveals_count - liars_count;
+        let (liars_count, errors_count) = calculate_errors_and_liars_count(errors, liars);
+
+        let collateral = if dr_output.collateral == 0 {
+            collateral_minimum
+        } else {
+            dr_output.collateral
+        };
+
         // Collateral division rest goes for the miner
+        let non_reveals_count = commits_count - reveals_count;
         let (reward, _rest) = calculate_witness_reward(
             commits_count,
             reveals_count,
-            honests_count,
-            &dr_output,
-            collateral_minimum,
+            liars_count + non_reveals_count,
+            errors_count,
+            dr_output.witness_reward,
+            collateral,
         );
 
         let mut outputs: Vec<ValueTransferOutput> = if reveals_count > 0 {
             revealers
                 .iter()
-                .zip(liars.iter())
-                .filter_map(|(&revealer, &liar)| {
-                    // Only reward reveals in consensus
-                    if liar {
-                        None
-                    } else {
+                .zip(liars.iter().zip(errors.iter()))
+                .filter_map(|(&revealer, (liar, error))| match (liar, error) {
+                    // If an out-of-consensus commitment was an error report, collateral is refunded
+                    (true, true) => {
+                        let vt_output = ValueTransferOutput {
+                            pkh: revealer,
+                            value: collateral,
+                            time_lock: 0,
+                        };
+                        error_committers.push(revealer);
+                        Some(vt_output)
+                    }
+                    // Case out-of-consensus value commitment
+                    (true, false) => None,
+                    // If the result of the tally is an error report,
+                    // commitments containing error reports are rewarded
+                    (false, true) => {
                         let vt_output = ValueTransferOutput {
                             pkh: revealer,
                             value: reward,
                             time_lock: 0,
                         };
-                        slashed_witnesses.remove(&revealer);
+                        out_of_consensus.remove(&revealer);
+                        error_committers.push(revealer);
+                        Some(vt_output)
+                    }
+                    // Case in consensus value
+                    (false, false) => {
+                        let vt_output = ValueTransferOutput {
+                            pkh: revealer,
+                            value: reward,
+                            time_lock: 0,
+                        };
+                        out_of_consensus.remove(&revealer);
                         Some(vt_output)
                     }
                 })
                 .collect()
         } else {
             // In case of no revealers, collateral returns to their owners
-            let reward = if dr_output.collateral == 0 {
-                collateral_minimum
-            } else {
-                dr_output.collateral
-            };
-
-            slashed_witnesses
+            out_of_consensus
                 .iter()
                 .map(|&committer| ValueTransferOutput {
                     pkh: committer,
-                    value: reward,
+                    value: collateral,
                     time_lock: 0,
                 })
                 .collect()
         };
 
+        let honests_count = reveals_count - liars_count - errors_count;
         let tally_change =
             calculate_tally_change(commits_count, reveals_count, honests_count, &dr_output);
         if tally_change > 0 {
@@ -487,13 +520,14 @@ where
         }
 
         let tally_bytes = Vec::try_from(report)?;
-        let slashed_witnesses = slashed_witnesses.into_iter().collect();
+        let out_of_consensus = out_of_consensus.into_iter().collect();
 
         Ok(TallyTransaction::new(
             dr_pointer,
             tally_bytes,
             outputs,
-            slashed_witnesses,
+            out_of_consensus,
+            error_committers,
         ))
     } else {
         Err(TransactionError::NoTallyStage.into())
@@ -718,7 +752,7 @@ mod tests {
     }
 
     fn from_tally_to_storage(fake_block_hash: Hash, mut p: DataRequestPool, dr_pointer: Hash) {
-        let tally_transaction = TallyTransaction::new(dr_pointer, vec![], vec![], vec![]);
+        let tally_transaction = TallyTransaction::new(dr_pointer, vec![], vec![], vec![], vec![]);
 
         // There is nothing to be stored yet
         assert_eq!(p.to_be_stored.len(), 0);
--- a/data_structures/src/error.rs
+++ b/data_structures/src/error.rs
@@ -175,12 +175,6 @@ pub enum TransactionError {
     /// RadonReport not in Tally Stage
     #[fail(display = "RadonReport not in Tally Stage")]
     NoTallyStage,
-    /// Mismatching number of reveals and liars vector.
-    #[fail(
-        display = "Mismatching number of reveals ({}) and liars vector ({})",
-        reveals_n, inputs_n
-    )]
-    MismatchingLiarsNumber { reveals_n: usize, inputs_n: usize },
     /// Minimum consensus percentage is invalid
     #[fail(
         display = "Minimum consensus percentage {} is invalid. Must be >50 and <100",
@@ -201,12 +195,21 @@ pub enum TransactionError {
     /// Zero amount specified
     #[fail(display = "Cannot build transaction with zero value")]
     ZeroAmount,
-    /// Incorrect value of Rewarded witnesses in Tally
+    /// Incorrect count of out-of-consensus witnesses in Tally
     #[fail(
-        display = "Incorrect value of Slashed witnesses in Tally. Expected: {:?}, found: {:?}",
+        display = "Incorrect count of out-of-consensus witnesses in Tally. Expected: {:?}, found: {:?}",
         expected, found
     )]
-    MismatchingSlashedWitnesses {
+    MismatchingOutOfConsensusCount {
+        expected: Vec<PublicKeyHash>,
+        found: Vec<PublicKeyHash>,
+    },
+    /// Incorrect count of witnesses with errors in Tally
+    #[fail(
+        display = "Incorrect count of witnesses with errors in Tally. Expected: {:?}, found: {:?}",
+        expected, found
+    )]
+    MismatchingErrorCount {
         expected: Vec<PublicKeyHash>,
         found: Vec<PublicKeyHash>,
     },
--- a/data_structures/src/radon_report.rs
+++ b/data_structures/src/radon_report.rs
@@ -148,6 +148,8 @@ pub struct TallyMetaData {
     /// This follows a reverse logic: `false` is truth and `true` is lie.
     /// A liar is an out-of-consensus value
     pub liars: Vec<bool>,
+    /// An error is a RadonError value (or considered as an error due to a RadonError consensus)
+    pub errors: Vec<bool>,
     /// Proportion between total reveals and "truthers" count:
     /// `liars.iter().filter(std::ops::Not).count() / reveals.len()`
     pub consensus: f32,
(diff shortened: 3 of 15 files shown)


wallet/src/repository/wallet/mod.rs
```diff
@@ -1086,7 +1086,8 @@ fn build_tally_report(
                 .collect::<Result<HashMap<PublicKeyHash, model::Reveal>>>()?;
 
             // Set not `in_consensus` reveals
-            for pkh in &tally.slashed_witnesses {
+            // TODO: Review that part for wallet guys (now out of consensus is no the same that slashed)
```
lrubiorod, line 1089: @mariocao, @aesedepece could you review that part?

src/cli/node/json_rpc_client.rs
```diff
@@ -719,7 +719,10 @@ pub fn data_request_report(
                         let honest = match dr_info.tally.as_ref() {
                             None => format!(""),
                             Some(tally) => {
-                                if tally.slashed_witnesses.contains(&pkh) {
+                                // TODO: Need review by tmpolaczyk
```
lrubiorod, line 722: @tmpolaczyk could you review that part?

data_structures/src/transaction.rs
```diff
@@ -430,8 +430,9 @@ pub struct TallyTransaction {
     pub dr_pointer: Hash, // DTTransaction hash
     // Outputs
     pub tally: Vec<u8>,
-    pub outputs: Vec<ValueTransferOutput>, // Witness rewards
-    pub slashed_witnesses: Vec<PublicKeyHash>,
+    pub outputs: Vec<ValueTransferOutput>,    // Witness rewards
```
tmpolaczyk, line 433: Please put this comment in a separated line, and add a comment to the `error_committers` field.

```suggestion
    /// Witness rewards
    pub outputs: Vec<ValueTransferOutput>,
```

wallet/src/repository/wallet/mod.rs
aesedepece: image

src/cli/node/json_rpc_client.rs
tmpolaczyk: This condition should be "if the witness was not rewarded". But now there are 3 possible outcomes, so we should re-think the `dataRequestReport` method.

The current behavior is:
* Honest: "+reward"
* Liar: red text "-0"

I propose:

* Honest: "+reward"
* Liar: red text "-0"
* Error: red text "+reward"

This way will be relatively simple to implement but it may be difficult for users to understand the difference, unless we explain all the consensus logic in `dataRequestReport --help`.

So yeah, we should add an additional case to this if. Will see if I can write a commit later.

node/src/actors/chain_manager/mod.rs
```diff
@@ -887,17 +887,23 @@ pub fn process_validations(
...
+    pub truths: u32,
+    pub lies: u32,
+    pub errors: u32,
+}
+
 #[derive(Debug, Default)]
 struct ReputationInfo {
     // Counter of "witnessing acts".
     // For every data request with a tally in this block, increment alpha_diff
     // by the number of witnesses specified in the data request.
```
tmpolaczyk, line 902: Unrelated to this PR, but this comment is outdated
```suggestion
    // by the number of reveals present in the tally.
```

validations/src/validations.rs
```diff
@@ -383,6 +383,7 @@ pub fn evaluate_tally_precondition_clause(
     reveals: Vec<RadonReport<RadonTypes>>,
     minimum_consensus: f64,
     num_commits: usize,
+    errors: &mut Vec<bool>,
```
tmpolaczyk, line 386: Any reason this is a separate argument and not part of `TallyPreconditionClauseResult`?

data_structures/src/data_request.rs
```diff
@@ -370,33 +370,42 @@ pub fn calculate_tally_change(
 pub fn calculate_witness_reward(
     commits_count: usize,
     reveals_count: usize,
-    honests_count: usize,
-    dr_output: &DataRequestOutput,
-    collateral_minimum: u64,
+    liars_count: usize,
+    errors_count: usize, // it is an out of consensus error
```
tmpolaczyk, line 374: Perhaps we can simplify the API by using something similar to `RequestResult` as an input to `calculate_witness_reward`.

src/cli/node/json_rpc_client.rs
lrubiorod: we consider an error if it is out of consensus and an error, so it should not have a reward. And I think we have to include -collateral for the liars?

lrubiorod: An error in consensus should be considered as honest

tmpolaczyk: > And I think we have to include -collateral for the liars?

Currently, `dataRequestReport` never shows `-collateral` when you lose the collateral, but that would make sense.

tmpolaczyk: Proposal number 2:
* Honest: "+reward"
* Liar: red text "-collateral"
* Error: "+0"

data_structures/src/data_request.rs
```diff
@@ -424,57 +435,83 @@ where
...
+                        // In case of an out of consensus error, collateral is returned
+                        let vt_output = ValueTransferOutput {
+                            pkh: revealer,
+                            value: collateral,
+                            time_lock: 0,
+                        };
+                        error_committers.push(revealer.clone());
+                        Some(vt_output)
+                    } else if error {
+                        // In case of an in-consensus error, it obtains the reward
```
aesedepece, line 482:
```suggestion
                        // If the result of the tally is an error report, commitments containing error reports are rewarded 
```

data_structures/src/error.rs
```diff
@@ -201,7 +207,7 @@ pub enum TransactionError {
     /// Zero amount specified
     #[fail(display = "Cannot build transaction with zero value")]
     ZeroAmount,
-    /// Incorrect value of Rewarded witnesses in Tally
+    /// Incorrect value of Slashed witnesses in Tally
```
aesedepece, line 210:
```suggestion
    /// Incorrect count of slashed witnesses in Tally
```

data_structures/src/error.rs
```diff
@@ -201,7 +207,7 @@ pub enum TransactionError {
...
     #[fail(
         display = "Incorrect value of Slashed witnesses in Tally. Expected: {:?}, found: {:?}",
```
aesedepece, line 212:
```suggestion
        display = "Incorrect count of slashed witnesses in Tally. Expected: {:?}, found: {:?}",
```

validations/src/validations.rs
```diff
@@ -952,18 +974,34 @@ pub fn validate_tally_transaction<'a>(
...
+    let sorted_expected_out_of_consensus = expected_ta_tx
+        .out_of_consensus
         .into_iter()
         .sorted()
         .collect();
 
     // Validation of slashed witnesses
-    if sorted_expected_slashed != sorted_slashed {
+    if sorted_expected_out_of_consensus != sorted_out_of_consensus {
         return Err(TransactionError::MismatchingSlashedWitnesses {
```
aesedepece, line 986:
```suggestion
        return Err(TransactionError::MismatchingOutOfConsensusCount {
```

validations/src/validations.rs
```diff
@@ -1001,14 +1037,21 @@ pub fn validate_tally_transaction<'a>(
         &dr_state.data_request,
     );
 
+    let collateral = if dr_state.data_request.collateral == 0 {
+        collateral_minimum
+    } else {
+        dr_state.data_request.collateral
+    };
```
aesedepece, line 1044:
```suggestion
    let collateral = max(dr_state.data_request.collateral, collateral_minimum);
```

validations/src/validations.rs
lrubiorod: good point

node/src/actors/chain_manager/mod.rs
```diff
@@ -1026,22 +1034,25 @@ fn update_pools(
...
 {
     let mut honests = vec![];
     let mut liars = vec![];
-    for (pkh, num_lies) in rep_info {
-        if num_lies == V::default() {
+    let mut errors = vec![];
+    for (pkh, result) in rep_info {
+        if result.lies > 0 {
+            liars.push((pkh, result.lies));
+        // TODO: Decide which percentage would be fair enough
```
tmpolaczyk, line 1047: The behavior implemented in this PR is to only give reputation to nodes whose number of errors is smaller than the number of "honests".

This opens an attack vector where an attacker can flood the network with requests that always return errors, and nobody will be able to earn reputation except of course the attacker nodes, which will avoid participating in their own requests and will slowly accumulate all the reputation.

data_structures/src/data_request.rs
```diff
@@ -424,57 +435,83 @@ where
...
         let mut outputs: Vec<ValueTransferOutput> = if reveals_count > 0 {
             revealers
                 .iter()
-                .zip(liars.iter())
-                .filter_map(|(&revealer, &liar)| {
+                .zip(liars.iter().zip(errors.iter()))
+                .filter_map(|(&revealer, (&liar, &error))| {
                     // Only reward reveals in consensus
-                    if liar {
+                    if liar && !error {
```
tmpolaczyk, line 470: This is a match

node/src/actors/chain_manager/mod.rs
lrubiorod: This attack need several wits to pay the rewards, because witness will earn the wits unless they do not earn reputation. The other option, give reputation to error_committers could produce incentives for node operators to commit errors without the risk to lose reputation, but also wining it

data_structures/src/data_request.rs
```diff
@@ -424,57 +435,83 @@ where
...
+            .into());
+        }
+
+        let (liars_count, errors_count) = calculate_errors_and_liars_count(errors, liars);
+
+        let collateral = if dr_output.collateral == 0 {
+            collateral_minimum
+        } else {
+            dr_output.collateral
+        };
```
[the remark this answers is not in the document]
lrubiorod: That it is a different approach. We consider that if there are not a collateral value, we put the collateral minimum. But if the requester specify a quantity for collateral, and is less than minimum I think we shouldnt change and wait for the data request error in validation. 

Maybe someone put 50_000 because he thinks is in wits and need a big amount of collateral, it is better that the request will be never included, and the requester notice his own error than publish the request with a very low value of collateral

validations/src/validations.rs
lrubiorod: same as before

src/cli/node/json_rpc_client.rs
tmpolaczyk: Proposal number 3:
* Honest: "+reward" "+rep"
* Liar: red text "-collateral"
* Error: "+reward"

Although I'm not sure if it is possible to get the amount of reputation that has been rewarded to witnesses.

lrubiorod: applied!!

data_structures/src/data_request.rs
```diff
@@ -412,69 +425,87 @@ where
...
             revealers
                 .iter()
-                .zip(liars.iter())
-                .filter_map(|(&revealer, &liar)| {
-                    // Only reward reveals in consensus
-                    if liar {
-                        None
-                    } else {
+                .zip(liars.iter().zip(errors.iter()))
+                .filter_map(|(&revealer, x)| match x {
```
tmpolaczyk, line 459: `match (liar, error)` would be easier to understand, now it's confusing, what's `(true, false)`?

And does it make sense to order the branches somehow? For example `(false, false)` first, `(true, true)` last, or vice versa.

data_structures/src/error.rs
```diff
@@ -181,6 +181,12 @@ pub enum TransactionError {
         reveals_n, inputs_n
     )]
     MismatchingLiarsNumber { reveals_n: usize, inputs_n: usize },
+    /// Mismatching number of reveals and errors vector.
+    #[fail(
+        display = "Mismatching number of reveals ({}) and errors vector ({})",
+        reveals_n, inputs_n
+    )]
+    MismatchingErrorsNumber { reveals_n: usize, inputs_n: usize },
```
tmpolaczyk, line 189: This error is unused.

data_structures/src/error.rs
```diff
@@ -181,6 +181,12 @@ pub enum TransactionError {
         reveals_n, inputs_n
     )]
     MismatchingLiarsNumber { reveals_n: usize, inputs_n: usize },
```
tmpolaczyk, line 183: This error is unused.

data_structures/src/radon_report.rs
```diff
@@ -148,6 +148,8 @@ pub struct TallyMetaData {
     /// This follows a reverse logic: `false` is truth and `true` is lie.
     /// A liar is an out-of-consensus value
     pub liars: Vec<bool>,
+    /// An error is a RadonError value
```
tmpolaczyk, line 151:
```suggestion
    /// An error is a RadonError value when the consensus is not a RadonError
```

data_structures/src/transaction.rs
```diff
@@ -426,12 +426,16 @@ impl RevealTransactionBody {
...
+    pub dr_pointer: Hash,
+    /// Tally result
     pub tally: Vec<u8>,
-    pub outputs: Vec<ValueTransferOutput>, // Witness rewards
-    pub slashed_witnesses: Vec<PublicKeyHash>,
+    /// Witness rewards
+    pub outputs: Vec<ValueTransferOutput>,
+    /// Addresses that are out of consensus (non revealers included)
+    pub out_of_consensus: Vec<PublicKeyHash>,
+    /// Addresses that commit a RadonError
```
tmpolaczyk, line 437:
```suggestion
    /// Addresses that commit a RadonError when the consensus is not a RadonError
```

validations/src/tests.rs
```diff
@@ -3470,31 +3471,36 @@ fn create_commit_reveal(
 //   - A CommitTransaction vector
 //   - A RevealTransaction vector
 //   - A vector with the liar publicKeys
```
tmpolaczyk, line 3473:
```suggestion
//   - A vector with the liar publicKeys
//   - A vector with the error publicKeys 
```

validations/src/validations.rs
```diff
@@ -943,6 +948,24 @@ fn create_expected_tally_transaction(
     Ok((ta_tx, dr_state.clone()))
 }
 
+pub fn calculate_errors_and_liars_count_from_tally(
```
tmpolaczyk, line 951: This funcion is named `calculate_errors_and_liars` but it returns `(liars, errors)`

validations/src/validations.rs
```diff
@@ -943,6 +948,24 @@ fn create_expected_tally_transaction(
...
+    tally_tx: &TallyTransaction,
+    dr_state: &DataRequestState,
+) -> (usize, usize) {
+    tally_tx
+        .out_of_consensus
+        .iter()
+        .fold((0, 0), |(l_count, e_count), x| {
+            if !dr_state.info.reveals.contains_key(x) {
```
tmpolaczyk, line 959: What is this branch? If an identity is out of consensus and is not a revealer? Shouldn't that be a lie?

data_structures/src/radon_report.rs
lrubiorod: is an error in both cases

data_structures/src/transaction.rs
lrubiorod: is an error in both cases

validations/src/validations.rs
lrubiorod: the function calculate_witness_rewards uses liars_count as a node that commit and reveal a value and it is a lie, so in that case we do not take into account the non-reveal that are calculated using another values... but yes, it could be misunderstanded

wallet/src/repository/wallet/mod.rs
mariocao: seems ok... maybe it could have some indirect consequence in the way `non-conensus` reveals are shown in Sheikah

wallet/src/repository/wallet/mod.rs
```diff
@@ -1086,7 +1086,8 @@ fn build_tally_report(
...
+            for pkh in &tally.out_of_consensus {
                 let liar = reveals.get_mut(&pkh).cloned();
                 if let Some(mut reveal) = liar {
```
mariocao, line 1092:
```suggestion
                if let Some(mut reveal) = outlier {
```

mariocao: Now they are not considered `dishonest` but outliers to the consensus.

wallet/src/repository/wallet/mod.rs
lrubiorod: perfect, but in that case it could be solved in a little fix in the future if we notice that

data_structures/src/data_request.rs
```diff
@@ -412,69 +426,87 @@ where
...
```
tmpolaczyk, line 460: I meant this:

```suggestion
                .filter_map(|(&revealer, (liar, error))| match (liar, error) {
```

node/src/actors/chain_manager/mod.rs
tmpolaczyk: Currently if one identity commits one error and one "honest", it will not be rewarded any reputation. Maybe we can change the `>` into `>=` to be less aggressive.

node/src/actors/chain_manager/mining.rs
```diff
@@ -939,14 +940,23 @@ fn build_block(
             tally_txns.push(ta_tx.clone());
             let commits_count = dr_state.info.commits.len();
             let reveals_count = dr_state.info.reveals.len();
-            let honests_count = commits_count - ta_tx.slashed_witnesses.len();
+
+            let (liars_count, errors_count) = calculate_liars_and_errors_count_from_tally(&ta_tx);
```
tmpolaczyk, line 944: This operation is only needed to calculate the `extra_tally_fee`. If we could refactor this function from accepting `tally_transactions: &[TallyTransaction]` into `tally_transactions_and_fees: &[(TallyTransaction, u64)]` we could skip this calculation.

But we would need to add that information to `create_tally_transactions`, and I'm not sure if this change will be small enough to be included in this PR :P

lrubiorod: uff, I want to merge as soon as possible, maybe we could open an issue to optimize some functions later

tmpolaczyk: That's fine

validations/src/tests.rs
```diff
@@ -4219,18 +4339,119 @@ fn tally_valid_3_reveals_dr_liar_invalid() {
...
     );
 }
 
+#[test]
+fn tally_valid_5_reveals_1_liar_1_error() {
+    // Reveal value: integer(0)
+    let reveal_value = vec![0x00];
+    let liar_value = vec![0x0a];
+
+    // Create a DataRequestPool with 3 reveals (one of them is a lie from the data requester)
```
tmpolaczyk, line 4361: This comment is wrong

validations/src/validations.rs
```diff
@@ -514,31 +521,32 @@ pub fn construct_report_from_clause_result(
     script: &RADTally,
     reports_len: usize,
 ) -> RadonReport<RadonTypes> {
+    // This TallyMetadata would be included in case of Error Result, in that case, anyone has to be
+    // classified as a lier, but as an error
```
tmpolaczyk, line 525:
```suggestion
    // classified not as a liar, but as an error

```

validations/src/validations.rs
```diff
@@ -952,18 +970,34 @@ pub fn validate_tally_transaction<'a>(
...
+    }
+
+    let sorted_error = ta_tx.error_committers.iter().cloned().sorted().collect();
+    let sorted_expected_error = expected_ta_tx
+        .error_committers
         .into_iter()
         .sorted()
         .collect();
 
     // Validation of slashed witnesses
```
tmpolaczyk, line 996:
```suggestion
    // Validation of error witnesses
```

node/src/actors/chain_manager/mod.rs
tmpolaczyk: Should we open an issue to further assess which percentage would be fair enough?

lrubiorod:
> Investigate issue with random_source.json:
> 
> [2020-06-09T11:20:38Z ERROR witnet_node::actors::chain_manager::mining] Error trying to mine a block: Invalid witness reward found: 1000001000. Expected value: 2000001000
> 
> Logs:
> [pr-1282-fail.zip](https://github.com/witnet/witnet-rust/files/4751830/pr-1282-fail.zip)

Fixed

Issue #1088, opened 2020-03-16
Unknown panic on TCP write


aesedepece:
```
thread 'main' panicked at 'no Task is currently running', /root/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-0.1.29/src/task_impl/mod.rs:47:9

stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at ./cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/libunwind.rs:88
   1: backtrace::backtrace::trace_unsynchronized
             at ./cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:77
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:59
   4: core::fmt::write
             at src/libcore/fmt/mod.rs:1052
   5: std::io::Write::write_fmt
             at src/libstd/io/mod.rs:1426
   6: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:62
   7: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:49
   8: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:204
   9: std::panicking::default_hook
             at src/libstd/panicking.rs:224
  10: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:472
  11: std::panicking::begin_panic
  12: futures::task_impl::Task::will_notify_current
  13: <tokio_sync::task::atomic_task::CurrentTask as tokio_sync::task::atomic_task::Register>::register
  14: tokio_sync::task::atomic_task::AtomicTask::register
  15: tokio_reactor::registration::Registration::poll_ready
  16: tokio_reactor::registration::Registration::poll_write_ready
  17: tokio_reactor::poll_evented::PollEvented<E>::poll_write_ready
  18: <tokio_reactor::poll_evented::PollEvented<E> as std::io::Write>::write
  19: <tokio_tcp::stream::TcpStream as std::io::Write>::write
  20: <tokio_io::split::WriteHalf<T> as std::io::Write>::write
  21: <actix::io::FramedWrite<T,U> as core::ops::drop::Drop>::drop
[...]

PR #1270, opened 2020-05-30
fix: ensure that `system.run()` lives inside a Task

aesedepece:
this is believed to help with #1088

### DIFF
--- a/node/src/actors/node.rs
+++ b/node/src/actors/node.rs
@@ -61,8 +61,11 @@ pub fn run(config: Config, callback: fn()) -> Result<(), failure::Error> {
     let json_rpc_server_addr = JsonRpcServer::default().start();
     SystemRegistry::set(json_rpc_server_addr);
 
-    // Run system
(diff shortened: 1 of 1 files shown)


node/src/actors/node.rs
```diff
@@ -62,7 +62,9 @@ pub fn run(config: Config, callback: fn()) -> Result<(), failure::Error> {
     SystemRegistry::set(json_rpc_server_addr);
 
     // Run system
```
tmpolaczyk, line 64: Maybe add a short comment to the code explaining the change?

```suggestion
    // Run system inside a future
    // This is believed to help with issue #1088
```

aesedepece: Applied!

tmpolaczyk: The important line is the one that references the issue, and I think it got lost?

aesedepece: Don't worry, thanks for sending it once again. Updated!

PR #1271, opened 2020-05-31
New folders layout

aesedepece:
- divides the `.witnet` folder into `.witnet/config` and `.witnet/storage`
- updates `witnet.toml` so it uses the new folders
- makes the `witnet-rust` Docker image migrate old storage files into the new storage folder
- allows the host to directly modify `witnet.toml` of a existing container
- copies `./witnet` into `/usr/local/bin`, so that `docker exec` commands also work as `docker exec witnet_node witnet node ...`
- makes `masterKeyExport` write to config folder, not storage folder, so that you don't delete our master key backup if a storage wipe is needed

Existing non-Docker setups should keep working, provided that they use a custom `witnet.toml` file that points directly to their storage folder.

### DIFF
--- a/src/cli/node/with_node.rs
+++ b/src/cli/node/with_node.rs
@@ -1,4 +1,8 @@
-use std::{net::SocketAddr, path::PathBuf, time::Duration};
+use std::{
+    net::SocketAddr,
+    path::{Path, PathBuf},
+    time::Duration,
+};
 
 use structopt::StructOpt;
 
@@ -7,7 +11,11 @@ use witnet_node as node;
 
 use super::json_rpc_client as rpc;
 
-pub fn exec_cmd(command: Command, mut config: Config) -> Result<(), failure::Error> {
+pub fn exec_cmd(
+    command: Command,
+    config_path: Option<PathBuf>,
+    mut config: Config,
+) -> Result<(), failure::Error> {
     match command {
         Command::GetBlock { node, hash } => {
             rpc::get_block(node.unwrap_or(config.jsonrpc.server_address), hash)
@@ -145,8 +153,8 @@ pub fn exec_cmd(command: Command, mut config: Config) -> Result<(), failure::Err
             let write_to_path = match (write, write_to) {
                 // Don't write
                 (false, None) => None,
-                // Write to default storage path
-                (true, None) => Some(config.storage.db_path),
+                // Write to the same folder where the config file is located
+                (true, None) => config_path.and_then(|path| path.parent().map(Path::to_path_buf)),
                 // Write to custom path
                 (_, Some(path)) => Some(path),
             };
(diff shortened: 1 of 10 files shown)


tmpolaczyk:
Can we also unify the wallet storage path with this one? The wallet is stored by default in `~/.local/share/witnet` while the node is in `~/.witnet`.

tmpolaczyk:
Also, sheikah uses `~/.sheikah`, do we want to unify that as well?

aesedepece:
@tmpolaczyk That's interesting. I'll see what can be done now and what can be left for another moment.

src/cli/node/with_node.rs
```diff
@@ -145,8 +153,8 @@ pub fn exec_cmd(command: Command, mut config: Config) -> Result<(), failure::Err
             let write_to_path = match (write, write_to) {
                 // Don't write
                 (false, None) => None,
-                // Write to default storage path
-                (true, None) => Some(config.storage.db_path),
+                // Write to the same folder where the config file is located
+                (true, None) => config_path.and_then(|path| path.parent().map(Path::to_path_buf)),
```
tmpolaczyk, line 157: If `config_path` is `None` this silently ignores the `--write` flag. We should return an error instead.

aesedepece: Sounds fair to me.

aesedepece: @tmpolaczyk commit 3f671b61e037e6d3cbb0f3415107dfd68f82acdf implements your suggestion

src/cli/node/with_node.rs
```diff
@@ -145,8 +153,12 @@ pub fn exec_cmd(command: Command, mut config: Config) -> Result<(), failure::Err
...
                 // Don't write
                 (false, None) => None,
-                // Write to default storage path
-                (true, None) => Some(config.storage.db_path),
+                // Write to the same folder where the config file is located
+                // Fail if using default configuration (not sourced from a file in the filesystem)
+                (true, None) => config_path
+                    .expect("Cannot guess a file system path for writing the master key file. Please specify a valid path right after the `--write` flag.")
+                    .parent()
+                    .map(Path::to_path_buf),
```
tmpolaczyk, line 161: Same issue, if the config path is "/", `.parent()` will return `None` and the master key will be silently not-written.

lrubiorod:
My approve is for testing, not for code review so... take into account :P

src/cli/node/with_node.rs
aesedepece: Oops, I merged cos' it was approved, but then saw your comment @tmpolaczyk. I'll do a little PR adding an `expect` after `parent()`. Sorry for pulling the trigger so quickly!

PR #1276, opened 2020-06-01
Match punctuation in subcommands description

clbartoli:
When running the command `cargo run -- node --help` some of the descriptions have a dot and some don't. This PR erases the dots in with_node.rs and with_wallet.rs.

### DIFF
--- a/src/cli/node/with_node.rs
+++ b/src/cli/node/with_node.rs
@@ -178,7 +178,7 @@ pub fn exec_cmd(command: Command, mut config: Config) -> Result<(), failure::Err
 
 #[derive(Debug, StructOpt)]
 pub enum Command {
-    #[structopt(name = "server", about = "Run a Witnet node server.", alias = "run")]
+    #[structopt(name = "server", about = "Run a Witnet node server", alias = "run")]
     Run(ConfigParams),
     #[structopt(
         name = "raw",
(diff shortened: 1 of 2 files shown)


src/cli/node/with_node.rs
```diff
@@ -255,7 +255,7 @@ pub enum Command {
         /// Socket address of the Witnet node to query.
         #[structopt(short = "n", long = "node")]
         node: Option<SocketAddr>,
-        /// Show all the information about utxos
+        /// Show all the information about utxos.
```
tmpolaczyk, line 258: I thought we wanted to remove dots?

clbartoli: I thought dots in comments were fine since those messages are not going to be printed

tmpolaczyk: These comments are used to document the flags of each subcommand, if you run `witnet node getUtxoInfo --help` you will see a detailed description of each. So yes, they are printed, but I'm not sure if they should end with a dot or not.

clbartoli: ahhh ok, no dots then :)

tmpolaczyk: As an example, `git --help` does not have dots, but `git add --help` has dots when explaining the flags.

src/cli/node/with_node.rs
```diff
@@ -341,10 +341,10 @@ pub enum Command {
         about = "Create a value transfer transaction that join utxos"
```
aesedepece, line 341:
```suggestion
        about = "Create a value transfer transaction that joins UTXOs"
```

src/cli/node/with_node.rs
```diff
@@ -314,10 +314,10 @@ pub enum Command {
         about = "Create a value transfer transaction that split utxos"
```
aesedepece, line 314:
```suggestion
        about = "Create a value transfer transaction that splits UTXOs"
```

PR #1277, opened 2020-06-01
feat(rad): make `RadonReport` include `partial_results`

### LINKED ISSUE #1244: [Wallet] Detail data request result for each operator
As a user trying a data request with the wallet, I would like to receive the trace with the result of each operator applied.

For instance, if I try a data request with which contains: `(1) IntegerSum` and then `(2) IntegerSum`,
I would like to receive the result of operation (1) and operation(2).

aesedepece:
This allows tracing the execution of a RADON script step by step.

Fix #1244

### DIFF
--- a/data_structures/src/radon_report.rs
+++ b/data_structures/src/radon_report.rs
@@ -1,6 +1,8 @@
 use std::convert::TryFrom;
 use std::time::{Duration, Instant};
 
+use serde::Serialize;
+
 use crate::radon_error::ErrorLike;
 
 /// A high level data structure aimed to be used as the return type of RAD executor methods:
@@ -10,7 +12,7 @@ use crate::radon_error::ErrorLike;
 /// It contains a RAD result paired with metadata specific to the stage of the script being executed.
 ///
 /// `RT` is the generalization of `RadonTypes`, and `IE` is the generalization of `RadError`
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Serialize)]
 pub struct RadonReport<RT>
 where
     RT: TypeLike,
@@ -82,6 +84,23 @@ where
     }
 }
 
+/*impl<RT> Serialize for RadonReport<RT>
+where
+    RT: TypeLike + Serialize,
+{
+    fn serialize<S>(&self, serializer: S) -> Result<<S as Serializer>::Ok, <S as Serializer>::Error>
+    where
+        S: Serializer,
+    {
+        let mut state = serializer.serialize_struct("RadonReport", 4)?;
+        state.serialize_field("metadata", &self.metadata)?;
+        state.serialize_field("partial_results", &self.partial_results)?;
+        state.serialize_field("result", &self.result)?;
+        state.serialize_field("running_time", &self.running_time)?;
+        state.end()
+    }
+}*/
+
 /// This trait identifies a RADON-compatible type system, i.e. most likely an `enum` with different
 /// cases for different data types.
 pub trait TypeLike: Clone + Sized {
@@ -135,7 +154,7 @@ impl ReportContext {
 }
 
 /// Tell different stage-specific metadata structures from each other.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Serialize)]
 pub enum Stage {
     /// Metadata for contextless execution of RADON scripts.
     Contextless,
@@ -163,7 +182,7 @@ impl Default for Stage {
 //pub struct AggregationMetaData {}
 
 /// Tally-specific metadata structure.
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug, Default, Serialize)]
 pub struct TallyMetaData {
     /// A positional vector of "truthers" and "liars", i.e. reveals that passed all the filters vs.
     /// those which were filtered out.
@@ -209,11 +228,11 @@ mod tests {
     use std::fmt;
 
     use failure::Fail;
+    use serde_cbor::Value as SerdeCborValue;
 
     use crate::radon_error::{ErrorLike, RadonError, RadonErrors};
 
     use super::*;
-    use serde_cbor::Value as SerdeCborValue;
 
     #[test]
     fn test_encode_not_cbor() {
--- a/data_structures/src/radon_report.rs
+++ b/data_structures/src/radon_report.rs
@@ -50,10 +50,10 @@ where
         context: &ReportContext,
     ) -> Self {
         let intercepted: Vec<RT> = partial_results.into_iter().map(RT::intercept).collect();
-        let result = match intercepted.last() {
-            None => unreachable!("Partial result vectors always contain at least 1 item"),
-            Some(x) => (*x).clone(),
-        };
+        let result = (*intercepted
+            .last()
+            .expect("Partial result vectors must always contain at least 1 item"))
+        .clone();
 
         RadonReport {
             metadata: context.stage.clone(),
@@ -84,7 +84,7 @@ where
 
 /// This trait identifies a RADON-compatible type system, i.e. most likely an `enum` with different
 /// cases for different data types.
-pub trait TypeLike: std::clone::Clone + std::marker::Sized {
+pub trait TypeLike: Clone + Sized {
     type Error: ErrorLike;
 
     fn encode(&self) -> Result<Vec<u8>, Self::Error>;
--- a/data_structures/src/radon_report.rs
+++ b/data_structures/src/radon_report.rs
@@ -17,6 +17,8 @@ where
 {
     /// Stage-specific metadata.
     pub metadata: Stage,
+    /// Vector of partial results (the results in between each of the operators in a script)
+    pub partial_results: Vec<RT>,
     /// This the intercepted result of the script execution: any `IE` raised in runtime has already
     /// been mapped into a `RT` (e.g. `RadError` -> `RadonTypes::RadonError`.
     pub result: RT,
@@ -30,13 +32,27 @@ where
     RT: TypeLike,
 {
     /// Factory for constructing a `RadonReport` from the `Result` of something that could be
-    /// `ErrorLike` plus a `ReportContext`.
+    /// `TypeLike` or `ErrorLike` plus a `ReportContext`.
     pub fn from_result(result: Result<RT, RT::Error>, context: &ReportContext) -> Self {
-        let intercepted = RT::intercept(result);
+        Self::from_partial_results(vec![result], context)
+    }
+
+    /// Factory for constructing a `RadonReport` from a vector of partial results, which could be
+    /// `TypeLike` or `ErrorLike`, plus a `ReportContext`.
+    pub fn from_partial_results(
+        partial_results: Vec<Result<RT, RT::Error>>,
+        context: &ReportContext,
+    ) -> Self {
+        let intercepted: Vec<RT> = partial_results.into_iter().map(RT::intercept).collect();
+        let result = match intercepted.last() {
+            None => unreachable!("Partial result vectors always contain at least 1 item"),
+            Some(x) => (*x).clone(),
+        };
 
         RadonReport {
-            result: intercepted,
             metadata: context.stage.clone(),
+            partial_results: intercepted,
+            result,
             running_time: context.duration(),
         }
     }
@@ -62,7 +78,7 @@ where
 
 /// This trait identifies a RADON-compatible type system, i.e. most likely an `enum` with different
 /// cases for different data types.
-pub trait TypeLike: std::marker::Sized {
+pub trait TypeLike: std::clone::Clone + std::marker::Sized {
     type Error: ErrorLike;
 
     fn encode(&self) -> Result<Vec<u8>, Self::Error>;
@@ -77,8 +93,8 @@ pub struct ReportContext {
     pub call_argument_index: Option<u8>,
     pub call_index: Option<u8>,
     pub call_operator: Option<u8>,
-    pub stage: Stage,
     pub completion_time: Option<Instant>,
+    pub stage: Stage,
     pub start_time: Option<Instant>,
     pub script_index: Option<u8>,
 }
--- a/data_structures/src/radon_report.rs
+++ b/data_structures/src/radon_report.rs
@@ -84,23 +84,6 @@ where
     }
 }
 
-/*impl<RT> Serialize for RadonReport<RT>
-where
-    RT: TypeLike + Serialize,
-{
-    fn serialize<S>(&self, serializer: S) -> Result<<S as Serializer>::Ok, <S as Serializer>::Error>
-    where
-        S: Serializer,
-    {
-        let mut state = serializer.serialize_struct("RadonReport", 4)?;
-        state.serialize_field("metadata", &self.metadata)?;
-        state.serialize_field("partial_results", &self.partial_results)?;
-        state.serialize_field("result", &self.result)?;
-        state.serialize_field("running_time", &self.running_time)?;
-        state.end()
-    }
-}*/
-
 /// This trait identifies a RADON-compatible type system, i.e. most likely an `enum` with different
 /// cases for different data types.
 pub trait TypeLike: Clone + Sized {
--- a/data_structures/src/radon_report.rs
+++ b/data_structures/src/radon_report.rs
@@ -18,7 +18,7 @@ where
     /// Stage-specific metadata.
     pub metadata: Stage,
     /// Vector of partial results (the results in between each of the operators in a script)
-    pub partial_results: Vec<RT>,
+    pub partial_results: Option<Vec<RT>>,
     /// This the intercepted result of the script execution: any `IE` raised in runtime has already
     /// been mapped into a `RT` (e.g. `RadError` -> `RadonTypes::RadonError`.
     pub result: RT,
@@ -34,7 +34,13 @@ where
     /// Factory for constructing a `RadonReport` from the `Result` of something that could be
     /// `TypeLike` or `ErrorLike` plus a `ReportContext`.
     pub fn from_result(result: Result<RT, RT::Error>, context: &ReportContext) -> Self {
-        Self::from_partial_results(vec![result], context)
+        let intercepted = RT::intercept(result);
+        RadonReport {
+            metadata: context.stage.clone(),
+            partial_results: None,
+            result: intercepted,
+            running_time: context.duration(),
+        }
     }
 
     /// Factory for constructing a `RadonReport` from a vector of partial results, which could be
@@ -51,7 +57,7 @@ where
 
         RadonReport {
             metadata: context.stage.clone(),
-            partial_results: intercepted,
+            partial_results: Some(intercepted),
             result,
             running_time: context.duration(),
         }
--- a/node/src/actors/rad_manager/handlers.rs
+++ b/node/src/actors/rad_manager/handlers.rs
@@ -14,6 +14,7 @@ use witnet_validations::validations::{
 use crate::actors::messages::{ResolveRA, RunTally};
 
 use super::RadManager;
+use witnet_rad::script::RadonScriptExecutionSettings;
 
 impl Handler<ResolveRA> for RadManager {
     type Result = ResponseFuture<RadonReport<RadonTypes>, RadError>;
@@ -52,7 +53,11 @@ impl Handler<ResolveRA> for RadManager {
                     // Perform aggregation on the values that made it to the output vector after applying the
                     // source scripts (aka _normalization scripts_ in the original whitepaper) and filtering out
                     // failures.
-                    witnet_rad::run_aggregation_report(values, &aggregator)
+                    witnet_rad::run_aggregation_report(
+                        values,
+                        &aggregator,
+                        RadonScriptExecutionSettings::all_but_partial_results(),
+                    )
                 }
                 Ok(TallyPreconditionClauseResult::MajorityOfErrors { errors_mode }) => {
                     Ok(RadonReport::from_result(
--- a/node/src/actors/rad_manager/handlers.rs
+++ b/node/src/actors/rad_manager/handlers.rs
@@ -5,7 +5,7 @@ use futures::Future;
 use tokio::util::FutureExt;
 
 use witnet_data_structures::radon_report::{RadonReport, ReportContext};
-use witnet_rad::{error::RadError, types::RadonTypes};
+use witnet_rad::{error::RadError, script::RadonScriptExecutionSettings, types::RadonTypes};
 use witnet_validations::validations::{
     construct_report_from_clause_result, evaluate_tally_precondition_clause,
     TallyPreconditionClauseResult,
@@ -14,7 +14,6 @@ use witnet_validations::validations::{
 use crate::actors::messages::{ResolveRA, RunTally};
 
 use super::RadManager;
-use witnet_rad::script::RadonScriptExecutionSettings;
 
 impl Handler<ResolveRA> for RadManager {
     type Result = ResponseFuture<RadonReport<RadonTypes>, RadError>;
--- a/node/tests/data_request_examples.rs
+++ b/node/tests/data_request_examples.rs
@@ -6,6 +6,7 @@ use std::{
 };
 use witnet_data_structures::chain::DataRequestOutput;
 use witnet_node::actors::messages::BuildDrt;
+use witnet_rad::script::RadonScriptExecutionSettings;
 use witnet_rad::types::{
     float::RadonFloat, integer::RadonInteger, string::RadonString, RadonTypes,
 };
@@ -44,7 +45,11 @@ fn run_dr_locally_with_data(
     let mut retrieval_results = vec![];
     for (r, d) in dr.data_request.retrieve.iter().zip(data.iter()) {
         log::info!("Running retrieval for {}", r.url);
-        retrieval_results.push(witnet_rad::run_retrieval_with_data(r, (*d).to_string())?);
+        retrieval_results.push(witnet_rad::run_retrieval_with_data(
+            r,
+            (*d).to_string(),
+            RadonScriptExecutionSettings::disable_all(),
+        )?);
     }
 
     log::info!("Running aggregation with values {:?}", retrieval_results);
--- a/node/tests/data_request_examples.rs
+++ b/node/tests/data_request_examples.rs
@@ -49,7 +49,7 @@ fn run_dr_locally_with_data(
         log::info!("Running retrieval for {}", r.url);
         retrieval_results.push(witnet_rad::run_retrieval_with_data(
             r,
-            (*d).to_string(),
+            *d,
             RadonScriptExecutionSettings::disable_all(),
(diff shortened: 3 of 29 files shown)


aesedepece:
Sorry for the ludicrous test, but this should definitely guarantee that it works hahaha.

data_structures/src/radon_report.rs
```diff
@@ -17,6 +17,8 @@ where
 {
     /// Stage-specific metadata.
     pub metadata: Stage,
+    /// Vector of partial results (the results in between each of the operators in a script)
+    pub partial_results: Vec<RT>,
```
tmpolaczyk, line 21: Could we make this field optional? If we always store the partial results we will be using extra memory for no reason, because we don't need partial results when executing data requests in the node, right?

I would keep the old behavior and add a new method `execute_radon_script_with_partial_results` or similar, and maybe even pass `partial_results: &mut Vec<_>` as an input argument, to avoid having to handle an optional field in `RadonReport`, if that doesn't complicate thinks in the wallet API.

aesedepece: I made some changes in that line, but through an approach that I believe that will give us better extensibility in the future. The extension of the changes is a little bigger than I wanted, but at least it has zero impact to any existing APIs.

rad/src/script.rs
```diff
@@ -189,11 +264,201 @@ fn test_execute_radon_script() {
...
+    let mut context = ReportContext::default();
+    let output = execute_radon_script(
+        input,
+        &script,
+        &mut context,
+        RadonScriptExecutionSettings::enable_all(),
+    )
+    .unwrap();
+    let partial_expected = vec![
+        RadonTypes::from(RadonString::from(
```
tmpolaczyk, line 283: Reuse variable?

```suggestion
        input
```

data_structures/src/radon_report.rs
```diff
@@ -30,13 +32,33 @@ where
...
+
+    /// Factory for constructing a `RadonReport` from a vector of partial results, which could be
+    /// `TypeLike` or `ErrorLike`, plus a `ReportContext`.
+    pub fn from_partial_results(
+        partial_results: Vec<Result<RT, RT::Error>>,
+        context: &ReportContext,
+    ) -> Self {
+        let intercepted: Vec<RT> = partial_results.into_iter().map(RT::intercept).collect();
+        let result = match intercepted.last() {
+            None => unreachable!("Partial result vectors always contain at least 1 item"),
```
tmpolaczyk, line 54: I can trigger this branch using `RadonReport::from_partial_results(vec![], context)`, therefore it is not unreachable. 

```suggestion
            None => panic!("Partial result vectors must always contain at least 1 item"),
```

data_structures/src/radon_report.rs
```diff
@@ -62,7 +84,7 @@ where
 
 /// This trait identifies a RADON-compatible type system, i.e. most likely an `enum` with different
 /// cases for different data types.
-pub trait TypeLike: std::marker::Sized {
+pub trait TypeLike: std::clone::Clone + std::marker::Sized {
```
tmpolaczyk, line 87: No need to be explicit:

```suggestion
pub trait TypeLike: Clone + Sized {
```

rad/src/script.rs
aesedepece: lol true

rad/src/script.rs
```diff
@@ -20,27 +21,98 @@ use crate::{
...
 ) -> Result<RadonReport<RadonTypes>, RadError> {
     // Set the execution timestamp
-    context.start();
-    // Run the execution
+    if settings.timing {
+        context.start();
+    }
+
+    // Initialize a vector for storing the partial results
+    let mut partial_results = vec![Ok(input.clone())];
```
tmpolaczyk, line 82: You can avoid this clone when `setings.partial_results` is disabled, by setting partial_results to `None` here:

```suggestion
    let mut partial_results = if settings.partial_results { Some(vec![Ok(input.clone())]) } else { None };
```

And later using `if let Some(partial_results) = &mut partial_results` {}

rad/src/script.rs
```diff
@@ -20,27 +21,98 @@ use crate::{
 pub type RadonCall = (RadonOpCodes, Option<Vec<Value>>);
 pub type RadonScript = Vec<RadonCall>;
 
+/// A set of flags for telling the RADON executor what execution features to enable and what
+/// metadata to collect.
+#[derive(Clone, Copy)]
+pub struct RadonScriptExecutionSettings {
+    pub partial_results: bool,
+    pub timing: bool,
+    pub breakpoints: bool,
```
tmpolaczyk, line 30: A short description of each flag would be nice

rad/src/script.rs
```diff
@@ -20,30 +21,115 @@ use crate::{
...
-    context.complete();
+
+    // Set the completion timestamp, if enabled by `timing` settings
+    if settings.timing {
+        context.complete();
+    }
 
     // Return a report as constructed from the result and the context
-    Ok(RadonReport::from_result(result, context))
+    Ok(if settings.partial_results {
```
tmpolaczyk, line 122: You can also use `let Some` here
```suggestion
    Ok(if let Some(partial_results) = partial_results {
```

tmpolaczyk:
Now that the running time is not calculated in the node, we can change this log line so that it only shows the result:

```
[2020-06-03T10:39:23Z DEBUG witnet_rad] Result for URL https://api.blockchair.com/bitcoin/stats: Ok(RadonReport { metadata: Retrieval, partial_results: None, result: String(RadonString { value: "0000000000000000000a8f8f15025d0e538c92e16c2e4cede2d6fe65bd172a98" }), running_time: 0ns })
```

rad/src/script.rs
aesedepece: True, applied!

aesedepece:
> Now that the running time is not calculated in the node, we can change this log line so that it only shows the result:
> 
> ```
> [2020-06-03T10:39:23Z DEBUG witnet_rad] Result for URL https://api.blockchair.com/bitcoin/stats: Ok(RadonReport { metadata: Retrieval, partial_results: None, result: String(RadonString { value: "0000000000000000000a8f8f15025d0e538c92e16c2e4cede2d6fe65bd172a98" }), running_time: 0ns })
> ```

Fair point. Please check if it makes more sense to you know. This is what I did:
image

rad/src/lib.rs
```diff
@@ -126,7 +126,16 @@ pub async fn run_retrieval_report(
             let result =
                 run_retrieval_with_data_report(retrieve, response_string, context, settings);
 
-            log::debug!("Result for URL {}: {:?}", retrieve.url, result);
+            match &result {
+                Ok(report) => {
+                    log::debug!(
+                        "Successful result for source {}: {:?}",
+                        retrieve.url,
+                        report
```
tmpolaczyk, line 134: This change got lost somewhere?

```suggestion
                        report.as_ref().map(|report| report.result)
```

node/tests/data_request_examples.rs
```diff
@@ -6,6 +6,7 @@ use std::{
 };
 use witnet_data_structures::chain::DataRequestOutput;
 use witnet_node::actors::messages::BuildDrt;
+use witnet_rad::script::RadonScriptExecutionSettings;
```
lrubiorod, line 9: fmt

aesedepece: Why?

rad/src/lib.rs
```diff
@@ -109,9 +123,19 @@ pub async fn run_retrieval_report(retrieve: &RADRetrieve) -> Result<RadonReport<
...
+
+            match &result {
+                Ok(report) => {
+                    log::debug!(
+                        "Successful result for source {}: {:?}",
+                        retrieve.url,
+                        report
+                    );
+                }
+                Err(e) => log::debug!("Failed result for source {}: {:?}", retrieve.url, e),
```
tmpolaczyk, line 137: I'm unable to see this line, I think it only happens if the RADON script is malformed, but they are checked when sending requests... And since HTTP errors are handled above, they are not printed here. So currently an HTTP error is not logged and if we see an HTTP error we must guess the URL. But let's leave that for another issue, or this PR will never be merged :laughing:

node/tests/data_request_examples.rs
lrubiorod: I would combine the witnet_rad::script... and witnet_rad::types in witnet_rad::{}

tmpolaczyk:
Sheikah cannot use this method yet because the `RadonReport` is serialized as a string, not as a JSON object. This is because of this line in `run_rad_req.rs`:

```rust
#[derive(Debug, Serialize)]
pub struct RunRadReqResponse {
    #[serde(serialize_with = "debug_serialize")]
    pub result: types::RadonReport<types::RadonTypes>,
}
```

We need to remove `#[serde(serialize_with = "debug_serialize")]` and implement `Serialize` for `RadonReport`. If it's not possible to implement `Serialize` for some field of `RadonReport`, maybe we can use the `debug_serialize` hack with that field.

node/tests/data_request_examples.rs
aesedepece: hahaha I was not seeing the next line from the issue comments view, sorry!

aesedepece:
@gabaldon @Tommytrg now you get something like this:
```json
{
   "metadata":"Contextless",
   "partial_results":[
      "{\"coord\":{\"lon\":13.41,\"lat\":52.52},\"weather\":[{\"id\":600,\"main\":\"Snow\",\"description\":\"light snow\",\"icon\":\"13n\"}],\"base\":\"stations\",\"main\":{\"temp\":-4,\"pressure\":1013,\"humidity\":73,\"temp_min\":-4,\"temp_max\":-4},\"visibility\":10000,\"wind\":{\"speed\":2.6,\"deg\":90},\"clouds\":{\"all\":75},\"dt\":1548346800,\"sys\":{\"type\":1,\"id\":1275,\"message\":0.0038,\"country\":\"DE\",\"sunrise\":1548313160,\"sunset\":1548344298},\"id\":2950159,\"name\":\"Berlin\",\"cod\":200}",
      {
         "clouds":{
            "all":75.0
         },
         "wind":{
            "deg":90.0,
            "speed":2.6
         },
         "base":"stations",
         "coord":{
            "lat":52.52,
            "lon":13.41
         },
         "visibility":10000.0,
         "weather":[
            {
               "main":"Snow",
               "description":"light snow",
               "id":600.0,
               "icon":"13n"
            }
         ],
         "dt":1548346800.0,
         "cod":200.0,
         "main":{
            "temp_min":-4.0,
            "humidity":73.0,
            "temp":-4.0,
            "pressure":1013.0,
            "temp_max":-4.0
         },
         "sys":{
            "message":0.0038,
            "sunrise":1548313160.0,
            "country":"DE",
            "type":1.0,
            "sunset":1548344298.0,
            "id":1275.0
         },
         "id":2950159.0,
         "name":"Berlin"
      },
      {
         "temp_min":-4.0,
         "humidity":73.0,
         "temp":-4.0,
         "pressure":1013.0,
         "temp_max":-4.0
      },
      -4.0
   ],
   "result":-4.0,
   "running_time":{
      "secs":0,
      "nanos":716252
   }
}
```

aesedepece:
I'm undecided about this format, because we are losing the types of each field :shrug:

aesedepece:
What do you think about something like:
```json
{
    "type": "RadonFloat",
    "value": -4.0
}
```

aesedepece:
My rationale for going the other way was about making it easier for Sheikah users to explore the API results. But maybe it's better for the wallet to give as much information as possible, and then Sheikah can decide whether to optionally show some details or not.

tmpolaczyk:
> I'm undecided about this format, because we are losing the types of each field shrug

Another option is tagged objects:

```
{ "RadonInteger": 1 }
```

aesedepece:
> > I'm undecided about this format, because we are losing the types of each field shrug
> 
> Another option is tagged objects:
> 
> ```
> { "RadonInteger": 1 }
> ```

That one I like too.

tmpolaczyk:
Add tests for the conversion of `RadonTypes` and `RadError` to JSON

data_structures/src/radon_report.rs
```diff
@@ -60,9 +84,26 @@ where
     }
 }
 
+/*impl<RT> Serialize for RadonReport<RT>
```
tmpolaczyk, line 87: Remove before merging

aesedepece: Done

wallet/tests/data_requests.rs
```diff
@@ -63,34 +59,93 @@ fn test_radon_error_json_serialization() {
...
-        RadonScriptExecutionSettings::enable_all(),
-    )
-    .unwrap();
-
-    let json = serde_json::ser::to_string(&report).unwrap();
-
-    assert_eq!(json, "");
+#[actix_rt::test]
+#[cfg(feature = "side_effected")]
+async fn test_data_request_report_json_serialization() {
```
tmpolaczyk, line 66: We have similar tests in `node/tests/data_request_examples.rs`. They avoid side effects by storing the result of the URL locally. Perhaps we can integrate this test with the other ones.

aesedepece: This one is explicitly testing the `try_data_request_locally` method, which provides no option for operating on static data, and directly calls the `run_retrieval_report` function that has the same issue.

tmpolaczyk: If it really can't be avoided, I would prefer to run a local webserver and handle the data requests using that, rather than querying external APIs.

aesedepece: Which one do you prefer from these options for moving forward with this?
- leaving it like it is, provided that `side_effected` is not a default option and thus this test will only be run if explicitly asked for.
- making `try_data_request_locally` and `run_retrieval_report` accept data injection through its signature.
- removing this test altogether.

tmpolaczyk: A test that is only run if explicitly asked for is useless, we will merge changes that break that test and not realize it.

> * making try_data_request_locally and run_retrieval_report accept data injection through its signature.

This sounds cool.

> removing this test altogether.

This also sounds good as long as there is a test somewhere that serializes a `RadonReport` to JSON, the `RadonReport` can be mocked or generated using a different function but we need to make sure that the API does not change.

rad/src/lib.rs
```diff
@@ -27,34 +28,54 @@ pub mod types;
...
+        .map(|report| report.into_inner())
         .collect();
 
-    let aggregation_response =
-        run_aggregation_report(retrieve_responses, &request.aggregate, settings)
-            .unwrap_or_else(|error| RadonReport::from_result(Err(error), context))
-            .into_inner();
+    let aggregation_report = run_aggregation_report(retrieval_values, &request.aggregate, settings)
+        .unwrap_or_else(|error| RadonReport::from_result(Err(error), context));
+    let aggregation_value = aggregation_report.clone().into_inner();
```
tmpolaczyk, line 69: No need to clone the whole result:

```suggestion
    let aggregation_value = aggregation_report.result.clone();
```

Same for `retrieval_reports`

wallet/tests/data_requests.rs
```diff
@@ -1,12 +1,15 @@
 use std::collections::HashMap;
 
-use serde_cbor::Value;
-
-use witnet_data_structures::{radon_error::RadonError, radon_report::ReportContext};
+use witnet_data_structures::chain::RADType;
+use witnet_data_structures::{
+    chain::{RADAggregate, RADRequest, RADRetrieve, RADTally},
```
tmpolaczyk:
```suggestion
    chain::{RADAggregate, RADRequest, RADRetrieve, RADTally, RADType},
```

rad/src/lib.rs
aesedepece: But "use of moved value" ya tal...

wallet/tests/data_requests.rs
aesedepece: Done, at last!

rad/src/lib.rs
```diff
@@ -42,22 +42,30 @@ pub struct RADRequestExecutionReport {
...
 ) -> RADRequestExecutionReport {
     let context = &mut ReportContext::default();
 
-    let retrieve_responses_fut = request
-        .retrieve
-        .iter()
-        .map(|retrieve| run_retrieval_report(retrieve, settings));
-
-    let retrieval_reports: Vec<RadonReport<RadonTypes>> =
-        futures::future::join_all(retrieve_responses_fut)
```
tmpolaczyk: Now the big question: was the join_all needed to run the retrievals in parallel, or are they run in parallel by default?

tmpolaczyk: Well, it looks like this only affects the `tryDataRequest` method, the node uses a different method to execute the data requests.

tmpolaczyk: Yes, it looks like the `join_all` was needed to run the retrievals in parallel, can you open an issue and add a FIXME?

rad/src/lib.rs
```diff
@@ -26,56 +28,97 @@ pub mod types;
...
-    let retrieve_responses_fut = request
-        .retrieve
-        .iter()
-        .map(|retrieve| run_retrieval_report(retrieve));
+    let mut retrieve_responses = Vec::new();
 
-    let retrieve_responses: Vec<RadonTypes> = futures::future::join_all(retrieve_responses_fut)
-        .await
+    if let Some(inputs) = inputs_injection {
+        for (retrieve, input) in request.retrieve.iter().zip(inputs.iter()) {
```
tmpolaczyk, line 52: Using `zip` has a problem, it will stop on the shortest iterator. This may introduce hard to debug errors. Since this branch is mainly used for tests, I vote for adding a simple assert:

```suggestion
        assert_eq!(request.retrieve.len(), inputs.len());
        for (retrieve, input) in request.retrieve.iter().zip(inputs.iter()) {
```

And in the documentation of `try_data_request` add a short note about `inputs_injection`: it's used to avoid running the HTTP queries, it must be the same length as the number of sources, etc.

rad/src/lib.rs
```diff
@@ -26,56 +28,97 @@ pub mod types;
...
-            retrieve
-                .unwrap_or_else(|error| RadonReport::from_result(Err(error), context))
-                .into_inner()
+            retrieve.unwrap_or_else(|error| RadonReport::from_result(Err(error), context))
         })
         .collect();
+    let retrieval_values: Vec<RadonTypes> = retrieval_reports
+        .clone()
+        .into_iter()
+        .map(|report| report.into_inner())
```
tmpolaczyk, line 72: By using a clone here you are creating a temporary vector that is not needed, always prefer `.iter().cloned()`, or in this case:

```suggestion
        .iter()
        .map(|report| report.result.clone())
```

rad/src/lib.rs
tmpolaczyk: I tested this change and it does not give any errors, you can think of it as `AggregationReport::clone(&aggregation_report.result)` no moved values here, only references.

rad/src/lib.rs
tmpolaczyk: Created issue #1289

rad/src/types/array.rs
```diff
@@ -13,9 +14,9 @@ use crate::{
     types::{RadonType, RadonTypes},
 };
 
-pub const RADON_ARRAY_TYPE_NAME: &str = "RadonArray";
+const RADON_ARRAY_TYPE_NAME: &str = "RadonArray";
 
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq, Serialize)]
```
tmpolaczyk, line 19: I think this is unused, as the `RadonArray` serialization is implemented in `impl Serialize for RadonTypes`. Probably the same applies to all the other radon types.

```suggestion
#[derive(Clone, Debug, PartialEq)]
```

wallet/Cargo.toml
```diff
@@ -33,3 +33,7 @@ witnet_crypto = { path = "../crypto", features = ["with-serde"] }
 witnet_protected = { path = "../protected", features = ["with-serde"] }
 witnet_data_structures = { path = "../data_structures" }
 witnet_validations = { path = "../validations" }
+
+[dev-dependencies]
+actix-rt = "1.1.1"
+serde_cbor = "0.11.1"
```
tmpolaczyk, line 39: This is no longer used?

```suggestion
```

aesedepece: True

wallet/tests/data_requests.rs
```diff
@@ -0,0 +1,159 @@
...
+    for (index, retrieve_report) in report.retrieve.iter().enumerate() {
+        // Each retrieval result must match last item in each retrieval partial results
+        assert_eq!(
+            &retrieve_report.result,
+            retrieve_report
+                .partial_results
+                .clone()
+                .unwrap()
+                .last()
+                .unwrap()
```
tmpolaczyk, line 124: Clone bad, as_ref good

```suggestion
                .as_ref()
                .unwrap()
                .last()
                .as_ref() // maybe this as_ref is not needed
                .unwrap()
```

wallet/tests/data_requests.rs
```diff
@@ -0,0 +1,159 @@
...
+        &request,
+        RadonScriptExecutionSettings::enable_all(),
+        Some(&inputs),
+    )
+    .await;
+    let aggregate_report = report.aggregate.clone();
+    let tally_report = report.tally.clone();
+
+    // Number of retrieval reports should match number of sources
+    assert_eq!(&report.retrieve.len(), &request.retrieve.len());
```
tmpolaczyk, line 113: `assert_eq` already uses references internally.

```suggestion
    assert_eq!(report.retrieve.len(), request.retrieve.len());
```

wallet/tests/data_requests.rs
```diff
@@ -0,0 +1,159 @@
...
+        r#"{"time":{"updated":"Jun 9, 2020 16:42:00 UTC","updatedISO":"2020-06-09T16:42:00+00:00","updateduk":"Jun 9, 2020 at 17:42 BST"},"disclaimer":"This data was produced from the CoinDesk Bitcoin Price Index (USD). Non-USD currency data converted using hourly conversion rate from openexchangerates.org","chartName":"Bitcoin","bpi":{"USD":{"code":"USD","symbol":"&#36;","rate":"9,724.8354","description":"United States Dollar","rate_float":9724.8354},"GBP":{"code":"GBP","symbol":"&pound;","rate":"7,692.6949","description":"British Pound Sterling","rate_float":7692.6949},"EUR":{"code":"EUR","symbol":"&euro;","rate":"8,635.1092","description":"Euro","rate_float":8635.1092}}}"#,
+    ];
+    let report = try_data_request(
+        &request,
+        RadonScriptExecutionSettings::enable_all(),
+        Some(&inputs),
+    )
+    .await;
+    let aggregate_report = report.aggregate.clone();
+    let tally_report = report.tally.clone();
```
tmpolaczyk, line 110:
```suggestion
    let aggregate_report = &report.aggregate;
    let tally_report = &report.tally;
```

wallet/tests/data_requests.rs
```diff
@@ -0,0 +1,159 @@
...
+        );
+
+        // Number of partial results for each source should match source's script length + 1
+        assert_eq!(
+            retrieve_report.partial_results.clone().unwrap().len(),
+            unpack_radon_script(&request.retrieve.get(index).unwrap().script)
+                .unwrap()
+                .len()
+                + 1
+        );
```
tmpolaczyk, line 134:
```suggestion
        assert_eq!(
            retrieve_report.partial_results.as_ref().unwrap().len(),
            unpack_radon_script(&request.retrieve.get(index).unwrap().script)
                .unwrap()
                .len()
                + 1
        );
```

rad/src/lib.rs
```diff
@@ -39,26 +40,37 @@ pub struct RADRequestExecutionReport {
...
             ))
         }
+
+        retrieve_responses
     } else {
-        for retrieve in &request.retrieve {
-            retrieve_responses.push(run_retrieval_report(retrieve, settings).await)
-        }
-    }
+        block_on(join_all(
```
tmpolaczyk, line 66: Using a blocking function inside an async context, this used to panic at runtime.

But we can as remove the `async` keyword from `pub async fn` and it will should work fine.

tmpolaczyk: Oh wait, this has an easier solution, just use `join_all(...).await`

rad/src/lib.rs
```diff
@@ -26,56 +29,108 @@ pub mod types;
...
-            retrieve
-                .unwrap_or_else(|error| RadonReport::from_result(Err(error), context))
-                .into_inner()
+            retrieve.unwrap_or_else(|error| RadonReport::from_result(Err(error), context))
         })
         .collect();
+    let retrieval_values: Vec<RadonTypes> = retrieval_reports
+        .iter()
+        .cloned()
+        .map(|report| report.into_inner())
```
tmpolaczyk, line 84: This is cloning the entire report, which may be expensive when the report includes the list of partial results. We just need to clone the final result:

```suggestion
        .map(|report| report.result.clone())
```

wallet/src/actors/worker/methods.rs
```diff
@@ -39,11 +39,7 @@ impl Worker {
     pub fn run_rad_request(&self, request: types::RADRequest) -> types::RADRequestExecutionReport {
         // Block on data request retrieval because the wallet was designed with a blocking run retrieval in mind.
         // This can be made non-blocking by returning a future here and updating.
```
tmpolaczyk, line 41: This comment is no longer relevant
```suggestion
```

rad/src/lib.rs
```diff
@@ -26,56 +29,107 @@ pub mod types;
...
-    let retrieve_responses_fut = request
-        .retrieve
-        .iter()
-        .map(|retrieve| run_retrieval_report(retrieve));
+    let retrieve_responses = if let Some(inputs) = inputs_injection {
+        assert_eq!(inputs.len(), request.retrieve.len(), "Tried to locally run a data request with a number of injected sources different than the number of retrieval paths ({} != {})", inputs.len(), request.retrieve.len());
 
-    let retrieve_responses: Vec<RadonTypes> = futures::future::join_all(retrieve_responses_fut)
-        .await
+        let mut retrieve_responses = Vec::new();
```
tmpolaczyk, line 57: Now you don't need to manually push to the vector anymore:

```suggestion
        request.retrieve.iter().zip(inputs.iter()).map(|(retrieve, input)| {
            run_retrieval_with_data_report(
                retrieve, input, context, settings,
            )
        }).collect()
```

node/src/actors/rad_manager/handlers.rs
```diff
@@ -14,6 +14,7 @@ use witnet_validations::validations::{
 use crate::actors::messages::{ResolveRA, RunTally};
 
 use super::RadManager;
+use witnet_rad::script::RadonScriptExecutionSettings;
```
tmpolaczyk, line 17: This import can be merged with the other ones

PR #1281, opened 2020-06-03
feat(chain_manager): improve / implement banners for synced / live states

### LINKED ISSUE #1280: Print a banner when the node goes into Live state
It is certainly confusing for many users that they get quite flashy feedback when the node goes into Synced state, but then when it moves into Live it does so in a really quite manner.

We should implement a flashy banner for Live state too (and potentially remove the the Synced one?)

aesedepece:
Fix #1280

### DIFF
--- a/node/src/actors/chain_manager/mod.rs
+++ b/node/src/actors/chain_manager/mod.rs
@@ -97,9 +97,16 @@ pub enum ChainManagerError {
     #[fail(display = "ChainManager is not ready yet")]
     ChainNotReady,
     /// The node attempted to do an action that is only allowed while `ChainManager`
-    /// is in `Live` state.
-    #[fail(display = "The node is not yet in `Live` state")]
-    NotLive,
+    /// is in `Synced` state.
+    #[fail(
+        display = "The node is not yet in `Synced` state (current state is {:?})",
+        current_state
+    )]
+    NotSynced {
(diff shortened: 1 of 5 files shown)


tmpolaczyk:
As I stated in issue #1280:

> I vote for renaming `SYNCED` to `ALMOST_SYNCED`, renaming `LIVE` to `SYNCED`, and showing the `SYNCED` banner in the new `SYNCED` state.

My reasoning is that we shouldn't bother users with implementation details, as far as they are concerned the node only has two states: synced and not synced. Adding more banners is simply confusing. They should see one banner when the node is ready to be used.

node/src/actors/chain_manager/mod.rs
```diff
@@ -112,12 +119,12 @@ pub enum StateMachine {
     WaitingConsensus,
     /// Second state, ChainManager synchronization process
     Synchronizing,
-    /// Third state, `ChainManager` is synced and ready to start consolidating block
-    /// candidates.
-    Synced,
+    /// Third state, `ChainManager` has all the blocks in the chain ans is ready to start
```
tmpolaczyk, line 122:
```suggestion
    /// Third state, `ChainManager` has all the blocks in the chain and is ready to start
```

tmpolaczyk:
Captura de pantalla de 2020-06-04 13-17-05

Cool!

PR #1282, opened 2020-06-03
Ars merkelization

girazoki:
This PR:

- Stores the prior to last ARS.
- Stores the prior to last bn256 ARS keys ordered by reputation. In case of tie, identities are ordered by pkh.
- Computes the superblock with the prior to last oordered bn256 ARS keys.
- Checks the validity of a b2n56 key before inserting it to AltKeys.
- Adds the ARS length to the superblock

### DIFF
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -2620,16 +2620,20 @@ pub struct ChainState {
     pub reputation_engine: Option<ReputationEngine>,
     /// Node mining stats
     pub node_stats: NodeStats,
-    /// Alternative public key mapping
+    /// Last ARS members ordered by reputation
     pub alt_keys: AltKeys,
+    /// Last ARS alt keys
+    pub last_ars: AltKeys,
+    /// Last ARS keys vector ordered by reputation
+    pub last_ars_ordered_keys: Vec<Bn256PublicKey>
 }
 
 /// Alternative public key mapping: maps each secp256k1 public key hash to
 /// different public keys in other curves
-#[derive(Debug, Default, Clone, Serialize, Deserialize, PartialEq)]
+#[derive(Debug, Default, Clone, Serialize, Deserialize, PartialEq, Eq)]
 pub struct AltKeys {
     /// BN256 curve
-    bn256: HashMap<PublicKeyHash, Bn256PublicKey>,
+    pub bn256: HashMap<PublicKeyHash, Bn256PublicKey>,
 }
 
 impl AltKeys {
@@ -2664,6 +2668,19 @@ impl AltKeys {
             }
         }
     }
+
+    pub fn get_rep_ordered_bn256_list(&self, reputation_set: &TotalReputationSet<PublicKeyHash, Reputation, Alpha>) -> Vec<Bn256PublicKey> {
+        let mut pkhs: Vec<PublicKeyHash>= self.bn256.iter().map(|(pkh, _)| pkh.clone()).collect();
+        pkhs.sort();
+        let ars_rep: Vec<Reputation>= pkhs.iter().map(|pkh| reputation_set.get(pkh)).collect();
+
+        let ordered_pkhs: Vec<_> = pkhs.into_iter().enumerate().sorted_by_key(|(i, _x)| ars_rep[*i]).collect();
+
+        let ordered_alts: Vec<Bn256PublicKey> = ordered_pkhs.into_iter().filter_map(|(_, pkh)| self.get_bn256(&pkh).cloned()).collect();
+
+        ordered_alts
+    }
+    
 }
 
 impl ChainState {
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -1184,6 +1184,14 @@ impl Bn256PublicKey {
         Ok(uncompressed)
     }
 
+    pub fn is_valid(self) -> bool {
+        // Verify that the provided key is a valid public key
+        match bn256::PublicKey::from_compressed(&self.public_key.clone()){
+            Ok(_) => true,
+            _ => false,
+        }
+    }
+
     pub fn to_bytes(&self) -> Vec<u8> {
         self.public_key.clone()
     }
@@ -2666,12 +2674,30 @@ impl AltKeys {
     pub fn insert_keys_from_block(&mut self, block: &Block) {
         // Add miner bn256 public keys
         if let Some(value) = block.block_header.bn256_public_key.clone() {
-            self.insert_bn256(block.block_header.proof.proof.pkh(), value);
+            // If the provided key is valid, insert it in altkeys
+            if value.clone().is_valid(){
+                self.insert_bn256(block.block_header.proof.proof.pkh(), value);
+            }
+            else{
+                log::warn!(
+                    "Ignoring invalid bn256 public key {:02x?} inserted by PKH {:02x?} in the block",
+                    value, block.block_header.proof.proof.pkh()
+                 );
+            }
         }
         // Add bn256 public keys from commitment transactions
         for commit in &block.txns.commit_txns {
             if let Some(value) = commit.body.bn256_public_key.clone() {
-                self.insert_bn256(commit.body.proof.proof.pkh(), value);
+                // If the provided key is valid, insert it in altkeys
+                if value.clone().is_valid() {
+                    self.insert_bn256(commit.body.proof.proof.pkh(), value);
+                }
+                else{
+                    log::warn!(
+                        "Ignoring invalid bn256 public key {:02x?} inserted by PKH {:02x?} in the commit",
+                        value, block.block_header.proof.proof.pkh()
+                    );
+                }
             }
         }
     }
@@ -4232,5 +4258,32 @@ mod tests {
         let ordered_keys = alt_keys.get_rep_ordered_bn256_list(&trs);
 
         assert_eq!(expected_order, ordered_keys);
+    }
+
+    #[test]
+    fn test_is_valid_true() {
+        let bls_pk =
+            Bn256PublicKey::from_secret_key(&Bn256SecretKey::from_slice(&[3; 32]).unwrap())
+                .unwrap();
+        assert!(bls_pk.is_valid());
+    }
+
+    #[test]
+    fn test_is_valid_false_invalid_length() {
+        let bls_pk =
+            Bn256PublicKey::from_secret_key(&Bn256SecretKey::from_slice(&[3; 32]).unwrap())
+                .unwrap();
+        let bls_pk_2 = Bn256PublicKey {
+            public_key: bls_pk.public_key[1..63].to_vec(),
+        };
+        assert_eq!(bls_pk_2.is_valid(), false);
+    }
+
+    #[test]
+    fn test_is_valid_false() {
+        let bls_pk = Bn256PublicKey {
+            public_key: vec![1; 65],
+        };
+        assert_eq!(bls_pk.is_valid(), false);
     }
 }
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -2620,12 +2620,12 @@ pub struct ChainState {
     pub reputation_engine: Option<ReputationEngine>,
     /// Node mining stats
     pub node_stats: NodeStats,
-    /// Last ARS members ordered by reputation
+    /// Alternative public key mapping
     pub alt_keys: AltKeys,
-    /// Last ARS alt keys
-    pub last_ars: AltKeys,
+    /// Last ARS pkhs
+    pub last_ars: Vec<PublicKeyHash>,
     /// Last ARS keys vector ordered by reputation
-    pub last_ars_ordered_keys: Vec<Bn256PublicKey>
+    pub last_ars_ordered_keys: Vec<Bn256PublicKey>,
 }
 
 /// Alternative public key mapping: maps each secp256k1 public key hash to
@@ -2633,7 +2633,7 @@ pub struct ChainState {
 #[derive(Debug, Default, Clone, Serialize, Deserialize, PartialEq, Eq)]
 pub struct AltKeys {
     /// BN256 curve
-    pub bn256: HashMap<PublicKeyHash, Bn256PublicKey>,
+    bn256: HashMap<PublicKeyHash, Bn256PublicKey>,
 }
 
 impl AltKeys {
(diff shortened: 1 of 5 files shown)


data_structures/src/chain.rs
```diff
@@ -2568,6 +2572,28 @@ impl AltKeys {
...
+        reputation_set: &TotalReputationSet<PublicKeyHash, Reputation, Alpha>,
+    ) -> Vec<Bn256PublicKey> {
+        let mut pkhs: Vec<PublicKeyHash> = self.bn256.iter().map(|(pkh, _)| pkh.clone()).collect();
+        pkhs.sort();
+        let ars_rep: Vec<Reputation> = pkhs.iter().map(|pkh| reputation_set.get(pkh)).collect();
+
+        let ordered_pkhs: Vec<_> = pkhs
+            .into_iter()
+            .enumerate()
+            .sorted_by_key(|(i, _x)| ars_rep[*i])
```
tmpolaczyk, line 2587: `ars_rep` is not needed, you can get the reputation directly here

```suggestion
            .sorted_by_key(|pkh| reputation_set.get(pkh))
```

tmpolaczyk: Similarly you also don't need `ordered_alts` because you can add `filter_map` right after the `sorted_by_key`.

girazoki: RUST101

data_structures/src/chain.rs
```diff
@@ -2570,7 +2578,10 @@ impl AltKeys {
     pub fn insert_keys_from_block(&mut self, block: &Block) {
         // Add miner bn256 public keys
         if let Some(value) = block.block_header.bn256_public_key.clone() {
-            self.insert_bn256(block.block_header.proof.proof.pkh(), value);
+            // If the provided key is valid, insert it in altkeys
+            if value.clone().is_valid(){
+                self.insert_bn256(block.block_header.proof.proof.pkh(), value);
+            }
```
tmpolaczyk, line 2584: Maybe we should print a warning in the `else` case. And the same validation needs to be done for commit keys (just below). And add tests :eyes:

data_structures/src/chain.rs
```diff
@@ -4037,4 +4092,103 @@ mod tests {
         assert_eq!(calculate_backup_witnesses(11, 3), 22);
         assert_eq!(calculate_backup_witnesses(11, 4), 44);
     }
+
+    #[test]
+    fn test_ordered_alts_no_tie() {
+        let mut trs = TotalReputationSet::new();
+        let mut alt_keys = AltKeys::default();
+
+        let p1 = PublicKey::from_bytes([1; 33]);
```
tmpolaczyk, line 4101: This can be confusing because the keys are ordered by pkh, not by public key. So how about creating the pkh directly?

```suggestion
        let p1_pkh = PublicKeyHash::from_bytes(&[0x01 as u8; 20]).unwrap();
```

girazoki: very true

node/src/actors/chain_manager/mod.rs
```diff
@@ -473,6 +474,21 @@ impl ChainManager {
...
 
+                // Store the ARS and the order of the keys
+                let trs = reputation_engine.trs();
+                let current_ars = reputation_engine
+                    .ars()
+                    .active_identities()
+                    .cloned()
+                    .sorted()
+                    .collect();
+                let alt_keys = self.chain_state.alt_keys.clone();
```
tmpolaczyk, line 485:
```suggestion
                let alt_keys = &self.chain_state.alt_keys;
```

node/src/actors/chain_manager/superblock.rs
```diff
@@ -84,8 +87,7 @@ impl SuperBlockState {
             Some(x) if x == sbv.superblock_index => self
                 .previous_ars_identities
                 .as_ref()
-                .map(|x| x.contains(&sbv.secp256k1_signature.public_key.pkh())),
-            // If the index is not the same as the current one, but it is within an acceptable range
+                .map(|x| x.contains(&sbv.secp256k1_signature.public_key.pkh())), // If the index is not the same as the current one, but it is within an acceptable range
```
tmpolaczyk, line 90: Accidentally deleted newline?

node/src/actors/chain_manager/superblock.rs
```diff
@@ -187,23 +192,30 @@ pub fn mining_build_superblock(
...
         last_block,
         last_block_in_previous_superblock,
     })
 }
 
+/// Takes a set of keys and calculates their hashes roots to be used as leaves.
+pub fn hash_key_leaves(ars_ordered_keys: &[Bn256PublicKey]) -> Vec<Hash> {
+    ars_ordered_keys
+        .iter()
+        .map(|bn256| Hash::SHA256(calculate_sha256(&bn256.clone().to_uncompressed().unwrap()).0))
```
tmpolaczyk, line 210: Do you think we should move this logic to `impl Hashable for Bn256PublicKey`? In other words, do we want to always hash the uncompressed public key, or do we want to sometimes also hash the compressed public key? I would prefer to always hash the uncompressed so we don't run into errors, but hashing the compressed key should be faster (because we don't have to uncompress it).

node/src/actors/chain_manager/superblock.rs
girazoki: yep

node/src/actors/chain_manager/superblock.rs
girazoki: what we could run into is that we need to hash both, and create a merkle root with both. But in both cases I think we benefit  from moving it to hashable

node/src/actors/chain_manager/mining.rs
```diff
@@ -136,18 +135,14 @@ impl ChainManager {
             // FIXME: After Reputation Merkelitation, only the ARS Members from the previous consolidated
             // Superblock will be added, to avoid include addresses that could be wrong later by
             // block reorganization
```
tmpolaczyk, line 137: Is this fixed now? We use `last_ars` to create the superblock, but is `last_ars` the ARS from the previous superblock?

tmpolaczyk: It looks like `last_ars` is the ARS from the previous block, so that's not implemented yet, but it should be easy, right? Just add a `ars_from_last_superblock` field to `ChainState`.

data_structures/src/chain.rs
```diff
@@ -1162,11 +1164,32 @@ impl Bn256PublicKey {
...
+    pub fn to_uncompressed(&self) -> Result<Vec<u8>, failure::Error> {
+        // Verify that this slice is a valid public key
+        let uncompressed =
+            bn256::PublicKey::from_compressed(&self.public_key)?.to_uncompressed()?;
+        Ok(uncompressed)
+    }
+
+    pub fn is_valid(&self) -> bool {
+        // Verify that the provided key is a valid public key
+        match bn256::PublicKey::from_compressed(&self.public_key) {
```
lrubiorod, line 1176:
```suggestion
        bn256::PublicKey::from_compressed(&self.public_key).is_ok()
```

data_structures/src/chain.rs
```diff
@@ -2559,15 +2586,48 @@ impl AltKeys {
...
+                }
             }
         }
     }
+
+    pub fn get_rep_ordered_bn256_list(
+        &self,
+        reputation_set: &TotalReputationSet<PublicKeyHash, Reputation, Alpha>,
+    ) -> Vec<Bn256PublicKey> {
+        let mut pkhs: Vec<PublicKeyHash> = self.bn256.iter().map(|(pkh, _)| *pkh).collect();
```
lrubiorod, line 2619: why you sort 2 times?

data_structures/src/chain.rs
```diff
@@ -2559,15 +2586,48 @@ impl AltKeys {
...
+        pkhs.sort();
+
+        let ordered_alts: Vec<_> = pkhs
+            .into_iter()
+            .enumerate()
```
lrubiorod, line 2624: if you later do not use the index, you do not need to use enumerate

data_structures/src/chain.rs
```diff
@@ -2559,15 +2586,48 @@ impl AltKeys {
...
+            .sorted_by_key(|(_, pkh)| reputation_set.get(pkh))
```
lrubiorod, line 2625: Here the reputation is sorted from lower to higher, but I thought it is used to stop when we reach 2/3 of reputation. So...  would be it better sort from high to low??

node/src/actors/chain_manager/mod.rs
```diff
@@ -484,6 +485,21 @@ impl ChainManager {
...
+                    .active_identities()
+                    .cloned()
+                    .sorted()
+                    .collect();
+                let alt_keys = &self.chain_state.alt_keys;
+
+                let ordered_alts: Vec<Bn256PublicKey> = alt_keys.get_rep_ordered_bn256_list(trs);
+                // last ars with previous block ars info
+                self.chain_state.last_ars = current_ars;
+                self.chain_state.last_ars_ordered_keys = ordered_alts;
```
lrubiorod, line 501: last_ars is also ordered, so that names could be misunderstanding. Because the main difference is that one are bls and the other secp??

node/src/actors/chain_manager/superblock.rs
```diff
@@ -246,17 +256,17 @@ mod tests {
         };
 
         let expected_superblock = SuperBlock {
+            ars_length: 1,
             data_request_root: dr_merkle_root_1,
             tally_root: tally_merkle_root_1,
-            ars_root: PublicKeyHash::default().hash(),
+            ars_root: Hash::default(),
```
lrubiorod, line 262: you has a variable name called default_hash that you could use
```suggestion
            ars_root: default_hash,
```

node/src/actors/chain_manager/superblock.rs
```diff
@@ -246,17 +256,17 @@ mod tests {
...
             index: 0,
             last_block: block.hash(),
             last_block_in_previous_superblock: default_hash,
         };
 
         let superblock =
-            mining_build_superblock(&[block], &[PublicKeyHash::default()], 0, default_hash)
-                .unwrap();
+            mining_build_superblock(&[block], &[Hash::default()], 0, default_hash).unwrap();
```
lrubiorod, line 269:
```suggestion
            mining_build_superblock(&[block], &[default_hash], 0, default_hash).unwrap();
```

node/src/actors/chain_manager/superblock.rs
```diff
@@ -311,21 +321,18 @@ mod tests {
         };
 
         let expected_superblock = SuperBlock {
+            ars_length: 1,
             data_request_root: expected_superblock_dr_root,
             tally_root: expected_superblock_tally_root,
-            ars_root: PublicKeyHash::default().hash(),
+            ars_root: Hash::default(),
```
lrubiorod, line 327:
```suggestion
            ars_root: default_hash,
```

node/src/actors/chain_manager/superblock.rs
```diff
@@ -311,21 +321,18 @@ mod tests {
...
 
-        let superblock = mining_build_superblock(
-            &[block_1, block_2],
-            &[PublicKeyHash::default()],
-            0,
-            default_hash,
-        )
-        .unwrap();
+        let superblock =
+            mining_build_superblock(&[block_1, block_2], &[Hash::default()], 0, default_hash)
```
lrubiorod, line 334:
```suggestion
            mining_build_superblock(&[block_1, block_2], &[default_hash], 0, default_hash)
```

data_structures/src/chain.rs
```diff
@@ -2559,15 +2586,48 @@ impl AltKeys {
     pub fn insert_keys_from_block(&mut self, block: &Block) {
         // Add miner bn256 public keys
         if let Some(value) = block.block_header.bn256_public_key.clone() {
-            self.insert_bn256(block.block_header.proof.proof.pkh(), value);
+            // If the provided key is valid, insert it in altkeys
+            if value.is_valid() {
```
lrubiorod, line 2590: should this validation be done inside of the insert_bn256 method?

node/src/actors/chain_manager/superblock.rs
```diff
@@ -578,4 +680,31 @@ mod tests {
         assert_eq!(sbs.add_vote(&v3), AddSuperBlockVote::MaybeValid);
         assert_eq!(sbs.votes_on_local_superlock, HashSet::new());
     }
+
+    #[test]
+    fn hash_uncompressed_leaves_() {
```
aesedepece, line 685: Weird function name.

data_structures/src/chain.rs
girazoki: it does not really matter I guess, as long as you also compute low to high in Ethereum

node/src/actors/chain_manager/mod.rs
girazoki: the truth is last_ars does not need to be actually sorted. I will fix that.

data_structures/src/chain.rs
girazoki: So that all nodes end up with the same reputation based sorting. We first sort by pkh, get the reputation of each node, then sort by reputation. In the case of two members having the same reputation, these will be sorted by pkh

lrubiorod: if you use sorted_by instead of sorted_by_key you has to use a Ordering ( the result of a cmp function). And if you do that, you can concat several comparitions using ".then()". And with that you avoid to sort 2 times.

image

tmpolaczyk: You can use `sorted_by_key(|x| (x.a, x.b))` to sort by `x.a`, and in case of tie sort by `x.b`.

Since we want to sort by reputation and in case of tie by pkh, it should be `(reputation, pkh)`. But in any case it's easier to just add tests and see.

node/src/actors/chain_manager/mining.rs
girazoki: I think it is fixed, the`'is_valid` method checks against the `previous_ars_identities` in superblock.rs and I think that is the method checked to see if you need to forward right? If that is the case then `previous_ars_identities` is now the previous superblock ARS so it should be fixed

data_structures/src/chain.rs
girazoki: I do not follow you guys

girazoki: ` let ordered_alts: Vec<_> = pkhs
            .into_iter()
            .sorted_by_key(|&pkh| (reputation_set.get(&pkh), pkh))
            .filter_map(|pkh| self.get_bn256(&pkh).cloned())
            .collect();
`
Would this be sufficient then? Sort by rep, if tie sort by pkh

data_structures/src/chain.rs
girazoki: makes sense

data_structures/src/chain.rs
```diff
@@ -573,7 +572,17 @@ pub struct SuperBlock {
...
+        let superblock_data = [
+            &self.ars_length.to_be_bytes()[..],
+            self.ars_root.as_ref(),
+            self.data_request_root.as_ref(),
+            &self.index.to_be_bytes()[..],
+            self.last_block.as_ref(),
+            self.last_block_in_previous_superblock.as_ref(),
+            self.tally_root.as_ref(),
+        ]
+        .concat();
```
tmpolaczyk, line 584: Maybe move this to a method of `SuperBlock`, so it is decoupled from the `Hashable` impl. And add some documentation there, something along the lines of "serialize the superblock by converting all the fields to bytes and concatenating them". Also note that this serialization method is super simple, but it only works if all the fields have a known fixed size.

```suggestion
        let superblock_data = self.serialize_as_bytes();
```

tmpolaczyk: So add more tests, and also explain why aren't we using protocol buffers (because this way it's easier to verify superblocks in the EVM)

data_structures/src/chain.rs
```diff
@@ -2558,16 +2600,31 @@ impl AltKeys {
...
+
+    pub fn get_rep_ordered_bn256_list(
+        &self,
+        reputation_set: &TotalReputationSet<PublicKeyHash, Reputation, Alpha>,
+    ) -> Vec<Bn256PublicKey> {
+        let pkhs: Vec<PublicKeyHash> = self.bn256.iter().map(|(pkh, _)| *pkh).collect();
+
+        let ordered_alts: Vec<_> = pkhs
+            .into_iter()
+            .sorted_by_key(|&pkh| (reputation_set.get(&pkh), pkh))
```
tmpolaczyk, line 2622: Now you don't need this collect, you can combine the two iterators into one:

```suggestion
        self.bn256.iter().map(|(pkh, _)| *pkh)
            .sorted_by_key(|&pkh| (reputation_set.get(&pkh), pkh)) 
```

node/src/actors/chain_manager/mining.rs
```diff
@@ -135,22 +134,12 @@ impl ChainManager {
         let superblock_period = u32::from(chain_info.consensus_constants.superblock_period);
         // Everyone creates superblocks, but only ARS members sign and broadcast them
         if current_epoch % superblock_period == 0 {
-            // FIXME(#1236): ARS Members have to include the BLS signature instead of PublicKeyHash
-            // FIXME: After Reputation Merkelitation, only the ARS Members from the previous consolidated
             // Superblock will be added, to avoid include addresses that could be wrong later by
             // block reorganization
```
tmpolaczyk, line 138: This should also be removed?
```suggestion
```

PR #1283, opened 2020-06-04
Attach node address to telemetry reports

aesedepece:
This attaches node addresses to telemetry reports for any errors or panics happening in the node runtime, after the node address has been already loaded into `ChainManager`.

This also allows compiling with no telemetry features at all.

### DIFF
--- a/node/src/actors/chain_manager/actor.rs
+++ b/node/src/actors/chain_manager/actor.rs
@@ -284,8 +284,13 @@ impl ChainManager {
             })
             .and_then(|res, act, _ctx| {
                 act.own_pkh = Some(res);
-                log::debug!("Public key hash successfully loaded from signature manager");
-                log::info!("PublicKeyHash: {}", res);
+                log::debug!("Address successfully loaded from signature manager");
+                log::info!("Node identity / address: {}", res);
+
+                // Telemetry is configured here to avoid a race condition if configured directly
+                // upon actor start because of asynchronous nature of `get_key` function.
+                act.configure_telemetry_scope();
+
                 actix::fut::ok(())
             })
             .wait(ctx);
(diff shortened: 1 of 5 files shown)


node/src/actors/chain_manager/actor.rs
```diff
@@ -308,4 +317,19 @@ impl ChainManager {
             })
             .wait(ctx);
     }
+
+    /// Put some basic information into the scope of the telemetry service (Sentry)
+    #[cfg(feature = "telemetry")]
+    fn configure_telemetry_scope(&mut self) {
+        log::debug!("Configuring telemetry scope");
+        println!("{:?}", self.own_pkh.unwrap());
```
tmpolaczyk, line 325: Remove before merging
```suggestion
```

node/src/actors/chain_manager/actor.rs
```diff
@@ -47,6 +47,10 @@ impl Actor for ChainManager {
             .ok();
 
         self.secp = Some(CryptoEngine::new());
+
+        // Telemetry is configured here to avoid a race condition if configured directly
+        // upon actor start because of asynchronous nature of `get_key` function.
+        self.configure_telemetry_scope();
```
tmpolaczyk, line 53: Isn't the comment contradictory? This is the actor start

node/src/actors/chain_manager/actor.rs
```diff
@@ -308,4 +317,19 @@ impl ChainManager {
             })
             .wait(ctx);
     }
+
+    /// Put some basic information into the scope of the telemetry service (Sentry)
+    #[cfg(feature = "telemetry")]
+    fn configure_telemetry_scope(&mut self) {
```
tmpolaczyk, line 323: What happens if the node panics before calling this function?

node/src/actors/chain_manager/actor.rs
aesedepece: HAHAHAHA

node/src/actors/chain_manager/actor.rs
aesedepece: No info about node identity, but we get the report anyway.

node/src/actors/chain_manager/actor.rs
aesedepece: Ooops I forgot to remove that function call from there. It belongs in `get_pkh`.

node/src/actors/chain_manager/actor.rs
tmpolaczyk: Perfect

PR #1284, opened 2020-06-04
Rename CLI commands

clbartoli:
This PR renames some of the CLI commands so that the _get_ prefix disappear.
The old commands still working as an alias of the new ones.

The docs/ folder is deleted


docs/interface/json-rpc.md
```diff
@@ -70,15 +70,15 @@ Response:
 {"jsonrpc":"2.0","result":true,"id":1}
 ```
 
-#### getBalance
+#### balance
```
tmpolaczyk, line 73: I think the documentation in `docs/` is deprecated

aesedepece: That's correct. We should simply remove that folder.

src/cli/node/with_node.rs
```diff
@@ -248,7 +258,8 @@ pub enum Command {
         node: Option<SocketAddr>,
     },
     #[structopt(
-        name = "getUtxoInfo",
+        name = "utxoInfo",
+        alias = "getUtxoInfo",
```
aesedepece, line 262:
```suggestion
        name = "utxos"
        alias = "coins"
        alias = "utxoInfo",
        alias = "getUtxoInfo",
```

tmpolaczyk: `getCoins`?

PR #1285, opened 2020-06-04
log(node): remove periodic reputation logging

lrubiorod:
Due to we have the getReputation cli method, and currently there are so many nodes. Do you think is needed to keep showing the rep every epoch?


aesedepece:
I'm for removing it too, provided that sorting the TRS on the spot is not free, and that every single line we log is a clear I/O side effected operation that surely has a performance penalty.

PR #1295, opened 2020-06-12
[wip] feat(node): implement only minted utxos are allowed to collateralize

lrubiorod:


lrubiorod:
Closed according to #1500

PR #1297, opened 2020-06-15
feat(docker): make the node keep running even after crashing

### LINKED ISSUE #1296: Add auto-restart in Docker image runner script

aesedepece:
(More info in the comments inside the commit changes)

Close #1296

### DIFF
--- a/docker/witnet-rust/runner.sh
+++ b/docker/witnet-rust/runner.sh
@@ -1,13 +1,54 @@
 #!/bin/bash
 
+ERROR_BANNER="
+
+ ██████╗██████╗  █████╗ ███████╗██╗  ██╗██╗
+██╔════╝██╔══██╗██╔══██╗██╔════╝██║  ██║██║
+██║     ██████╔╝███████║███████╗███████║██║
+██║     ██╔══██╗██╔══██║╚════██║██╔══██║╚═╝
+╚██████╗██║  ██║██║  ██║███████║██║  ██║██╗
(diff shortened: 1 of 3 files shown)


docker/witnet-rust/runner.sh
```diff
@@ -1,13 +1,22 @@
...
+# Run in a loop so it is automatically restarted upon crashing
+while true; do
+    # Run the public address detector if enabled
+    if [[ "$PUBLIC_ADDR_DISCOVERY" == "true" ]]; then
+        ./ip_detector.sh
+    fi
+
+    # Run the node itself, using configuration from the default directory and passing down any arguments that may be
+    # appended when running "docker run"
+    /tmp/witnet-raw -c /.witnet/config/witnet.toml "$@" && wait
```
tmpolaczyk, line 21: What is `wait`?

```suggestion
    /tmp/witnet-raw -c /.witnet/config/witnet.toml "$@"
```

tmpolaczyk:
This can be a problem when the node panics on purpose. For example, if there is some error parsing the configuration file, or reading the database. How can we make sure that the node operator will notice that the node crashed? Maybe adding a banner and a sleep 30 after the node crashes will be enough. Or maybe disable the automatic node restarting for the first 90 seconds or so, so that any panics during startup are not ignored. Not sure what is the current user experience regarding that.

aesedepece:
The assumption here is that node operators using Docker are unlikely to take action upon panics, as they are in a quite controlled environment where if anything goes wrong, it's rather a bug than a misconfiguration.

I'm totally in for the banner + delay for the very reasons you explained related to early panics on start.

Issue #1293, opened 2020-06-12
The bridge does not get out of including state if posting DR fails


girazoki:
Apparently if the bridge fails to post a DR in the WItnet network (i.e., not enough mature UTXOs) the internal state of the DR stays in "including" and does not retry at a later point in time.

tmpolaczyk:
The bridge assumes that `sendRequest` will never fail, and the `DataRequestTransaction` will eventually be included in a witnet block. Currently the workaround is to restart the bridge node, or to run a different bridge that will claim that request and post its proof of inclusion.

There are two situations in which the bridge can get stuck in this state:

If the `sendRequest` call fails (for example when the node has not enough balance or is not synced), the data request should be retried later, as long as the data request is still claimed by this node. This can be easily detected as there is a line in the code that says "ERROR: send request".

If the `sendRequest` call succeeds but the transaction is never included in a witnet block (because of too small fees or some double spend error), the data request should also be retried later, but we need to decide a timeout. How many witnet blocks do we want to wait before sending the transaction again, potentially with a higher fee? How do we avoid creating two transactions and getting the two included into blocks, do we need to implement a `replaceByFee` method in the node?

Ideally, the bridge should not claim any data requests until the node is synchronized, and after that it should only claim data requests if it can pay for them (and if it's profitable). But that does not solve this issue, we need to add a timeout or something that will make the node send the data request transactions to witnet again.

Conclusion: unclear, let's just add a timeout

tmpolaczyk:
This is fixed but only if `post_to_witnet_more_than_once = true`. It would be nice to detect when the `sendRequest` call fails (for example when the node has not enough balance or is not synced), or even refrain from claiming requests if the node does not have enough balance.

Tommytrg:
This issue belongs to the old bridge implementation

PR #1311, opened 2020-06-16
feat(validations): include validation for vt and dr that exceed the b…

### LINKED ISSUE #1314: Return validation result in JSON-rpc when a transaction is created
Currently, when we create a transaction using CLI methods, the transaction created is validated inside of the AddTransaction handler, and not in the BuildVtt or BuildDrt handler (they send a "notify" message to AddTransaction). It means that a possible error that happen in the validation methods is notified to the node, but not to the JSON-rpc, so the client could think that the transaction is properly added to the network when it is not.

lrubiorod:
…lock limit

Also Close #1314

### DIFF
--- a/node/src/actors/chain_manager/handlers.rs
+++ b/node/src/actors/chain_manager/handlers.rs
@@ -839,19 +839,27 @@ impl Handler<BuildVtt> for ChainManager {
             Ok(vtt) => {
                 let fut = transaction_factory::sign_transaction(&vtt, vtt.inputs.len())
                     .into_actor(self)
-                    .then(|s, _act, ctx| match s {
+                    .then(|s, act, ctx| match s {
                         Ok(signatures) => {
                             let transaction =
                                 Transaction::ValueTransfer(VTTransaction::new(vtt, signatures));
                             let tx_hash = transaction.hash();
-                            ctx.notify(AddTransaction { transaction });
-
-                            actix::fut::ok(tx_hash)
+                            Box::new(
+                                act.handle(AddTransaction { transaction }, ctx)
+                                    .map(move |_, _, _| tx_hash),
+                            )
                         }
                         Err(e) => {
                             log::error!("Failed to sign value transfer transaction: {}", e);
 
-                            actix::fut::err(e)
+                            let res: Box<
+                                dyn ActorFuture<
+                                    Item = Hash,
+                                    Error = failure::Error,
+                                    Actor = ChainManager,
+                                >,
+                            > = Box::new(actix::fut::err(e));
+                            res
                         }
                     });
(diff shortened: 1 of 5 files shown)


validations/src/tests.rs
```diff
@@ -1801,7 +1893,10 @@ fn data_request_value_overflow() {
     });
     assert_eq!(
         x.unwrap_err().downcast::<TransactionError>().unwrap(),
-        TransactionError::FeeOverflow,
+        TransactionError::DataRequestWeightLimitExceeded {
+            weight: 41680557,
+            max_weight: 80000
+        },
```
tmpolaczyk, line 1899: I think you can avoid this change by increasing the max dr weight used inside `test_drtx`

tmpolaczyk:
Let's not forget about this in issue #1293: the bridge needs to handle data request transactions that might never be included in a block because of their size.

node/src/actors/chain_manager/handlers.rs
```diff
@@ -839,14 +842,39 @@ impl Handler<BuildVtt> for ChainManager {
...
+                                epoch_constants,
+                                &mut signatures_to_verify,
+                                act.consensus_constants().max_vt_weight,
+                            ) {
+                                Ok(_) => {
+                                    let transaction = Transaction::ValueTransfer(vtt_tx);
+                                    let tx_hash = transaction.hash();
 
-                            actix::fut::ok(tx_hash)
+                                    ctx.notify(AddTransaction { transaction });
```
tmpolaczyk, line 868: Wouldn't it be better to use .send here? The validation is already done in AddTransaction, this way we will validate twice.

node/src/actors/chain_manager/handlers.rs
```diff
@@ -894,19 +902,27 @@ impl Handler<BuildDrt> for ChainManager {
...
-                            ctx.notify(AddTransaction { transaction });
-
-                            actix::fut::ok(tx_hash)
+                            Box::new(
+                                act.handle(AddTransaction { transaction }, ctx)
+                                    .map(move |_, _, _| tx_hash),
+                            )
                         }
                         Err(e) => {
-                            log::error!("Failed to sign data request transaction: {}", e);
```
tmpolaczyk, line 907: Copy-paste error

node/src/actors/chain_manager/handlers.rs
```diff
@@ -894,19 +902,27 @@ impl Handler<BuildDrt> for ChainManager {
...
+                            log::error!("Failed to sign value transfer transaction: {}", e);
```
tmpolaczyk, line 916: Copy paste error

Issue #1291, opened 2020-06-11
Discussion: Add a JSON RPC method to check the current data request pool


drcpu-github:
One thing I wanted to add to my blockchain explorer project is a visualization / list of data request transactions that are still pending (not currently included in any block yet, e.g., like in [1]). I don't think it is currently possible to get those using a JSON RPC call to your witnet node.

I believe dataRequestReport will only work for requests that have been added to a block? Would this be a useful addition? Is this feasible to implement?

[1] https://www.blockchain.com/nl/btc/unconfirmed-transactions

tmpolaczyk:
> I believe dataRequestReport will only work for requests that have been added to a block?

Correct. You can use `getTransaction` to see pending transactions if you know the transaction hash. The response will contain `"blockHash":"pending"`:

    {"jsonrpc":"2.0","result":{"blockHash":"pending","transaction":{"DataRequest": ... 

But there is no method to get all the pending transactions, but that would certainly be useful, because it can help us to test things when we implement the block size limit.

So yes, we can implement a method like `getMempool` or `getPendingTransactions` that returns a list of pending transaction hashes. Since in witnet we have different transaction types, the returned transactions will be categorized by type (the names of the fields may change):

```json
{
  "value_transfers": ["407b635a160bb981ab1572fb6c525654469cc71e822404a78e8da17c744d7fbc", "2fa34a6294d95b67ee9703bc474ed704fa1ae50de91ae4a1d141dc12485d421a"],
  "data_requests": ["c12aa15ab0e16c81ed763aaca54a22a426e19f77702f8e4de702e2793b7bda86"],
  "commits": [],
  "reveals": []
}
```

And then to get more information about one transaction, one can use `getTransaction` with the corresponding transaction hash.

In the case of commit and reveal transactions, it is not clear whether we want to include them or not. Because commits expire after 1 epoch = 45 seconds, and reveals are usually included in the next block or two. I think we can not include them for now, just in case we end up having thousands of commits per epoch, making this method return useless data. On the other hand, the number of commits seen by the node the last epoch can be a useful statistic, but that's something related to the number of active nodes that can be included in a different method.

Alternative: create one JSON-RPC method for each transaction type: `getPendingVtt`, `getPendingDrt`.

[...]

PR #1312, opened 2020-06-16
feat(jsonrpc): implement getMempool method

tmpolaczyk:
Basic support for #1291 

TODO: rename to `getPendingTransactions`?


aesedepece:
I'm for `pendingTransactions` plus `mempool` as an alias, plus the typical `get` alisases.

PR #1313, opened 2020-06-16
Add synchronization progress to nodeStatus command

### LINKED ISSUE #1278: Make `nodeStats` report `ChainManager`'s status and synchronization progress
## Issue statement
Current users find it quite frustrating when they have to "guess" the state and synchronization progress of their own node by skimming through the logs or issuing commands like `getBalance`, which is intended for a different purpose and prompts the users with errors they are expected to make sense of.

## Proposed solution
We should make `nodeStats` report `ChainManager`'s status (`WaitingConsensus` / `Syncing` / `Synced` / `Live`) and synchronization progress.

tmpolaczyk:
Close #1278 

Add synchronization progress to nodeStats command

```
Block mining stats:
- Proposed blocks: 0
- Blocks included in the block chain: 0
Data Request mining stats:
- Times with eligibility to mine a data request: 0
- Proposed commits: 0
- Accepted commits: 0
- Slashed commits: 0
Synchronization progress: 99.99% (186960/186964)
```

```
Block mining stats:
- Proposed blocks: 1
- Blocks included in the block chain: 0
Data Request mining stats:
- Times with eligibility to mine a data request: 0
- Proposed commits: 0
- Accepted commits: 0
- Slashed commits: 0
The node is synchronized and the current epoch is 186966
```

### DIFF
--- a/node/src/actors/json_rpc/json_rpc_methods.rs
+++ b/node/src/actors/json_rpc/json_rpc_methods.rs
@@ -28,7 +28,7 @@ use self::mock_actix::SystemService;
 use crate::{
     actors::{
         chain_manager::{ChainManager, ChainManagerError, StateMachine},
-        epoch_manager::EpochManager,
+        epoch_manager::{EpochManager, EpochManagerError},
         inventory_manager::{InventoryManager, InventoryManagerError},
         messages::{
             AddCandidates, AddTransaction, BuildDrt, BuildVtt, GetBalance, GetBlocksEpochRange,
@@ -705,8 +705,8 @@ pub fn send_value(params: Result<BuildVtt, jsonrpc_core::Error>) -> JsonRpcResul
 pub struct SyncStatus {
     /// The hash of the top consolidated block and the epoch of that block
     pub chain_beacon: CheckpointBeacon,
-    /// The current epoch
-    pub current_epoch: u32,
+    /// The current epoch, or None if the epoch 0 is in the future
+    pub current_epoch: Option<u32>,
     /// Is the node synchronized?
     pub synchronized: bool,
 }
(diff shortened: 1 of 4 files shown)


src/cli/node/json_rpc_client.rs
```diff
@@ -869,6 +869,35 @@ pub fn get_node_stats(addr: SocketAddr) -> Result<(), failure::Error> {
         node_stats.slashed_count
     );
 
+    let request = r#"{"jsonrpc": "2.0","method": "syncStatus", "id": "1"}"#;
+    let response = send_request(&mut stream, &request)?;
+    let sync_status: SyncStatus = parse_response(&response)?;
+
+    if sync_status.synchronized {
+        println!(
```
lrubiorod, line 877: Should we include the nodeStats call if the node is Synchronized?

node/src/actors/json_rpc/json_rpc_methods.rs
```diff
@@ -700,21 +699,27 @@ pub fn send_value(params: Result<BuildVtt, jsonrpc_core::Error>) -> JsonRpcResul
...
+pub struct SyncStatus {
+    /// The hash of the top consolidated block and the epoch of that block
+    pub chain_beacon: CheckpointBeacon,
+    /// The current epoch
+    pub current_epoch: u32,
+    /// Num peers inbound
+    // TODO: remove? we have getPeers
+    pub num_peers_inbound: u32,
+    /// Num peers outbound
+    // TODO: remove? we have getPeers
```
lrubiorod, line 713: agree, it would have to be in only one CLI method, if we would include here I think we should remove the other method. But I prefer different methods

PR #1318, opened 2020-06-17
feat: add transaction hashes to getBlock RPC call

drcpu-github:
Added transaction hashes as separate vectors to the JSON object returned by the getBlock RPC call. I use this in my blockchain explorer to parse and add all transactions to my database.

EDIT: Forgot this, but apparently I should add my witnet ID to pull requests -- WIT_IBQ0Z


drcpu-github:
Fixing the formatter check failures in a new pull request.

PR #1325, opened 2020-06-22
feat(node): disable superblock creation

tmpolaczyk:
This is a temporal workaround to reduce the message overhead in testnet 0.9.0c


node/src/actors/chain_manager/mining.rs
```diff
@@ -137,7 +137,7 @@ impl ChainManager {
 
         let superblock_period = u32::from(chain_info.consensus_constants.superblock_period);
         // Everyone creates superblocks, but only ARS members sign and broadcast them
-        if current_epoch % superblock_period == 0 {
+        if self.create_superblocks && current_epoch % superblock_period == 0 {
```
lrubiorod, line 140: add comment to disable this later

PR #1326, opened 2020-06-22
feat: remove consensus constants from witnet.toml file

lrubiorod:


config/src/defaults.rs
```diff
@@ -82,11 +82,15 @@ pub trait Defaults {
     }
 
     /// Timestamp at the start of epoch 0
-    fn consensus_constants_checkpoint_zero_timestamp(&self) -> i64;
+    fn consensus_constants_checkpoint_zero_timestamp(&self) -> i64 {
+        // Testnet configuration
+        // Friday, 19-Jun-2020, 09:00 UTC
+        1592557200
+    }
```
tmpolaczyk, line 89: This should not be changed here, but on the `impl Defaults for Testnet`. It's currently set to:

        // June 1st, 2019
        1_559_347_200

PR #1330, opened 2020-06-23
feat(docker): stop overwriting witnet.toml config file

### LINKED ISSUE #1329: Make docker image not overwrite witnet.toml config file
After #1326 has been merged, we can safely modify the docker image so that it does not overwrite the `witnet.toml` config file.

Is simply should try to copy the default file over the path where a existing file may exist, but fail silently if that's the case. This is exactly the behavior of the `no-clobber` mode of the `cp` command:

```
       -n, --no-clobber
              do not overwrite an existing file (overrides a previous -i option)
```

aesedepece:
Fix #1329

### DIFF
--- a/docker/witnet-rust/migrator.sh
+++ b/docker/witnet-rust/migrator.sh
@@ -19,8 +19,9 @@ function migrate {
   mkdir -p "$WITNET_CONFIG_FOLDER" &&
   log "Ensuring that storage folder '$WITNET_STORAGE_FOLDER' does exist" &&
   mkdir -p "$WITNET_STORAGE_FOLDER" &&
-  log "Moving configuration files into configuration folder '$WITNET_CONFIG_FOLDER'" &&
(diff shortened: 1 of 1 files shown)


docker/witnet-rust/migrator.sh
```diff
@@ -20,7 +20,7 @@ function migrate {
   log "Ensuring that storage folder '$WITNET_STORAGE_FOLDER' does exist" &&
   mkdir -p "$WITNET_STORAGE_FOLDER" &&
   log "Moving configuration files into configuration folder '$WITNET_CONFIG_FOLDER'" &&
-  mv "$CONFIG_FILE" "$WITNET_CONFIG_FOLDER/witnet.toml" &&
+  mv -n "$CONFIG_FILE" "$WITNET_CONFIG_FOLDER/witnet.toml" &&
```
tmpolaczyk, line 23: Would it be possible to add a log line indicating that the config file was not overwritten?

aesedepece: Sure!

PR #1331, opened 2020-06-23
chore: update consensus constants for T9c reboot

lrubiorod:


config/src/defaults.rs
```diff
@@ -356,7 +356,7 @@ impl Defaults for Testnet {
     }
 
     fn consensus_constants_checkpoint_zero_timestamp(&self) -> i64 {
-        // Friday, 19-Jun-2020, 09:00 UTC
-        1_592_557_200
+        // Tuesday, 24-Jun-2020, 09:00 UTC
```
tmpolaczyk, line 359:
```suggestion
        // Wednesday, 24-Jun-2020, 09:00 UTC
```

PR #1332, opened 2020-06-23
feat(config): stop reading consensus constants from config file

aesedepece:
From now on, consensus constants shall be only defined at compile time.

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -369,23 +369,8 @@ impl Config {
             Environment::Testnet => &Testnet,
         };
 
-        let consensus_constants = match config.environment {
-            // When in mainnet, ignore the [consensus_constants] section of the configuration
-            Environment::Mainnet => {
-                let consensus_constants_no_changes = PartialConsensusConstants::default();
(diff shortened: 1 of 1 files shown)


config/src/config.rs
```diff
@@ -369,23 +369,8 @@ impl Config {
...
 
-        let consensus_constants = match config.environment {
-            // When in mainnet, ignore the [consensus_constants] section of the configuration
-            Environment::Mainnet => {
-                let consensus_constants_no_changes = PartialConsensusConstants::default();
-                // Warn the user if the config file contains a non-empty [consensus_constant] section
-                if config.consensus_constants != consensus_constants_no_changes {
-                    log::warn!(
-                        "Consensus constants in the configuration are ignored when running mainnet"
-                    );
```
tmpolaczyk, line 380: It would be nice to keep this warning, but removing "when running mainnet"

tmpolaczyk: Not possible because logging is enabled after parsing the config, so any logs here will not be displayed.

PR #1335, opened 2020-06-24
chore: update known_peers for 0.9.0c

aesedepece:


aesedepece:
This is what was used for `0.9.0c`, published here for the sake of transparency.

For `9.1`, I'd like to collect a list of community nodes that we can all rely upon. It is important that they have a decent uptime and they stick with the network for a while, ideally preserving their IPs across multiple network reboots.

bertux:
> This is what was used for `0.9.0c`, published here for the sake of transparency.
> 
> For `9.1`, I'd like to collect a list of community nodes that we can all rely upon. It is important that they have a decent uptime and they stick with the network for a while, ideally preserving their IPs across multiple network reboots.

I would be glad to join this list of community nodes. My goal is to have 10 nodes distributed in 10 different regions.

PR #1339, opened 2020-06-24
Prioritize transactions

lrubiorod:


data_structures/src/chain.rs
```diff
@@ -1363,19 +1363,21 @@ impl RADTally {
     }
 }
 
-type WeightedHash = (u64, Hash);
-type WeightedVTTransaction = (u64, VTTransaction);
+type PrioritizedHash = (u64, Hash);
```
tmpolaczyk, line 1366: This `u64` is `nanowits / weight`, which looks like it should be a `f64`, but we cannot use floats as keys because they cannot be sorted. There is this crate that provides sortable floats, if we decide to use `f64` instead of `u64`:

https://docs.rs/ordered-float/1.0.2/ordered_float/struct.OrderedFloat.html

data_structures/src/chain.rs
```diff
@@ -1363,19 +1363,21 @@ impl RADTally {
...
 /// [`Hash`](Hash) and iteration over the
 /// transactions sorted from by transactions with bigger fees to
 /// transactions with smaller fees.
 #[derive(Debug, Default, Clone)]
 pub struct TransactionsPool {
-    vt_transactions: HashMap<Hash, WeightedVTTransaction>,
-    sorted_index: BTreeSet<WeightedHash>,
+    vt_transactions: HashMap<Hash, PrioritizedVTTransaction>,
+    sorted_vt_index: BTreeSet<PrioritizedHash>,
     // Currently transactions related with data requests don't use weight
```
tmpolaczyk, line 1378: Outdated comment :D

```suggestion
```

data_structures/src/chain.rs
```diff
@@ -1471,7 +1474,7 @@ impl TransactionsPool {
     ///
     /// assert_eq!(pool.vt_len(), 0);
     ///
-    /// pool.insert(transaction);
+    /// pool.insert(transaction,0);
```
tmpolaczyk, line 1477: May I ask for 1 extra space in all the doc examples?

```suggestion
    /// pool.insert(transaction, 0);
```

node/src/actors/chain_manager/mining.rs
```diff
@@ -898,11 +899,11 @@ fn build_block(
...
             );
             value_transfer_txns.push(vt_tx.clone());
-            transaction_fees += transaction_fee;
-            vt_weight += transaction_weight;
+            transaction_fees = transaction_fees.saturating_add(transaction_fee);
+            vt_weight = new_vt_weight;
         }
 
-        if new_vt_weight > max_vt_weight - min_vt_weight {
+        if vt_weight > max_vt_weight.saturating_sub(min_vt_weight) {
```
tmpolaczyk, line 906: Sorry, what is this line for? I don't understand

lrubiorod: If the free space in the block for VTTransactions is less than the minimum weight for a VTTransactions, it not make sense to continue checking the rest of pending transactions

data_structures/src/chain.rs
```diff
@@ -1798,21 +1806,28 @@ impl TransactionsPool {
...
-        let weight = 0; // TODO: weight = transaction-fee / transaction-weight
+    #[allow(clippy::cast_precision_loss)]
+    pub fn insert(&mut self, transaction: Transaction, fee: u64) {
         let key = transaction.hash();
 
         match transaction {
             Transaction::ValueTransfer(vt_tx) => {
-                self.vt_transactions.insert(key, (weight, vt_tx));
-                self.sorted_index.insert((weight, key));
+                let weight = std::cmp::max(u64::from(vt_tx.weight()), 1) as f64;
```
tmpolaczyk, line 1819: I think that if you remove this `u64::from` you can also remove `     #[allow(clippy::cast_precision_loss)]`
```suggestion
                let weight = std::cmp::max(vt_tx.weight(), 1) as f64;
```

tmpolaczyk: Also, dividing by 0 is not a problem with floats, `1.0/0.0` is `+inf`, so a transaction with 0 weight will have maximum priority.

node/src/actors/chain_manager/mining.rs
```diff
@@ -1353,13 +1319,23 @@ mod tests {
...
+        transaction_pool.insert(transaction_3, 10);
+        assert_eq!(transaction_pool.vt_len(), 3);
+
+        let mut unspent_outputs_pool = UnspentOutputsPool::default();
+        let output1 = ValueTransferOutput {
+            time_lock: 0,
+            pkh: MY_PKH_1.parse().unwrap(),
+            value: 1_000_000,
+        };
+        //let tx_output1 = VTTransactionBody::new(vec![], vec![output1.clone()]);
```
tmpolaczyk, line 1334: There are comments like this in the other tests as well

data_structures/src/chain.rs
aesedepece: _Infinite priority_

data_structures/src/chain.rs
```diff
@@ -1873,7 +1888,7 @@ impl TransactionsPool {
     /// // TODO: assert!(tx1.weight() >= tx2.weight());
```
tmpolaczyk, line 1888: Remove this TODO, we would need to assert that `tx1.fee() / tx1.weight() >= tx2.fee() / tx2.weight()` but that's not easy.
```suggestion
```

PR #1341, opened 2020-06-25
Several fixes to p2p / bucketing

aesedepece:
Namely:
- fix(p2p): evict peer addresses from `tried` buckets upon connection failure
- refactor(p2p): improve readability of `remove_from_tried`
- tests(p2p): add tests for new and tried buckets
- refactor(p2p): avoid redundant hashmap lookups
- feat(node): refuse creating multiple inbound sessions for similar IP ranges

### DIFF
--- a/node/src/actors/connections_manager/mod.rs
+++ b/node/src/actors/connections_manager/mod.rs
@@ -1,17 +1,19 @@
+use std::net::SocketAddr;
+
 use actix::prelude::*;
 use futures::Stream;
 use tokio::net::{TcpListener, TcpStream};
 
-use crate::actors::{
-    messages::{Create, InboundTcpConnect, ResolverResult},
-    sessions_manager::SessionsManager,
-};
-
-use crate::config_mngr;
-
-use crate::actors::peers_manager::PeersManager;
-use std::net::SocketAddr;
 use witnet_p2p::sessions::SessionType;
+
+use crate::{
+    actors::{
+        messages::{Create, InboundTcpConnect, ResolverResult},
+        sessions_manager::SessionsManager,
+    },
+    config_mngr,
+    peers_manager::PeersManager,
+};
 
 mod actor;
 mod handlers;
--- a/node/src/actors/connections_manager/mod.rs
+++ b/node/src/actors/connections_manager/mod.rs
@@ -9,6 +9,8 @@ use crate::actors::{
 
 use crate::config_mngr;
 
+use crate::actors::peers_manager::PeersManager;
+use std::net::SocketAddr;
 use witnet_p2p::sessions::SessionType;
 
 mod actor;
@@ -70,6 +72,7 @@ impl ConnectionsManager {
     fn process_connect_addr_response(
         response: Result<ResolverResult, MailboxError>,
         session_type: SessionType,
+        address: &SocketAddr,
     ) -> actix::fut::FutureResult<(), (), Self> {
         // Process the Result<ResolverResult, MailboxError>
         match response {
@@ -81,7 +84,14 @@ impl ConnectionsManager {
                 // Process the ResolverResult
                 match res {
                     Err(error) => {
-                        log::debug!("Failed to connect to a peer with error: {:?}", error);
+                        log::debug!(
+                            "Failed to connect to peer address {} with error: {:?}",
+                            address,
+                            error
+                        );
+                        // Try to evict this address from `tried` buckets
+                        PeersManager::remove_address_from_tried(address);
+
                         actix::fut::err(())
                     }
                     Ok(stream) => {
--- a/node/src/actors/peers_manager/mod.rs
+++ b/node/src/actors/peers_manager/mod.rs
@@ -5,6 +5,7 @@ use actix::{
     SystemService, WrapFuture,
 };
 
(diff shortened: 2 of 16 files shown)


aesedepece:
I'm remaking the last feature so it is more efficient and we can double check when consolidating.

p2p/src/sessions/mod.rs
```diff
@@ -372,3 +423,59 @@ pub struct GetConsolidatedPeersResult {
     /// The address shown here can be used to connect to this peers in the future.
     pub outbound: Vec<SocketAddr>,
 }
+
+/// A convenient wrapper around a collection of network ranges.
+/// This enables efficient tracking of the inbound connections we have so as to prevent sybil
+/// machines from monopolizing our inbound peers table.
+#[derive(Default)]
+pub struct NetworkRangesCollection {
+    inner: BTreeSet<Vec<u8>>,
```
tmpolaczyk, line 432: This would be much more efficient if we know the size of the ip range and use a fixed sized array instead. In ipv4 it's 2 bytes but in ipv6 it's 8 bytes. Since I'm not sure if we currently support ipv6, how about setting it to 2 bytes:

```suggestion
    inner: BTreeSet<[u8; 2]>,
```

And we can use

```rust
let (_, range, _) = split_socket_addresses(address);
let range = [range[0], range[1]];
```

tmpolaczyk, line 432: Unless we want the ip ranges to be sorted, using a `HashSet` is more efficient than using a `BTreeSet`:

```suggestion
    inner: HashSet<Vec<u8>>,
```

p2p/tests/buckets.rs
```diff
@@ -0,0 +1,176 @@
...
+        let peer_addr = ip("192.168.1.1:21337");
+
+        let new_index_sk_1 = tried_bucket_index(sk_1, &peer_addr);
+        let new_index_sk_2 = tried_bucket_index(sk_2, &peer_addr);
+
+        assert_ne!(new_index_sk_1, new_index_sk_2);
+    }
+}
+
+fn new_bucket_index(sk: u64, socket_addr: &SocketAddr, src_socket_addr: &SocketAddr) -> u16 {
```
tmpolaczyk, line 159: This looks like it shouldn't be here, why not use `new_bucket_index` from `p2p` crate?

p2p/tests/sessions.rs
```diff
@@ -568,3 +568,50 @@ fn p2p_sessions_register_more_than_limit() {
...
 }
+
+#[test]
+/// Check that peer addresses are considered "similar" if the first two octets of the IP are
+/// matching.
+fn p2p_peer_address_is_similar_to_inbound_session() {
+    // Create sessions struct
+    let mut sessions = Sessions::<String>::default();
+
+    let inbound_address_1 = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 8002);
```
tmpolaczyk, line 579: I prefer just parse and unwrap:
```suggestion
    let inbound_address_1 = "127.0.0.1:8002".parse().unwrap();
```

p2p/tests/buckets.rs
aesedepece: > This looks like it shouldn't be here, why not use `new_bucket_index` from `p2p` crate?

Because the other is a method and takes `&self`.

p2p/tests/sessions.rs
aesedepece: Me too, but I simply copy-pasted from the rest of tests in the crate :rofl:

p2p/tests/sessions.rs
```diff
@@ -568,3 +568,50 @@ fn p2p_sessions_register_more_than_limit() {
     assert_eq!(sessions.get_num_outbound_sessions(), 3);
     assert_eq!(sessions.num_missing_outbound(), 0);
 }
+
+#[test]
+/// Check that peer addresses are considered "similar" if the first two octets of the IP are
+/// matching.
+fn p2p_peer_address_is_similar_to_inbound_session() {
```
tmpolaczyk, line 575: Can we have a similar test that checks that after calling unregister_session, the ip range is available again?

node/src/actors/peers_manager/mod.rs
```diff
@@ -5,6 +5,7 @@ use actix::{
     SystemService, WrapFuture,
 };
 
+use crate::actors::messages::RemoveAddressesFromTried;
 use crate::{
     actors::{connections_manager::ConnectionsManager, messages::OutboundTcpConnect, storage_keys},
```
tmpolaczyk, line 10: Merge this import and run `just fmt`:

```suggestion
use crate::{
    actors::{connections_manager::ConnectionsManager, messages::{OutboundTcpConnect, RemoveAddressesFromTried}, storage_keys},
```

node/src/actors/connections_manager/mod.rs
```diff
@@ -9,6 +9,8 @@ use crate::actors::{
 
 use crate::config_mngr;
 
+use crate::actors::peers_manager::PeersManager;
```
tmpolaczyk, line 12: Merge import

PR #1344, opened 2020-06-26
chore: level down some spammy logging lines

aesedepece:
As suggested by @lrubiorod

### DIFF
--- a/node/src/actors/session/actor.rs
+++ b/node/src/actors/session/actor.rs
@@ -63,7 +63,7 @@ impl Actor for Session {
                 .into_actor(self)
                 .then(|res, act, ctx| match res {
                     Ok(Ok(_)) => {
-                        log::debug!(
+                        log::trace!(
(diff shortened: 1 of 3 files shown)


node/src/actors/session/actor.rs
```diff
@@ -153,7 +153,7 @@ impl Session {
                     }
                     Ok(Err(CheckpointZeroInTheFuture(zero))) => {
                         let date = pretty_print(zero, 0);
-                        log::warn!("Checkpoint zero is in the future ({:?}). Delaying chain bootstrapping until then.", date);
+                        log::warn!("Network bootstrapping is scheduled for {:?}. The node will remain idle and delay chain bootstrapping until then. Wait for it!", date);
```
tmpolaczyk, line 156: This line is duplicated in `node/src/actors/chain_manager/actor.rs`, not sure why.

PR #1357, opened 2020-06-29
feat(config): add initial difficulty as a consensus constant

### LINKED ISSUE #1345: Add an initial difficulty in the consensus constants
Currently we have an initial difficulty in the code as a patch for Testnet 0.9.0-c. We should have this difficulty as a parameter in the ConsensusConstants

lrubiorod:
Close #1345

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -437,6 +437,14 @@ pub fn consensus_constants_from_partial(
             .reputation_issuance
             .to_owned()
             .unwrap_or_else(|| defaults.consensus_constants_reputation_issuance()),
+        initial_difficulty: config
+            .initial_difficulty
+            .to_owned()
+            .unwrap_or_else(|| defaults.consensus_constants_initial_difficulty()),
+        epochs_with_initial_difficulty: config
+            .epochs_with_initial_difficulty
+            .to_owned()
+            .unwrap_or_else(|| defaults.consensus_constants_epochs_with_initial_difficulty()),
         reputation_issuance_stop: config
             .reputation_issuance_stop
             .to_owned()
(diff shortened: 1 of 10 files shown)


aesedepece:
We may apply something like this to all config settings that are expressing duration:

```rust
    #[partial_struct(serde(
        default,
        deserialize_with = "from_secs",
        rename = "initial_difficulty_duration"
    ))]
```

lrubiorod:
> We may apply something like this to all config settings that are expressing duration:
> 
> ```rust
>     #[partial_struct(serde(
>         default,
>         deserialize_with = "from_secs",
>         rename = "initial_difficulty_duration"
>     ))]
> ```
I called duration, but is the number of epochs... Should we use another less misleading name?

aesedepece:
I'd try to be explicit about what the unit of measure is.

tmpolaczyk:
Let's name it "proof_of_work_epochs" :P

tmpolaczyk:
Remove ignore from test in validations/src/validations.rs:

    // FIXME: remove this hack

tmpolaczyk:
A few things that I would like to see documented somewhere:

* initial_difficulty ignores the backup factor. So if initial_difficulty = 1000 and the backup factor is 8, the probability to mine a block will be 1/1000 and not 8/1000

* epochs_with_initial_difficulty includes the epoch that it is set to. So if epochs_with_initial_difficulty = 1000, the block for epoch 1000 will also have the initial difficulty. The first block without initial difficulty will be the block for epoch 1001. This also means that setting epochs_with_initial_difficulty = 0 will keep this feature enabled for epoch 0, but that shouldn't be an issue because the block for epoch 0 is the genesis block, and so it follows a different validation.

Issue #1346, opened 2020-06-29
Implement an "ice bucket" for peer addresses that shouldn't be retried in a while


aesedepece:
- The "ice bucket" should contain peer addresses for which we want to make sure we don't try to establish a session for a while.
- Any address that is removed from the existing buckets (`new` & `tried`) should be inserted into the ice bucket.
- The ice bucket should be cleaned up periodically, removing all addresses that have been sitting there for more time than set in the configuration file.
- Peers that fall out of "hard consensus" (base superblocks) may also go into the ice bucket.

aesedepece:
Solved by #1358

PR #1358, opened 2020-06-29
feat(peers): implement ice bucket

aesedepece:
Ice bucket is not being used yet, but data structure is already in place.

Progress towards solving #1346

### DIFF
--- a/p2p/src/peers/mod.rs
+++ b/p2p/src/peers/mod.rs
@@ -4,10 +4,9 @@ use rand::{thread_rng, Rng};
 use serde::{Deserialize, Serialize};
 use std::{
     cmp,
-    collections::{HashMap, HashSet},
+    collections::HashMap,
     convert::TryFrom,
     fmt,
-    hash::{Hash, Hasher},
     net::{IpAddr, Ipv4Addr, SocketAddr},
     time::Duration,
 };
@@ -17,7 +16,7 @@ use witnet_crypto::hash::calculate_sha256;
 use witnet_util::timestamp::get_timestamp;
 
 /// Peer information being used while listing available Witnet peers
-#[derive(Debug, Deserialize, Eq, Serialize)]
+#[derive(Debug, Deserialize, Serialize)]
 pub struct PeerInfo {
     /// The socket address for a potential peer
     pub address: SocketAddr,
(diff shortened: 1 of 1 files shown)


aesedepece:
The basic functionality is already there, but it's missing tests :wink:

p2p/tests/buckets.rs
```diff
@@ -4,6 +4,71 @@ use witnet_p2p::peers::{
...
+
+        assert!(can_ice);
+
+        let is_iced = peers.ice_bucket_contains(&address);
+
+        assert!(is_iced)
+    }
+
+    #[test]
+    fn test_icing_blocks_the_entire_ip() {
```
tmpolaczyk, line 28: What's the reasoning behind this?

So, assuming that an ip is "iced" when there is no node running there, if I run two nodes on the same ip but later stop one, there is a chance that a new node will try to connect to the old node address, see that there is no node running there, and block the other perfectly valid node?

Also, an attacker can broadcast peers with an invalid port to move that ip into the ice bucket, essentially banning a node from connecting to certain ips. So if there is a node running on 192.168.1.1:21337, an attacker can broadcast 192.168.1.1:1 as a peer address, and other nodes will try to connect, see that there is no node there, and "ice" this ip.

Also, assuming that nodes running on different consensus constants are "iced", this may interfere with private testnets: 2 nodes running in different testnets in the same ip may cause other nodes to block this ip range.

p2p/src/peers/mod.rs
```diff
@@ -64,6 +148,37 @@ impl Peers {
...
+        let lumped = LumpedPeerInfo::from(addr);
+        let (contains, needs_removal) = self
+            .ice_bucket
+            .get(&lumped)
+            .map(|entry| {
+                // If the address was iced more than `ice_period` seconds ago, we can remove it from
+                // the ice bucket and pretend it was not even there in the first place.
+                let needs_removal = entry.0.timestamp < timestamp.saturating_sub(ice_period);
+
+                (!needs_removal, needs_removal)
```
lrubiorod, line 171: This line blows my mind

p2p/src/peers/mod.rs
```diff
@@ -201,7 +320,10 @@ impl Peers {
                 {
                     log::trace!("Removed a tried peer address: \n{}", self);
 
-                    self.tried_bucket.remove(&bucket_index)
+                    self.tried_bucket.remove(&bucket_index).map(|entry| {
+                        self.ice_peer_address(&entry.address);
```
lrubiorod, line 324: In the try_peer method (https://github.com/witnet/witnet-rust/blob/master/node/src/actors/peers_manager/mod.rs#L106), there is a hack where the tried peer is removed, but is not wrong.

In this case, when a new peer wants to use a tried index that is already in use, but it is not tested from a long time, first of all, that peer is removed from tried, and before to try the new peer, the old one is tried. Currently the index is empty, so if the old one success in the handshake, it will be added again to the tried bucket but with a updated timestamp, if not, the index will be free for the next filler try adding a new peer in that index

p2p/src/peers/mod.rs
```diff
@@ -159,15 +277,18 @@ impl Peers {
...
+    /// If an address is in the ice bucket, it gets ignored
     /// Returns all the overwritten or rejected addresses
     pub fn add_to_tried(
         &mut self,
         address: SocketAddr,
     ) -> Result<Option<SocketAddr>, failure::Error> {
-        // Insert address
-        let result = if !address.ip().is_unspecified() {
+        // Insert address, silently ignoring unspecified addresses and "iced" addresses
+        let result = if !address.ip().is_unspecified() && !self.ice_bucket_contains(&address) {
```
girazoki, line 287: Wondering whether this is something the feeler can do. Assuming peers in the iced bucket are not announced, if all peers have a particular peer in the ice bucket it might end up in the situation where the peer can not get out from the ice bucket

p2p/tests/buckets.rs
aesedepece: True, the implications of this are quite serious. Will modify it so it only ices the specific port.

p2p/src/peers/mod.rs
aesedepece: Peers should be able get out of the ice bucket as soon as they get announced to someone that didn't iced them, or someone that iced them more than some certain time ago ("ice has melted").

p2p/src/peers/mod.rs
aesedepece: :trollface:

p2p/src/peers/mod.rs
```diff
@@ -4,43 +4,113 @@ use rand::{thread_rng, Rng};
...
 }
 
 /// Peers TBD
-#[derive(Default, Serialize, Deserialize)]
+#[derive(Default, Deserialize, Serialize)]
 pub struct Peers {
-    /// Bucket for tried addresses
-    tried_bucket: HashMap<u16, PeerInfo>,
+    /// Bucket for "iced" addresses (will not be tried in a while)
+    ice_bucket: HashSet<PeerInfo>,
```
tmpolaczyk, line 94: This looks like it should be a `HashMap<SocketAddr, u64>`, where `u64` is the timestamp. I would prefer to do it that way to so we can remove the custom implementation of `PartialEq`, because that's dangerous and it may cause us problems in the future. Not sure about the performance differences between `HashSet<PeerInfo>` + custom `PartialEq`, and `HashMap<SocketAddr, u64>`.

aesedepece: True, that's not needed anymore :+1:

PR #1359, opened 2020-06-29
Check beacon during handshake

### LINKED ISSUE #1348: Don't peer with nodes on incompatible chain tips
- Current protocol requires peers to periodically exchange beacon messages (`LAST_BEACON`) in a declaration of their local chain tip.
- These beacons are currently only used for building a rough idea of network consensus by comparing beacons from all your peers.
- We can also use them to prevent nodes from peering in the first place if they declare incompatible chain tips.
- For this to work efficiently, `LAST_BEACON` needs to be exchanged during the protocol handshake.
- Beacons that signal a different block for the same epoch are considered incompatible.
- Beacons that belong in the future (but no further than current date) are only acceptable on inbound sessions. Outbound sessions should only be established with peers that are in perfect synchrony with us, or ahead of us.
- Dropped sessions may trigger the remote address to be put in an ice bucket (#1346 )

tmpolaczyk:
Close #1348

BREAKING CHANGE

This PR adds a `LastBeacon` field to the `Version` message. This field is used in the handshake to ensure that a node never connects to outbound peers whose beacon is behind its own, or whose beacon has a different block hash. This is also used by the feeler process to ensure that the peers that are added to tried are in consensus with the node.

### DIFF
--- a/node/src/actors/messages.rs
+++ b/node/src/actors/messages.rs
@@ -23,6 +23,7 @@ use witnet_data_structures::{
     },
     radon_report::RadonReport,
     transaction::{CommitTransaction, RevealTransaction, Transaction},
+    types::LastBeacon,
 };
 use witnet_p2p::{
     error::SessionsError,
@@ -865,6 +866,17 @@ pub struct LogMessage {
 
 impl Message for LogMessage {
     type Result = SessionsUnitResult;
+}
+
+/// Set the LastBeacon
+#[derive(Clone, Debug)]
+pub struct SetLastBeacon {
+    /// Current tip of the chain
+    pub beacon: LastBeacon,
+}
+
+impl Message for SetLastBeacon {
+    type Result = ();
 }
 
 // JsonRpcServer messages (notifications)
--- a/node/src/actors/session/handlers.rs
+++ b/node/src/actors/session/handlers.rs
@@ -601,7 +601,7 @@ fn handshake_version(session: &mut Session, sender_address: &Address) -> Vec<Wit
             session.magic_number,
             session.public_addr,
             session.remote_addr,
-            session.current_epoch,
+            session.last_beacon.clone(),
         );
         responses.push(version);
     }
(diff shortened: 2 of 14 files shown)


node/src/actors/messages.rs
```diff
@@ -867,6 +868,17 @@ impl Message for LogMessage {
     type Result = SessionsUnitResult;
 }
 
+/// Get number of inbound and outbound sessions
```
girazoki, line 871: I do not get this comment

node/src/actors/session/handlers.rs
```diff
@@ -575,16 +589,75 @@ fn handshake_verack(session: &mut Session) {
...
+    current_beacon: &LastBeacon,
+    received_beacon: &LastBeacon,
+) -> Result<(), HandshakeError> {
+    match current_beacon
+        .highest_block_checkpoint
+        .checkpoint
+        .cmp(&received_beacon.highest_block_checkpoint.checkpoint)
+    {
+        // current_checkpoint < received_checkpoint: received beacon is ahead of us
+        Ordering::Less => Ok(()),
```
girazoki, line 602: Can we also check that the provided epoch is not ahead of the epoch we truly should be at?

node/src/actors/session/handlers.rs
```diff
@@ -575,16 +589,75 @@ fn handshake_verack(session: &mut Session) {
     flags.verack_rx = true;
 }
 
+fn check_beacon_compatibility(
```
girazoki, line 592: Also tests for this?

node/src/actors/session/handlers.rs
tmpolaczyk: Yes, but we need to make sure that we always set `highest_block_checkpoint` to the current epoch, except when the node is not synced, than the `highest_block_checkpoint` should be the epoch of the last block.

tmpolaczyk: Ok, this will make it impossible to bootstrap a testnet after the epoch 0. This is because new nodes always use (0, genesis_hash) as the beacon, so if the current epoch is not 0 other peers will reject them. And since all the nodes are in the same state, they will never be able to start to mine blocks.

girazoki: OK forget about that then!

tmpolaczyk: Oh right, I check that the epoch is equal, but instead we can check if the epoch is equal or lower. This will allow to bootstrap after epoch 0 because nodes will be able to peer with other nodes that are on a compatible beacon. But is that what we want?

PR #1361, opened 2020-06-30
Almost synched if no consensus

### LINKED ISSUE #1349: Upon consensus failure, go `AlmostSynced` instead of `WaitingConsensus`
- Current implementation of the `ChainManager` state machine triggers a state transition from `Synced` to `WaitingConsensus` every time we detect consensus failure among the beacons received from our outbound peers.
- Such transition forces an involuntary desync from whatever the chain tip is in reality, because real time block consolidation is paused.
- Not signaling the majority chain tip makes connections to be dropped by your remote inbound peers, which in turn leaves them in a bad position and most likely also cause them to transition into `WaitingConsensus`.
- After a critical mass of nodes are in `WaitingConsensus` state, there's no way for the network to recover and get on the same page for resuming blocks production.
- It is believed that we can safely modify the state machine so that upon consensus failure, it transitions into `AlmostSynced` state — as in this state the nodes do not propose blocks, but still validate and consolidate candidates in real time, so they will not be perceived as "out of consensus" or "lagging behind" to other nodes that may rely on them as outbound peers.

### LINKED ISSUE #1351: While `AlmostSynced`, do not drop peers that seem to be in consensus with my own consolidation
- Please read #1349 carefully.
- While being in `AlmostSynced`, there's no real need to drop outbound sessions with peers that are signaling the same chain tip as we do. Moreover, dropping those sessions is dangerous because it can lead us to a poorer view of what the network consensus is.
- The proposed change is that, exclusively when the `ChainManager` state machine is in `AlmostSynced` state, we shall not drop peers that signal the same chain tip as the last one we consolidated (under the assumption that we do validate and consolidate block candidates while in `AlmostSynced` state)
- Sessions with outbound peers that report incompatible chain tips (see #1348) should be dropped as they are today, and their addresses may be put into an ice bucket as per #1346 .

girazoki:
This PR:

- Goes to almost synced state if while synched we lost consensus on our peers.
- While in almost synced state, if drops only peers whose beacons are different from ours.

Close #1351 
Close #1349

### DIFF
--- a/node/src/actors/chain_manager/handlers.rs
+++ b/node/src/actors/chain_manager/handlers.rs
@@ -790,7 +790,7 @@ impl Handler<PeersBeacons> for ChainManager {
                 // we need to rewind one epoch
                 if pb_len == 0 {
                     log::warn!("[CONSENSUS]: We have not received any beacons for this epoch");
-                    self.sm_state = StateMachine::WaitingConsensus;
+                    self.sm_state = StateMachine::AlmostSynced;
                 }
 
                 let our_beacon = self.get_chain_beacon();
@@ -809,7 +809,7 @@ impl Handler<PeersBeacons> for ChainManager {
                         self.initialize_from_storage(ctx);
                         log::info!("Restored chain state from storage");
 
-                        self.sm_state = StateMachine::WaitingConsensus;
+                        self.sm_state = StateMachine::AlmostSynced;
 
                         Ok(peers_to_unregister)
                     }
@@ -821,7 +821,7 @@ impl Handler<PeersBeacons> for ChainManager {
                             our_beacon
                         );
 
-                        self.sm_state = StateMachine::WaitingConsensus;
+                        self.sm_state = StateMachine::AlmostSynced;
 
                         // Unregister all peers to try to obtain a new set of trustworthy peers
                         Ok(peers_to_unregister)
--- a/node/src/actors/chain_manager/handlers.rs
+++ b/node/src/actors/chain_manager/handlers.rs
@@ -1,7 +1,9 @@
 use actix::{fut::WrapFuture, prelude::*};
 use futures::future::Future;
 use itertools::Itertools;
-use std::{cmp::Ordering, collections::BTreeMap, collections::HashMap, convert::TryFrom};
+use std::{
+    cmp::Ordering, collections::BTreeMap, collections::HashMap, convert::TryFrom, net::SocketAddr,
+};
 
 use witnet_data_structures::{
     chain::{
@@ -625,6 +627,15 @@ impl PeersBeacons {
         // Some(None) (most of the peers did not send a beacon)
         .and_then(|x| *x)
     }
+
+    /// Collects the peers to unregister based on the beacon they reported and the beacon to be compared it with
+    pub fn decide_peers_to_unregister(&self, beacon: CheckpointBeacon) -> Vec<SocketAddr> {
+        // Unregister peers which have a different beacon
+        (&self.pb)
+            .iter()
+            .filter_map(|(p, b)| if *b != Some(beacon) { Some(*p) } else { None })
+            .collect()
+    }
 }
 
 impl Handler<PeersBeacons> for ChainManager {
@@ -646,9 +657,8 @@ impl Handler<PeersBeacons> for ChainManager {
         // Calculate the consensus, or None if there is no consensus
         let consensus_threshold = self.consensus_c as usize;
         let consensus = peers_beacons.consensus(consensus_threshold);
-        let pb = peers_beacons.pb;
         let outbound_limit = peers_beacons.outbound_limit;
-        let pb_len = pb.len();
+        let pb_len = peers_beacons.pb.len();
         let peers_needed_for_consensus = outbound_limit
             .map(|x| {
                 // ceil(x * consensus_threshold / 100)
@@ -656,16 +666,7 @@ impl Handler<PeersBeacons> for ChainManager {
             })
             .unwrap_or(1);
         let peers_to_unregister = if let Some(consensus_beacon) = consensus {
-            // Consensus: unregister peers which have a different beacon
-            pb.into_iter()
-                .filter_map(|(p, b)| {
-                    if b != Some(consensus_beacon) {
-                        Some(p)
-                    } else {
-                        None
-                    }
-                })
-                .collect()
+            peers_beacons.decide_peers_to_unregister(consensus_beacon)
         } else if pb_len < peers_needed_for_consensus {
             // Not enough outbound peers, do not unregister any peers
             log::debug!(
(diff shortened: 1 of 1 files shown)


node/src/actors/chain_manager/handlers.rs
```diff
@@ -617,9 +627,9 @@ impl Handler<PeersBeacons> for ChainManager {
         // Calculate the consensus, or None if there is no consensus
         let consensus_threshold = self.consensus_c as usize;
         let consensus = peers_beacons.consensus(consensus_threshold);
-        let pb = peers_beacons.pb;
+        //let pb = peers_beacons.pb;
```
aesedepece, line 630: Hey this.

node/src/actors/chain_manager/handlers.rs
```diff
@@ -617,9 +627,9 @@ impl Handler<PeersBeacons> for ChainManager {
...
         let outbound_limit = peers_beacons.outbound_limit;
-        let pb_len = pb.len();
+        let pb_len = peers_beacons.pb.len();
         let peers_needed_for_consensus = outbound_limit
             .map(|x| {
                 // ceil(x * consensus_threshold / 100)
```
aesedepece, line 635: Is this line dead too?

node/src/actors/chain_manager/handlers.rs
```diff
@@ -646,9 +648,15 @@ impl Handler<PeersBeacons> for ChainManager {
             );
             vec![]
         } else {
-            // No consensus: unregister all peers
-            log::warn!("No consensus: unregister all peers");
-            pb.into_iter().map(|(p, _b)| p).collect()
+            // No consensus: if state is AlmostSynced unregister those that are not coincident with ours.
+            // Else, unregister all peers
+            if self.sm_state == StateMachine::AlmostSynced {
+                log::warn ! ("No consensus while Almost Synched: unregister peers that do not coincide with our last beacon");
```
aesedepece, line 654:
```suggestion
                log::warn ! ("Lack of peer consensus while state is `AlmostSynced`: peers that do not coincide with our last beacon will be unregistered");
```

node/src/actors/chain_manager/handlers.rs
```diff
@@ -646,9 +648,15 @@ impl Handler<PeersBeacons> for ChainManager {
...
+                peers_beacons.decide_peers_to_unregister(self.get_chain_beacon())
+            } else {
+                log::warn!("No consensus: unregister all peers");
```
aesedepece, line 657:
```suggestion
                log::warn!("Lack of peer consensus: all peers will be unregistered");
```

node/src/actors/chain_manager/handlers.rs
```diff
@@ -628,15 +638,7 @@ impl Handler<PeersBeacons> for ChainManager {
             .unwrap_or(1);
         let peers_to_unregister = if let Some(consensus_beacon) = consensus {
             // Consensus: unregister peers which have a different beacon
```
aesedepece, line 640: Feels to me like this comment does not belong in here anymore.

node/src/actors/chain_manager/handlers.rs
```diff
@@ -1273,4 +1281,60 @@ mod tests {
         };
         assert_eq!(peers_beacons.consensus(60), Some(beacon1));
     }
+    #[test]
+    fn test_unregister_peers() {
```
aesedepece, line 1285: Is it maybe worth testing missing beacons?

```rust
cons = PeersBeacons {
    pb: vec![
        ("127.0.0.1:10001".parse().unwrap(), None),
        ("127.0.0.1:10002".parse().unwrap(), Some(beacon1)),
        ("127.0.0.1:10003".parse().unwrap(), None),
        ("127.0.0.1:10004".parse().unwrap(), Some(beacon2)),
    ],
    outbound_limit: Some(4),
};
assert_eq!(
    peers_beacons.decide_peers_to_unregister(beacon2),
    [
        "127.0.0.1:10001".parse().unwrap(),
        "127.0.0.1:10002".parse().unwrap(),
        "127.0.0.1:10003".parse().unwrap(),
    ]
);
```

node/src/actors/chain_manager/handlers.rs
girazoki: nope, pb_len is used in subsequent loops

node/src/actors/chain_manager/handlers.rs
```diff
@@ -33,6 +33,7 @@ use crate::{
     storage_mngr,
     utils::mode_consensus,
 };
+use std::net::SocketAddr;
```
lrubiorod, line 36: fmt

node/src/actors/chain_manager/handlers.rs
```diff
@@ -646,9 +646,15 @@ impl Handler<PeersBeacons> for ChainManager {
             );
             vec![]
         } else {
-            // No consensus: unregister all peers
-            log::warn!("No consensus: unregister all peers");
-            pb.into_iter().map(|(p, _b)| p).collect()
+            // No consensus: if state is AlmostSynced unregister those that are not coincident with ours.
+            // Else, unregister all peers
+            if self.sm_state == StateMachine::AlmostSynced {
```
lrubiorod, line 651:
```suggestion
            if self.sm_state == StateMachine::AlmostSynced ||  self.sm_state == StateMachine::Synced {
```

node/src/actors/chain_manager/handlers.rs
```diff
@@ -780,7 +786,7 @@ impl Handler<PeersBeacons> for ChainManager {
                         self.initialize_from_storage(ctx);
                         log::info!("Restored chain state from storage");
 
-                        self.sm_state = StateMachine::WaitingConsensus;
+                        self.sm_state = StateMachine::AlmostSynced;
```
lrubiorod, line 789: If your peers has a consensus and you are out of them, you have to restore from storage and move to WaitingConsensus not AlmostSynced

girazoki: veeery true

node/src/actors/chain_manager/handlers.rs
```diff
@@ -761,7 +769,7 @@ impl Handler<PeersBeacons> for ChainManager {
                 // we need to rewind one epoch
                 if pb_len == 0 {
                     log::warn!("[CONSENSUS]: We have not received any beacons for this epoch");
-                    self.sm_state = StateMachine::WaitingConsensus;
+                    self.sm_state = StateMachine::AlmostSynced;
                 }
```
tmpolaczyk, line 773: This `if` may be redundant because in the case with 0 beacons the consensus is `None` and that's handled below.

node/src/actors/chain_manager/handlers.rs
```diff
@@ -596,6 +598,15 @@ impl PeersBeacons {
         // Some(None) (most of the peers did not send a beacon)
         .and_then(|x| *x)
     }
+
+    /// Collects the peers to unregister based on the beacon they reported and the beacon to be compared it with
+    pub fn decide_peers_to_unregister(&self, beacon: CheckpointBeacon) -> Vec<SocketAddr> {
+        // Unregister peers which have a different beacon
+        (&self.pb)
+            .iter()
+            .filter_map(|(p, b)| if *b != Some(beacon) { Some(*p) } else { None })
```
aesedepece, line 607: This is believed to be more efficient, as it short-circuits as soon as `p == None`:
```suggestion
            .filter_map(|(peer, b)| *b.filter(|b| b != beacon).map(|_| *peer))
```

girazoki: thanks!

tmpolaczyk: Be careful, this is not equivalent. If `b` is `None`, it should return `Some(*p)` because we want to unregister peers that have sent no beacon.

PR #1362, opened 2020-06-30
fix(rad): tell integers from floats when converting from JSON to CBOR

### LINKED ISSUE #1146: Function `json_to_cbor` mistakenly manages all JSON numbers as Float
As seen here, the RAD manager is currently managing all numeric values coming from JSON as `RadonFloat`:
https://github.com/witnet/witnet-rust/blob/2c0e9d614028867ab9c942bcd90c02124693cf2b/rad/src/operators/string.rs#L132

This leads to really poor UX for devs who are expecting those values to be operable using `RadonInteger` operators.

We should detect whether the numeric value has decimals, and cast to `RadonInteger` or `RadonFloat` automatically.

aesedepece:
Fix #1146

### DIFF
--- a/rad/src/operators/string.rs
+++ b/rad/src/operators/string.rs
@@ -115,6 +115,10 @@ pub fn string_match(input: &RadonString, args: &[Value]) -> Result<RadonTypes, R
         .unwrap_or(Ok(temp_def))
 }
 
+/// Converts a JSON value (`json::JsonValue`) into a CBOR value (`serde_cbor::value::Value`).
+/// Some conversions are totally straightforward, but some others  need some more logic (e.g.
+/// telling apart integers from floats).
+#[allow(clippy::cast_possible_truncation)]
 fn json_to_cbor(value: &json::JsonValue) -> Value {
     match value {
         json::JsonValue::Array(value) => Value::Array(value.iter().map(json_to_cbor).collect()),
@@ -126,8 +130,19 @@ fn json_to_cbor(value: &json::JsonValue) -> Value {
             Value::Map(entries)
         }
         json::JsonValue::Short(value) => Value::Text(String::from(value.as_str())),
+        json::JsonValue::Number(value) => {
+            let (_, _, exponent) = value.as_parts();
+            let floating = f64::from(*value);
+            // Cast the float into an integer if it has no fractional part and its value will fit
+            // into the range of `i128` (38 is the biggest power of 10 that `i128` can safely hold)
+            if floating.fract() == 0.0 && exponent.abs() < 38 {
+                // This cast is assumed to be safe as per the previous guard
+                Value::Integer(floating as i128)
+            } else {
+                Value::Float(floating)
+            }
+        }
         json::JsonValue::String(value) => Value::Text(String::from(value.as_str())),
-        json::JsonValue::Number(value) => Value::Float((*value).into()),
         _ => Value::Null,
     }
 }
@@ -177,9 +192,9 @@ mod tests {
         let output = parse_json_array(&json_array).unwrap();
 
         let expected_output = RadonArray::from(vec![
-            RadonTypes::Float(RadonFloat::from(1)),
-            RadonTypes::Float(RadonFloat::from(2)),
-            RadonTypes::Float(RadonFloat::from(3)),
+            RadonTypes::Integer(RadonInteger::from(1)),
+            RadonTypes::Integer(RadonInteger::from(2)),
+            RadonTypes::Integer(RadonInteger::from(3)),
         ]);
 
         assert_eq!(output, expected_output);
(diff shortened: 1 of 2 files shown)


rad/src/script.rs
```diff
@@ -273,14 +276,14 @@ fn test_execute_radon_script() {
             Some(vec![Value::Text(String::from("main"))]),
         ),
         (
-            RadonOpCodes::MapGetFloat,
+            RadonOpCodes::MapGetInteger,
```
tmpolaczyk, line 279: This is exactly the issue we are trying to solve, now if the temperature changes from -4ºC to -4.5ºC, this data request breaks. So we should find a way to accept both -4 and -4.5 as valid floats, instead of modifying this test to work on integers.

aesedepece: If you want that behavior, you can just do `MapGetFloat` instead of `MapGetInteger` and it works as well, I guess. 

Do you prefer if I leave the script in the test just as it was? Maybe the change I made is too reckless and a bad example because it can break easily.

aesedepece: I've updated the test to illustrate my point :wink:

tmpolaczyk: I expected the test to not change to make sure we don't break anything. I guess this test was not using `MapGetInteger` because that didn't work, so we also need to add a new test using `MapGetInteger`.

rad/src/operators/string.rs
```diff
@@ -127,7 +130,27 @@ fn json_to_cbor(value: &json::JsonValue) -> Value {
         }
         json::JsonValue::Short(value) => Value::Text(String::from(value.as_str())),
         json::JsonValue::String(value) => Value::Text(String::from(value.as_str())),
-        json::JsonValue::Number(value) => Value::Float((*value).into()),
+        json::JsonValue::Number(value) => {
+            // This trick allows to tell whether a number has decimals by checking whether the
+            // mantissa contains more significant bits than the exponent can express.
```
tmpolaczyk, line 135: I thought you may be reimplementing `value.as_i64()`, but then I realized we are using `json`, not `serde_json`. No idea why.

https://docs.serde.rs/serde_json/struct.Number.html#method.as_i64

rad/src/operators/string.rs
```diff
@@ -127,7 +130,27 @@ fn json_to_cbor(value: &json::JsonValue) -> Value {
...
+            let (positive, mantissa, exponent) = value.as_parts();
+            // Factor is 10 ^ |exponent|
+            let factor = u64::pow(
+                10,
+                u32::try_from(exponent.abs())
+                    .expect("Exponent of JsonValue::Number should fit in u32"),
+            );
```
tmpolaczyk, line 142: `u64::pow` will panic if the result does not fit in a u64, for example when the exponent is 100.

rad/src/script.rs
aesedepece: Let me see what I can do.

rad/src/operators/string.rs
aesedepece: Don't think we will ever need 100 as an exponent, because such think doesn't exist in JSON anyway, as it is subject to the limitations of double precision  / binary64 as put forth by IEEE 754 (minimum = `-(2^53 - 1)`, maximum = `2^53 - 1`, exponent_bits = `11`)

tmpolaczyk: What do you mean? `1e100` is valid JSON, and `1` followed by 100 zeros is also valid JSON. Add this test:

```rust
        let json = JsonValue::Number(Number::from(1e100));
        let resulting_cbor = json_to_cbor(&json);
        let expected_cbor = serde_cbor::Value::Float(1e100);
        assert_eq!(resulting_cbor, expected_cbor);
```

aesedepece: That works now! :wink:

tmpolaczyk: Cool, let's add a few tests near `i128::MAX` and `i128::MIN` to make sure it works as expected.

tmpolaczyk:
Also we need to make sure that this fixes the issue on sheikah's end. @Tommytrg can you provide an example data request that doesn't work without this PR?

aesedepece:
> Also we need to make sure that this fixes the issue on sheikah's end. @Tommytrg can you provide an example data request that doesn't work without this PR?

Sounds like he can't provide such data request, so I guess this is OK to merge? :stuck_out_tongue:

Tommytrg:
I've been trying this PR whit sheikah and it worked correctly retrieving `integer` and `float`. On the other hand, there is something odd when applying `getFloat` from an integer: it returns the a float wihout decimal. Is this the expected behavior or its out of the scope of this PR?

aesedepece:
> I've been trying this PR whit sheikah and it worked retrieving `integer`s and `float`s. On the other hand, there is something odd when applying `getFloat` from an integer: it returns the a float wihout decimal. Is this the expected behavior or its out of the scope of this PR?

aesedepece:
@witnet/devs anyone against merging this?

aesedepece:
> I've been trying this PR whit sheikah and it worked retrieving `integer`s and `float`s. On the other hand, there is something odd when applying `getFloat` from an integer: it returns the a float wihout decimal. Is this the expected behavior or its out of the scope of this PR?

This is perfectly fine. If it's an integer, you are just expected to do `getInteger`.

tmpolaczyk:
> @witnet/devs anyone against merging this?

imagen

I would just like to see how this PR changes the result of a data request

tmpolaczyk:
> I've been trying this PR whit sheikah and it worked retrieving `integer`s and `float`s. On the other hand, there is something odd when applying `getFloat` from an integer: it returns the a float wihout decimal. Is this the expected behavior or its out of the scope of this PR?

This is because `1.0.to_string()` returns `"1"`. Floats are already annotated so I don't know if adding `.0` will improve UI. Unless we want to always show floats using scientific notation or something, then `1.0` would be `1e0`. But yeah, let's discuss this in a different PR.

aesedepece:
> I would just like to see how this PR changes the result of a data request
Any DR that worked before and stopped working with these changes was malformed anyway because the former implementation of the JSON → CBOR decoder was not conforming to the RADON "standard" :rofl:

Tommytrg:
@tmpolaczyk You can modify a `get` of [this data request](https://github.com/witnet/sheikah/blob/master/examples/subscript.json)  to use `getFloat` or `getIntenger`

Issue #1308, opened 2020-06-16
Ensure that invalid transactions are eventually cleared from memory


tmpolaczyk:
Currently, transactions are only removed from the in-memory transactions pool when they are seen inside a consolidated block. This means that invalid transactions (because of a double spend for example) will never be included in a block, they will stay in the transactions pool forever (or until the node is restarted). This may explain some issues reported by node operators about memory usage increasing over time.

The simple solution is to remove invalid transactions when mining a block. This has some problems because as the network grows, each node will mine less frequently.

Another simple solution is to periodically perform "garbage collection": validate all the pending transactions and remove the invalid ones.

Decide what to do, and check if it fixes the memory usage.

tmpolaczyk:
> Another simple solution is to periodically perform "garbage collection": validate all the pending transactions and remove the invalid ones.

This has the problem that validating all the transactions at once may be a bad idea if the transactions pool has thousands of transactions. As an alternative, we could re-validate n transactions every epoch, where n is small (2 or 10).

Another idea is to perform this periodical validation when inserting a new transaction: for each new transaction, validate 1 or 2 old transactions and remove them if they are invalid. This has the problem that when there is a peak of new transactions, validation will take 2x or 3x more time.

And I think that the "correct" way to fix this problem is to remove all the transactions that have inputs that are included in a transaction that is included in a consolidated block. In other words, if two transactions try to spend the same UTXO, including one of them in a block removes the other one from the mempool. This would require an additional index like `HashMap<utxo, Vec<tx_hash>>`, which ironically implies more memory usage.

aesedepece:
Even if a transaction is valid with regards of its proper construction and usage of unspent inputs, we will need to make sure that transactions don't live in the mempool forever. Otherwise, it is very poor UX for people that messes miner fees, because the UTXOs they put as input would be locked forever.

In other words, there needs to be a "mempool timeout". This seems to be the dominant approach in Bitcoin. AFAIK, miners discard transactions that have been in the mempool for more than 2 weeks. 

With that in mind, I think we could keep lists of what transactions were inserted in some time window (e.g. separated by hours), so that after the timeout we can delete the entries from that list that may still remain without having to iterate over the entire pool.

Discarding transactions that were originally valid but became invalid at a later point in time because their inputs got spent is obviously a separate problem in its own.

[...]

PR #1363, opened 2020-07-01
feat(data_structures): improve remove method in TransactionsPool to r…

lrubiorod:
…emove all the transactiosn with the same input

### DIFF
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -1413,6 +1413,8 @@ pub struct TransactionsPool {
     // A hashset of recently received transactions hashes
     // Used to avoid validating the same transaction more than once
     pending_transactions: HashSet<Hash>,
+    // Map to avoid double spending issues
+    output_pointer_map: HashMap<OutputPointer, Vec<Hash>>,
 }
 
 impl TransactionsPool {
@@ -1447,6 +1449,7 @@ impl TransactionsPool {
             sorted_vt_index: BTreeSet::new(),
             sorted_dr_index: BTreeSet::new(),
             pending_transactions: HashSet::new(),
+            output_pointer_map: HashMap::with_capacity(capacity),
         }
     }
 
@@ -1664,6 +1667,9 @@ impl TransactionsPool {
             .unwrap_or(Ok(false))
     }
 
+    /// Remove a value transfer transaction from the pool and make sure that other transactions
+    /// that may try to spend the same UTXOs are also removed.
+    ///
     /// Returns an `Option` with the value transfer transaction for the specified hash or `None` if not exist.
     ///
     /// The `key` may be any borrowed form of the hash, but `Hash` and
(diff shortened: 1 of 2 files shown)


data_structures/src/chain.rs
```diff
@@ -1671,6 +1674,24 @@ impl TransactionsPool {
...
+                    }
+                }
+            }
+        }
+
+        transaction
+    }
+
+    /// Value Transfer remove inner
+    fn vt_remove_inner(&mut self, key: &Hash) -> Option<VTTransaction> {
```
aesedepece, line 1694: Quite opaque of a function name and comment :rofl: 
```suggestion
    /// Remove a value transfer transaction from the pool without any further guarantee about removing other transactions that may try to spend the same UTXOs.
    fn vt_remove_inner(&mut self, key: &Hash) -> Option<VTTransaction> {
```

data_structures/src/chain.rs
```diff
@@ -1671,6 +1674,24 @@ impl TransactionsPool {
     /// assert!(!pool.vt_contains(&transaction.hash()));
     /// ```
     pub fn vt_remove(&mut self, key: &Hash) -> Option<VTTransaction> {
```
aesedepece, line 1676:
```suggestion
    /// Remove a value transfer transaction from the pool and make sure that other transactions that may try to spend the same UTXOs are also removed.
    pub fn vt_remove(&mut self, key: &Hash) -> Option<VTTransaction> {
```

data_structures/src/chain.rs
```diff
@@ -1701,6 +1722,24 @@ impl TransactionsPool {
     /// assert!(!pool.dr_contains(&transaction.hash()));
     /// ```
     pub fn dr_remove(&mut self, key: &Hash) -> Option<DRTransaction> {
```
aesedepece, line 1724: /// Remove a data request transaction from the pool and make sure that other transactions that may try to spend the same UTXOs are also removed.

data_structures/src/chain.rs
```diff
@@ -1701,6 +1722,24 @@ impl TransactionsPool {
...
+                    }
+                }
+            }
+        }
+
+        transaction
+    }
+
+    /// Data Request remove inner
+    fn dr_remove_inner(&mut self, key: &Hash) -> Option<DRTransaction> {
```
aesedepece, line 1742:
```suggestion
    /// Remove a data request transaction from the pool without any further guarantee about removing other transactions that may try to spend the same UTXOs.
    fn dr_remove_inner(&mut self, key: &Hash) -> Option<DRTransaction> {
```

tmpolaczyk:
As mentioned in #1308, this will increase the memory usage quite a bit.

We would need to test this in a real world scenario to see if it is worth it.

lrubiorod:
> As mentioned in #1308, this will increase the memory usage quite a bit.
> 
> We would need to test this in a real world scenario to see if it is worth it.

Totally agree. Moreover, unless we are increasing the memory usage quite a bit, when we combine this solution with the possibility to validate the transactions once, and not every time that we see in a block, that will implies less validations and less memory usage

aesedepece:
Memory pools can be limited in size anyway, as they are only a nice-to-have for mining and speeding up validations, but not strictly needed for keeping a node up to date.

PR #1370, opened 2020-07-03
feat: add development environment

tmpolaczyk:
This environment allows to set consensus_constants in witnet.toml

### DIFF
--- a/config/src/defaults.rs
+++ b/config/src/defaults.rs
@@ -334,11 +334,41 @@ pub trait Defaults {
     }
 }
 
+/// Struct that will implement all the development defaults
+pub struct Development;
+
 /// Struct that will implement all the mainnet defaults
 pub struct Mainnet;
 
 /// Struct that will implement all the testnet defaults
 pub struct Testnet;
+
+impl Defaults for Development {
+    fn connections_server_addr(&self) -> SocketAddr {
+        SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 21337)
+    }
+
+    fn jsonrpc_server_address(&self) -> SocketAddr {
+        SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 21338)
+    }
+
+    fn storage_db_path(&self) -> PathBuf {
+        PathBuf::from(".witnet")
+    }
+
+    fn connections_bootstrap_peers_period(&self) -> Duration {
+        Duration::from_secs(15)
+    }
+
+    fn consensus_constants_checkpoint_zero_timestamp(&self) -> i64 {
+        // Wednesday, 24-Jun-2020, 11:00 UTC
+        1_592_996_400
+    }
+
+    fn connections_reject_sybil_inbounds(&self) -> bool {
+        false
+    }
+}
 
 impl Defaults for Mainnet {
     fn connections_server_addr(&self) -> SocketAddr {
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -65,14 +65,18 @@ pub struct ChainInfo {
 }
 
 /// Possible values for the "environment" configuration param.
+// The variants are explicitly tagged so that bincode serialization does not break
 #[derive(Serialize, Deserialize, Copy, Clone, Debug, PartialEq)]
 pub enum Environment {
     /// "mainnet" environment
     #[serde(rename = "mainnet")]
-    Mainnet,
+    Mainnet = 0,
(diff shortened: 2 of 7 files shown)


config/src/defaults.rs
```diff
@@ -328,12 +328,38 @@ pub trait Defaults {
...
+
+    fn connections_bootstrap_peers_period(&self) -> Duration {
+        Duration::from_secs(15)
+    }
+
+    fn consensus_constants_checkpoint_zero_timestamp(&self) -> i64 {
+        // Wednesday, 24-Jun-2020, 11:00 UTC
+        1_592_996_400
+    }
+}
```
aesedepece, line 361: Great opportunity for `connections.reject_sybil_inbounds = false` (as otherwise, local chain testing is _impossibru_)

data_structures/src/chain.rs
```diff
@@ -93,12 +97,22 @@ impl fmt::Display for Environment {
...
     pub fn bech32_prefix(&self) -> &str {
         match self {
             Environment::Mainnet => "wit",
             _ => "twit",
         }
     }
+
+    /// Returns true if the consensus constants can be modified in this environment.
+    /// This is only allowed in a development environment.
+    pub fn can_modify_consensus_constants(&self) -> bool {
```
aesedepece, line 110: Maybe `override` is more explicit about the fact that there are some defaults that you can optionally customize?
```suggestion
    pub fn can_override_consensus_constants(&self) -> bool {
```

config/src/defaults.rs
tmpolaczyk: Very true

tmpolaczyk:
This currently breaks existing nodes with:

    thread 'main' panicked at 'Mismatching environments: tried to run a node on environment "Testnet" with a chain that was initialized with environment "Mainnet".', node/src/actors/chain_manager/actor.rs:140:29

Because adding an additional variant to the `Environment` enum is not safu.

data_structures/src/chain.rs
```diff
@@ -93,12 +97,22 @@ impl fmt::Display for Environment {
 }
 
 impl Environment {
+    /// Returns the Bech32 prefix used in this environment: "wit" in mainnet and "twit" in testnet
```
girazoki, line 100:
```suggestion
    /// Returns the Bech32 prefix used in this environment: "wit" in mainnet and "twit" elsewhere
```

tmpolaczyk:
> This currently breaks existing nodes with:
> 
> ```
> thread 'main' panicked at 'Mismatching environments: tried to run a node on environment "Testnet" with a chain that was initialized with environment "Mainnet".', node/src/actors/chain_manager/actor.rs:140:29
> ```
> 
> Because adding an additional variant to the `Environment` enum is not safu.

Fixed this by setting the discriminant manually. Note that it is still not possible to "migrate" the storage from one environment to another, so a node that was started in "Development" cannot later change to "Testnet". That could be easily implemented but I can't find any use case, so it will stay unsupported.

aesedepece:
> > This currently breaks existing nodes with:
> > ```
> > thread 'main' panicked at 'Mismatching environments: tried to run a node on environment "Testnet" with a chain that was initialized with environment "Mainnet".', node/src/actors/chain_manager/actor.rs:140:29
> > ```
> > 
> > 
> > Because adding an additional variant to the `Environment` enum is not safu.
> 
> Fixed this by setting the discriminant manually. Note that it is still not possible to "migrate" the storage from one environment to another, so a node that was started in "Development" cannot later change to "Testnet". That could be easily implemented but I can't find any use case, so it will stay unsupported. Note that if the consensus constants change, the node does not panic.

This makes sense to me. As we know for sure that mainnet will use a separate environment that will not be available until then, we can rest assured that no testnet leftovers are present in mainnet storage. (Of course I'm assuming that this won't affect the master key)

tmpolaczyk:
> Of course I'm assuming that this won't affect the master key. Or is this assumption not right??? @tmpolaczyk

It does not affect the master key but note that when migrating from testnet to mainnet, all the testnet data (chain state, utxo set, blocks, transactions) will stay in the database as there is no clean-up process.

aesedepece:
> > Of course I'm assuming that this won't affect the master key. Or is this assumption not right??? @tmpolaczyk
> 
> It does not affect the master key but note that when migrating from testnet to mainnet, all the testnet data (chain state, utxo set, blocks, transactions) will stay in the database as there is no clean-up process.

Yeah, just as expected. If everything goes fine with the mainnet launch, we can make the software do some testnet data clean-up in a further release, just for the sake of reclaiming some disk space.

PR #1376, opened 2020-07-07
refactor: optimize how we deal with config when starting actors

aesedepece:


p2p/tests/buckets.rs
```diff
@@ -12,8 +12,11 @@ mod ice {
 
     #[test]
     fn test_can_ice_an_address() {
-        let mut peers = Peers::new();
-        peers.set_ice_period(Duration::from_secs(1000));
+        let ice_period = Duration::from_secs(1000);
+        let mut peers = Peers {
```
aesedepece, line 16: I suspect mutability is not needed here anymore. Let's see what clippy has to say.

aesedepece: It was needed indeed because the `contains` method does write-back when the ice "melts".

PR #1377, opened 2020-07-07
feat (p2p): melt addresses inserted through addPeers

girazoki:
This PR adds methods to remove addresses from the ice bucket, and removes those addresses added manually thtough addPeers method

### DIFF
--- a/node/src/actors/peers_manager/handlers.rs
+++ b/node/src/actors/peers_manager/handlers.rs
@@ -18,7 +18,7 @@ impl Handler<AddPeers> for PeersManager {
         // The easiest way to check if a peer was added manually is by using msg.src_address,
         // which is set to None when using the addPeers JSON-RPC method and also when reading peers from config
         if msg.src_address.is_none() {
-            self.peers.ice_peer_addresses_remove(&msg.addresses);
+            self.peers.remove_many_from_ice(&msg.addresses);
         }
         self.peers.add_to_new(msg.addresses, msg.src_address)
     }
--- a/node/src/actors/peers_manager/handlers.rs
+++ b/node/src/actors/peers_manager/handlers.rs
@@ -14,6 +14,12 @@ impl Handler<AddPeers> for PeersManager {
     fn handle(&mut self, msg: AddPeers, _: &mut Context<Self>) -> Self::Result {
         // Insert address
         log::trace!("Adding the following peer addresses: {:?}", msg.addresses);
+        // Remove peers added manually from the ice bucket to ensure that they are always added
+        // The easiest way to check if a peer was added manually is by using msg.src_address,
+        // which is set to None when using the addPeers JSON-RPC method and also when reading peers from config
+        if msg.src_address.is_none() {
+            self.peers.ice_peer_addresses_remove(&msg.addresses);
+        }
         self.peers.add_to_new(msg.addresses, msg.src_address)
     }
 }
--- a/p2p/src/peers/mod.rs
+++ b/p2p/src/peers/mod.rs
@@ -256,6 +256,11 @@ impl Peers {
         addrs
             .iter()
             .filter_map(|address| {
+                // Remove from iced
+                if ice {
+                    self.ice_peer_address(address);
+                }
+
                 let bucket_index = self.tried_bucket_index(&address);
                 let bucket_entry = self.tried_bucket.get(&bucket_index);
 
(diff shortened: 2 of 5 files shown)


node/src/actors/peers_manager/handlers.rs
```diff
@@ -14,6 +14,9 @@ impl Handler<AddPeers> for PeersManager {
     fn handle(&mut self, msg: AddPeers, _: &mut Context<Self>) -> Self::Result {
         // Insert address
         log::trace!("Adding the following peer addresses: {:?}", msg.addresses);
+        if msg.src_address == None {
```
tmpolaczyk, line 17: You can add a comment explaining why do we need this:

    // Remove peers added manually from the ice bucket to ensure that they are always added
    // The easiest way to check if a peer was added manually is by using msg.src_address,
    // which is set to None when using the addPeers JSON-RPC method and also when reading peers from config

Also, clippy prefers `if msg.src_address.is_none()`

p2p/src/peers/mod.rs
```diff
@@ -374,6 +374,21 @@ impl Peers {
...
     }
+
+    /// Remove address from the ice bucket
+    pub fn ice_peer_address_remove(&mut self, address: &SocketAddr) {
+        log::trace!("Removing peer address {} from the ice bucket", address);
+
+        self.ice_bucket.remove(address);
+    }
+
+    /// Remove a list of addresses from the ice bucket
```
tmpolaczyk, line 385: Mention that this does not take into account the ice_period, and always removes the peers

```suggestion
    /// Remove a list of addresses from the ice bucket, ignoring the ice_period
```

p2p/src/peers/mod.rs
```diff
@@ -374,6 +374,21 @@ impl Peers {
 
         true
     }
+
+    /// Remove address from the ice bucket
+    pub fn ice_peer_address_remove(&mut self, address: &SocketAddr) {
```
aesedepece, line 379: For uniformity with the other buckets:
```suggestion
    pub fn remove_from_ice(&mut self, address: &SocketAddr) {
```

p2p/src/peers/mod.rs
```diff
@@ -374,6 +374,21 @@ impl Peers {
...
+        log::trace!("Removing peer address {} from the ice bucket", address);
+
+        self.ice_bucket.remove(address);
+    }
+
+    /// Remove a list of addresses from the ice bucket
+    pub fn ice_peer_addresses_remove(&mut self, addresses: &[SocketAddr]) {
```
aesedepece, line 386:
```suggestion
    pub fn remove_many_from_ice(&mut self, addresses: &[SocketAddr]) {
```

p2p/src/peers/mod.rs
```diff
@@ -241,6 +241,11 @@ impl Peers {
         addrs
             .iter()
             .filter_map(|address| {
+                // Remove from iced
```
tmpolaczyk, line 244: No, this is "Add to iced"

girazoki: damn it

p2p/src/peers/mod.rs
```diff
@@ -389,6 +388,21 @@ impl Peers {
...
+    /// Remove address from the ice bucket, ignoring the ice period
+    pub fn remove_from_ice(&mut self, address: &SocketAddr) {
+        log::trace!("Removing peer address {} from the ice bucket", address);
+
+        self.ice_bucket.remove(address);
+    }
+
+    /// Remove a list of addresses from the ice bucket, ignoring the ice_period
+    pub fn remove_many_from_ice(&mut self, addresses: &[SocketAddr]) {
+        log::trace!("Removing peer address {:?} from the ice bucket", addresses);
```
aesedepece, line 401:
```suggestion
        log::trace!("Removing the following peer addresses from the ice bucket: {:?} ", addresses);
```

p2p/src/peers/mod.rs
```diff
@@ -256,6 +256,11 @@ impl Peers {
         addrs
             .iter()
             .filter_map(|address| {
+                // Add to iced
+                if ice {
+                    self.ice_peer_address(address);
+                }
```
tmpolaczyk, line 262: Can we add a test for this change? Check that when trying to remove an address that did not exist, this address will be iced.

girazoki: sure

PR #1378, opened 2020-07-07
feat(p2p): use hardcoded ice period (60s) before network bootstrapping

aesedepece:


p2p/src/peers/mod.rs
```diff
@@ -28,6 +28,8 @@ pub struct PeerInfo {
 /// Peers TBD
 #[derive(Deserialize, Serialize)]
 pub struct Peers {
+    /// Flag whether the chain is bootstrapped, specially with regards to ice period
+    pub bootstrapped: bool,
```
tmpolaczyk, line 32: This is only set to true during actor initialization, so if the actor starts before epoch 0 it will be set to false until the node is restarted.

aesedepece: True, I was missing updating upon bootstrapping!

PR #1379, opened 2020-07-07
getConsensusConstants JSON-RPC method

### LINKED ISSUE #1342: Fix bug dataRequestReport
The CLI method dataRequestReport returns an error when we try to check a DataRequest whose collateral value is by default (0)

### LINKED ISSUE #1374: `getConsensusConstants` method in JSON-RPC interface
As required for #1342 and some other over-the-top features and independent pieces of software (e.g. block explorer, wallet, etc.), it is now needed that the JSON-RPC interface exposes a method for listing the consensus constants that the node is using.

This need was introduced after merging #1332.

tmpolaczyk:
Close #1374
Fix #1342 

Sample output:

    {"jsonrpc":"2.0","id":"1","method":"getConsensusConstants"}

    {"jsonrpc":"2.0","result":{"activity_period":1000,"bootstrap_hash":"00000000000000000001000000000000000000007769746e65742d302e372e30","checkpoint_zero_timestamp":1594018864,"checkpoints_period":30,"collateral_age":5,"collateral_minimum":1000000000,"epochs_with_initial_difficulty":1000,"extra_rounds":3,"genesis_hash":"a0b6496d8de53c5998b89a72bb233c600822e14a3b6709154553c74f90604f23","initial_difficulty":10,"max_dr_weight":8000,"max_vt_weight":2000,"mining_backup_factor":8,"mining_replication_factor":3,"reputation_expire_alpha_diff":20000,"reputation_issuance":1,"reputation_issuance_stop":1048576,"reputation_penalization_factor":0.5,"superblock_period":10},"id":"1"}

### DIFF
--- a/node/src/actors/json_rpc/json_rpc_methods.rs
+++ b/node/src/actors/json_rpc/json_rpc_methods.rs
@@ -40,7 +40,7 @@ use crate::{
         peers_manager::PeersManager,
         sessions_manager::SessionsManager,
     },
-    signature_mngr,
+    config_mngr, signature_mngr,
 };
 
 type JsonRpcResultAsync = Box<dyn Future<Item = Value, Error = jsonrpc_core::Error> + Send>;
@@ -77,6 +77,9 @@ pub fn jsonrpc_io_handler(
     io.add_method("knownPeers", |_params: Params| known_peers());
     io.add_method("nodeStats", |_params: Params| node_stats());
     io.add_method("getMempool", |params: Params| get_mempool(params.parse()));
+    io.add_method("getConsensusConstants", |params: Params| {
+        get_consensus_constants(params.parse())
+    });
 
     // Enable methods that assume that JSON-RPC is only accessible by the owner of the node.
     // A method is sensitive if it touches in some way the master key of the node.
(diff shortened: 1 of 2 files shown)


tmpolaczyk:
In the future, we may need a way to return the environment (development/testnet/mainnet). In that case, we should consider creating a more general getConfig method.

node/src/actors/json_rpc/json_rpc_methods.rs
```diff
@@ -77,6 +77,9 @@ pub fn jsonrpc_io_handler(
     io.add_method("knownPeers", |_params: Params| known_peers());
     io.add_method("nodeStats", |_params: Params| node_stats());
     io.add_method("getMempool", |params: Params| get_mempool(params.parse()));
+    io.add_method("getConsensusConstants", |params: Params| {
+        get_consensus_constants(params.parse())
```
tmpolaczyk, line 81: Should this method be marked as sensitive?

lrubiorod: why do you think that could be marked as sensitive?

tmpolaczyk: It allows you to know to which network is the node connected to (testnet or mainnet), but I'm not sure if that is sensitive or not.

lrubiorod: I think it is not sensitive. It is suppose that most of the nodes will use the same consensus constants

tmpolaczyk: True, I thought it can be sensitive if we add other config values to the output, but that will be a different `getConfig` method, not this one. So let's leave this as not sensitive.

lrubiorod (approved):

PR #1380, opened 2020-07-07
feat(peers): melt peers during a specified period

lrubiorod:


node/src/actors/messages.rs
```diff
@@ -538,6 +538,13 @@ impl Message for AddPeers {
     type Result = PeersSocketAddrsResult;
 }
 
+/// Message to add melted peer addresses to the list
```
tmpolaczyk, line 541:
```suggestion
/// Message to remove all the melted peer addresses from the ice bucket and add them back to the new bucket
```

node/src/actors/peers_manager/handlers.rs
```diff
@@ -18,6 +18,22 @@ impl Handler<AddPeers> for PeersManager {
     }
 }
 
+/// Handler for AddPeers message
+impl Handler<MeltPeers> for PeersManager {
+    type Result = Result<(), failure::Error>;
+
+    fn handle(&mut self, _msg: MeltPeers, _: &mut Context<Self>) -> Self::Result {
+        // Melt addresses
+        let addresses = self.peers.extract_melted_peers_from_ice_bucket();
```
tmpolaczyk, line 27: The method name is more descriptive than the comment :laughing: 

```suggestion
        let addresses = self.peers.extract_melted_peers_from_ice_bucket();
```

node/src/actors/sessions_manager/mod.rs
```diff
@@ -10,6 +10,7 @@ use ansi_term::Color::Cyan;
 use witnet_p2p::sessions::{SessionType, Sessions};
 
 use self::beacons::Beacons;
+use crate::actors::messages::MeltPeers;
```
tmpolaczyk, line 13: fmt

p2p/src/peers/mod.rs
```diff
@@ -112,6 +112,28 @@ impl Peers {
         contains
     }
 
+    /// Extract all melted peers included in the ice bucket
+    pub fn extract_melted_peers_from_ice_bucket(&mut self) -> Vec<SocketAddr> {
+        let mut addresses = vec![];
+        let new_ice_bucket = self
```
tmpolaczyk, line 118: You should probably use [Vec::retain](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.retain) instead of clone + filter_map

node/src/actors/sessions_manager/mod.rs
```diff
@@ -129,13 +130,18 @@ impl SessionsManager {
...
+        let peers_manager_addr = PeersManager::from_registry();
         // Schedule the discovery_peers with a given period
         ctx.run_later(discovery_peers_period, move |act, ctx| {
             // Send Anycast(GetPeers) message
             ctx.notify(Anycast {
                 command: SendGetPeers {},
                 safu: false,
             });
+
+            // Send MeltPeers message
```
tmpolaczyk, line 142: Explain what it does and why do it here:

```suggestion
            // Remove peers from ice bucket that have been "melted"
            // This peers will be added to new again, to ensure that they never get stuck in the ice bucket
```

p2p/src/peers/mod.rs
```diff
@@ -179,6 +195,10 @@ impl Peers {
                     // If the index points to the same address that it is already
                     // in tried, we don't include in new bucket
                     if elem.is_none() || (elem.unwrap().address != address) {
+                        // If the source address that sent us this peer addresses is None, use the same address
+                        // that we want to add. This will make all the peer addresses that were added using manual methods
+                        // go to the same bucket that if it was announced by that address.
+                        let src_address = src_address.unwrap_or_else(|| address);
```
tmpolaczyk, line 201: This change may be dangerous, not sure. Perhaps we should replace `Option<SocketAddr>` with an new enum:

```rust
enum PeerSource {
    Config,
    JsonRpc,
    GetPeers(SocketAddr),
    InboundConnection(SocketAddr),
    IceBucket,
}
```

tmpolaczyk: At the end of the day, bucketing is only useful when the list of peers is not trusted. The peers in the ice bucket may or may not be trusted, but with the proposed change where `src_address = address`, 1 node sending us 100 peers will be the same as 100 nodes sending us 1 peer each.

node/src/actors/peers_manager/handlers.rs
```diff
@@ -18,6 +18,22 @@ impl Handler<AddPeers> for PeersManager {
...
+        if !addresses.is_empty() {
+            log::debug!("Melt those addresses: {:?}", addresses);
```
aesedepece, line 30:
```suggestion
            log::debug!("Melt these addresses: {:?}", addresses);
```

p2p/src/peers/mod.rs
aesedepece: I'm not totally sure about this either. But I think that the most important property is that the position that a peer address takes in the buckets is deterministic. In other words, I believe that this change is of very little advantage for an attacker that may abuse the system so that many of their addresses go iced and then melt into `new`, because the bucket and index are still one-shot.

aesedepece: @girazoki ?

girazoki: this actually breaks my PR I think haha. Still the most important thing is that from what I remember a source ip can only go to a subset of the buckets, so from that perspective with this change with a single attacker source address can fill more buckets. However, since attacker IPs will be propagated by potentially all peers in the network I do believe that this change might be of little advantage to him.

node/src/actors/sessions_manager/mod.rs
```diff
@@ -129,13 +129,19 @@ impl SessionsManager {
...
         // Schedule the discovery_peers with a given period
         ctx.run_later(discovery_peers_period, move |act, ctx| {
             // Send Anycast(GetPeers) message
             ctx.notify(Anycast {
                 command: SendGetPeers {},
                 safu: false,
             });
+
+            // Remove peers from ice bucket that have been "melted"
+            // This peers will be added to new again, to ensure that they never get stuck in the ice bucket
```
aesedepece, line 142:
```suggestion
            // These peers will be added to `new` again, so as to ensure that they never get stuck in the `ice` bucket
```

node/src/actors/peers_manager/handlers.rs
```diff
@@ -18,6 +19,21 @@ impl Handler<AddPeers> for PeersManager {
     }
 }
 
+/// Handler for AddPeers message
```
girazoki, line 22:
```suggestion
/// Handler for MeltPeers message
```

node/src/actors/peers_manager/handlers.rs
```diff
@@ -18,6 +19,21 @@ impl Handler<AddPeers> for PeersManager {
...
+        if !addresses.is_empty() {
+            log::debug!("Melt these addresses: {:?}", addresses);
+            self.peers.add_to_new(addresses, None).map(|_| ())
+        } else {
+            Ok(())
+        }
+    }
+}
+
 /// Handler for AddPeers message
```
girazoki, line 37:
```suggestion
/// Handler for AddConsolidatedPeer message
```

node/src/actors/sessions_manager/mod.rs
```diff
@@ -129,13 +129,19 @@ impl SessionsManager {
...
+        let peers_manager_addr = PeersManager::from_registry();
         // Schedule the discovery_peers with a given period
         ctx.run_later(discovery_peers_period, move |act, ctx| {
             // Send Anycast(GetPeers) message
             ctx.notify(Anycast {
                 command: SendGetPeers {},
                 safu: false,
             });
+
+            // Remove peers from ice bucket that have been "melted"
```
aesedepece, line 141:
```suggestion
            // Remove peers from `ice` bucket that have "melted", i.e. they have been in that bucket for as long as specified by the `connections.bucketing_ice_period_seconds` setting in the configuration
```

p2p/src/peers/mod.rs
```diff
@@ -127,6 +127,26 @@ impl Peers {
         contains
     }
 
+    /// Extract all melted peers included in the ice bucket
```
aesedepece, line 130:
```suggestion
    /// Extract (collect and remove) all melted peers from the `ice` bucket
```

node/src/actors/sessions_manager/mod.rs
```diff
@@ -136,10 +136,26 @@ impl SessionsManager {
...
                 safu: false,
             });
+
             act.discovery_peers(ctx, discovery_peers_period);
         });
     }
 
+    /// Method to periodically melt peers
+    fn melt_peers(&self, ctx: &mut Context<Self>, melt_peers_period: Duration) {
+        let peers_manager_addr = PeersManager::from_registry();
```
aesedepece, line 146: Do we know if this call has any overhead? This function is somehow recursive, so we could pass `peers_manager_addr` as an `Option` argument and only do `from_registry` if it is `None`, so it's only called once.

Here of course I'm assuming that actors don't change addresses for the lifetime of the Actix `System`.

If on the contrary actors can change addresses, then there's a little problem with the current implementation because the address is obtained before `run_later`, but then used in the callback. During that time, `PeersManager` may have changed address, and therefore the message will likely go to "dead letter" and one iteration of `MeltPeers` will be missed. Not a big deal, I guess.

Speaking of what, I wonder if this is real recursion or `ctx.run_later` does actually use something like the actor mailbox at all. Why is this important? Taking into account that Rust doesn't have tail call optimization, if this is causing real recursion, we may easily exhaust the stack after a certain number of re-entries.

I imagine (but haven't confirmed at all) that clippy probably has some guards against dangerous recursion, but being Actix in the middle, I don't know if there could be something that it misses.

node/src/actors/messages.rs
```diff
@@ -538,6 +538,14 @@ impl Message for AddPeers {
...
 }
 
+/// Message to remove all the melted peer addresses from the ice bucket
+/// and add them back to the new bucket
+pub struct MeltPeers;
+
+impl Message for MeltPeers {
+    type Result = Result<(), failure::Error>;
+}
+
```
aesedepece, line 548: This is no longer used, is it?

node/src/actors/peers_manager/actor.rs
```diff
@@ -89,6 +86,9 @@ impl Actor for PeersManager {
                 // Start the storage peers process on SessionsManager start
                 act.persist_peers(ctx, storage_peers_period);
 
+                // The peers melt process begins upon SessionsManager's start
```
tmpolaczyk, line 89:
```suggestion
                // The peers melt process begins upon PeersManager's start
```

node/src/actors/peers_manager/mod.rs
```diff
@@ -141,6 +141,24 @@ impl PeersManager {
...
+    fn melt_peers(&self, ctx: &mut Context<Self>, check_melted_peers_period: Duration) {
+        // Schedule the melt_peers with a given period
+        ctx.run_later(check_melted_peers_period, move |act, ctx| {
+            // Remove peers from `ice` bucket that have "melted", i.e. they have been in that bucket
+            // for as long as specified by the `connections.bucketing_ice_period_seconds` setting in
+            // the configuration. These peers will be added to `new` again, so as to ensure that
+            // they never get stuck in the `ice` bucket
+            let addresses = act.peers.extract_melted_peers_from_ice_bucket();
+            if !addresses.is_empty() {
+                log::debug!("Melt these addresses: {:?}", addresses);
```
aesedepece, line 154: One of these is maybe nicer:
```suggestion
                log::debug!("Melted these addresses: {:?}", addresses);
```
```suggestion
                log::debug!("Melting these addresses: {:?}", addresses);
```

PR #1385, opened 2020-07-09
Transactions in mempool can be assumed to be valid

### LINKED ISSUE #1352: Transactions in mempool can be assumed to be valid
- Read the comments in #1308 carefully.
- This depends on and is blocked by #1308.
- Under the assumption that transactions are validated in real time as they are introduced into the in-memory transaction pool (mempool), and that #1308 is sorted out, we can safely assume that transactions that do exist in the mempool are valid and don't need any further validation when they are found in a block we are validating, or when we are collecting transactions for anchoring them into a block candidate that we are proposing.

lrubiorod:
Close #1352

### DIFF
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -1459,8 +1459,8 @@ impl RADTally {
 }
 
 type PrioritizedHash = (OrderedFloat<f64>, Hash);
-type PrioritizedVTTransaction = (OrderedFloat<f64>, VTTransaction);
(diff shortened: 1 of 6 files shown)


tmpolaczyk:
I think this change may be dangerous. We could test this by running a few nodes with this features on the testnet, but since the major problems will arise when we implement reorganizations, we cannot test this right now.

So I would prefer to add this optimization later, after we have other consensus critical features ready. And maybe combine this change with storing the utxo set in a database instead of in ram, and other similar optimizations.

lrubiorod:
tagged in #1352

PR #1387, opened 2020-07-09
fix(rad): add missing RADON operators

### LINKED ISSUE #1386: Missing mapValues operator for each type
Currently, witnet-rust only expects one `mapValues` operator. However, there are `mapValues` operator for each type:
- MapValuesArray = 0x69
- MapValuesBoolean = 0x6a
- MapValuesBytes = 0x6b
- MapValuesInteger = 0x6c
- MapValuesFloat = 0x6d
- MapValuesMap = 0x6e
- MapValuesString = 0x6f

Tommytrg:
Add missing `mapValues` operators

fix #1386

### DIFF
--- a/rad/src/operators/mod.rs
+++ b/rad/src/operators/mod.rs
@@ -93,7 +93,13 @@ pub enum RadonOpCodes {
     MapGetMap = 0x66,
     MapGetString = 0x67,
     MapKeys = 0x68,
-    MapValues = 0x69,
+    MapValuesArray = 0x69,
+    //    MapValuesBoolean= 0x6a,
+    //    MapValuesBytes = 0x6b,
+    //    MapValuesInteger = 0x6c,
+    //    MapValuesFloat = 0x6d,
+    //    MapValuesMap = 0x6d,
+    //    MapValuesString = 0x6e,
     ///////////////////////////////////////////////////////////////////////
     // String operator codes (start at 0x70)
     StringAsBoolean = 0x70,


Tommytrg:
D you think we should keep `0x69` as `mapValues` in order to reflect the current behavior or update its name to the correct one as I've made?

tmpolaczyk:
> D you think we should keep `0x69` as `mapValues` in order to reflect the current behavior or update its name to the correct one as I've made?

I think the current `mapValues` operator is implemented as a generic operator that returns an array and does not care about the inner type of the array

tmpolaczyk:
The most simple way to implement this is to make `MapValuesArray` do the same as the old `MapValues`, but check the type of the resulting array. If the resulting array is not of the correct type, return an error. Of course that doesn't work with non-homogeneous arrays, for example, should this work or should it be an error?

`[1, 2, "3"] => MapValuesArray => [[1], [2], ["3"]]`
`[1, 2, "3"] => MapValuesInteger => [1, 2, 3]`
`[1, 2, "3"] => MapValuesString => ["1", "2", "3"]`

Tommytrg:
IMO, if the casting is possible it shouldn't throw an errror.

On the other hand, could do you handle a map with heterogeneus values like: `{ a: 1, b: true }`? Should the values be casted into a string such all values can be represented as string :facepalm: ?

aesedepece:
The assumption here is that `MapValuesXXX` can only be applied on homogeneous maps.

Homogeneus maps? What's that? Well, consider that with regards to the current RADON interpreter, maps function roughly as arrays of tuples where the first element (key) is a string and the second element (value) can take virtually any type.

The issue with `MapValuesXXX` is that it tries to put those values of unknown types into an array, implying that if the type of those values varies, then we are left with a non-homogeneous array, which is painful to work with.

Regarding the scope of this specific PR, I'm any definition of `MapValuesArray` acting as `RadonMap<T> -> RadonArray<T>`.  IMO `MapValuesArray` should be `RadonArray<RadonArray<T>>` just like `MapValuesInteger` is expected to return `RadonArray<Integer>` and not simply `Integer`. If you want the other behavior, you can explicitly flatten. 

So @tmpolaczyk 's reasoning seems correct to me. And I wouldn't be against generic `MapValues` or `MapValuesAny` method either, with a signature such that it performs `RadonMap<T> -> RadonArray<T>`.

aesedepece:
Closing this PR as the immediate actions regarding RADON operators are to be taken in #1458 and #1459.

PR #1388, opened 2020-07-10
little fixes propedeutic to testnet 9.1

lrubiorod:


node/src/actors/peers_manager/mod.rs
```diff
@@ -141,6 +144,15 @@ impl PeersManager {
...
+    pub fn current_check_melted_peers_period(&self) -> Duration {
+        if self.peers.bootstrapped {
+            self.check_melted_peers_period
+        } else {
+            Duration::from_secs(60)
+        }
+    }
+
     /// Method to periodically melt peers
     fn melt_peers(&self, ctx: &mut Context<Self>, check_melted_peers_period: Duration) {
```
tmpolaczyk, line 157: If the period is calculated using `act.current_check_melted_peers_period()`, you don't need this argument.

```suggestion
    fn melt_peers(&self, ctx: &mut Context<Self>) {
```

node/src/actors/peers_manager/actor.rs
```diff
@@ -69,7 +69,7 @@ impl Actor for PeersManager {
                     .send(GetEpoch)
                     .into_actor(act)
                     .then(move |epoch, act, ctx| {
-                        if epoch.is_ok() {
+                        if epoch.expect("Error when asking for current epoch while initializing PeersManager").is_ok() {
```
aesedepece, line 72:
```suggestion
                        if epoch.expect("Error when asking for current epoch while initializing PeersManager: {:?}", epoch).is_ok() {
```

PR #1400, opened 2020-07-13
feat(superblock): track who voted what

### LINKED ISSUE #1394: Track who voted what and how many votes each candidate gets

### LINKED ISSUE #1396: Discard superblock double-votes

tmpolaczyk:
Close #1394 
Close #1396

### DIFF
--- a/data_structures/src/superblock.rs
+++ b/data_structures/src/superblock.rs
@@ -35,8 +35,12 @@ pub struct SuperBlockState {
     // Current superblock index, used to limit the range of broadcasted votes to
     // [index - 1, index + 1]. So if index is 10, only votes with index 9, 10, 11 will be broadcasted
     current_superblock_index: Option<u32>,
+    // Map of identities that voted more than once. This votes are considered invalid.
+    identities_that_voted_more_than_once: HashMap<PublicKeyHash, Vec<SuperBlockVote>>,
     // Set of ARS identities that can currently send superblock votes
     previous_ars_identities: Option<HashSet<PublicKeyHash>>,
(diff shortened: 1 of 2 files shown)


data_structures/src/superblock.rs
```diff
@@ -192,6 +202,16 @@ impl SuperBlockState {
             hash_prev_block: self.current_superblock_hash?,
         })
     }
+
+    /// Returns the superblock hash and the number of votes of the most voted superblock.
+    /// In case of tie, returns one of the superblocks with the most votes.
```
aesedepece, line 207: Sounds spooky at first. But then you realize that if we got the same number of votes, it is impossible that they'd be able to meet the 2/3 requirement anyway.

This should be perfectly fine then, unless at some point we act differently depending on whether our locally built superblock matches the one resulting from this function.

tmpolaczyk: Yeah that was my though as well. In the future we can add here the logic to check if the most voted superblock has more than 2/3 of the votes, and return `None` otherwise.

data_structures/src/superblock.rs
```diff
@@ -35,19 +37,58 @@ pub struct SuperBlockState {
...
+
+            self.identities_that_voted_more_than_once
+                .insert(pkh, vec![old_sbv, sbv]);
+
+            false
+        } else {
+            self.votes_on_each_superblock
+                .entry(sbv.superblock_hash)
+                .or_default()
+                .push(sbv);
```
aesedepece, line 86: Funny that this will always go into `or_default` :grin:

PR #1402, opened 2020-07-14
fix(rad): return partial results for subscripts

### LINKED ISSUE #1401: Detail data request result for each subscript operator
The rad request result when trying a data request should include the partial results of the operators within a subscript argument. 

As for example: 
```
partial_results: [
 0: { RadonString: '' },
 1: { RadonMap: {} },
 2: { RadonMap: {} },
 3: { RadonArray: [] },
 4: { script: [ 0: { RadonError: '' } ], result: { RadonError: '' }},
]
```

aesedepece:
Fix #1401

### DIFF
--- a/data_structures/src/radon_report.rs
+++ b/data_structures/src/radon_report.rs
@@ -4,7 +4,6 @@ use std::time::{Duration, Instant};
 use serde::Serialize;
 
 use crate::radon_error::ErrorLike;
-use crate::radon_report::Stage::Contextless;
 
 /// A high level data structure aimed to be used as the return type of RAD executor methods:
 ///
@@ -95,18 +94,35 @@ pub trait TypeLike: Clone + Sized {
 }
 
 /// A generic structure for bubbling up any kind of metadata that may be generated during the
-/// execution of a RADON script.
+/// execution of a RADON script. This is specially useful for tracing errors back to specific calls
+/// in scripts.
+///
+/// The word "processed" in the structure member descriptions implies that the execution was
+/// successful up to the preceding call or argument, and that the parsing and execution of the
+/// reported call or argument index did at least began.
+///
+/// Scripts that are executed successfully in their full extent will therefore report here the
+/// index of the last call in the script. On the contrary, scripts that fail will report the index
+/// of the call that failed.
 pub struct ReportContext<RT>
 where
     RT: TypeLike,
 {
+    /// The arguments of the last call that has been processed.
     pub call_arguments: Option<Vec<serde_cbor::Value>>,
+    /// The index of the last argument in a call that has been processed.
     pub call_argument_index: Option<u8>,
+    /// The index of the last call that has been processed.
     pub call_index: Option<u8>,
+    /// The operator in the last call that has been processed.
     pub call_operator: Option<u8>,
+    /// The timestamp when the execution of the script finished.
     pub completion_time: Option<Instant>,
+    /// Metadata that is specific to the stage of the script.
     pub stage: Stage<RT>,
+    /// The timestamp when the execution of the script began.
     pub start_time: Option<Instant>,
+    /// The index of the last script or subscript in a stage that has been processed.
     pub script_index: Option<u8>,
 }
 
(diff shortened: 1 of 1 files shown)


rad/src/operators/array.rs
```diff
@@ -111,23 +115,40 @@ pub fn map(input: &RadonArray, args: &[Value]) -> Result<RadonTypes, RadError> {
...
-        )?);
+        let report = execute_radon_script(item.clone(), subscript.as_slice(), context, settings)?;
+
+        results.push(item.clone());
+        reports.push(report);
     }
 
-    Ok(RadonArray::from(result).into())
+    if let Stage::Retrieval(metadata) = &mut context.stage {
+        let mut partial_results = vec![];
```
tmpolaczyk, line 130: I think this piece of code is duplicated in `ArrayMap` and `ArrayFilter`. We can extract it into a function and give it a name.

The benefit of using iter and collect instead of for loop and push is that you get a capacity estimate for free. So try to do some iterator magic, or simply use `Vec::with_capacity(subscript.len() + 1)` here.

It should be possible to implement this without using clone, maybe using `std::mem::take` or something like that.

I don't think the `if let` are needed, the subscripts results must all have the same length, right? And report.partial_results cannot be `None`?. In that case just use unwrap or expect, that will help us to find issues in tests.

tmpolaczyk:
Should we add some limit to the length of the partial results? In some cases it may blow up pretty quickly, for example when using filter on large arrays. We could limit it to only store the partial results of the first 10 elements of an array, but then there is no way to get the other results.

rad/src/operators/array.rs
aesedepece: All of these comments were very well received and have been already addressed. Thanks bro.

The only thing I couldn't address is about using `clone`. I will always end up having a `&mut RadonTypes` from which it is impossible to `mem::take` because `RadonTypes` does not implement `Default` and, even if it did, constructing a default value just to replace the mutable reference by means of `mem::take` will roughly equal the cost of simply cloning it.

aesedepece:
> Should we add some limit to the length of the partial results? In some cases it may blow up pretty quickly, for example when using filter on large arrays. We could limit it to only store the partial results of the first 10 elements of an array, but then there is no way to get the other results.

This sounds to me like is worth a separate issue, as such limit would most likely need to be implemented in some other places.

rad/src/operators/array.rs
```diff
@@ -155,29 +165,49 @@ pub fn filter(
             };
             let subscript = unpack_subscript(first_arg).map_err(subscript_err)?;
 
-            let mut result = vec![];
+            let mut reports = vec![];
+            let mut results = vec![];
+
+            let settings = RadonScriptExecutionSettings::enable_all();
```
tmpolaczyk, line 171: This should depend on the context, right? If partial results are disabled they should remain disabled, no need to always accumulate the partial reports.

rad/src/operators/array.rs
tmpolaczyk: Ok, good enough.

> constructing a default value just to replace the mutable reference by means of mem::take will roughly equal the cost of simply cloning it.

That's not true if the type is a large `RadonArray` or `RadonMap`.

rad/src/operators/array.rs
aesedepece: True. I should start tagging my TODOs as such so that I don't forget to change stuff once I get them working :sweat_smile:

aesedepece: Done. Commit 5c4d20e01284b600e41f7b2b404747381bb23419 adds a factory that creates instances of `RadonScriptExecutionSettings` that are appropriate for each specific `Stage`.

rad/src/lib.rs
```diff
@@ -19,6 +19,7 @@ use crate::{
     types::{array::RadonArray, string::RadonString, RadonTypes},
     user_agents::UserAgent,
 };
+use witnet_data_structures::radon_report::RetrievalMetadata;
```
girazoki, line 22: fmt

rad/src/operators/array.rs
aesedepece: Partial results was an optional feature in the first place :sweat_smile:

aesedepece:
@devs I'd love to get this merged

PR #1403, opened 2020-07-15
feat (config manager): add bootstrapping_committee

girazoki:
This PR adds a bootstrapping committee in charge of signing the first superblocks until collateral_age plus activity_period epoch is reached.

### DIFF
--- a/data_structures/src/superblock.rs
+++ b/data_structures/src/superblock.rs
@@ -32,6 +32,8 @@ pub enum AddSuperBlockVote {
 pub struct SuperBlockState {
     // Set of ARS identities that will be able to send superblock votes in the next superblock epoch
     current_ars_identities: Option<HashSet<PublicKeyHash>>,
+    // Subset of ARS in charge of signing the next superblock
+    current_signing_committee: Option<HashSet<PublicKeyHash>>,
     // Current superblock hash created by this node
     current_superblock_hash: Option<Hash>,
     // Current superblock index, used to limit the range of broadcasted votes to
@@ -112,7 +114,9 @@ impl SuperBlockState {
                     }
                 }
                 Some(false) => {
-                    if Some(sbv.superblock_index) == self.current_superblock_index {
+                    if Some(sbv.superblock_index) == self.current_superblock_index
+                        || self.previous_ars_identities.is_none()
+                    {
                         AddSuperBlockVote::NotInArs
                     } else {
                         AddSuperBlockVote::InvalidIndex
@@ -128,19 +132,21 @@ impl SuperBlockState {
     /// Returns true, false, or unknown
     fn is_valid(&self, sbv: &SuperBlockVote) -> Option<bool> {
         match self.current_superblock_index {
-            // We do not know the current index, we cannot know if the vote is valid
-            None => None,
+            // The superblock state is not initialized therefore no one should sign
+            None => Some(false),
             // If the index is the same as the current one, the vote is valid if it is signed by a
-            // member of the ARS
+            // member of the ARS. If no ARS is set, return false
             Some(x) if x == sbv.superblock_index => self
-                .previous_ars_identities
+                .current_signing_committee
                 .as_ref()
-                .map(|x| x.contains(&sbv.secp256k1_signature.public_key.pkh())),
+                .map_or(Some(false), |x| {
+                    Some(x.contains(&sbv.secp256k1_signature.public_key.pkh()))
+                }),
             // If the index is not the same as the current one, but it is within an acceptable range
             // of [x-1, x+1], broadcast the vote without checking if it is a member of the ARS, as
             // the ARS may have changed and we do not keep older copies of the ARS in memory
             Some(x) => {
-                // Check [x-1, x+1] range with overflow prevention
+                // Check [x-1,  x+1] range with overflow prevention
                 if ((x.saturating_sub(1))..=(x.saturating_add(1))).contains(&sbv.superblock_index) {
                     None
                 } else {
@@ -198,6 +204,8 @@ impl SuperBlockState {
                     hs.clear();
                     hs.extend(ars_pkh_keys.iter().cloned());
                     self.previous_ars_ordered_keys = ars_ordered_bn256_keys.to_vec();
+                    // For the current index, update the signing committee
+                    self.update_superblock_signing_committee(superblock_index);
                 }
 
                 // This replace is needed because the for loop below needs unique access to self,
@@ -246,6 +254,12 @@ impl SuperBlockState {
     /// Check if we had already received this superblock vote
     pub fn contains(&self, sbv: &SuperBlockVote) -> bool {
         self.received_superblocks.contains(sbv)
+    }
+
+    /// Updates the current superblock signing committee for a given superblock index
+    /// #FIXME This function should be update with issue 1395
+    pub fn update_superblock_signing_committee(&mut self, _current_index: u32) {
+        self.current_signing_committee = self.previous_ars_identities.clone();
     }
 }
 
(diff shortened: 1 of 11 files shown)


data_structures/src/superblock.rs
```diff
@@ -422,19 +435,19 @@ mod tests {
     #[test]
     fn superblock_state_default_add_votes() {
         // When the superblock state is initialized to default (for example when starting the node),
-        // all the received superblock votes are marked as `MaybeValid` or `AlreadySeen`
+        // all the received superblock votes are marked as `NotInArs'
```
aesedepece, line 438:
```suggestion
        // all the received superblock votes are marked as `NotInArs`
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -970,7 +970,25 @@ impl ChainManager {
...
         .map_err(|e, _, _| log::error!("Superblock building failed: {:?}", e))
         .and_then(move |(block_headers, last_hash), act, _ctx| {
-            let ars_members = &act.chain_state.last_ars;
+            let chain_info = act.chain_state.chain_info.as_ref().unwrap();
+            let x;
+            let ars_members = {
+                // Before reaching the epoch activity_period + collateral_age the bootstrap committee signs the superblock
+                if act.current_epoch.unwrap()
+                    > chain_info.consensus_constants.collateral_age
+                        + chain_info.consensus_constants.activity_period
```
aesedepece, line 979: `collateral_age` is measured in blocks, while `activity_period` is measured in epochs. Are we OK to add these together?

node/src/actors/chain_manager/mod.rs
```diff
@@ -970,7 +970,25 @@ impl ChainManager {
...
+                } else {
+                    x = chain_info
+                        .consensus_constants
+                        .bootstrapping_committee
+                        .iter()
+                        .map(|add| add.parse().unwrap())
+                        .collect();
+                    &x
+                }
+            };
```
aesedepece, line 991: By making `ars_members` not a reference (but referencing later when using it as an argument), we can get rid of the `x` binding altogether.

node/src/actors/chain_manager/mod.rs
lrubiorod: The idea is to leave decentralized when we are sure that the ARS doesnt contain shitty nodes. So the idea was to wait until requests are living for 1 activity period (stressed net). But as Adan said, if there are empty epochs, the collateral age is not the same, because it is counted in blocks, so should we use an offset?

girazoki: I see your concern with a small ARS by the time this timeframe elapses. But was not it the reason why we decided to put this timeframe in the first place? Perhaps we can double it if we feel safer

data_structures/src/superblock.rs
```diff
@@ -57,6 +57,18 @@ pub struct SuperBlockState {
 }
 
 impl SuperBlockState {
+    // Initialize the superblock state
+    pub fn new(superblock_genesis_hash: Hash, bootsrap_committee: Vec<PublicKeyHash>) -> Self {
+        let mut hs = HashSet::new();
+        hs.extend(bootsrap_committee.iter().cloned());
```
aesedepece, line 63:
```suggestion
    pub fn new(superblock_genesis_hash: Hash, bootstrap_committee: Vec<PublicKeyHash>) -> Self {
        let hs: HashSet<PublicKeyHash> = bootsrap_committee.iter().cloned().collect();
```

data_structures/src/superblock.rs
```diff
@@ -422,19 +431,19 @@ mod tests {
     #[test]
     fn superblock_state_default_add_votes() {
         // When the superblock state is initialized to default (for example when starting the node),
-        // all the received superblock votes are marked as `MaybeValid` or `AlreadySeen`
+        // all the received superblock votes are marked as `NotInArs`
         let mut sbs = SuperBlockState::default();
```
tmpolaczyk, line 435: You could add a test similar to this one but using a non-empty bootstrapping committee. Check that for the first superblock only votes from members of the committee are accepted.

node/src/actors/chain_manager/actor.rs
```diff
@@ -175,11 +176,20 @@ impl ChainManager {
                             },
                         };
 
+                        let bootstrap_committee = chain_info
+                            .consensus_constants
+                            .bootstrapping_committee
+                            .iter()
+                            .map(|add| add.parse().unwrap())
```
tmpolaczyk, line 183:
```suggestion
                            .map(|add| add.parse().expect("Malformed bootstrapping committee"))
```

data_structures/src/superblock.rs
girazoki: great, doing that now

node/src/actors/chain_manager/actor.rs
```diff
@@ -12,6 +12,7 @@ use crate::{
     config_mngr, signature_mngr, storage_mngr,
 };
 use witnet_crypto::key::CryptoEngine;
+use witnet_data_structures::superblock::SuperBlockState;
```
tmpolaczyk, line 15: fmt

node/src/actors/chain_manager/mod.rs
```diff
@@ -970,12 +963,30 @@ impl ChainManager {
...
+                    > chain_info.consensus_constants.collateral_age
+                        + chain_info.consensus_constants.activity_period
+                {
+                    act.chain_state.last_ars.clone()
+                } else {
+                    chain_info
+                        .consensus_constants
+                        .bootstrapping_committee
+                        .iter()
+                        .map(|add| add.parse().unwrap())
```
tmpolaczyk, line 981:
```suggestion
                            .map(|add| add.parse().expect("Malformed bootstrapping committee"))
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -1657,7 +1668,7 @@ mod tests {
             chain_manager.get_superblock_beacon(),
             CheckpointBeacon {
                 checkpoint: 0,
-                hash_prev_block: genesis_hash,
+                hash_prev_block: Hash::default(),
```
tmpolaczyk, line 1671: I think this test is supposed to check that the superblock genesis hash is set correctly. And it is not, because the `SuperBlockState` is not initialized. You could try to do `chain_manager.chain_state.superblock_state = SuperBlockState::new(...)` and set the genesis hash there.

data_structures/src/superblock.rs
```diff
@@ -127,26 +142,19 @@ impl SuperBlockState {
...
     fn is_valid(&self, sbv: &SuperBlockVote) -> Option<bool> {
-        match self.current_superblock_index {
-            // We do not know the current index, we cannot know if the vote is valid
-            None => None,
-            // If the index is the same as the current one, the vote is valid if it is signed by a
-            // member of the ARS
-            Some(x) if x == sbv.superblock_index => self
-                .previous_ars_identities
+        if self.current_superblock_index == sbv.superblock_index {
+            self.current_signing_committee
```
tmpolaczyk, line 146: Hey, don't remove the comments
```suggestion
            // If the index is the same as the current one, the vote is valid if it is signed by a
            // member of the current signing committee
            self.current_signing_committee
```

data_structures/src/superblock.rs
```diff
@@ -224,13 +228,12 @@ impl SuperBlockState {
...
     }
 
-    /// Returns an option with the last superblock hash and index if none of these fields is None.
-    /// else, returns None
-    pub fn get_beacon(&self) -> Option<CheckpointBeacon> {
-        Some(CheckpointBeacon {
-            checkpoint: self.current_superblock_index?,
-            hash_prev_block: self.current_superblock_hash?,
-        })
+    /// Returns ahe last superblock hash and index.
```
aesedepece, line 231:
```suggestion
    /// Returns the last superblock hash and index.
```

data_structures/src/superblock.rs
```diff
@@ -509,9 +546,9 @@ mod tests {
         );
 
         // The only think that should change is the superblock_index
-        expected_sbs.current_superblock_index = Some(1);
-        // And the superblock_hash, which will be set to None
-        expected_sbs.current_superblock_hash = None;
+        expected_sbs.current_superblock_index = 1;
+        // And the superblock_hash, which will be set to the previous superblock
+        expected_sbs.current_superblock_hash = _sb1.hash();
```
tmpolaczyk, line 551: You should remove the `_` if the variable is no longer unused.

```suggestion
        expected_sbs.current_superblock_hash = sb1.hash();
```

data_structures/src/superblock.rs
aesedepece: Is he?

data_structures/src/superblock.rs
```diff
@@ -127,26 +142,21 @@ impl SuperBlockState {
...
+                .map_or(Some(false), |x| {
+                    Some(x.contains(&sbv.secp256k1_signature.public_key.pkh()))
+                })
+        } else if ((self.current_superblock_index.saturating_sub(1))
+            ..=self.current_superblock_index.saturating_add(1))
+            .contains(&sbv.superblock_index)
+        {
+            None
+        } else {
+            Some(false)
```
tmpolaczyk, line 159:
```suggestion
            // If the index is not the same as the current one, but it is within an acceptable range 
            // of [x-1, x+1], broadcast the vote without checking if it is a member of the ARS, as 
            // the ARS may have changed and we do not keep older copies of the ARS in memory 
            None
        } else {
            // If the index is outside the [x-1, x+1] range, it is considered not valid
            Some(false)
```

PR #1407, opened 2020-07-20
Superblock signing committee

### LINKED ISSUE #1391: Vote for superblocks only if our ARS position passes voting requirements
- Take into account voting requirements as defined by #1395
- Implement method for collecting identities of ARS members that pass the requirements
- Guard superblock signing and broadcasting from being done if we don't pass the requirements

### LINKED ISSUE #1393: Reject superblock votes if ARS position fails requirements

### LINKED ISSUE #1395: Define requirements for ARS subsets
Define the subset of the ARS in charge of proposing superblocks.
This includes:
- choose the size of the the subset
- order the ARS by reputation
- Iterate the list to get the subset

clbartoli:
This PR creates the method `update_superblock_signing_committee` that returns the subset of the ARS in charge of signing a specific superblock. In particular:
- Orders the ARS with the method `get_rep_ordered_pkh_list` and stores it in the `chain_state.last_ars`
- Adds the size of the superblock signing committee in the consensus constants, and sets it to 100 as default. 
- Creates the function `update_superblock_signing_committee`, this function is called when building the superblock using the previous superblock_hash.
- Adds previous_ordered_ars_identities and current_ordered_ars_identities in the superblockState

Close #1395 
Close #1391 
Close #1393

### DIFF
--- a/config/src/defaults.rs
+++ b/config/src/defaults.rs
@@ -133,6 +133,11 @@ pub trait Defaults {
             .unwrap()
     }
 
+    /// Default period between superblocks
+    fn consensus_constants_superblock_signing_committee_size(&self) -> u32 {
+        100
+    }
+
     /// JSON-RPC server enabled by default
     fn jsonrpc_enabled(&self) -> bool {
         true
--- a/config/src/defaults.rs
+++ b/config/src/defaults.rs
@@ -133,7 +133,7 @@ pub trait Defaults {
             .unwrap()
     }
 
-    /// Default period between superblocks
+    /// Default size of the superblock signing committee
     fn consensus_constants_superblock_signing_committee_size(&self) -> u32 {
         100
     }
--- a/data_structures/src/superblock.rs
+++ b/data_structures/src/superblock.rs
@@ -23,7 +23,7 @@ pub enum AddSuperBlockVote {
     /// unverifiable vote because we do not have the required ARS state
     MaybeValid,
     /// vote from a peer not in the ARS
-    NotInArs,
+    NotInSigningCommittee,
     /// valid vote but with different hash
     ValidButDifferentHash,
     /// valid vote with identical hash
@@ -135,7 +135,7 @@ impl SuperBlockState {
                     if sbv.superblock_index == self.current_superblock_index
                         || self.previous_ars_identities.is_none()
                     {
-                        AddSuperBlockVote::NotInArs
+                        AddSuperBlockVote::NotInSigningCommittee
                     } else {
                         AddSuperBlockVote::InvalidIndex
                     }
@@ -497,19 +497,19 @@ mod tests {
     #[test]
     fn superblock_state_default_add_votes() {
         // When the superblock state is initialized to default (for example when starting the node),
-        // all the received superblock votes are marked as `NotInArs`
+        // all the received superblock votes are marked as `NotInSigningCommittee`
         let mut sbs = SuperBlockState::default();
 
         let v1 = SuperBlockVote::new_unsigned(Hash::SHA256([1; 32]), 0);
-        assert_eq!(sbs.add_vote(&v1), AddSuperBlockVote::NotInArs);
+        assert_eq!(sbs.add_vote(&v1), AddSuperBlockVote::NotInSigningCommittee);
 
         let v2 = SuperBlockVote::new_unsigned(Hash::SHA256([2; 32]), 0);
-        assert_eq!(sbs.add_vote(&v2), AddSuperBlockVote::NotInArs);
+        assert_eq!(sbs.add_vote(&v2), AddSuperBlockVote::NotInSigningCommittee);
 
         // Before building the first superblock locally we do not know the current superblock_index,
-        // so all the superblock votes will be "NotInArs"
+        // so all the superblock votes will be "NotInSigningCommittee"
         let v3 = SuperBlockVote::new_unsigned(Hash::SHA256([3; 32]), 33);
-        assert_eq!(sbs.add_vote(&v3), AddSuperBlockVote::NotInArs);
+        assert_eq!(sbs.add_vote(&v3), AddSuperBlockVote::NotInSigningCommittee);
     }
 
     #[test]
@@ -537,7 +537,7 @@ mod tests {
         let mut v1 = SuperBlockVote::new_unsigned(Hash::default(), 0);
         v1.secp256k1_signature.public_key = p2;
 
-        assert_eq!(sbs.add_vote(&v1), AddSuperBlockVote::NotInArs);
+        assert_eq!(sbs.add_vote(&v1), AddSuperBlockVote::NotInSigningCommittee);
     }
 
     #[test]
@@ -564,7 +564,7 @@ mod tests {
             )
             .unwrap();
         let v1 = SuperBlockVote::new_unsigned(sb1.hash(), 0);
-        assert_eq!(sbs.add_vote(&v1), AddSuperBlockVote::NotInArs);
+        assert_eq!(sbs.add_vote(&v1), AddSuperBlockVote::NotInSigningCommittee);
     }
 
     #[test]
@@ -648,7 +648,7 @@ mod tests {
         let mut sbs = SuperBlockState::default();
 
         let v0 = SuperBlockVote::new_unsigned(Hash::SHA256([1; 32]), 0);
-        assert_eq!(sbs.add_vote(&v0), AddSuperBlockVote::NotInArs);
+        assert_eq!(sbs.add_vote(&v0), AddSuperBlockVote::NotInSigningCommittee);
         assert_eq!(sbs.add_vote(&v0), AddSuperBlockVote::AlreadySeen);
 
         let block_headers = vec![BlockHeader::default()];
@@ -670,7 +670,7 @@ mod tests {
             )
             .unwrap();
         // After building a new superblock the cache is invalidated but the previous ARS is still empty
-        assert_eq!(sbs.add_vote(&v0), AddSuperBlockVote::NotInArs);
+        assert_eq!(sbs.add_vote(&v0), AddSuperBlockVote::NotInSigningCommittee);
         assert_eq!(sbs.add_vote(&v0), AddSuperBlockVote::AlreadySeen);
 
         let _sb2 = sbs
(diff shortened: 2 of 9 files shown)


node/src/actors/chain_manager/mod.rs
```diff
@@ -896,6 +892,11 @@ impl ChainManager {
         let consensus_constants = self.consensus_constants();
 
         let superblock_period = u32::from(consensus_constants.superblock_period);
+        let signing_committee_size =
```
girazoki, line 895: twice

data_structures/src/superblock.rs
```diff
@@ -257,11 +264,45 @@ impl SuperBlockState {
...
 
-    /// Updates the current superblock signing committee for a given superblock index
-    /// #FIXME This function should be update with issue 1395
-    pub fn update_superblock_signing_committee(&mut self, _current_index: u32) {
-        self.current_signing_committee = self.previous_ars_identities.clone();
+    /// Updates the current superblock signing committee for a given superblock hash
+    pub fn update_superblock_signing_committee(
+        &mut self,
+        _signing_committee_size: u32,
+    ) -> Option<HashSet<PublicKeyHash>> {
```
girazoki, line 271: Internally updating self.current_signing_committee and also returning it seems a bit odd to me, although I understand this is for testing purposes. Maybe we can just make this agnostic of superblock_state and make all the necessary parameters be inserted as arguments? We could just call this function something like calculate_superblock_signing_committee

data_structures/src/superblock.rs
```diff
@@ -257,11 +264,41 @@ impl SuperBlockState {
...
+        } else {
+            // Get the number of subsets of _signing_committee_size members
+            let n = ars_ordered.len() as u64 / u64::from(_signing_committee_size);
+            // Start counting the members of the subset from the superblock_hash
+            let a = u64::from(*self.current_superblock_hash.as_ref().get(0).unwrap());
+            let b = a % u64::from(_signing_committee_size);
+            let first_member = b % n;
+            // Get the subset
+            let subset = magic_partition(&ars_ordered.to_vec(), first_member, n);
+            let hs: HashSet<PublicKeyHash> = subset.iter().cloned().collect();
```
aesedepece, line 286: This could be an alternative to the `magic_partition` thing:
```suggestion
            let subset: HashSet<PublicKeyHash> = ars_ordered.iter()
                .cloned()
                .enumerate()
                .filter_map(|(index, member)| {
                    if (index + offset) % possible_subsets == 0 {
                        Some(member)
                    } else {
                        None
                    }
                })
                .collect()
```
(Where `offset` was `a` and `possible_subsets` was `n` before)

data_structures/src/superblock.rs
```diff
@@ -257,11 +264,41 @@ impl SuperBlockState {
...
+        if ars_ordered.len() < usize::try_from(_signing_committee_size).unwrap() {
+            self.current_signing_committee = self.previous_ars_identities.clone();
+            self.current_signing_committee.clone()
+        } else {
+            // Get the number of subsets of _signing_committee_size members
+            let n = ars_ordered.len() as u64 / u64::from(_signing_committee_size);
+            // Start counting the members of the subset from the superblock_hash
+            let a = u64::from(*self.current_superblock_hash.as_ref().get(0).unwrap());
+            let b = a % u64::from(_signing_committee_size);
+            let first_member = b % n;
```
aesedepece, line 283: I have the concern that applying modulo on a uniform number makes the resulting number non-uniform. For example, `x % 100 for x in [0, 255]` gives 2 points for `[0, 199]` but 3 for `[200, 255]`:

image

This may translate into some positions in the ARS being "luckier" than others.

aesedepece: Here's a good example about modulo bias:
http://martin.upspin.org/2017/12/14/modulo-bias.html

aesedepece: Properly speaking, bias is introduced whenever you want to map from set A to set B where the cardinality of set B is not a factor or multiple of set A :sunglasses:

aesedepece: I believe this snippet actually gives uniform values in a specific range out of uniform values in a different range:
```rust
use rand::{Rng, rngs::StdRng};
use rand_core::SeedableRng;

pub fn main() {
    let hash = [0;32];
    let rng = &mut StdRng::from_seed(hash);
    let offset = rng.gen_range(0, 255);
    assert_eq!(offset, 173);
}
```

aesedepece: But the problem is that that code is consensus-critical but hard to port to other platforms (very Rust specific), so we may have to find something else.

clbartoli: The magic_partition function has been changed and it shouldn't have that problem anymore. As an example, now given a vector [0, 1, 2, 3, 4], if you specify to have a subset starting from 3 and length 2, the function returns the subset [3, 0].

config/src/defaults.rs
```diff
@@ -133,6 +133,11 @@ pub trait Defaults {
             .unwrap()
     }
 
+    /// Default size of the superblock signing committee
+    fn consensus_constants_superblock_signing_committee_size(&self) -> u32 {
+        100
+    }
```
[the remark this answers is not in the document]
aesedepece: Dismissing my suggestion after being proven quite inconvenient

aesedepece:
Let me suggest the usage of `Arc` for making `PublicKeyHash` cloning cheaper in terms of memory footprint, usage of the heap and most likely also CPU cycles when deriving some of the fields in `SuperBlockState` from other of its fields.

Here's a commit that you can put on top of your PR if you like the idea:
https://github.com/aesedepece/witnet-rust/commit/a8bf58c057c271ad461700e895db32e394af1baa

data_structures/src/superblock.rs
```diff
@@ -32,6 +35,8 @@ pub enum AddSuperBlockVote {
 pub struct SuperBlockState {
     // Set of ARS identities that will be able to send superblock votes in the next superblock epoch
     current_ars_identities: HashSet<PublicKeyHash>,
+    // current_ars_identities ordered by reputation
+    current_ordered_ars_identities: Vec<PublicKeyHash>,
```
lrubiorod, line 39: why is the reason to have two similar fields, one ordered and other not. It should be possible to keep only the ordered one?

data_structures/src/superblock.rs
```diff
@@ -43,6 +48,8 @@ pub struct SuperBlockState {
     identities_that_voted_more_than_once: HashMap<PublicKeyHash, Vec<SuperBlockVote>>,
     // Set of ARS identities that can currently send superblock votes
     previous_ars_identities: Option<HashSet<PublicKeyHash>>,
+    // previous_ars_identities ordered by reputation
+    previous_ordered_ars_identities: Vec<PublicKeyHash>,
```
lrubiorod, line 52: same concern here

data_structures/src/superblock.rs
lrubiorod: best performance of the hashset :P

data_structures/src/superblock.rs
```diff
@@ -256,14 +272,54 @@ impl SuperBlockState {
...
+            first.try_into().unwrap(),
+            signing_committee_size.try_into().unwrap(),
+        );
+        let hs: HashSet<PublicKeyHash> = subset.iter().cloned().collect();
+        Some(hs)
     }
 }
 
+// Take size element out of v.len() starting with element at index first:
+// magic_partition(v, 3, 3), v=[]0, 1, 2, 3, 4, 5].
```
lrubiorod, line 303:
```suggestion
// magic_partition(v, 3, 3), v=[0, 1, 2, 3, 4, 5].
```

data_structures/src/superblock.rs
```diff
@@ -931,6 +1141,196 @@ mod tests {
         assert_eq!(sbs.votes_on_each_superblock, HashMap::new());
     }
 
+    #[test]
+    fn superblock_voted_by_singing_committee() {
```
lrubiorod, line 1145:
```suggestion
    fn superblock_voted_by_signing_committee() {
```

data_structures/src/superblock.rs
```diff
@@ -931,6 +1141,196 @@ mod tests {
...
+        // When adding a superblock vote, it should be valid only by members of the
+        // superblock singning commiittee.
```
lrubiorod, line 1147:
```suggestion
        // superblock signing committee.
```

data_structures/src/superblock.rs
```diff
@@ -931,6 +1141,196 @@ mod tests {
...
+                &block_headers,
+                &ars_identities,
+                &ars_identities,
+                &ordered_ars,
+                100,
+                0,
+                genesis_hash,
+            )
+            .unwrap();
+        sbs.previous_ordered_ars_identities = vec![p1.pkh(), p2.pkh(), p3.pkh()];
```
lrubiorod, line 1250:
```suggestion
        sbs.previous_ordered_ars_identities = ars_identities.clone();
```

data_structures/src/superblock.rs
```diff
@@ -931,6 +1141,196 @@ mod tests {
...
+        sbs.previous_ordered_ars_identities = vec![
+            p1.pkh(),
+            p2.pkh(),
+            p3.pkh(),
+            p4.pkh(),
+            p5.pkh(),
+            p6.pkh(),
+            p7.pkh(),
+            p8.pkh(),
+        ];
```
lrubiorod, line 1319:
```suggestion
        sbs.previous_ordered_ars_identities = ars_identities.clone();
```

data_structures/src/superblock.rs
```diff
@@ -256,14 +272,54 @@ impl SuperBlockState {
...
+            signing_committee_size.try_into().unwrap(),
+        );
+        let hs: HashSet<PublicKeyHash> = subset.iter().cloned().collect();
+        Some(hs)
     }
 }
 
+// Take size element out of v.len() starting with element at index first:
+// magic_partition(v, 3, 3), v=[]0, 1, 2, 3, 4, 5].
+// Will return elements at index 3, 5, 1.
```
lrubiorod, line 304: Should be great also add tests for the magic_partition function

aesedepece:
> Let me suggest the usage of `Arc` for making `PublicKeyHash` cloning cheaper in terms of memory footprint, usage of the heap and most likely also CPU cycles when deriving some of the fields in `SuperBlockState` from other of its fields.
> 
> Here's a commit that you can put on top of your PR if you like the idea:
> [aesedepece@a8bf58c](https://github.com/aesedepece/witnet-rust/commit/a8bf58c057c271ad461700e895db32e394af1baa)

I just run some benchmarks... and surprise: there's no big advantage in the case of `PublicKeyHash`. The optimization and the `Arc`'s intrinsic overhead basically cancel each other out :shrug:

PR #1428, opened 2020-07-30
feat(node): send superblock votes at the end of the synchronization

### LINKED ISSUE #1427: Send gathered votes for last superblock when sending the last block batch

mariocao:
Closes #1427

### DIFF
--- a/node/src/actors/session/handlers.rs
+++ b/node/src/actors/session/handlers.rs
@@ -29,9 +29,9 @@ use crate::actors::{
     messages::{
         AddBlocks, AddCandidates, AddConsolidatedPeer, AddPeers, AddSuperBlockVote, AddTransaction,
         CloseSession, Consolidate, EpochNotification, GetBlocksEpochRange,
-        GetHighestCheckpointBeacon, GetItem, PeerBeacon, RemoveAddressesFromTried, RequestPeers,
-        SendGetPeers, SendInventoryAnnouncement, SendInventoryItem, SendLastBeacon,
-        SendSuperBlockVote, SessionUnitResult,
+        GetHighestCheckpointBeacon, GetItem, GetSuperBlockVotes, PeerBeacon,
+        RemoveAddressesFromTried, RequestPeers, SendGetPeers, SendInventoryAnnouncement,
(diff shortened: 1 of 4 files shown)


node/src/actors/session/handlers.rs
```diff
@@ -223,9 +223,23 @@ impl StreamHandler<BytesMut, Error> for Session {
                             .into_actor(self)
                             .map_err(|e, _, _| log::error!("Inventory request error: {}", e))
                             .and_then(move |item_responses, session, _| {
+                                let mut send_superblock_votes = false;
                                 for (i, item_response) in item_responses.into_iter().enumerate() {
                                     match item_response {
-                                        Ok(item) => send_inventory_item_msg(session, item),
+                                        Ok(item) => {
+                                            if let InventoryItem::Block(block) = &item {
+                                                if block.hash()
```
tmpolaczyk, line 231: Can we compare the checkpoints instead? 

```suggestion
                                                if block.block_header.beacon.checkpoint
```

node/src/actors/session/handlers.rs
```diff
@@ -244,14 +258,38 @@ impl StreamHandler<BytesMut, Error> for Session {
...
                                             break;
                                         }
                                     }
                                 }
 
-                                actix::fut::ok(())
+                                actix::fut::ok(send_superblock_votes)
+                            })
+                            .and_then(|send_superblock_votes, session, _ctx| {
+                                let chain_manager_addr = ChainManager::from_registry();
```
tmpolaczyk, line 270: This section needs a comment explaining what are we doing and why do we need to do it. For example:

```suggestion
                            .and_then(|send_superblock_votes, session, _ctx| {
                                // If this is the last batch, send to the peer all the superblock votes that are currently stored in
                                // ChainManager. This allows faster synchronization in some cases, because if a node has not
                                // received enough votes, it will revert to the last consolidated superblock and start the
                                // synchronization again.
                                // Note that it is not strictly needed as part of the protocol.
                                let chain_manager_addr = ChainManager::from_registry();
```

mariocao: Sounds right! :+1:

PR #1429, opened 2020-07-30
Reenable and improve NTP support

### LINKED ISSUE #1189: Print NTP drift in milliseconds
The NTP time drift is currently being printed in seconds, which is not granular enough in most cases. Using milliseconds will make this metric much more meaningful.

### LINKED ISSUE #1193: Experiment and troubleshoot reenabling fast clock adjustment using NTP
When we implemented NTP there were some adverse side effects that led us to sort of disable such feature by setting a huge readjustment period so it would only be triggered when the node starts up.

We have made significant improvements to the epoch manager and chain manager ever since, which could mean that clock readjustment could be re-enabled safely by reducing the NTP query period again.

This issue entails experimenting with fast periods and troubleshooting any problem that may be found.

aesedepece:
This reenables NTP, adds a ±5s variance on its polling period and prints drift with milliseconds.

Fix #1193 
Fix #1189

### DIFF
--- a/config/src/defaults.rs
+++ b/config/src/defaults.rs
@@ -318,7 +318,7 @@ pub trait Defaults {
     }
 
     fn ntp_update_period(&self) -> Duration {
-        Duration::from_secs(600)
+        Duration::from_secs(1024)
     }
 
     fn ntp_server(&self) -> Vec<String> {
--- a/node/src/actors/epoch_manager/mod.rs
+++ b/node/src/actors/epoch_manager/mod.rs
@@ -235,7 +235,19 @@ impl EpochManager {
     }
 
     /// Method to monitor checkpoints and execute some actions on each
+    ///
+    /// This function internally introduces a small random variance (±5s) on the update period to
+    /// prevent multiple nodes that were started at the same time from bunching at the NTP servers,
+    /// just like `ntpd` does.
     fn update_ntp_timestamp(&self, ctx: &mut Context<Self>, period: Duration, addr: String) {
+        // Variance is not applied on periods that are smaller than 10 seconds
+        let period = if period > Duration::from_secs(10) {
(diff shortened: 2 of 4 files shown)


tmpolaczyk:
Looks good, but the problem we had with NTP was missing epoch notifications, so I wouldn't enable NTP until we are sure that that no longer happens.

aesedepece:
> Looks good, but the problem we had with NTP was missing epoch notifications, so I wouldn't enable NTP until we are sure that that no longer happens.

See this: https://github.com/witnet/witnet-rust/issues/1193#issuecomment-666237346

I don't know if that's enough evidence, though. What experiment do you propose for assessing the safety of this?

Also, I kind of remember that shortly after we disabled NTP, we also implemented a check for missing epochs, because we were still having the same issue even without NTP. So I'm somehow expecting that, being that protection in place, if NTP causes a node to time travel from one epoch to the previous or following one, the `EpochManager` will be able to manage the situation elegantly.

aesedepece:
At a deeper look at the logs, once in a while I can see this strange epoch logging ordering, which will sound familiar to you:

image

However, as mentioned before, this seem not to be causing any issue at all, most probably because of the missing epochs check (it is even noticeable in the screenshot that blocks are being consolidated in a timely and orderly fashion)

tmpolaczyk:
> What experiment do you propose for assessing the safety of this?

Run a few nodes on the current testnet with `ntp_period = 60` and see if they stay synced all the time.

config/src/defaults.rs
```diff
@@ -318,7 +318,7 @@ pub trait Defaults {
     }
 
     fn ntp_update_period(&self) -> Duration {
-        Duration::from_secs(600)
+        Duration::from_secs(1024)
```
mariocao, line 321: why 1024? :)

PR #1430, opened 2020-07-30
feat(reputation): do not update reputation if no block

mariocao:


reputation/src/ars.rs
```diff
@@ -115,15 +113,16 @@ where
     where
         M: IntoIterator<Item = K>,
     {
-        if self.queue.len() >= self.capacity {
+        let identities: HashSet<K> = identities.into_iter().collect();
+
+        if !identities.is_empty() && self.queue.len() >= self.capacity {
```
tmpolaczyk, line 118: Why check `!identities.is_empty()`? If we want to assume that `identities` is never empty we should add `assert!(identities.is_empty())`.

reputation/src/ars.rs
```diff
@@ -449,32 +431,29 @@ mod tests {
...
+    #[should_panic]
+    fn update_ars_empty_should_fail() {
+        let mut ars = ActiveReputationSet::new(3);
+        let id1 = "Alice".to_string();
         assert_eq!(ars.contains(&id1), false);
-        assert_eq!(ars.contains(&id2), true);
-        assert_eq!(ars.contains(&id3), true);
-        assert_eq!(ars.last_update, 21);
-        assert_eq!(ars.active_identities_number(), 2);
+        ars.update(vec![], 1);
```
tmpolaczyk, line 457:
```suggestion
        let _ = ars.update(vec![], 1);
```

PR #1432, opened 2020-08-03
Split blocks at batch function

girazoki:
This PR provides functionality for split_blocks_batch function which splits a block batch in:

- Blocks needed to construct a consolidated superblock
- Blocks needed to potentially construct an intermediate superblock

### DIFF
--- a/node/src/actors/chain_manager/handlers.rs
+++ b/node/src/actors/chain_manager/handlers.rs
@@ -1182,7 +1182,7 @@ where
     });
     if let Some(wrong_index) = wrong_index {
         // We received blocks that do not match the current epoch and the last consolidated superblock.
-        // as an example, if the last consolidated superblock has the block 9 inside, and we are in epoch 50,
+        // As an example, if the last consolidated superblock has the block 9 inside, and we are in epoch 50,
         // it means we reverted somehow. Thus Blocks between 10 and 49 cannot exist.
         return Err(ChainManagerError::WrongBlocksForSuperblock {
             wrong_index: key(&blocks[wrong_index]),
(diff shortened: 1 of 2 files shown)


node/src/actors/chain_manager/mod.rs
```diff
@@ -135,6 +148,18 @@ impl Default for StateMachine {
...
 }
 
+/// Synchronization target determined by the beacons received from outbound peers
+#[derive(Clone, Debug)]
+pub struct SyncTarget {
+    // TODO: the target block must be set, but the node will not assume that it is valid
+    block: CheckpointBeacon,
+    // The target superblock must always be set. Here we only know the superblock index and hash,
+    // we do not know the block hash. The block index can be derived from the superblock index.
+    // This must be a superblock beacon consolidated with more than 2/3 of the votes, and it we be
```
clbartoli, line 158:
```suggestion
    // This must be a superblock beacon consolidated with more than 2/3 of the votes, and it must be
```

node/src/actors/chain_manager/handlers.rs
```diff
@@ -1336,4 +1437,143 @@ mod tests {
...
+    fn test_split_blocks_batch() {
+        use BlockBatches::*;
+        let mut sync_target = SyncTarget {
+            block: Default::default(),
+            superblock: Default::default(),
+        };
+        let superblock_period = 10;
+
+        // TODO: remove all the calls to b(x)
+        let b = |x| x;
```
mariocao, line 1450: As the comment says, we could remove these lines and remove all calls to b(x) :smile:

node/src/actors/chain_manager/handlers.rs
```diff
@@ -1140,6 +1141,109 @@ impl Handler<GetMempool> for ChainManager {
...
+        - ((current_superblock_index - sync_target.superblock.checkpoint) % 2))
+        * superblock_period;
+
+    let wrong_index = blocks.iter().position(|block| {
+        key(block) >= sync_target.superblock.checkpoint * superblock_period
+            && key(block) < first_valid_block
+    });
+    if let Some(wrong_index) = wrong_index {
+        // We received blocks that do not match the current epoch and the last consolidated superblock.
+        // as an example, if the last consolidated superblock has the block 9 inside, and we are in epoch 50,
```
clbartoli, line 1185:
```suggestion
        // As an example, if the last consolidated superblock has the block 9 inside, and we are in epoch 50,
```

PR #1420, opened 2020-07-27
[WIP] Superblock sync and consensus

tmpolaczyk:

PR #1435, opened 2020-08-06
feat: implement new consensus based on superblocks

mariocao:
This PR supersedes the WIP PR #1420 with code from @tmpolaczyk and @girazoki .

In summary it replaces the previous consensus algorithm with a new one based on superblocks and ARS committee votes:

- updates the last beacon structure with additional superblock information in `SessionsManager`.
- implements new synchronization logic in `ChainManager`.

This PR has been tested under the following conditions:

1. Initial sync of 2 nodes that are in the initial signing committee: they reached Synced state and start voting superblocks as expected.
2. Additional node synchronization to an already bootstrapped chain.
3. Revert the chain if not sufficient superblock votes are broadcasted.
4. Recovering a initial sync node after a failure (storage was not removed).
5. Having a small activity period, thus checking if changes in the ARS are also working properly.

The current implementation also has some known limitations:

- Initial synchronization requires that `checkpoint_zero_timestamp` has to be in the future.
- Turning of a signing committee member and removing its storage leads to a change in the BLS key. If the node votes twice for the same superblock, it will be considered invalid.

Additionally, there are potential code improvements that were not addressed in this PR:

- Verify that superblock creation does not start until all the blocks have been persisted in the `Inventory` (e.g. line 387 in `chain_manager/handlers.rs`).
- Some errors in the `AddBlocks` handler are not being elevated so that hey could be logged properly (e.g. returning `return actix::fut::err(());` in line 478 of `chain_manager/handlers.rs`).
- Refactor in functions `process_blocks_batch` and `process_first_batch`.
- In `superblock_consensus` function in `chain_manager/handlers.rs` line 708, the `block_consensus` is retrieved from the first element. Maybe be there is a better way to choose one (e.g. random?).

### DIFF
--- a/node/src/actors/chain_manager/handlers.rs
+++ b/node/src/actors/chain_manager/handlers.rs
@@ -22,7 +22,7 @@ use witnet_validations::validations::validate_rad_request;
 use super::{transaction_factory, ChainManager, ChainManagerError, StateMachine, SyncTarget};
 use crate::{
     actors::{
-        chain_manager::BlockCandidate,
+        chain_manager::{handlers::BlockBatches::*, BlockCandidate},
         messages::{
             AddBlocks, AddCandidates, AddCommitReveal, AddSuperBlockVote, AddTransaction,
             Broadcast, BuildDrt, BuildVtt, EpochNotification, GetBalance, GetBlocksEpochRange,
@@ -343,38 +343,42 @@ impl Handler<AddBlocks> for ChainManager {
                     superblock_period,
                 );
 
-                use BlockBatches::*;
                 match block_batches {
                     // TargetNotReached:
                     // 1. process blocks (not yet ready for consolidation)
                     // 2. requests block batch -> revert to WaitingConsensus
-                    Ok(TargetNotReached(blocks1)) => {
+                    Ok(TargetNotReached(blocks)) => {
                         let (batch_succeeded, num_processed_blocks) =
-                            self.process_first_batch(ctx, &sync_target, &blocks1);
+                            self.process_first_batch(ctx, &sync_target, &blocks);
                         if !batch_succeeded {
                             return;
                         }
-                        log_sync_progress(&sync_target, &blocks1, num_processed_blocks, 1);
 
-                        log::debug!("1 Target not reached, request blocks batch");
+                        log::debug!("TargetNotReached: superblock target #{} not reached, requesting more blocks. ({} processed blocks)",
+                            sync_target.superblock.checkpoint, num_processed_blocks);
                         self.request_blocks_batch(ctx);
                     }
                     // SyncWithoutCandidate:
                     // 1. process blocks
                     // 2. construct consolidated superblock (if needed)
                     // 3. handle remaining blocks
-                    Ok(SyncWithoutCandidate(blocks1, blocks2)) => {
+                    Ok(SyncWithoutCandidate(consolidate_blocks, remainig_blocks)) => {
                         let (batch_succeeded, num_processed_blocks) =
-                            self.process_first_batch(ctx, &sync_target, &blocks1);
+                            self.process_first_batch(ctx, &sync_target, &consolidate_blocks);
                         if !batch_succeeded {
                             return;
                         }
-                        log_sync_progress(&sync_target, &blocks1, num_processed_blocks, 1);
+                        log_sync_progress(
+                            &sync_target,
+                            &consolidate_blocks,
+                            num_processed_blocks,
+                            "SyncWithoutCandidate(consolidation)",
+                        );
 
                         if let Some(consolidate_superblock_epoch) = self.superblock_consolidation_is_needed(&sync_target, superblock_period) {
                             // We need to persist blocks in order to be able to construct the
                             // superblock
-                            self.persist_blocks_batch(ctx, blocks1);
+                            self.persist_blocks_batch(ctx, consolidate_blocks);
                             let to_be_stored =
                                 self.chain_state.data_request_pool.finished_data_requests();
                             self.persist_data_requests(ctx, to_be_stored);
@@ -391,11 +395,11 @@ impl Handler<AddBlocks> for ChainManager {
                             .and_then(move |(), act, ctx| {
                                 act.sm_state = StateMachine::WaitingConsensus;
                                 // Process remaining blocks
-                                let (batch_succeeded, num_processed_blocks) = act.process_blocks_batch(ctx, &sync_target, &blocks2);
+                                let (batch_succeeded, num_processed_blocks) = act.process_blocks_batch(ctx, &sync_target, &remainig_blocks);
                                 if !batch_succeeded {
                                     return actix::fut::err(());
                                 }
-                                log_sync_progress(&sync_target, &blocks2, num_processed_blocks, 2);
+                                log_sync_progress(&sync_target, &remainig_blocks, num_processed_blocks, "SyncWithoutCandidate(remaining)");
 
                                 actix::fut::ok(())
                             })
@@ -406,18 +410,27 @@ impl Handler<AddBlocks> for ChainManager {
                     // 2. construct consolidated superblock (if needed)
                     // 3. build and vote new candidate superblock
                     // 4. process remaining blocks
-                    Ok(SyncWithCandidate(blocks1, blocks2, blocks3)) => {
+                    Ok(SyncWithCandidate(
+                        consolidate_blocks,
+                        candidate_blocks,
+                        remaining_blocks,
+                    )) => {
                         let (batch_succeeded, num_processed_blocks) =
-                            self.process_first_batch(ctx, &sync_target, &blocks1);
+                            self.process_first_batch(ctx, &sync_target, &consolidate_blocks);
                         if !batch_succeeded {
                             return;
                         }
-                        log_sync_progress(&sync_target, &blocks1, num_processed_blocks, 1);
+                        log_sync_progress(
+                            &sync_target,
+                            &consolidate_blocks,
+                            num_processed_blocks,
+                            "SyncWithCandidate(consolidation)",
+                        );
 
                         if let Some(consolidate_superblock_epoch) = self.superblock_consolidation_is_needed(&sync_target, superblock_period) {
                             // We need to persist blocks in order to be able to construct the
                             // superblock
-                            self.persist_blocks_batch(ctx, blocks1);
+                            self.persist_blocks_batch(ctx, consolidate_blocks);
                             let to_be_stored =
                                 self.chain_state.data_request_pool.finished_data_requests();
                             self.persist_data_requests(ctx, to_be_stored);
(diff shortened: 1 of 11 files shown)


node/src/actors/chain_manager/handlers.rs
```diff
@@ -525,7 +640,44 @@ impl PeersBeacons {
     /// The beacons are `Option<CheckpointBeacon>`, so peers that have not
     /// sent us a beacon are counted as `None`. Keeping that in mind, we
     /// reach consensus as long as consensus_threshold % of peers agree.
-    pub fn consensus(&self, consensus_threshold: usize) -> Option<CheckpointBeacon> {
+    pub fn block_consensus(&self, consensus_threshold: usize) -> Option<CheckpointBeacon> {
```
mariocao, line 643: Maybe we could remove this function.

node/src/actors/chain_manager/mod.rs
```diff
@@ -258,14 +262,28 @@ impl ChainManager {
...
+        self.last_chain_state
+            .chain_info
+            .as_mut()
+            .unwrap()
+            .highest_superblock_checkpoint = self
+            .chain_state
+            .chain_info
+            .as_ref()
+            .unwrap()
+            .highest_superblock_checkpoint;
```
lrubiorod, line 285: there is a function called get_superblock_beacon() that returns exactly the same in line 688 in chain_managers/mod.rs

node/src/actors/chain_manager/mod.rs
```diff
@@ -691,6 +702,74 @@ impl ChainManager {
...
+                if act.chain_state.superblock_state.contains(&superblock_vote) {
+                    return actix::fut::ok(());
+                }
+                // Validate vote: the identity should be able to vote
+                // We broadcast all superblock votes with valid secp256k1 signature, signed by members
+                // of the ARS, even if the superblock hash is different from our local superblock hash.
+                // If the superblock index is different from the current one we cannot check ARS membership,
+                // so we broadcast it if the index is within an acceptable range (not too old).
+                let _should_broadcast =
+                    match act.chain_state.superblock_state.add_vote(&superblock_vote) {
```
lrubiorod, line 733: _should_broadcast is not used, it looks like a copy past from other part, but it is not clear to me why it is used here more than add_vote and show a log message

node/src/actors/chain_manager/handlers.rs
```diff
@@ -306,118 +302,234 @@ impl Handler<AddBlocks> for ChainManager {
...
+                    return;
+                }
 
-                            let beacon = self.get_chain_beacon();
-                            show_sync_progress(
-                                beacon,
-                                target_beacon,
-                                self.epoch_constants.unwrap(),
-                            );
+                let sync_target = self.sync_target.clone().unwrap();
```
girazoki, line 322: comment on why this cannot fail?

node/src/actors/chain_manager/mod.rs
lrubiorod: Due to there are 2 copies of that part of code, should we create a function?

node/src/actors/chain_manager/handlers.rs
```diff
@@ -306,118 +302,234 @@ impl Handler<AddBlocks> for ChainManager {
...
 
-                            if beacon == target_beacon {
-                                break;
-                            }
+                use BlockBatches::*;
+                match block_batches {
+                    // TargetNotReached:
+                    // 1. process blocks
+                    // 2. requests block batch -> revert to WaitingConsensus
+                    Ok(TargetNotReached(blocks1)) => {
```
girazoki, line 350: lets just change the naming of blocks 1 blocks 2 and blocks 3 maybe? in case 1 this can be something like non_consolidable_blocks

node/src/actors/chain_manager/handlers.rs
```diff
@@ -306,118 +302,234 @@ impl Handler<AddBlocks> for ChainManager {
...
+                                }
+                            })
+                            .and_then(move |candidate_superblock_epoch, act, ctx| {
+                                act.build_and_vote_candidate_superblock(ctx, candidate_superblock_epoch)
+                            })
+                            .and_then(move |_, act, ctx| {
+                                // Process remaining blocks
+                                let (batch_succeeded, num_processed_blocks) = act.process_blocks_batch(ctx, &sync_target, &blocks3);
+                                if !batch_succeeded {
+                                    log::error!("3 Received invalid blocks batch");
```
girazoki, line 473:
```suggestion
                                    log::error!("Received invalid blocks batch");
```

node/src/actors/chain_manager/mod.rs
mariocao: Nice one!

node/src/actors/chain_manager/mod.rs
```diff
@@ -715,6 +798,11 @@ impl ChainManager {
         })
         .into_actor(self)
         .and_then(move |(), act, _ctx| {
+            // Check if we already received this vote (again, because this future can be executed
+            // by multiple tasks concurrently)
+            if act.chain_state.superblock_state.contains(&superblock_vote) {
+                return actix::fut::err(());
```
lrubiorod, line 804: In lines 710 and 725 we returns Ok, now we have similar cases (789, 804) but we return an Error here, is it make sense?

node/src/actors/chain_manager/handlers.rs
```diff
@@ -306,118 +302,234 @@ impl Handler<AddBlocks> for ChainManager {
...
 
+fn log_sync_progress(
+    sync_target: &SyncTarget,
+    blocks: &[Block],
+    num_processed_blocks: usize,
+    iteration: usize,
+) {
+    // TODO: refactor Logs
+    if num_processed_blocks == 0 {
+        log::debug!("{} Sync done, 0 blocks processed", iteration);
```
girazoki, line 525: This log is not very descriptive

node/src/actors/chain_manager/mod.rs
```diff
@@ -903,7 +992,120 @@ impl ChainManager {
...
+                let mut superblock_vote =
+                    SuperBlockVote::new_unsigned(superblock_hash, superblock.index);
+                let bn256_message = superblock_vote.bn256_signature_message();
+
+                signature_mngr::bn256_sign(bn256_message)
+                    .map_err(|e| {
+                        log::error!("Failed to sign superblock with bn256 key: {}", e);
+                    })
+                    .and_then(move |bn256_keyed_signature| {
+                        // Actually, we don't need to include the BN256 public key because
```
lrubiorod, line 1021:
```suggestion
                        // TODO: Actually, we don't need to include the BN256 public key because
```

node/src/actors/chain_manager/mod.rs
mariocao: It is not the best way to do it, but it seems it is used for signalling that the vote shouldn't be broadcasted again.

node/src/actors/chain_manager/mod.rs
```diff
@@ -691,6 +702,74 @@ impl ChainManager {
...
+                            log::debug!(
+                                "Not forwarding superblock vote: invalid superblock index: {}",
+                                superblock_vote.superblock_index
+                            );
+
+                            false
+                        }
+                        AddSuperBlockVote::NotInSigningCommittee => {
+                            log::debug!(
+                                "Not forwarding superblock vote: identity not in ARS: {}",
```
clbartoli, line 754:
```suggestion
                                "Not forwarding superblock vote: identity not in Signing Committee: {}",
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -646,14 +654,17 @@ impl ChainManager {
                                 transaction: Transaction::Reveal(reveal),
                             })
                         }
+                        // Persist blocks and transactions but do not persist chain_state, it will
+                        // be persisted on superblock consolidation
+                        // TODO: this means that after a reorganization a call to getBlock or
```
girazoki, line 659: lets create a issue for this! quite importante IMHO

node/src/actors/chain_manager/handlers.rs
mariocao: Actually, the clone is no longer required! :)

node/src/actors/chain_manager/mod.rs
```diff
@@ -924,112 +1132,155 @@ impl ChainManager {
...
+                        log::info!("Before update: superblock {:?}", act.get_superblock_beacon());
+                        act.chain_state.chain_info.as_mut().unwrap().highest_superblock_checkpoint =
+                            act.chain_state.superblock_state.get_beacon();
+                        if act.sm_state == StateMachine::Synced {
+                            // Persist previous_chain_state with current superblock_state
+                            act.persist_chain_state(ctx);
+                        }
+                        log::info!("Consensus! Superblock {:?}", act.get_superblock_beacon());
+                        log::info!("Current tip of the chain: {:?}", act.get_chain_beacon());
+                        log::info!(
```
lrubiorod, line 1210: should it be in info level? It looks like debug level

node/src/actors/chain_manager/mod.rs
girazoki: FIXME: 1437

node/src/actors/chain_manager/mod.rs
```diff
@@ -924,112 +1132,155 @@ impl ChainManager {
...
 
-            let ars_members = {
-                // Before reaching the epoch activity_period + collateral_age the bootstrap committee signs the superblock
-                // collateral_age is measured in blocks instead of epochs, but this only means that the period in which
-                // the bootstrap committee signs is at least epoch activity_period + collateral_age
+                        actix::fut::ok(superblock)
+                    }
+                    SuperBlockConsensus::Different(target_superblock_hash) => {
+                        // No consensus: move to waiting consensus and restore chain_state from storage
+                        // TODO: start syncronizing with target superblock hash
```
lrubiorod, line 1253: This TODO, is necessary?

node/src/actors/chain_manager/handlers.rs
```diff
@@ -306,118 +302,234 @@ impl Handler<AddBlocks> for ChainManager {
...
+                    msg.blocks,
+                    self.current_epoch.unwrap(),
+                    &sync_target,
+                    superblock_period,
+                );
 
-                            if beacon == target_beacon {
-                                break;
-                            }
+                use BlockBatches::*;
```
lrubiorod, line 345: this import has to be here?

node/src/actors/chain_manager/mod.rs
mariocao: The comment was confusing. Actually, it wasn't TODO but an explanation. :rofl:

node/src/actors/chain_manager/handlers.rs
```diff
@@ -306,118 +302,234 @@ impl Handler<AddBlocks> for ChainManager {
...
+                                    act.sm_state = StateMachine::WaitingConsensus;
+                                    act.sync_waiting_for_add_blocks_since = None;
+
+                                    return actix::fut::err(());
+                                }
+                                log_sync_progress(&sync_target, &blocks3, num_processed_blocks, 3);
+
+                                log::info!("Block sync target achieved, go to WaitingConsensus state");
+                                // Target achieved, go back to state 1
+                                act.sm_state = StateMachine::WaitingConsensus;
```
lrubiorod, line 483: In lines 457-459 we already do the same, it is necessary?

mariocao: I think so :thinking:

node/src/actors/chain_manager/mod.rs
```diff
@@ -1104,6 +1313,65 @@ impl ChainManager {
...
+        (batch_succeeded, num_processed_blocks)
+    }
+
+    fn superblock_consolidation_is_needed(
+        &self,
+        sync_target: &SyncTarget,
+        superblock_period: u32,
+    ) -> Option<u32> {
+        if sync_target.superblock.checkpoint
+            == self.chain_state.superblock_state.get_beacon().checkpoint
```
girazoki, line 1368:
```suggestion
            != self.chain_state.superblock_state.get_beacon().checkpoint
```

mariocao: I changed

mariocao: I changed the `if` and `else` branches :)

Issue #1258, opened 2020-05-21
`getUtxoInfo` CLI method could also provide total amount of collateralizable UTXOs


aesedepece:
Currently, the `getUtxoInfo` CLI method provides useful insights on these counts:
- Total number of utxos
- Total number of utxos bigger than collateral minimum
- Total number of utxos older than collateral coinage

In addition to the total number, it'd be great if you could also get the sum of the amount of those UTXos, so that you can get a clear idea of how many wits can be collateralized right now and how many are waiting for maturity.

PR #1440, opened 2020-08-07
feat(jsonrpc): Add sum of the utxo amounts to getUtxoInfo and format it as a table

drcpu-github:
This pull request solves #1258. I took the liberty to add some extra categories (and rephrase others) for ease of use. I also formatted the output as a table similar to the output of `getUtxoInfo --long`. 

The categories I added are "Utxos smaller than collateral minimum", which I think is useful to indicate if you should do joinTransactions (previously you'd have to calculate that yourself). I also added "Utxos bigger than and not ready for collateral" for completeness.

If some of these changes are unwanted, I can obviously change them.

PS: I don't know why the formatting check requires me to put the utxos_table columns names on one line (comparing line 155 to 160 and line 215), but I did it anyway.

### DIFF
--- a/src/cli/node/json_rpc_client.rs
+++ b/src/cli/node/json_rpc_client.rs
@@ -148,47 +148,101 @@ pub fn get_utxo_info(
     let utxo_info = parse_response::<UtxoInfo>(&response)?;
 
     let utxos_len = utxo_info.utxos.len();
+    let mut utxo_sum = 0;
 
-    if long {
-        let mut table = Table::new();
-        table.set_format(*prettytable::format::consts::FORMAT_NO_BORDER_LINE_SEPARATOR);
-        table.set_titles(row![
-            "OutputPointer",
-            "Value (in wits)",
-            "Time lock",
-            "Ready for collateral"
-        ]);
-        for utxo_metadata in utxo_info
-            .utxos
-            .into_iter()
-            .sorted_by_key(|um| (um.value, um.output_pointer.clone()))
-        {
+    let mut utxo_too_small_counter = 0;
+    let mut utxo_too_small_sum = 0;
+
+    let mut utxo_not_ready_counter = 0;
+    let mut utxo_not_ready_sum = 0;
+
+    let mut utxo_ready_counter = 0;
+    let mut utxo_ready_sum = 0;
+
+    let mut table = Table::new();
+    table.set_format(*prettytable::format::consts::FORMAT_NO_BORDER_LINE_SEPARATOR);
+    table.set_titles(row![
+        "OutputPointer",
+        "Value (in wits)",
+        "Time lock",
+        "Ready for collateral"
+    ]);
+
+    for utxo_metadata in utxo_info
+        .utxos
+        .into_iter()
+        .sorted_by_key(|um| (um.value, um.output_pointer.clone()))
+    {
+        let ready_for_collateral: bool = (utxo_metadata.value >= utxo_info.collateral_min)
+            && utxo_metadata.utxo_mature
+            && utxo_metadata.timelock == 0;
+
+        if long {
             let value = Wit::from_nanowits(utxo_metadata.value).to_string();
             let time_lock = if utxo_metadata.timelock == 0 {
                 "Ready".to_string()
             } else {
-                let date = pretty_print(utxo_metadata.timelock as i64, 0);
-                format!("{:?}", date)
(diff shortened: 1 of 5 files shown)


lrubiorod:
image
Something failing in utxos small than minimum?

src/cli/node/json_rpc_client.rs
```diff
@@ -174,16 +189,52 @@ pub fn get_utxo_info(
...
+        {
+            utxo_sum += utxo_metadata.value;
+            if utxo_metadata.ready_for_collateral {
+                utxo_ready_sum += utxo_metadata.value;
+            } else if utxo_metadata.timelock == 0 {
+                utxo_too_small_sum += utxo_metadata.value;
+            } else {
+                utxo_not_ready_sum += utxo_metadata.value;
+            }
+        }
```
lrubiorod, line 206: If you are going to use the for in the two parts of the if condition clause, I would include the table in the "if long clause" to not repeating the same code lines

drcpu-github: I incorrectly assumed timelock would be set for blocks that were recently minted. In order to solve this I would have to compare utxo_metadata.value with collateral_min, but I don't know how to access the consensus constants in json_rpc_client.rs.

Are you asking to put the for loop outside of the if/else statement?

lrubiorod: Timelock is used to create transactions that can not be spent until the specified timestamp, so collateral age is only for being used as a collateral in a data request, not for a normal transaction

I was thinking that in the past, some UtxoInfo fields as "bigger_than_min_counter" were used in other parts of the code, so that counters were useful to do not have to calculate everytime, buuuuut now it is only used for the CLI methods. So if we are going to iterate the array of UtxoMetadata's with and without the --long flag, those counters are useless, so maybe we can remove it and include more info in UtxoMetadata. Currently, in UtxoMetadata contains a field called "ready_for_collateral" which has 3 conditions (bigger than min, collateral mature and a valid timelock), so we can also remove it, and include other fields.

My proposal: 
- Remove counters from UtxoInfo struct and include the collateral_min, that could be removed in a future when we could get the ConsensusConstants from JSON-RPC
- Remove "ready_for_collateral" from UtxoMetadata struct and include "mature_utxo" field that only check the collateral age (aka "all_utxos.included_in_block_number(o).unwrap() <= block_number_limit" condition in create_utxo_metadata function)
- Calculate the readyness of the utxo in the part of the JSON-RPC client checking the 3 conditions:
    - Value > collateral_min
    - Maturity
    - Timelock == 0

What do you think?

According to the other question... yes :P

drcpu-github: That sounds pretty much like what I wanted to do (the only viable solution I saw at the moment), but I was not sure if you would support adding more data to the UtxoMetadata struct.

As for the collateral_min, I would rather add a boolean stating if it is bigger than the collateral_min, rather than the consensus constant itself (saves some space). But yes, both of these will become obsolete once we have access to getConsensusConstants (and I'll update that when this is the case).

I'll update the pull request ASAP.

PS: I actually tried integrating the getConsensusConstants RPC from @tmpolaczyk but am failing miserably (getting a "Method not found" error). I am missing something when implementing it, which I need to further look into.

drcpu-github:
I removed the bigger_than_min_counter and old_utxos_counter fields. I also removed and simplified all code that was doing bookkeeping for these variables.

I kept the commit separate from the previous one since it changes quite a lot, but for the merge with master, this should obviously be squashed. Let me know if you have any questions / comments / suggestions.

src/cli/node/json_rpc_client.rs
lrubiorod: Dont worry!! As soon as you can. And in a week more or less, @tmpolaczyk will come back from holiday, so let him fight with his PR :P

lrubiorod:
> I removed the bigger_than_min_counter and old_utxos_counter fields. I also removed and simplified all code that was doing bookkeeping for these variables.
> 
> I kept the commit separate from the previous one since it changes quite a lot, but for the merge with master, this should obviously be squashed. Let me know if you have any questions / comments / suggestions.

Great job!! Squash the commits and I will approve

image

lrubiorod:
clippy complains!!

drcpu-github:
> clippy complains!!

Yeah, that's my bad, I wanted to get rid of the quotation marks around the timelock date because it feels out-of-place. Should be fixed now, my local build is succesful.

PR #1442, opened 2020-08-10
Modify mining consensus

lrubiorod:


node/src/actors/chain_manager/mining.rs
```diff
@@ -1007,10 +1006,15 @@ fn calculate_mining_probability(
     let mut equal = 0;
     let mut less = 0;
     for &active_id in rep_engine.ars().active_identities() {
-        match rep_engine.trs().get(&active_id).cmp(&own_rep) {
-            Ordering::Greater => greater += 1,
-            Ordering::Equal => equal += 1,
-            Ordering::Less => less += 1,
+        let rep = rep_engine.trs().get(&active_id);
+        if rep.0 > 0 && own_rep.0 > 0 {
```
girazoki, line 1010:
```suggestion
match (rep.0 > 0, own_rep.0  > 0) {
                (true, false) => greater += 1,
                (false, true) => less += 1,
                _ => equal +=1,
            }
```

girazoki: do not know if this is more efficient but the code reads maybe better to me. WHat do you think?

lrubiorod: I agree

PR #1447, opened 2020-08-11
Reduced committee size

girazoki:
This PR adds functionality for reducing the superblock voting committee size if sufficient epochs elpase without consolidated superblock. The period after which the comittee gets reduced is configurable through consensus constants.

This PR aims at solving the problem of loss of availability of a percetange of the ARS members. 2 approaches were discussed to achieve this:

- Reduce the 2/3 votes required to consolidate a block
- Reduce the committee size, thus increasing the probability of achieving 2/3 of the votes.

The first clearly recovers the chain faster, but also in my view decreases the security of the system by breaking the 2/3 vote assumption. The latter spends more time recovering the chain (ideally, for each committee reduction, several iterations would need to be tried before a new reduction happens). However, it also maintains the 2/3 vote approach. In the catastrophic event of a high loss of availability, I believe it is preferable to halt the network for a higher amount of time but maintaining the security guarantees, specially for the bridge. One committee where an attacker achieves the necessary votes to consolidate a superblock is sufficient to break the Ethereum bridge forever, and as such, I believe it is preferable to opt for the second option.

### DIFF
--- a/config/src/defaults.rs
+++ b/config/src/defaults.rs
@@ -111,9 +111,14 @@ pub trait Defaults {
         10
     }
 
-    /// Default period between superblocks
-    fn consensus_constants_superblock_agreement_decreasing_period(&self) -> u32 {
-        100
+    /// Default period (in superblock periods) after which the committee should be reduced
+    fn consensus_constants_superblock_committee_decreasing_period(&self) -> u32 {
+        5
+    }
+
+    /// Default committee reduction step
+    fn consensus_constants_superblock_committee_decreasing_step(&self) -> u32 {
+        5
     }
 
     /// Default Hash value for the auxiliary bootstrap block
--- a/config/src/defaults.rs
+++ b/config/src/defaults.rs
@@ -111,6 +111,11 @@ pub trait Defaults {
         10
     }
 
+    /// Default period between superblocks
+    fn consensus_constants_superblock_agreement_decreasing_period(&self) -> u32 {
+        100
+    }
+
     /// Default Hash value for the auxiliary bootstrap block
     // TODO Decide an appropriate default value
     fn consensus_constants_bootstrap_hash(&self) -> Hash {
--- a/config/src/defaults.rs
+++ b/config/src/defaults.rs
@@ -274,7 +274,7 @@ pub trait Defaults {
     }
 
     /// First superblocks signing committee
-    fn consensus_constants_bootsrapping_committee(&self) -> Vec<String> {
+    fn consensus_constants_bootstrapping_committee(&self) -> Vec<String> {
         // FIXME(#1114): Choose a proper value for the committee
         // [] pkhs
 
--- a/data_structures/src/superblock.rs
+++ b/data_structures/src/superblock.rs
@@ -371,6 +371,7 @@ impl SuperBlockState {
         self.signing_committee = calculate_superblock_signing_committee(
             self.ars_previous_identities.clone(),
             signing_committee_size,
+            superblock_index,
             self.current_superblock_beacon.hash_prev_block,
         );
 
@@ -413,6 +414,7 @@ impl SuperBlockState {
 pub fn calculate_superblock_signing_committee(
     ars_identities: ARSIdentities,
     signing_committee_size: u32,
+    current_superblock_index: u32,
     superblock_hash: Hash,
 ) -> HashSet<PublicKeyHash> {
     // If the number of identities is lower than committee_size all the members of the ARS sign the superblock
@@ -420,7 +422,7 @@ pub fn calculate_superblock_signing_committee(
         ars_identities.identities
     } else {
         // Start counting the members of the subset from the superblock_hash
-        let mut first = u32::from(*superblock_hash.as_ref().get(0).unwrap());
+        let mut first = u32::from(*superblock_hash.as_ref().get(0).unwrap()) + current_superblock_index;
         first %= signing_committee_size;
         // Get the subset
         let subset = magic_partition(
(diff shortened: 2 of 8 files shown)


data_structures/src/superblock.rs
```diff
@@ -1432,7 +1435,7 @@ mod tests {
         // Receive three superblock votes for index 1
         // Since the signing_committee_size is 2, one of the votes will not be valid
         let mut v1 = SuperBlockVote::new_unsigned(sb2_hash, 1);
-        v1.secp256k1_signature.public_key = p1;
+        v1.secp256k1_signature.public_key = p3;
```
lrubiorod, line 1438: why you change the pk?

node/src/actors/chain_manager/mod.rs
```diff
@@ -76,6 +76,7 @@ use crate::{
     },
     signature_mngr, storage_mngr,
 };
+use std::cmp::max;
```
lrubiorod, line 79: fmt

node/src/actors/chain_manager/mod.rs
```diff
@@ -1195,10 +1196,22 @@ impl ChainManager {
...
+                        // Committee_size = expected_size - ((current_index-last_consolidated_index)/decrease_period))
+
+                        let committee_size = calculate_committee_size(
+                            consensus_constants
+                                .superblock_signing_committee_size,
+                            consensus_constants.superblock_agreement_decreasing_period,
+                            chain_info.highest_superblock_checkpoint.checkpoint,
+                                superblock_index
+                        );
+                        log::error!("{} is the size", committee_size);
```
lrubiorod, line 1209: You can modify this log to be a properly DEBUG log message

girazoki: This was not supposed to be part of the PR hahaha

data_structures/src/superblock.rs
girazoki: because with the index being taken into account, the public key that is valid is p3 and not p1.

node/src/actors/chain_manager/mod.rs
girazoki: but I will

lrubiorod:
I think the reduction factor should be specified as a consensus constant

config/src/defaults.rs
```diff
@@ -111,6 +111,11 @@ pub trait Defaults {
         10
     }
 
+    /// Default period between superblocks
+    fn consensus_constants_superblock_agreement_decreasing_period(&self) -> u32 {
+        100
```
lrubiorod, line 116: 100 superepochs are 12 hours, maybe it is too much

girazoki: agree

node/src/actors/chain_manager/mod.rs
```diff
@@ -1195,10 +1194,22 @@ impl ChainManager {
                         // Store the ARS and the order of the keys
                         let ars_identities = ARSIdentities::new(ars_members);
 
+                        // Committee size should decrease if sufficient epochs have elapsed since last confirmed superblock
+                        // Committee_size = expected_size - ((current_index-last_consolidated_index)/decrease_period))
+
```
clbartoli, line 1199:
```suggestion
                        // Committee size should decrease if sufficient epochs have elapsed since last confirmed superblock
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -1834,6 +1845,29 @@ fn show_sync_progress(
     );
 }
 
+// TODO: handle recovery cases after reduction
+fn calculate_committee_size(
```
clbartoli, line 1850: Maybe we could change the name to `update_committee_size`, since the committee size is a consensus constant and you don`t need to calculate it, does it make sense?

node/src/actors/chain_manager/mod.rs
```diff
@@ -1834,6 +1845,29 @@ fn show_sync_progress(
...
+    default_committee_size: u32,
+    decreasing_period: u32,
+    last_consolidated_checkpoint: u32,
+    current_checkpoint: u32,
+) -> u32 {
+    // If the last consolidated superblock is 0, we return the default committee size
```
clbartoli, line 1857:
```suggestion
    // If the last consolidated superblock is 0, return the default committee size
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -1834,6 +1845,29 @@ fn show_sync_progress(
...
+    if last_consolidated_checkpoint == 0 {
+        default_committee_size
+    } else {
+        // We calculate the difference between the last consolidated superblock checkpoint and the current one
+        // If this difference exceeds the decreasing_period, we reduce the committee size
+        // The minimum committee size is 1
```
clbartoli, line 1863:
```suggestion
        // Calculate the difference between the last consolidated superblock checkpoint and the current one
        // If this difference exceeds the decreasing_period, reduce the committee size
        // The minimum committee size is 1
```

node/src/actors/chain_manager/mod.rs
girazoki: Hmm I agree that calculate might not be the best name for it. But the truth is that based on your last consolidated epoch and the default committee size you will always have to calculate it... Update sounds a bit like if we were gonna update the consensus constant for me. Any more ideas for the naming?

config/src/defaults.rs
```diff
@@ -111,6 +111,16 @@ pub trait Defaults {
         10
     }
 
+    /// Default period after which the committee should be reduced
```
mariocao, line 114:
```suggestion
    /// Default period (in superblock periods) after which the committee should be reduced
```

data_structures/src/superblock.rs
```diff
@@ -9,6 +9,7 @@ use std::{
 
 use serde::{Deserialize, Serialize};
 
+use witnet_crypto::hash::calculate_sha256;
```
lrubiorod, line 12: fmt

PR #1448, opened 2020-08-12
feat(cli): add claim command to node CLI

### LINKED ISSUE #1445: Message signing through CLI
The node should support signing messages from the CLI.

mariocao:
This PR adds a new command to the node CLI called `claim` a node identity.

This command will be used in order to prove the ownership of a node identity by signing a given data (e.g. an identifier) with the node master key. The output will also contain the signed data, the public key and the signature itself.

```bash
USAGE:
    witnet node claim [FLAGS] [OPTIONS] --identifier <identifier>

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information
        --write      Write the signed claimed data to "storage_path/claim-<id>-<public_key>.txt"

OPTIONS:
    -i, --identifier <identifier>    Identifier to be claimed by the node (e.g. Witnet ID)
    -n, --node <node>                Socket address of the Witnet node to query
        --write-to <write-to>        Change the path to write storage_path/claim-<id>-<public_key>.txt". Implies --write
```

The command also supports writing the output into a file by using the option `--write` or by directly defining the output file with the `--write-to` option.

The output is formatted using JSON as follows:

```json
{
  "address": "twit1...",
  "identifier": "WITNET_XYZ",
  "public_key": "<hex_code>",
  "signature": "<hex_code>"
}
```

Closes #1445

### DIFF
--- a/src/cli/node/json_rpc_client.rs
+++ b/src/cli/node/json_rpc_client.rs
@@ -1239,4 +1239,45 @@ mod tests {
             )
         );
     }
+
+    #[test]
+    fn verify_claim_output() {
+        use witnet_crypto::{
+            secp256k1::Secp256k1,
+            signature::{verify, PublicKey as SecpPublicKey, Signature as SecpSignature},
+        };
+
+        let json_output = r#"
+        {
+          "address": "twit17k4tzsf9zs70q8ndur7qvavvhvrkfd8jkjrppw",
+          "identifier": "WITNET_000",
+          "public_key": "038f48d48aaa177c54809598649a037fb75a391449c8d0fee3f7d3b7f8fcd48239",
+          "signature": "a1a37548b1367dd683b87abf534aafa5c9c3c9c15fd4186d437180a61e7bd31e585cf36ff2fddbc6ad5bbdddb65c2195895f855b60a7b81f44a100288a821561"
+        }"#;
+
+        // Parse the string of data into serde_json::Value.
+        let signature_with_data: SignatureWithData = serde_json::from_str(json_output).unwrap();
+
+        // Check address is correctly derived from public key
+        let address = PublicKeyHash::from_public_key(
+            &PublicKey::try_from_slice(
+                &hex::decode(signature_with_data.public_key.clone()).unwrap(),
+            )
+            .unwrap(),
+        )
+        .bech32(Environment::Testnet);
+        assert_eq!(address, signature_with_data.address);
+
+        // Required fields for Secpk1 signature verification
+        let secp = Secp256k1::new();
+        let signed_data = calculate_sha256(signature_with_data.identifier.as_bytes().as_ref());
+        let public_key =
+            SecpPublicKey::from_slice(&hex::decode(signature_with_data.public_key).unwrap())
+                .unwrap();
+        let signature =
+            SecpSignature::from_compact(&hex::decode(signature_with_data.signature).unwrap())
+                .unwrap();
+
+        assert!(verify(&secp, &public_key, signed_data.as_ref(), &signature).is_ok());
+    }
 }
--- a/src/cli/node/json_rpc_client.rs
+++ b/src/cli/node/json_rpc_client.rs
@@ -14,11 +14,15 @@ use std::{
     str::FromStr,
 };
 
-use witnet_crypto::key::{CryptoEngine, ExtendedPK, ExtendedSK};
+use witnet_crypto::{
+    hash::calculate_sha256,
+    key::{CryptoEngine, ExtendedPK, ExtendedSK},
+};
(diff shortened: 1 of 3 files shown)


src/cli/node/json_rpc_client.rs
```diff
@@ -14,11 +14,16 @@ use std::{
     str::FromStr,
 };
 
-use witnet_crypto::key::{CryptoEngine, ExtendedPK, ExtendedSK};
+use std::fs::File;
```
lrubiorod, line 17: fmt

src/cli/node/json_rpc_client.rs
```diff
@@ -1028,6 +1033,64 @@ pub fn add_peers(addr: SocketAddr, peers: Vec<SocketAddr>) -> Result<(), failure
...
 }
 
+#[derive(Serialize, Deserialize)]
+struct SignatureWithData {
+    identifier: String,
+    public_key: String,
+    signature: String,
+}
+
+pub fn claim(
```
lrubiorod, line 1043: good boardgame :P

src/cli/node/json_rpc_client.rs
```diff
@@ -1028,6 +1033,64 @@ pub fn add_peers(addr: SocketAddr, peers: Vec<SocketAddr>) -> Result<(), failure
...
+        calculate_sha256(identifier.as_bytes()).as_ref(),
+    );
+
+    let mut stream = start_client(addr)?;
+    let response = send_request(&mut stream, &request)?;
+
+    let signature: KeyedSignature = parse_response(&response)?;
+    match serde_json::to_string_pretty(&SignatureWithData {
+        identifier: identifier.clone(),
+        public_key: PublicKeyHash::from_public_key(&signature.public_key).to_string(),
```
lrubiorod, line 1063: it is a public key or a public key hash?

lrubiorod: or and address?? xD

src/cli/node/with_node.rs
```diff
@@ -235,6 +259,20 @@ pub enum Command {
         #[structopt(long = "limit", allow_hyphen_values = true, default_value = "-50")]
         limit: i64,
     },
+    Claim {
```
clbartoli, line 262:
```suggestion
#[structopt(
        name = "claim",
        about = "Claim identity"
    )]
    Claim {
```

clbartoli: I think that adding this lines gives a description to the command. At the moment when writing --help it appears an empty line 
image

src/cli/node/json_rpc_client.rs
mariocao: This is the Display implementation for PublicKeyHash, i.e. a bech32 address

mariocao:
I will add a test for verifying that the output can be decoded  and the signature verified. :)

src/cli/node/with_node.rs
mariocao: Good one! I completely forgot it!

aesedepece:
Potential instructions for node operators:

1. While your node is running, execute the following command: `docker exec witnet_node claim -i YOUR_WIT_ID`
2. Copy the output of that command, and send it via email to `<>`

I'm not going the `--write` way because I'm assuming that most node operators who are using VPSs will find it quite cumbersome to copy the output flies to their PC using `scp` or anything similar. What do you think?

src/cli/node/with_node.rs
mariocao: I wrote this: "Claim a Witnet identity by signing the identifier with the node master key"

mariocao:
> I will add a test for verifying that the output can be decoded and the signature verified. :)

The test it is not really testing the function `claim()` because it relies on the JSON RPC response.
However, I thought it could be useful to leave there a check of an example output, so that we can be sure that verification can be done correctly (i.e. signature and address verification).

mariocao:
> Potential instructions for node operators:
> 
> 1. While your node is running, execute the following command: `docker exec witnet_node claim -i YOUR_WIT_ID`
> 2. Copy the output of that command, and send it via email to `<>`
> 
> I'm not going the `--write` way because I'm assuming that most node operators who are using VPSs will find it quite cumbersome to copy the output flies to their PC using `scp` or anything similar. What do you think?

Both options are fine:
- If node operators have a virtual volume, the `--write` option might be useful.
- If not, it is true that it is cumbersome to copy the files.

The main drawback of copying and pasting might be some parsing errors we might encounter. However, nothing that we cannot overcome. :smile:

mariocao:
Btw, I included the public key bytes because it's needed for the signature verification.
The address is included just for the user to double-check it! :)

PR #1449, opened 2020-08-13
Increment step by step reduced committee

girazoki:
This PR adds recovery steps when a reduced committee has been reached


node/src/actors/chain_manager/mod.rs
```diff
@@ -76,6 +76,7 @@ use crate::{
     },
     signature_mngr, storage_mngr,
 };
+use std::cmp::min;
```
lrubiorod, line 79: fmt

PR #1461, opened 2020-08-17
[wip] make tally creation mandatory and remove tally fee

lrubiorod:


data_structures/src/chain.rs
```diff
@@ -1071,9 +1073,8 @@ pub struct DataRequestOutput {
     pub data_request: RADRequest,
     pub witness_reward: u64,
     pub witnesses: u16,
-    pub commit_fee: u64,
-    pub reveal_fee: u64,
-    pub tally_fee: u64,
+    // This fee will be earn for the miner when include commits and/or reveals in the block
```
clbartoli, line 1076:
```suggestion
    // This fee will be earn by the miner when include commits and/or reveals in the block
```

data_structures/src/chain.rs
```diff
@@ -1087,30 +1088,25 @@ impl DataRequestOutput {
     /// Calculate the total value of a data request, return error on overflow
     ///
     /// ```ignore
-    /// total_value = (witness_reward + commit_fee + reveal_fee) * witnesses + tally_fee
+    /// total_value = (witness_reward + commit_and_reveal_fee + commit_and_reveal_fee) * witnesses
```
clbartoli, line 1091: why is it counted twice?

clbartoli: Ahh got it, I thought for a moment that commit_and_reveal_fee was the sum of both fees

lrubiorod: its the same for both, so now we have only one field to represents the 2 fees

data_structures/src/error.rs
```diff
@@ -381,6 +381,12 @@ pub enum BlockError {
         weight, max_weight
     )]
     TotalDataRequestWeightLimitExceeded { weight: u32, max_weight: u32 },
+    /// Tallies required
+    #[fail(
+        display = "There are {} tallies required not found in block {}",
```
aesedepece, line 386:
```suggestion
        display = "{} expected tally transactions are missing in block candidate {}",
```

data_structures/src/error.rs
```diff
@@ -381,6 +381,12 @@ pub enum BlockError {
...
+        count, block_hash
+    )]
+    TalliesRequired { count: usize, block_hash: Hash },
```
aesedepece, line 389:
```suggestion
    MissingExpectedTallies { count: usize, block_hash: Hash },
```

validations/src/validations.rs
```diff
@@ -1643,6 +1648,16 @@ pub fn validate_block_transactions(
     }
     let ta_hash_merkle_root = ta_mt.root();
 
+    // All tallies expected should have been removed when created
+    // If not, block is invalide due to have expected tallies not found
```
aesedepece, line 1652:
```suggestion
    // All data requests for which we expected tally transactions should have been removed upon creation of the tallies
    // If not, block is invalid due to missing expected tallies
```

aesedepece (changes requested):

PR #1462, opened 2020-08-19
Request with zero retrieval source set as invalid

### LINKED ISSUE #1288: Data requests with zero retrieval sources should be invalid
Currently, data requests with an empty `"retrievals": []` field are resolved with a `NoReveals` error. But they should be rejected during validation and not included in any blocks.

clbartoli:
This PR adds two checks in the function `validate_rad_request`:
 - If the data request has an empty retrieve vector it returns the error `UnsupportedReducerInAT`.
 - If the data request has an empty url source is returns the error `NoRetrievalSource`.

The tests have been changed since the data request used as example was not longer valid.

Closes #1288

### DIFF
--- a/data_structures/src/error.rs
+++ b/data_structures/src/error.rs
@@ -467,6 +467,8 @@ pub enum DataRequestError {
     NotTallyStage,
     #[fail(display = "Cannot persist unfinished data request (with no Tally)")]
     UnfinishedDataRequest,
+    #[fail(display = "The data request is not valid since it has no retrieval sources")]
+    NoRetrievalSource,
 }
 
 /// Possible errors when converting between epoch and timestamp
--- a/data_structures/src/error.rs
+++ b/data_structures/src/error.rs
@@ -468,7 +468,7 @@ pub enum DataRequestError {
     #[fail(display = "Cannot persist unfinished data request (with no Tally)")]
     UnfinishedDataRequest,
     #[fail(display = "The data request is not valid since it has no retrieval sources")]
-    NoRetrievalSource,
+    NoRetrievalSources,
 }
 
 /// Possible errors when converting between epoch and timestamp
--- a/validations/src/validations.rs
+++ b/validations/src/validations.rs
@@ -5,6 +5,8 @@ use std::{
     fmt,
 };
 
+//std::cmp::PartialEq<std::vec::Vec<_>>
+
 use itertools::Itertools;
 
 use witnet_crypto::{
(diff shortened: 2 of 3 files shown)


validations/src/validations.rs
```diff
@@ -5,6 +5,8 @@ use std::{
     fmt,
 };
 
+//std::cmp::PartialEq<std::vec::Vec<_>>
```
aesedepece, line 8: Redundant line

validations/src/validations.rs
```diff
@@ -293,7 +295,15 @@ pub fn validate_mint_transaction(
 /// Function to validate a rad request
 pub fn validate_rad_request(rad_request: &RADRequest) -> Result<(), failure::Error> {
     let retrieval_paths = &rad_request.retrieve;
+    // If the data requests has no retrive it is set as invalid
```
aesedepece, line 298:
```suggestion
    // If the data requests has no sources to retrieve, it is set as invalid
```

validations/src/validations.rs
```diff
@@ -293,7 +295,15 @@ pub fn validate_mint_transaction(
...
+    if retrieval_paths.is_empty() {
+        return Err(RadError::UnsupportedReducerInAT { operator: 0 }.into());
```
aesedepece, line 300: Isn't this supposed to be `NoRetrievalSource`?

aesedepece (changes requested):

validations/src/validations.rs
clbartoli: This error already existed for the case of empty retrieve vector, however the `NoRetrivalSource` is the error for the specific case in which there is no url parameter in the retrieve vector

lrubiorod: This error happens in that case, but it should be only for unsupported reducers, so as @aesedepece says, we could return the same error (aka NoRetrievalSources)

lrubiorod (approved):

data_structures/src/error.rs
```diff
@@ -467,6 +467,8 @@ pub enum DataRequestError {
     NotTallyStage,
     #[fail(display = "Cannot persist unfinished data request (with no Tally)")]
     UnfinishedDataRequest,
+    #[fail(display = "The data request is not valid since it has no retrieval sources")]
+    NoRetrievalSource,
```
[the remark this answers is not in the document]
aesedepece: Ping

validations/src/validations.rs
```diff
@@ -293,7 +293,15 @@ pub fn validate_mint_transaction(
 /// Function to validate a rad request
 pub fn validate_rad_request(rad_request: &RADRequest) -> Result<(), failure::Error> {
     let retrieval_paths = &rad_request.retrieve;
+    // If the data requests has no sources to retrive, it is set as invalid
```
aesedepece, line 296:
```suggestion
    // If the data requests has no sources to retrieve, it is set as invalid
```

aesedepece (changes requested):

aesedepece (approved):

PR #1466, opened 2020-08-24
fix(superblock): use active nodes in committee when there are not act…

lrubiorod:
…ive reputed nodes

### DIFF
--- a/node/src/actors/chain_manager/mod.rs
+++ b/node/src/actors/chain_manager/mod.rs
@@ -1234,7 +1234,14 @@ impl ChainManager {
                                     + chain_info.consensus_constants.activity_period
                             {
                                 let ars_members = reputation_engine.get_rep_ordered_ars_list();
-                                 reputed_ars(ars_members, &reputation_engine)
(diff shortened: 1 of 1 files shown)


mariocao (approved):

tmpolaczyk (approved):

node/src/actors/chain_manager/mod.rs
```diff
@@ -1181,7 +1181,14 @@ impl ChainManager {
                                     + chain_info.consensus_constants.activity_period
                             {
                                 let ars_members = reputation_engine.get_rep_ordered_ars_list();
-                                 reputed_ars(ars_members, &reputation_engine)
+                                let reputed = reputed_ars(&ars_members, &reputation_engine);
+
+                                // In case of non reputed nodes, returns all active nodes
```
aesedepece, line 1186:
```suggestion
                                // In case of no reputed nodes, return all active nodes
```

aesedepece (changes requested):

aesedepece (approved):

PR #1469, opened 2020-08-24
refactor(superblock): move superblock creation out of mining

### LINKED ISSUE #1450: SuperBlockSync: separate superblock synchronizing from block mining

lrubiorod:
Close #1450

### DIFF
--- a/node/src/actors/chain_manager/mod.rs
+++ b/node/src/actors/chain_manager/mod.rs
@@ -217,8 +217,6 @@ pub struct ChainManager {
     external_address: Option<PublicKeyHash>,
     /// Mint Percentage to share with the external address
     external_percentage: u8,
-    /// Enable superblock creation
-    create_superblocks: bool,
     /// List of superblock votes received while we are synchronizing
     temp_superblock_votes: Vec<SuperBlockVote>,
 }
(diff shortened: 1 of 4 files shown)


aesedepece (approved):

node/src/actors/chain_manager/mod.rs
```diff
@@ -671,6 +669,64 @@ impl ChainManager {
         }
     }
 
+    /// Create a superblock and broadcast it
+    fn superblock_creating_and_broadcasting(
```
mariocao, line 673:
```suggestion
    fn create_and_broadcast_superblock(
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -671,6 +669,64 @@ impl ChainManager {
         }
     }
 
+    /// Create a superblock and broadcast it
```
mariocao, line 672:
```suggestion
    /// Create a superblock, sign a superblock vote and broadcast it
```

Maybe we could add a FIXME to avoid signing superblocks in case the identity is not in the signing committee

mariocao (changes requested):

tmpolaczyk (approved):

mariocao (approved):

PR #1470, opened 2020-08-24
refactor(node): avoid mutation upon persisting previous chain state

aesedepece:
Co-authored-by: Mario Cao <mariocao@gmail.com>

### DIFF
--- a/node/src/actors/chain_manager/actor.rs
+++ b/node/src/actors/chain_manager/actor.rs
@@ -243,8 +243,8 @@ impl ChainManager {
                 });
 
                 // Persist current chain state
-                act.last_chain_state = act.chain_state.clone();
-                act.persist_chain_state(ctx);
+                act.previous_chain_state = act.chain_state.clone();
+                act.persist_previous_chain_state(ctx);
             }).wait(ctx);
     }
 
--- a/node/src/actors/chain_manager/actor.rs
+++ b/node/src/actors/chain_manager/actor.rs
@@ -242,8 +242,8 @@ impl ChainManager {
                     },
                 });
 
-                // Persist current chain state
-                act.previous_chain_state = act.chain_state.clone();
+                // Copy current chain state into previous chain state, and persist it
+                act.move_chain_state_forward();
                 act.persist_previous_chain_state(ctx);
             }).wait(ctx);
     }
--- a/node/src/actors/chain_manager/mod.rs
+++ b/node/src/actors/chain_manager/mod.rs
@@ -284,6 +284,11 @@ impl ChainManager {
             .wait(ctx);
     }
 
+    /// Replace `previous_chain_state` with current `chain_state`
+    fn move_chain_state_forward(&mut self) {
+        self.previous_chain_state = self.chain_state.clone();
+    }
+
     /// Method to Send items to Inventory Manager
     fn persist_items(&self, ctx: &mut Context<Self>, items: Vec<StoreInventoryItem>) {
         // Get InventoryManager address
@@ -1090,9 +1095,9 @@ impl ChainManager {
                             "Consensus while sync! Superblock {:?}",
                             act.get_superblock_beacon()
                         );
-                        // Persist current_chain_state with
-                        // current_superblock_state and the new superblock beacon
-                        act.previous_chain_state = act.chain_state.clone();
+
+                        // Copy current chain state into previous chain state, and persist it
+                        act.move_chain_state_forward();
                         act.persist_previous_chain_state(ctx);
 
                         actix::fut::ok(())
@@ -1222,6 +1227,7 @@ impl ChainManager {
                         if act.sm_state == StateMachine::Synced {
                             // Persist previous_chain_state with current superblock_state
                             act.persist_previous_chain_state(ctx);
+                            act.move_chain_state_forward();
                         }
 
                         log::info!("Consensus reached for Superblock #{:?}", superblock_beacon.checkpoint);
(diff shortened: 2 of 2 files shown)


node/src/actors/chain_manager/mod.rs
```diff
@@ -245,42 +245,47 @@ impl SystemService for ChainManager {}
...
 /// Auxiliary methods for ChainManager actor
 impl ChainManager {
-    /// Method to persist backup of chain_state into storage
-    fn persist_last_chain_state(&self, ctx: &mut Context<Self>) {
-        match self.last_chain_state.chain_info.as_ref() {
-            Some(x) => x,
+    /// Persist previous chain state into storage
+    fn persist_previous_chain_state(&mut self, ctx: &mut Context<Self>) {
+        // It makes no sense to persist empty chain states
+        match self.previous_chain_state.chain_info {
```
lrubiorod, line 251:
```suggestion
        if self.previous_chain_state.chain_info.is_none() {
```

aesedepece: Nice catch! Changed!

node/src/actors/chain_manager/actor.rs
```diff
@@ -246,8 +246,8 @@ impl ChainManager {
                 });
 
                 // Persist current chain state
-                act.last_chain_state = act.chain_state.clone();
-                act.persist_chain_state(ctx);
+                act.previous_chain_state = act.chain_state.clone();
+                act.persist_previous_chain_state(ctx);
```
tmpolaczyk, line 250: This is repeated twice, let's move the clone inside a method:

```suggestion
                act.persist_current_chain_state(ctx);
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -1152,30 +1154,36 @@ impl ChainManager {
...
+                        log::debug!("Before update: superblock {:?}", superblock_beacon);
+
+                        // Take last beacon from superblock state and use it in current chain_info
                         act.chain_state.chain_info.as_mut().unwrap().highest_superblock_checkpoint =
                             act.chain_state.superblock_state.get_beacon();
+
                         if act.sm_state == StateMachine::Synced {
                             // Persist previous_chain_state with current superblock_state
-                            act.persist_chain_state(ctx);
+                            act.persist_previous_chain_state(ctx);
```
tmpolaczyk, line 1170: Here we call `persist_previous_chain_state` but we don't update `self.previous_state`. This means that consecutive calls to `persist_previous_chain_state` will all persist the same state (the initial one). We must update `self.previous_chain_state` after persisting:

```suggestion
                            act.persist_previous_chain_state(ctx);
                            act.previous_chain_state = act.chain_state.clone();
```

Not sure if we should also add this line to the other places, where `self.previous_chain_state` is updated before.

node/src/actors/chain_manager/actor.rs
aesedepece: Done, it's `self.advance_chain_state` now

node/src/actors/chain_manager/mod.rs
aesedepece: Intredasting. @mariocao what do you think?

tmpolaczyk:
We could simplify the logic by giving names to the old copies of the chain state. For example when we know that there is consensus for superblock n we want to persist the chain state at superblock n to the storage. And when we are synchronizing up to superblock m, we want to persist the chain state at superblock m. The only problem is that the current chain state is n+2. 

So if instead of `previous_chain_state: ChainState` we use `previous_chain_states: HashMap<SuperEpoch, ChainState>`, then the logic becomes straight forward because we can just call `self.persist_chain_state(consensus_superepoch)`.

node/src/actors/chain_manager/mod.rs
mariocao: seems correct...

node/src/actors/chain_manager/mod.rs
```diff
@@ -1034,10 +1041,10 @@ impl ChainManager {
...
                             act.get_superblock_beacon()
                         );
-                        // Persist current_chain_state with
-                        // current_superblock_state and the new superblock beacon
-                        act.last_chain_state = act.chain_state.clone();
-                        act.persist_chain_state(ctx);
+
+                        // Copy current chain state into previous chain state, and persist it
+                        act.advance_chain_state();
+                        act.persist_previous_chain_state(ctx);
```
lrubiorod, line 1047: Here you are moving current_chain_state to previous and persist previous, so... you are persisting the current chain_state. I think it should be first persist the previous one and update the new one as previous
```suggestion
                        act.persist_previous_chain_state(ctx);
                        act.advance_chain_state();
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -1152,30 +1159,36 @@ impl ChainManager {
...
+                        log::debug!("Before update: superblock {:?}", superblock_beacon);
+
+                        // Take last beacon from superblock state and use it in current chain_info
                         act.chain_state.chain_info.as_mut().unwrap().highest_superblock_checkpoint =
                             act.chain_state.superblock_state.get_beacon();
+
                         if act.sm_state == StateMachine::Synced {
                             // Persist previous_chain_state with current superblock_state
-                            act.persist_chain_state(ctx);
+                            act.persist_previous_chain_state(ctx);
```
lrubiorod, line 1175:
```suggestion
                            act.persist_previous_chain_state(ctx);
                            act.advance_chain_state();
```
I agree

aesedepece: Done!

node/src/actors/chain_manager/mod.rs
aesedepece: Done!

tmpolaczyk (approved):
I found an edge case where the storage is not persisted, but will open a separate PR for that

lrubiorod (approved):

Issue #1456, opened 2020-08-17
Make RePoE chances non-linear, but trapezoidal instead


aesedepece:
This mechanism will ensure a fairer distribution of rewards among nodes with non-neutral reputation.

PR #1472, opened 2020-08-25
feat(cli): include eligibility in getReputation CLI method

lrubiorod:


aesedepece:
image

tmpolaczyk:
This is a breaking change for the JSON-API `getReputation` method, right?

lrubiorod:
> This is a breaking change for the JSON-API `getReputation` method, right?

Yep

node/src/actors/chain_manager/handlers.rs
```diff
@@ -1301,12 +1301,23 @@ impl Handler<GetReputation> for ChainManager {
...
+            is_active,
+        };
+
+        Ok((rep_stats, total_reputation))
     }
 }
 
 impl Handler<GetReputationAll> for ChainManager {
-    type Result = Result<HashMap<PublicKeyHash, (Reputation, bool)>, failure::Error>;
+    type Result = Result<(u64, HashMap<PublicKeyHash, ReputationStats>), failure::Error>;
```
tmpolaczyk, line 1320: `GetReputation` returns `(ReputationStats, u64)`, but `GetReputationAll` returns `(u64, HashMap)`. This is a bit confusing, so instead how about explicitly stating the result, like

```rust
struct GetReputationResult {
    stats: HashMap<PublicKeyHash, ReputationStats>,
    total_reputation: u64,
}
```

And we could unify the return types of `GetReputation` and `GetReputationAll`, `GetReputation` would return a `stats` with one element.

drcpu-github:
Is this the number of data requests you were eligible to solve and the number you did actually solve?

lrubiorod:
> Is this the number of data requests you were eligible to solve and the number you did actually solve?

This PR is related with this https://github.com/witnet/witnet-rust/issues/1456.
To avoid reputation oligarchies, we order nodes by reputation and we apply another distribution (trapezoidal distribution). So, if a node has 20000 of reputation and another only 1, it doesnt mean that the first one has a probability 20000 bigger than the other.

Currently in the testnet, your probability to mine a data request is: reputation/total_active_reputation * factor (this factor is related with the number of witnesses required). And with #1456, it will be: eligibility/total_active_reputation * factor, where eligibility depends on your reputation but it is not exactly the same.

And... this PR modifies the json methods for reputation so... you will need to update it in the blockexplorer

src/cli/node/json_rpc_client.rs
```diff
@@ -247,56 +247,53 @@ pub fn get_utxo_info(
...
+                log::info!("Node pkh: {}", node_pkh);
+
+                node_pkh
+            }
         }
     };
 
+    let params = (pkh, all);
     let request = format!(
         r#"{{"jsonrpc": "2.0","method": "getReputation", "params": [{}], "id": "1"}}"#,
```
tmpolaczyk, line 277: Here the request will be formatted as one of the two options, depending on the `--all` flag:

```json
{"jsonrpc": "2.0","method": "getReputation", "params": [["twit1fulan0j78dmdfa3gx0779kszj5sc5j35feljv5",false]], "id": "1"}
{"jsonrpc": "2.0","method": "getReputation", "params": [["twit1qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqkz4ay",true]], "id": "1"}
```

As you can see this is pretty strange, I think I prefer the old format, even if it means adding back the `getReputationAll` JSON-RPC method:

```json
{"jsonrpc": "2.0","method": "getReputation", "params": ["twit1fulan0j78dmdfa3gx0779kszj5sc5j35feljv5"], "id": "1"}
{"jsonrpc": "2.0","method": "getReputationAll", "id": "1"}
```

drcpu-github:
Oh, right, I did not link it to the new trapezoidal distribution for solving data requests.

Should not be too big of a change for the explorer, probably just add a bit of extra parsing logic.

node/src/actors/messages.rs
```diff
@@ -253,39 +253,37 @@ impl Message for GetUtxoInfo {
...
+/// GetReputation result
+#[derive(Clone, Debug, Default, Eq, PartialEq, Serialize, Deserialize)]
+pub struct GetReputationResult {
+    /// ReputationsStats hashmap
+    pub stats: HashMap<PublicKeyHash, ReputationStats>,
+    /// Total active reputation
+    pub total_reputation: u64,
+}
+
 /// Get Reputation of one pkh
```
tmpolaczyk, line 276:
```suggestion
/// Get reputation of one identity if `all` is set to `false`, or all identities if `all` is set to `true`
```

node/src/actors/messages.rs
```diff
@@ -253,39 +253,37 @@ impl Message for GetUtxoInfo {
...
+    /// Eligibility
+    pub eligibility: u32,
+    /// Is active flag
+    pub is_active: bool,
+}
+
+/// GetReputation result
+#[derive(Clone, Debug, Default, Eq, PartialEq, Serialize, Deserialize)]
+pub struct GetReputationResult {
+    /// ReputationsStats hashmap
```
tmpolaczyk, line 270:
```suggestion
    /// Map of identity public key hash to reputation stats
```

node/src/actors/messages.rs
```diff
@@ -253,39 +253,37 @@ impl Message for GetUtxoInfo {
     type Result = Result<UtxoInfo, failure::Error>;
 }
 
+/// Reputation info
+#[derive(Clone, Debug, Default, Eq, PartialEq, Serialize, Deserialize)]
+pub struct ReputationStats {
+    /// Reputation
+    pub reputation: Reputation,
+    /// Eligibility
```
tmpolaczyk, line 261: I would be good to add a short explanation on what's the difference between eligibility and reputation, or link to the function that calculates eligibility given reputation.

tmpolaczyk (approved):
I added some documentation suggestions

PR #1471, opened 2020-08-24
refactor(node): centralize state machine changes in ChainManager

aesedepece:
Centralizes all state machine changes in ChainManager so debugging is easier and every transition gets logged in a predictable format.

PR #1485, opened 2020-08-25
Improve chain_state persist logs and add panics

### LINKED ISSUE #1439: SuperBlock sync: Review last_chain_state persistance with superblocks

tmpolaczyk:
Fix #1439

Followup to #1471 

This PR adds a new data structure `chain_state_snapshot` to replace the old `previous_chain_state`. The functionality is the same, but the `chain_state_snapshot` also stores the superblock index of the previous chain state, and the superblock index of the last chain state that has been persisted to the storage.

This will prevent accidentally persisting the same chain state twice, or skipping a superblock in some edge cases. The node will panic if it:
* Stores in memory two different copies of the chain state for the same superblock index. This means that one of the copies is wrong.
* Tries to persist the chain state for superblock N, when there is no copy of the chain state for superblock N (because we forgot to call `self.move_chain_state_forward`)
* Tries to persist the chain state for a superblock older than the highest persisted superblock (overwritting the chain state with an older copy)

Also, this PR improves some logs and deletes some unused functions.

### DIFF
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -2756,36 +2756,6 @@ impl ChainState {
             .consensus_constants
             .clone()
     }
-
-    /// Method to check that all inputs point to unspent outputs
-    pub fn find_unspent_outputs(&self, inputs: &[Input]) -> bool {
-        inputs.iter().all(|tx_input| {
-            let output_pointer = tx_input.output_pointer();
-
-            self.unspent_outputs_pool.contains_key(&output_pointer)
-        })
-    }
-    /// Retrieve the output pointed by the output pointer in an input
-    pub fn get_output_from_input(&self, input: &Input) -> Option<&ValueTransferOutput> {
-        let output_pointer = input.output_pointer();
-
-        self.unspent_outputs_pool.get(&output_pointer)
-    }
-    /// Map a vector of inputs to a the vector of ValueTransferOutputs pointed by the inputs' output pointers
-    pub fn get_outputs_from_inputs(
-        &self,
-        inputs: &[Input],
-    ) -> Result<Vec<ValueTransferOutput>, Input> {
-        let v = inputs
-            .iter()
-            .map(|i| self.get_output_from_input(i))
-            .fuse()
-            .flatten()
-            .cloned()
-            .collect();
-
-        Ok(v)
-    }
 }
 
 /// Alternative public key mapping: maps each secp256k1 public key hash to
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -2726,6 +2726,68 @@ pub struct ChainState {
     pub superblock_state: SuperBlockState,
 }
 
+impl ChainState {
+    /// Return the number of consolidated blocks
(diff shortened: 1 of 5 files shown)


node/src/actors/chain_manager/mod.rs
```diff
@@ -161,6 +161,138 @@ pub struct SyncTarget {
...
+            .chain_info
+            .as_ref()
+            .expect("ChainInfo is None")
+            .highest_block_checkpoint;
+
+        let superblock_beacon = state
+            .chain_info
+            .as_ref()
+            .expect("ChainInfo is None")
+            .highest_superblock_checkpoint;
```
lrubiorod, line 189: There are 2 functions that do the same (get_beacon(), and get_superblock_beacon)

tmpolaczyk: Yes, sadly this is not the same as `self.get_beacon()`, because we are getting the beacon of the new `state`, and `state.get_beacon()` does not exist.

tmpolaczyk: Done, created `ChainState::get_chain_beacon` and `ChainState::get_superblock_beacon`

data_structures/src/chain.rs
```diff
@@ -2726,6 +2726,35 @@ pub struct ChainState {
     pub superblock_state: SuperBlockState,
 }
 
+impl ChainState {
+    /// Return the number of consolidated blocks
+    pub fn block_number(&self) -> u32 {
+        u32::try_from(self.block_chain.len()).unwrap()
+    }
+
+    pub fn get_chain_beacon(&self) -> CheckpointBeacon {
```
mariocao, line 2735: Missing doc

data_structures/src/superblock.rs
```diff
@@ -148,10 +148,32 @@ impl SuperBlockVotesMempool {
...
+                    .map(|vote| vote.secp256k1_signature.public_key.pkh())
+                    .map(|pkh| pkh.to_string())
+                    .collect();
+
+                format!("  {}: {} votes: {:?}\n", superblock_hash, pkhs.len(), pkhs)
+            })
+            .collect();
+
+        // Remove trailing "\n" if `s` is not empty
+        s.pop();
```
mariocao, line 172: nice!

node/src/actors/chain_manager/mod.rs
```diff
@@ -170,7 +170,7 @@ pub struct ChainManager {
     /// Blockchain state data structure
     chain_state: ChainState,
     /// Backup for ChainState
```
mariocao, line 172: Maybe we could add why we need the backup (for resetting the state after a reorganization)

mariocao (changes requested):

lrubiorod (approved):

data_structures/src/chain.rs
tmpolaczyk: The code should not compile with missing docs, opened PR #1491 to fix that.

mariocao (approved):

Issue #1478, opened 2020-08-25
Signal to JsonRpc clients consolidated blocks when a superblock is consolidated


mariocao:

mariocao:
Estimation included in #1477

PR #1488, opened 2020-08-25
feat(node): signal block consolidation upon creating next superblock

aesedepece:
Solve #1477
Solve #1478
Solve #1494
Solve #1501

### DIFF
--- a/net/src/client/tcp/actors/jsonrpc.rs
+++ b/net/src/client/tcp/actors/jsonrpc.rs
@@ -146,11 +146,12 @@ impl Actor for JsonRpcClient {
 impl Supervised for JsonRpcClient {}
 
 /// JSONRPC notification.
-///
-/// Contains:
-/// 1. Subscription ID
-/// 2. A JSON value
-pub struct Notification(pub String, pub Value);
+pub struct Notification {
+    /// This doubles as subscription ID or subscription topic, depending on where it is used.
+    pub id: String,
+    /// A JSON value.
+    pub value: Value,
+}
 
 impl Message for Notification {
     type Result = ();
@@ -272,7 +273,10 @@ impl Handler<Subscribe> for JsonRpcClient {
                             .map(move |value| {
                                 log::debug!("<< Forwarding notification from node to subscribers",);
                                 log::trace!("<< {:?}", value);
-                                Notification(id.clone(), value)
+                                Notification {
+                                    id: id.clone(),
+                                    value,
+                                }
                             })
                             .map_err(|err| Error::RequestFailed {
                                 message: err.to_string(),
@@ -303,14 +307,17 @@ impl Handler<Subscribe> for JsonRpcClient {
 impl StreamHandler<Notification, Error> for JsonRpcClient {
     fn handle(
         &mut self,
-        Notification(subscription_id, value): Notification,
+        Notification {
+            id: subscription_id,
+            value,
+        }: Notification,
         _ctx: &mut Self::Context,
     ) {
         if let Ok(subscriptions) = (*self.active_subscriptions).lock() {
             if let Some(Subscribe(ref request, ref recipient)) = subscriptions.get(&subscription_id)
             {
                 let method = subscription_topic_from_request(request);
-                if let Err(err) = recipient.do_send(Notification(method, value)) {
+                if let Err(err) = recipient.do_send(Notification { id: method, value }) {
                     log::error!("Client couldn't notify subscriber: {}", err);
                 }
             }
--- a/net/src/client/tcp/actors/jsonrpc.rs
+++ b/net/src/client/tcp/actors/jsonrpc.rs
@@ -1,7 +1,7 @@
 //! Defines a JsonRPC over TCP actor.
 //!
 //! See the `JsonRpcClient` struct for more information.
-use std::time::Duration;
+use std::{collections::HashMap, time::Duration};
 
 use actix::prelude::*;
 use async_jsonrpc_client::{
@@ -22,8 +22,7 @@ pub struct JsonRpcClient {
     socket: TcpSocket,
     url: String,
     retry_connect: bool,
-    subscriber: Option<SetSubscriber>,
-    subscription_id: Option<String>,
+    subscriptions: HashMap<String, Subscribe>,
 }
 
 impl JsonRpcClient {
@@ -36,8 +35,7 @@ impl JsonRpcClient {
             socket,
             url: url.to_owned(),
             retry_connect: false,
-            subscriber: None,
-            subscription_id: None,
+            subscriptions: Default::default(),
         };
 
         Ok(client.start())
@@ -53,7 +51,19 @@ impl JsonRpcClient {
         self._handle = _handle;
         self.socket = socket;
         self.retry_connect = false;
-        // TODO: re-subscribe
+    }
+
+    /// Renew all existing subscriptions.
+    pub fn resubscribe(&mut self, ctx: &mut <Self as Actor>::Context) {
+        log::debug!("Recovering {} subscriptions", self.subscriptions.len());
+        self.subscriptions
+            .clone()
+            .into_iter()
+            .for_each(|(_, subscribe)| {
+                let method = subscription_method_from_request(&subscribe.0);
+                log::debug!("Resubscribing to `{}` notifications", method);
+                <Self as Handler<Subscribe>>::handle(self, subscribe, ctx);
+            })
     }
 
     /// Send Json-RPC request.
@@ -82,12 +92,20 @@ impl JsonRpcClient {
 
 impl Actor for JsonRpcClient {
     type Context = Context<Self>;
+
+    fn stopped(&mut self, _ctx: &mut Self::Context) {
+        log::error!("JsonRpcClient actor stopped!")
+    }
 }
 
 impl Supervised for JsonRpcClient {}
 
-/// TODO: doc
-pub struct Notification(pub Value);
+/// JSONRPC notification.
+///
+/// Contains:
+/// 1. Subscription ID
+/// 2. A JSON value
+pub struct Notification(pub String, pub Value);
 
 impl Message for Notification {
     type Result = ();
@@ -148,7 +166,7 @@ impl Handler<Request> for JsonRpcClient {
     ) -> Self::Result {
         if self.retry_connect {
             self.reconnect();
-            ctx.notify(Subscribe);
+            self.resubscribe(ctx);
         }
 
         let fut = self
@@ -156,6 +174,7 @@ impl Handler<Request> for JsonRpcClient {
             .into_actor(self)
             .timeout(timeout, Error::RequestTimedOut(timeout.as_millis()))
             .map_err(move |err, _act, ctx| {
+                log::error!("JSONRPC Request error: {:?}", err);
                 if is_connection_error(&err) {
                     ctx.notify(RetryConnect);
                 }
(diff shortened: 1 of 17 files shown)


node/src/actors/chain_manager/mod.rs
```diff
@@ -1224,6 +1225,10 @@ impl ChainManager {
                         act.chain_state.chain_info.as_mut().unwrap().highest_superblock_checkpoint =
                             act.chain_state.superblock_state.get_beacon();
 
+                        // Let JSON-RPC clients know that the blocks in the previous superblock can now
+                        // be considered consolidated
+                        act.notify_previous_superblock_consolidation(ctx);
+
```
mariocao, line 1231: It is weird that there are some format issues. For example this match case is not properly indented. :)

node/src/actors/chain_manager/mod.rs
```diff
@@ -1483,6 +1488,32 @@ impl ChainManager {
...
+            .superblock_state
+            .get_beacon()
+            .checkpoint
+            .checked_mul(superblock_period)
+            .expect("Multiplying a superblock ID by `superblock_period` should never overflow");
+        let initial_epoch = final_epoch.saturating_sub(superblock_period);
+        let beacons = self.handle(
+            GetBlocksEpochRange::new_with_limit(initial_epoch..final_epoch, 0),
+            ctx,
+        );
```
mariocao, line 1507: Oh my... we are storing all block headers in memory! :rofl:

mariocao (changes requested):

net/src/client/tcp/actors/jsonrpc.rs
```diff
@@ -20,40 +24,81 @@ use super::Error;
...
+                log::debug!(
+                    "Processing pending subscription for topic {}: {:?}",
+                    topic,
+                    subscribe.0
+                );
+                ctx.notify(subscribe.clone());
+            });
+
+        // Clear up all subscriptions (will be pushed again if they keep failing)
+        active_subscriptions.clear();
```
tmpolaczyk, line 100: I think this will not clear `self.active_subscriptions`, it will only clear the clone created by `.map(|x| x.clone())`.

net/src/client/tcp/actors/jsonrpc.rs
```diff
@@ -82,12 +127,30 @@ impl JsonRpcClient {
...
 impl Supervised for JsonRpcClient {}
 
-/// TODO: doc
-pub struct Notification(pub Value);
+/// JSONRPC notification.
+///
+/// Contains:
+/// 1. Subscription ID
+/// 2. A JSON value
+pub struct Notification(pub String, pub Value);
```
tmpolaczyk, line 153: There is no need to keep unnamed fields

```suggestion
/// JSONRPC notification.
pub struct Notification {
    /// Subscription ID
    pub subscription_id: String,
    /// A JSON value
    pub value: Value,
}
```

net/src/client/tcp/actors/jsonrpc.rs
```diff
@@ -191,69 +246,73 @@ impl Message for Subscribe {
...
+        let topic = subscription_topic_from_request(&request);
+        log::debug!(
+            "Handling Subscribe message for topic {}. Request is {:?}",
+            topic,
+            request
         );
-        self.retry_connect = true;
+
+        ctx.address()
+            .send(request.clone())
```
tmpolaczyk, line 259: Use `ctx.notify()` to send messages to self:

```suggestion
        ctx.notify(request.clone())
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -1295,6 +1300,11 @@ impl ChainManager {
                         &act.chain_state.alt_keys,
                     );
 
+                    // Put the local superblock into chain state
+                    act.chain_state
+                        .superblock_state
+                        .set_current_superblock(superblock.clone());
```
tmpolaczyk, line 1306: Can we move this logic inside `act.chain_state.superblock_state.build_superblock`?

node/src/actors/chain_manager/mod.rs
tmpolaczyk: It's only a mapping of block_epoch to block_hash, 4 bytes + 32 bytes. That's about 35MB for 1 million blocks, so we can wait a few million blocks until it becomes a problem.

node/src/actors/chain_manager/mod.rs
```diff
@@ -1488,6 +1498,37 @@ impl ChainManager {
...
+            .checked_mul(superblock_period)
+            .expect("Multiplying a superblock ID by `superblock_period` should never overflow");
+        let initial_epoch = final_epoch.saturating_sub(superblock_period);
+        let beacons = self.handle(
+            GetBlocksEpochRange::new_with_limit(initial_epoch..final_epoch, 0),
+            ctx,
+        );
+
+        // Get superblock from chain state snapshot
+        let superblock = self.chain_state.superblock_state.get_current_superblock();
```
tmpolaczyk, line 1518: This is the current superblock, so this function will only be valid if called after the superblock votes are counted, but before the next superblock is created. To avoid future problems, I would prefer to pass `superblock` as a parameter to `notify_previous_superblock_consolidation`, and use the id of that superblock to calculate the block epoch range. That way this function can be called at any point in time, even if the superblock has been consolidated long ago.

node/src/actors/json_rpc/json_rpc_methods.rs
```diff
@@ -232,8 +232,11 @@ pub fn jsonrpc_io_handler(
                 };
 
                 match method_name.as_str() {
-                    "newBlocks" => {
```
tmpolaczyk, line 235: `newBlocks` is more clear than `blocks`. Unless you can now receive old blocks?

wallet/src/params.rs
```diff
@@ -76,4 +77,17 @@ impl NodeParams {
             }
         }
     }
+
+    /// Get the address of an existing JsonRpcClient actor.
+    pub fn get_client(&self) -> NodeClient {
+        let lock = self.client.clone();
```
tmpolaczyk, line 83: Will this clone the `NodeClient` or it will only increase the reference count and return an `Arc`? To avoid that question you can use `Arc::clone` explicitly, or annotate the type as `let lock: Arc<_> = ...`

```suggestion
        let lock = Arc::clone(&self.client);
```

wallet/src/params.rs
```diff
@@ -76,4 +77,17 @@ impl NodeParams {
...
+        log::trace!(
+            "Getting JsonRpcClient actor from worker ({} references)",
+            Arc::strong_count(&lock)
+        );
+
+        lock.read()
+            .map(|x| x.clone())
+            .expect("Node client lock should never be poisoned")
```
tmpolaczyk, line 91: This may be true because locks can only be poisoned on write, and I can't find where `self.client` is written to. But if it's not written to, you should be able to remove the `RwLock`.

net/src/client/tcp/actors/jsonrpc.rs
aesedepece: I think `ctx.notify` does `do_send` instead of `send` though, so we lose the chance to do the fancy stuff that we do here with the response.

aesedepece: I mean, I thought exactly the same, put `ctx.notify` in place, realized it had a totally different API and rolled back the change so it could be refactored in a different PR.

node/src/actors/json_rpc/json_rpc_methods.rs
aesedepece: Using your same logic, if the node can only notify new stuff as it comes, there's no point in tagging anything as "new" — we simply know it is.

wallet/src/params.rs
aesedepece: This clones the `Arc` indeed, but that fact is concealed by the magic dereference of the `Arc` when it is used afterwards. Great suggestion — the most explicit the code, the better.

wallet/src/params.rs
aesedepece: Yes, at this point it can be removed. It was there in the first place because I implemented a way for the client address to be swapped in case the `JsonRpcClient` actor died, but I removed that later because it didn't work well enough.

node/src/actors/chain_manager/mod.rs
```diff
@@ -1488,6 +1498,37 @@ impl ChainManager {
...
+    /// Let JSON-RPC clients know that the blocks in the previous superblock can now
+    /// be considered consolidated
+    fn notify_previous_superblock_consolidation(&mut self, ctx: &mut Context<ChainManager>) {
+        let superblock_period = u32::from(self.consensus_constants().superblock_period);
+        let final_epoch = self
+            .chain_state_snapshot
+            .highest_persisted_superblock
+            .checked_mul(superblock_period)
+            .expect("Multiplying a superblock ID by `superblock_period` should never overflow");
+        let initial_epoch = final_epoch.saturating_sub(superblock_period);
```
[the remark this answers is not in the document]
aesedepece: Can you point to any occurrences of this?

node/src/actors/json_rpc/json_rpc_methods.rs
tmpolaczyk: Still, I would prefer to leave `newBlocks` as an alias to `blocks` to avoid breaking changes.

node/src/actors/chain_manager/mod.rs
aesedepece: What's the advantage? My concern is that `build_superblock` looks like it's modeled in a rather self-contained way and it felt strange to me if it made any significant changes on `self.chain_state`.

net/src/client/tcp/actors/jsonrpc.rs
tmpolaczyk: The actix docs mention that sending messages to self is a bad idea and that's why `ctx.notify()` exists. I don't know why, maybe it's an anti-pattern in the actor world or something.

node/src/actors/chain_manager/mod.rs
tmpolaczyk: The advantage if we merge them is that we can be sure that `superblock_state.get_current_superblock()` returns the last superblock. Because the only way to change it is using `superblock_state.build_superblock()`. By adding a `set_current_superblock` method we must make sure that it is always called after `superblock_state.build_superblock()`, which is error prone. It's a similar argument to making fields public vs private.

net/src/client/tcp/actors/jsonrpc.rs
```diff
@@ -82,12 +127,31 @@ impl JsonRpcClient {
...
+    }
 }
 
 impl Supervised for JsonRpcClient {}
 
-/// TODO: doc
-pub struct Notification(pub Value);
+/// JSONRPC notification.
+pub struct Notification {
+    /// This doubles as subscription ID or subscription topic, depending on where it is used.
```
tmpolaczyk, line 150: Please don't. Create a separate `NotificationTopic` struct.

wallet/src/actors/app/params.rs
```diff
@@ -9,8 +9,14 @@ use crate::actors;
...
+    pub client: NodeClient,
     pub server_addr: SocketAddr,
     pub session_expires_in: Duration,
     pub requests_timeout: Duration,
 }
+
+#[derive(Clone)]
+pub struct NodeClient {
+    pub url: String,
+    pub actor: Addr<JsonRpcClient>,
```
tmpolaczyk, line 21: `Addr` is just an `Arc` internally so it be cloned easily, but `String` will perform a deep clone. So `url` should probably be `Arc<str>`. Also I think `url` is never used, so it can be removed.

tmpolaczyk: We could also simply use `Arc<NodeClient>` when clones are needed instead.

node/src/actors/json_rpc/server.rs
```diff
@@ -187,15 +187,53 @@ impl Handler<NewBlock> for JsonRpcServer {
...
+                    ctx.spawn(sink.notify(params).into_actor(self).then(move |res, _, _| {
+                        if let Err(e) = res {
+                            log::error!("Failed to send notification: {:?}", e);
+                        }
+
+                        actix::fut::ok(())
+                    }));
+                }
+            } else {
+                log::warn!("Failed to find a subscription for superblocks notifications");
```
tmpolaczyk, line 233: This warning will show when there are no clients connected. Remove it or move it to debug:
```suggestion
                log::debug!("No subscriptions for superblocks notifications");
```

mariocao (approved):

wallet/src/actors/app/params.rs
aesedepece: True, it's never used.

aesedepece: It's used inside a `json!` macro in `notify_client`.

PR #1502, opened 2020-08-31
Modify trapezoid heigh factor

### LINKED ISSUE #1496: Discuss new reputation algorithm (trapezoid) parameters
Discuss the parameters of the new reputation algorithm that directly affects the eligibility of identities.

lrubiorod:
Close #1496


data_structures/src/chain.rs
```diff
@@ -2913,7 +2913,18 @@ impl ReputationEngine {
...
-            .sorted_by_key(|&pkh| std::cmp::Reverse((self.trs.get(&pkh), pkh)))
+            .sorted_by_key(|&pkh| {
+                let [a0, a1, a2, a3] = self.current_alpha.0.to_be_bytes();
+                let pkh_mod_bytes = [
+                    pkh.hash[0] + a0,
+                    pkh.hash[1] + a1,
+                    pkh.hash[2] + a2,
+                    pkh.hash[3] + a3,
+                ];
+                let hash = Hash::from(calculate_sha256(&pkh_mod_bytes));
```
tmpolaczyk, line 2924: It may be expensive to calculate a SHA256 hash every time we compare two elements, but since it is only used as a tie breaker we can optimize this.

You can use `sorted_by(|a, b| { ...` which takes two elements as input and returns an `Ordering`: `FnMut(&T, &T) -> Ordering`. Then compare the reputation of a with the reputation of b and only calculate the hash when the reputation is equal.

https://doc.rust-lang.org/stable/std/primitive.slice.html#method.sort_by

tmpolaczyk: And add tests that show that the ordering changes when `current_alpha` changes.

validations/src/validations.rs
```diff
@@ -2733,89 +2733,89 @@ mod tests {
             .unwrap();
 
         let (_, p00) = calculate_reppoe_threshold(&rep_engine, &ids[0], 1);
```
tmpolaczyk, line 2735: Maybe we could simplify this test by adding a helper function that calculates the threshold of all the identities in order, like

```rust
let rep_thresholds = |thres| {
    vec![
        (calculate_reppoe_threshold(&rep_engine, &ids[0], thres).1 * 100_f64).round() as u32,
        (calculate_reppoe_threshold(&rep_engine, &ids[1], thres).1 * 100_f64).round() as u32,
        // TODO: use a for loop
    ]
};

let expected_rep_1 = vec![28, 24, 18, 9, 5, 14, 1, 1];
assert_eq!(rep_thresholds(1), expected_rep_1);
```

data_structures/src/chain.rs
```diff
@@ -2913,10 +2913,46 @@ impl ReputationEngine {
...
+    a: &PublicKeyHash,
+    b: &PublicKeyHash,
+    rep_eng: &ReputationEngine,
+) -> Ordering {
+    let rep_a = rep_eng.trs().get(a).0;
+    let rep_b = rep_eng.trs().get(b).0;
+
+    rep_a
+        .cmp(&rep_b)
+        .then({
```
tmpolaczyk, line 2934: This way the hash is still computed every time, but the `then_with` method does exactly what you want

```suggestion
        .then_with(|| {
```

data_structures/src/chain.rs
```diff
@@ -2913,10 +2913,64 @@ impl ReputationEngine {
...
+    let rep_a = rep_eng.trs().get(a).0;
+    let rep_b = rep_eng.trs().get(b).0;
+
+    rep_a.cmp(&rep_b).then_with(|| {
+        compare_vector_with_offset(&a.hash, &b.hash, rep_eng.current_alpha.0 as usize)
+    })
+}
+
+/// Auxiliary function that compares 2 vectors with an offset
+fn compare_vector_with_offset(a: &[u8], b: &[u8], offset: usize) -> Ordering {
```
tmpolaczyk, line 2937: I think this function should be discussed in a separated PR. It would be great if you could create an issue and change the body of this function to something super simple like

```rust
    // FIXME(#1506): implement better tie breaker
    a.cmp(b)
```

This way we can clearly state the design goals and the attack vectors.

I prefer to remove the magic function because I wrote a short script and with 1 hour of proof of work I was able to generate a PKH that is better than 77% of the random PKHs, so I assume that this function can be easily broken: [Playground](https://play.rust-lang.org/?version=stable&mode=release&edition=2018&gist=5de2247ee012c06975927ff7bf5fd580)

tmpolaczyk (approved):

PR #1503, opened 2020-08-31
Limit mempool size

### LINKED ISSUE #1495: TransactionsPool with size limit

tmpolaczyk:
Close #1495 

This PR adds a size limit to the `TransactionsPool`. The default is around 192M.

With this limit, when adding new transaction to a full `TransactionsPool`, the transactions with the smaller `fee / weight` ratio will be removed. Note that the limit only applies to value transfer and data request transactions. It does not apply to commits and reveals because they are implicitly limited by the number of data requests.

This limit can be configured in `witnet.toml`, as `mining.transactions_pool_total_weight_limit`. The units are "weight units" as described in https://github.com/witnet/WIPs/pull/32

Estimated overhead: total memory usage of a typical `TransactionsPool` with `transactions_pool_total_weight_limit = 192M`

* With only value transfer transactions with (1, 2, or 3) inputs and 2 outputs, the peak RAM usage is about 320MB.
* With only data request transactions with (1, 2, or 3) inputs, 2 outputs, and witnesses=2, the peak RAM usage is about 220MB. This does not include the potential commits and reveals, which are taken into account when calculating a data request weight.

Therefore, I think we can assume that the overhead will be less than x2, so the default `transactions_pool_total_weight_limit = 192M` will require at most 384 MB of RAM for the `TransactionsPool`. You can profile the memory usage using by compiling `data_structures/examples/transactions_pool_overhead.rs` and running it through a memory profiler like massif or heaptrack.

### DIFF
--- a/data_structures/examples/transactions_pool_overhead.rs
+++ b/data_structures/examples/transactions_pool_overhead.rs
@@ -1,0 +1,126 @@
+use rand::{thread_rng, Rng};
+use witnet_data_structures::chain::{
+    DataRequestOutput, Hash, Input, KeyedSignature, OutputPointer, PublicKeyHash, RADAggregate,
+    RADRequest, RADRetrieve, RADTally, RADType, Secp256k1Signature, Signature, TransactionsPool,
+    ValueTransferOutput,
+};
+use witnet_data_structures::transaction::{
+    DRTransaction, DRTransactionBody, Transaction, VTTransaction, VTTransactionBody,
+};
+
+fn random_request() -> RADRequest {
+    RADRequest {
+        time_lock: 0,
+        retrieve: vec![
+            RADRetrieve {
+                kind: RADType::HttpGet,
+                url: String::from("https://www.bitstamp.net/api/ticker/"),
+                script: vec![130, 24, 119, 130, 24, 100, 100, 108, 97, 115, 116],
+            },
+            RADRetrieve {
+                kind: RADType::HttpGet,
+                url: String::from("https://api.coindesk.com/v1/bpi/currentprice.json"),
+                script: vec![
+                    132, 24, 119, 130, 24, 102, 99, 98, 112, 105, 130, 24, 102, 99, 85, 83, 68,
+                    130, 24, 100, 106, 114, 97, 116, 101, 95, 102, 108, 111, 97, 116,
+                ],
+            },
+        ],
+        aggregate: RADAggregate {
+            filters: vec![],
+            reducer: 3,
+        },
+        tally: RADTally {
+            filters: vec![],
+            reducer: 3,
+        },
+    }
+}
+
+fn random_dr_output() -> DataRequestOutput {
+    let mut rng = thread_rng();
+
+    DataRequestOutput {
+        data_request: random_request(),
+        witness_reward: rng.gen(),
+        // The number of witnesses changes the RAM usage considerably
+        // More witnesses = more weight = less transactions in pool = less RAM usage
+        witnesses: 2,
+        commit_and_reveal_fee: rng.gen(),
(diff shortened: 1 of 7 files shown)


data_structures/src/chain.rs
```diff
@@ -1449,6 +1476,7 @@ impl TransactionsPool {
     /// # use witnet_data_structures::chain::TransactionsPool;
     /// let pool = TransactionsPool::with_capacity(20);
     /// ```
+    // TODO: assess if this method can be removed
```
aesedepece, line 1479: I think so.

data_structures/src/chain.rs
```diff
@@ -1877,8 +1923,53 @@ impl TransactionsPool {
...
+    /// Remove transactions until the size limit is satisfied.
+    /// Returns a list of all the removed transactions.
+    fn remove_transactions_for_size_limit(&mut self) -> Vec<Transaction> {
+        let mut removed_transactions = vec![];
+
+        while self.total_vt_size_bytes + self.total_dr_size_bytes > self.weight_limit {
+            // Try to split the memory between value transfer and data requests using the same
+            // ratio as the one used in blocks
+            // The ratio of vt to dr in blocks is currently 4:1
+            // TODO: calculate this from ConsensusConstants?
```
aesedepece, line 1935: Please!

node/src/actors/chain_manager/actor.rs
```diff
@@ -90,6 +90,8 @@ impl ChainManager {
                 // Get consensus parameter from config
                 act.consensus_c = config.connections.consensus_c;
 
+                let _removed_transactions = act.transactions_pool.set_total_weight_limit(config.mining.transactions_pool_total_weight_limit);
```
aesedepece, line 93: What about logging the evicted transactions at debug level?

tmpolaczyk: :+1:
Here `removed_transactions` should always be empty, but I added logs when inserting new transactions.

data_structures/src/chain.rs
tmpolaczyk: Done, but need tests to ensure that a ratio of 4:1 is 4.0 and not 0.25

data_structures/examples/transactions_pool_overhead.rs
```diff
@@ -0,0 +1,101 @@
...
+        if !limit_reached && !removed_transactions.is_empty() {
+            println!("Limit reached");
+            limit_reached = true;
+
+            /*
+            println!("Removing all transactions");
+            pool.set_total_weight_limit(0);
+            pool.assert_empty();
+            return;
+            */
```
lrubiorod, line 98: Is this part in WIP yet?

data_structures/examples/transactions_pool_overhead.rs
```diff
@@ -0,0 +1,133 @@
...
+
+    let mut limit_reached = false;
+
+    for i in 0..1_000_000 {
+        let (transaction, fee) = random_transaction();
+        let removed_transactions = pool.insert(transaction, fee);
+
+        if !limit_reached && !removed_transactions.is_empty() {
+            println!("Limit reached after {} transactions", i);
+            limit_reached = true;
```
lrubiorod, line 123: why do not use a break point to exit when limit is achieved?

tmpolaczyk: I also want to see that the usage stays constant after the limit is reached

data_structures/examples/transactions_pool_overhead.rs
tmpolaczyk: This was used to debug #1505, will remove now

PR #1505, opened 2020-09-02
fix(data_structures): avoid potential memory leak in transactions pool

tmpolaczyk:
This should almost never happen in the testnet, because this bug requires two transactions that try to spend the same UTXO, but one of them has at least one additional input.

Example: two transactions T1 and T2.
* T1 has input I1
* T2 has inputs I1 and I2
* T1 is included in a block, making T2 invalid. But the mapping (I2 => T2) is not removed, wasting memory
* However, if another transaction T3 with input I2 is included in a block, the memory is cleared

After PR #1503 this will trigger when removing any transaction from a full `TransactionsPool`, so it should be fixed before that.

### DIFF
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -1703,7 +1703,7 @@ impl TransactionsPool {
     /// assert!(!pool.vt_contains(&transaction.hash()));
     /// ```
     pub fn vt_remove(&mut self, key: &Hash) -> Option<VTTransaction> {
-        let transaction = self.vt_remove_inner(key);
+        let transaction = self.vt_remove_inner(key, true);
 
         if let Some(transaction) = &transaction {
             self.remove_inputs(&transaction.body.inputs);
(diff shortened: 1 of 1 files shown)


data_structures/src/chain.rs
```diff
@@ -1719,6 +1719,22 @@ impl TransactionsPool {
             .remove(key)
             .map(|(weight, transaction)| {
                 self.sorted_vt_index.remove(&(weight, *key));
+                // Remove the inputs from the output_pointer_map
```
lrubiorod, line 1722: Should you repeat in the dr_remove_inner method?

tmpolaczyk: I had a test for DR transactions but removed it because "it should be the same". So that was a mistake :laughing:

lrubiorod (approved):

girazoki (approved):

PR #1508, opened 2020-09-02
Superblock inclusion proofs

girazoki:


data_structures/src/chain.rs
```diff
@@ -635,6 +636,76 @@ impl SuperBlock {
...
+    /// and the blocks that the superblock contains
+    /// If the DR transaction is not found within those blocks, returns None
+    pub fn dr_proof_of_inclusion(
+        &self,
+        blocks: &[Block],
+        tx: DRTransaction,
+    ) -> Option<TxInclusionProof> {
+        // Get the PoI for the block root, if the data request is found on the list of blocks
+        // Obtain also the tally root of the block containing the Dr TX.
+        let (mut poi, current_root) = blocks.iter().find_map(|b| {
```
tmpolaczyk, line 650: You only use `current_root` to find the `dr_root_idx` later. You could get the index here directly:

```suggestion
        let (mut poi, dr_root_idx) = blocks.iter().enumerate().find_map(|(idx, b)| {
```

And return `Some((tx.data_proof_of_inclusion(b)?, idx))`

data_structures/src/chain.rs
```diff
@@ -635,6 +636,76 @@ impl SuperBlock {
...
         .concat()
     }
+
+    /// Generate a Proof of Inclusion for a given Superblock tally root, provided the DR transaction
+    /// and the blocks that the superblock contains
+    /// If the DR transaction is not found within those blocks, returns None
+    pub fn dr_proof_of_inclusion(
+        &self,
+        blocks: &[Block],
+        tx: DRTransaction,
```
tmpolaczyk, line 646: This should be a reference, it will save some clones from the tests.

```suggestion
        tx: &DRTransaction,
```

data_structures/src/transaction.rs
```diff
@@ -253,11 +253,31 @@ impl TxInclusionProof {
         }
     }
 
+    // Create a TX inclusion proof assuming the inputs are already Hashes
+    pub fn new_with_hashes(index: usize, leaves: Vec<Hash>) -> TxInclusionProof {
```
tmpolaczyk, line 257: This is basically the same as `TxInclusionProof::new`, so if you make it generic you will be able to implement `new` as a wrapper of `new_with_hashes`:

```suggestion
    pub fn new_with_hashes<I: IntoIterator<Item = Hash>>(
        index: usize,
        leaves: I,
    ) -> TxInclusionProof {
```

Then `TxInclusionProof::new` becomes one line:

```rust
        Self::new_with_hashes(index, leaves.into_iter().map(|t| t.hash()))
```

tmpolaczyk:
Does this create the expected proof when a superblock only contains 1 block? I think it should be the same as the first proof, skipping the `concat` stage, right?

girazoki:
Good point, I will try it out!

girazoki:
Added a test and passes. So everything well

data_structures/src/chain.rs
```diff
@@ -27,6 +27,7 @@ use witnet_protected::Protected;
 use witnet_reputation::{ActiveReputationSet, TotalReputationSet};
 use witnet_util::timestamp::get_timestamp;
 
+use crate::transaction::TxInclusionProof;
```
tmpolaczyk, line 30: fmt

data_structures/src/chain.rs
```diff
@@ -4989,4 +5052,572 @@ mod tests {
...
+    fn test_dr_merkle_root_superblock() {
+        let inputs: Vec<Input> = (1..4)
+            .map(|x| {
+                Input::new(OutputPointer {
+                    transaction_id: Hash::default(),
+                    output_index: x,
+                })
+            })
+            .collect();
+        let dr_txs: Vec<DRTransaction> = inputs
```
tmpolaczyk, line 5066: I think this is repeated in all tests, how about creating a function like `example_drs(n)` that returns n transactions

data_structures/src/chain.rs
```diff
@@ -4989,4 +5052,572 @@ mod tests {
...
+            &[b1.block_header.clone(), b2.block_header.clone()],
+            &[Hash::default()],
+            1,
+            Hash::default(),
+        );
+
+        let expected_indices = vec![0, 2, 2];
+        let expected_lemma_lengths = vec![3, 3, 2];
+
+        for index in 0..expected_indices.len() {
```
tmpolaczyk, line 5104: I think this loop is also common in all tests?

data_structures/src/chain.rs
```diff
@@ -635,6 +636,66 @@ impl SuperBlock {
...
+        let (mut poi, tally_root_idx) = blocks
+            .iter()
+            .enumerate()
+            .find_map(|(idx, b)| Some((tx.data_proof_of_inclusion(b)?, idx)))?;
+
+        // Collect all tally roots from the blocks
+        let tally_roots: Vec<Hash> = blocks
+            .iter()
+            .map(|b| b.block_header.merkle_roots.tally_hash_merkle_root)
+            .collect();
```
tmpolaczyk, line 689: This is no longer needed, you also need to remove `: Vec<Hash>`

```suggestion
```

data_structures/src/chain.rs
```diff
@@ -3422,13 +3480,107 @@ pub fn block_example() -> Block {
...
+        let proof = InclusionProof::sha256(result.index, lemma);
+        assert!(proof.verify(
+            tally_txs[index].data_poi_hash().into(),
+            sb.tally_root.into()
+        ));
+    }
+}
+
+pub fn build_test_dr_txs(length: u32) -> Vec<DRTransaction>{
+    let inputs: Vec<Input> = (1..length+1)
```
tmpolaczyk, line 3528: There's a syntax for that:

```suggestion
    let inputs: Vec<Input> = (1..=length)
```

Also, could you move this functions inside `mod tests`? Since they should only be used in tests.

tmpolaczyk (approved):

data_structures/src/chain.rs
```diff
@@ -3582,10 +3640,114 @@ pub fn block_example() -> Block {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::superblock::mining_build_superblock;
```
lrubiorod, line 3643: fmt

data_structures/src/chain.rs
```diff
@@ -3582,10 +3640,114 @@ pub fn block_example() -> Block {
...
     use crate::{
         superblock::ARSIdentities,
         transaction::{CommitTransactionBody, RevealTransactionBody, VTTransactionBody},
     };
+    use witnet_crypto::merkle::{merkle_tree_root, InclusionProof};
+
+    pub fn dr_root_superblock_loop_test(
```
mariocao, line 3650: I think these methods don't need to be public. :)

mariocao (changes requested):

data_structures/src/transaction.rs
```diff
@@ -242,7 +242,15 @@ impl TxInclusionProof {
         index: usize,
         leaves: I,
     ) -> TxInclusionProof {
-        let mt = FullMerkleTree::sha256(leaves.into_iter().map(|t| t.hash().into()).collect());
+        Self::new_with_hashes(index, leaves.into_iter().map(|t| t.hash()))
+    }
+
+    // Create a TX inclusion proof assuming the inputs are already Hashes
```
mariocao, line 248:
```suggestion
    /// Create a TX inclusion proof assuming the inputs are already Hashes
```

mariocao (changes requested):

data_structures/src/chain.rs
```diff
@@ -635,6 +635,64 @@ impl SuperBlock {
...
+    /// Generate a Proof of Inclusion for a given Superblock DR root, provided the DR transaction
+    /// and the blocks that the superblock contains
+    /// If the DR transaction is not found within those blocks, returns None
+    pub fn dr_proof_of_inclusion(
+        &self,
+        blocks: &[Block],
+        tx: &DRTransaction,
+    ) -> Option<TxInclusionProof> {
+        // Get the PoI for the block root, if the data request is found on the list of blocks
+        // Obtain also the index of the tally root of the block containing the Dr TX.
```
lrubiorod, line 648:
```suggestion
        // Obtain also the index of the DataRequest root of the block containing the Dr TX.
```

data_structures/src/chain.rs
```diff
@@ -635,6 +635,64 @@ impl SuperBlock {
...
+
+        // Collect all DR roots from the blocks
+        let dr_roots = blocks
+            .iter()
+            .map(|b| b.block_header.merkle_roots.dr_hash_merkle_root);
+
+        // Generate the second PoI, using the block DR root as a leave
+        let second_poi = TxInclusionProof::new_with_hashes(dr_root_idx, dr_roots);
+
+        // Concatenate one PoI with the second. This will update the index and append the syblings
```
tmpolaczyk, line 662:
```suggestion
        // Concatenate one PoI with the second. This will update the index and append the siblings
```

mariocao (approved):

lrubiorod (approved):

PR #1509, opened 2020-09-03
feat(reputation): implement a tie breaker for sort based on hash

### LINKED ISSUE #1507: Improve tie breakear in "get_rep_ordered_ars_list" function

lrubiorod:
Close #1507

### DIFF
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -2946,9 +2946,10 @@ impl ReputationEngine {
             .collect()
     }
 }
+
 /// Compare 2 PublicKeyHashes comparing:
 /// First: reputation
-/// Second: PublicKeyHash bytes from an offset
+/// Second: Hashes related to PublicKeyHash and alpha clock || PublicKeyHashes in case of 0 rep
 fn compare_reputed_pkh(
     a: &PublicKeyHash,
     b: &PublicKeyHash,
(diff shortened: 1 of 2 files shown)


data_structures/src/chain.rs
```diff
@@ -2929,8 +2930,21 @@ fn compare_reputed_pkh(
...
+
+            a_bytes.extend(alpha_bytes.clone());
+            b_bytes.extend(alpha_bytes);
+
+            let new_hash_a: Hash = calculate_sha256(&a_bytes).into();
+            let new_hash_b: Hash = calculate_sha256(&b_bytes).into();
+
+            new_hash_a.cmp(&new_hash_b)
+        } else {
+            a.cmp(&b)
```
tmpolaczyk, line 2946: Good idea. But explain why:

```suggestion
            // If both identities have 0 reputation their ordering is not important because they will have the same eligibility, so compare them by PublicKeyHash
            a.cmp(&b)
```

data_structures/src/chain.rs
```diff
@@ -2929,8 +2930,21 @@ fn compare_reputed_pkh(
     let rep_b = rep_eng.trs().get(b).0;
 
     rep_a.cmp(&rep_b).then_with(|| {
-        // FIXME(#1507): implement better tie breaker
-        a.cmp(b)
+        if rep_a > 0 {
+            let alpha_bytes = rep_eng.current_alpha.0.to_be_bytes().to_vec();
```
tmpolaczyk, line 2934: You should be able to remove `.to_vec()` here if you set the type to slice:

```suggestion
            let alpha_bytes: &[u8] = &rep_eng.current_alpha.0.to_be_bytes();
```

tmpolaczyk (approved):

PR #1510, opened 2020-09-03
fix(reputation): avoid substract with overflow

lrubiorod:


data_structures/src/chain.rs
```diff
@@ -3062,7 +3062,7 @@ fn trapezoidal_eligibility(
         calculate_trapezoid_triangle(m, k, active_reputed_ids_len);
 
     // To complete the trapezoid, an offset needs to be added (the rectangle at the base)
-    let remaining_reputation = total_active_rep - total_triangle_reputation;
+    let remaining_reputation = total_active_rep.saturating_sub(total_triangle_reputation);
```
tmpolaczyk, line 3065: Doesn't this imply that the reputation in the triangle can be greater than the total active reputation?

Maybe instead of `saturating_sub` we should modify the `magic_line` function to use floor instead of round. This way the triangle reputation should always be smaller than or equal to the total.

tmpolaczyk (approved):

girazoki (approved):

PR #1512, opened 2020-09-04
fix: clear TransactionsPool on reorganizations

tmpolaczyk:
This is a temporary workaround. See #1513 for the discussion

### DIFF
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -1512,6 +1512,39 @@ impl TransactionsPool {
             && self.re_transactions.is_empty()
     }
 
+    /// Remove all the transactions but keep the allocated memory for reuse.
+    pub fn clear(&mut self) {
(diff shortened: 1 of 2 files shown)


girazoki (approved):

lrubiorod (approved):

data_structures/src/chain.rs
```diff
@@ -1512,6 +1512,39 @@ impl TransactionsPool {
             && self.re_transactions.is_empty()
     }
 
+    /// Remove all the transactions but keep the allocated memory for reuse.
+    pub fn clear(&mut self) {
+        let TransactionsPool {
```
tmpolaczyk, line 1517: I'm using this syntax instead of `self.vt_transactions.clear();` because this way will stop compiling if we add new fields, so we won't forget to clear them.

PR #1520, opened 2020-09-07
fix(node): accept genesis block in AlmostSynced state

lrubiorod:


girazoki (approved):

node/src/actors/chain_manager/handlers.rs
```diff
@@ -289,7 +289,7 @@ impl Handler<AddBlocks> for ChainManager {
                     }
                 }
```
tmpolaczyk, line 290: We should add debug logs to detect unhandled `AddBlocks` messages. One here and one in `StateMachine::Synced => {}`.

```suggestion
                } else {
                    log::debug!("Unhandled AddBlocks message");
                }
```

mariocao (approved):

tmpolaczyk (approved):

PR #1523, opened 2020-09-08
Retrieval proxies / multiple transports / inconsistent source attack protection (aka "smart retrievals")

### LINKED ISSUE #1274: Assesment: Smart retrievals for preventing Inconsistence source attack

### LINKED ISSUE #1511: Discussion: How to avoid "malicious" requests? (Inconsistent source attack)
Currently, all the requests included in the network as considered as valid to solve by the node operators. But there are some cases where to solve a request has a risk to lose reputation, wits and/or anonymity:
- A request with a strict filter.
- A request with a source that retrieval random values.
- A request that expose our anonymity (an API that retrieval our IP)
- ...

The idea would be to discuss about how to improve the witnessing logic to detect and avoid this kind of requests

aesedepece:
Enables using proxies (HTTP, HTTPS, SOCKS4 and SOCKS5) when performing data retrieval. This allows retrieving data sources through different transports so as to ensure that the data sources are consistent and we are taking as small of a risk as
possible when committing to specially crafted data requests that may be potentially ill-intended.

Just add this to the `[connections]` section of `witnet.toml` to proxy your retrievals through 3 different open proxies:
```toml
retrieval_proxies = [
        "socks5://181.102.35.195:1080",
        "socks5://8.210.60.118:1080",
        "socks5://47.114.8.133:1080"
]
```

These addresses have been taken from [this public list of proxies][proxies].

Close #1274 
Close #1511 

[proxies]: https://spys.one/en/socks-proxy-list/

### DIFF
--- a/net/src/client/http/mod.rs
+++ b/net/src/client/http/mod.rs
@@ -1,0 +1,62 @@
+use std::str::FromStr;
+use std::sync::Arc;
+
+use futures03::future::BoxFuture;
+use surf::http::{self, Response, Uri};
+use surf::middleware::{Body, HttpClient as SurfHttpClient};
+use surf::Client;
+
+type Request = http::Request<Body>;
+
+/// A surf-alike HTTP client that additionally supports proxies (HTTP(S), SOCKS4 and SOCKS5)
+#[derive(Clone, Debug)]
+pub struct WitnetHttpClient {
+    client: Arc<isahc::HttpClient>,
+}
+
+impl WitnetHttpClient {
+    /// Create a new `WitnetHttpClient`
+    pub fn new(proxy: &Option<&str>) -> Self {
+        let proxy_address = proxy.map(|x| Uri::from_str(x).unwrap());
+        let client = Arc::new(
+            if let Some(proxy_address) = proxy_address {
+                isahc::HttpClient::builder().proxy(proxy_address)
+            } else {
+                isahc::HttpClient::builder()
+            }
+            .build()
+            .unwrap(),
+        );
+
+        Self { client }
+    }
+
+    /// Turn this `WitnetHttpClient` into a `surf`-compatible client
+    pub fn as_surf_client(&self) -> Client<Self> {
+        Client::with_client(self.clone())
+    }
+}
+
+impl SurfHttpClient for WitnetHttpClient {
+    type Error = isahc::Error;
+
+    fn send(&self, req: Request) -> BoxFuture<'static, Result<Response<Body>, Self::Error>> {
+        let client = self.client.clone();
+        Box::pin(async move {
+            // Compose HTTP request
+            let (parts, body) = req.into_parts();
+            let body = isahc::Body::reader(body);
+            let req: http::Request<isahc::Body> = http::Request::from_parts(parts, body);
+
+            // Send HTTP request and wait for response
+            let res = client.send_async(req).await?;
+
+            // Read HTTP response
+            let (parts, body) = res.into_parts();
+            let body = Body::from_reader(body);
+            let res = http::Response::from_parts(parts, body);
+
+            Ok(res)
+        })
+    }
+}
--- a/net/src/client/http/mod.rs
+++ b/net/src/client/http/mod.rs
@@ -16,8 +16,8 @@ pub struct WitnetHttpClient {
 
 impl WitnetHttpClient {
     /// Create a new `WitnetHttpClient`
-    pub fn new(proxy: &Option<&str>) -> Self {
-        let proxy_address = proxy.map(|x| Uri::from_str(x).unwrap());
+    pub fn new(proxy: &Option<String>) -> Self {
+        let proxy_address = proxy.as_ref().map(|x| Uri::from_str(x.as_str()).unwrap());
         let client = Arc::new(
             if let Some(proxy_address) = proxy_address {
                 isahc::HttpClient::builder().proxy(proxy_address)
--- a/node/src/actors/node.rs
+++ b/node/src/actors/node.rs
@@ -53,7 +53,8 @@ pub fn run(config: Arc<Config>, callback: fn()) -> Result<(), failure::Error> {
     SystemRegistry::set(inventory_manager_addr);
 
     // Start RadManager actor
-    let rad_manager_addr = RadManager::default().start();
+    let rad_manager_addr =
+        RadManager::with_proxies(config.connections.retrieval_proxies.clone()).start();
     SystemRegistry::set(rad_manager_addr);
 
     // Start JSON RPC server
--- a/rad/src/lib.rs
+++ b/rad/src/lib.rs
@@ -4,11 +4,14 @@ use futures::{executor::block_on, future::join_all};
 use serde::Serialize;
 pub use serde_cbor::to_vec as cbor_to_vec;
 pub use serde_cbor::Value as CborValue;
+use surf::Client;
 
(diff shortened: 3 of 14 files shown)


rad/src/lib.rs
```diff
@@ -201,11 +208,81 @@ pub async fn run_retrieval_report(
...
     // Disable all execution tracing features, as this is the best-effort version of this method
-    run_retrieval_report(retrieve, RadonScriptExecutionSettings::disable_all())
+    run_retrieval_report(retrieve, RadonScriptExecutionSettings::disable_all(), None)
         .await
         .map(RadonReport::into_inner)
 }
 
+/// Run retrieval using multiple transports, and only produce a positive result if the retrieved
+/// values pass the filter function from the tally stage.
+///
```
girazoki, line 218: Is there a reason to use the filter with the tally script instead of the aggregation script? I am assuming the first will be more restrictive and that is the idea, but it also sounds counter-intuitive to have a filter for different sources and apply a different filter to the same source.

rad/src/lib.rs
```diff
@@ -201,11 +208,81 @@ pub async fn run_retrieval_report(
...
+        .collect();
+
+    let values = join_all(futures).await;
+
+    evaluate_paranoid_retrieval(values, tally, settings)
+}
+
+/// Evaluate whether the values obtained when retrieving a data source through multiple transports
+/// are consistent, i.e. they all pass the filters from the tally stage.
+fn evaluate_paranoid_retrieval(
```
girazoki, line 248: Can we add a small test for this?

net/src/client/http/mod.rs
```diff
@@ -0,0 +1,62 @@
+use std::str::FromStr;
+use std::sync::Arc;
+
+use futures03::future::BoxFuture;
+use surf::http::{self, Response, Uri};
+use surf::middleware::{Body, HttpClient as SurfHttpClient};
+use surf::Client;
```
lrubiorod, line 7: fmt

rad/src/lib.rs
lrubiorod: The tally filter is the filter that will slash the nodes, so one of the risky cases would be a data request with a very strict filter in tally. So it has to be tested to declare the request as "safu"

rad/src/lib.rs
```diff
@@ -201,11 +208,81 @@ pub async fn run_retrieval_report(
...
         .map(RadonReport::into_inner)
 }
 
+/// Run retrieval using multiple transports, and only produce a positive result if the retrieved
+/// values pass the filter function from the tally stage.
+///
+/// The idea behind this is to avoid producing commitments for data requests with sources that act
+/// in an inconsistent way, i.e. they return very different values when queried through different
+/// HTTP transports at once.
+pub async fn run_paranoid_retrieval(
```
lrubiorod, line 222: Love the name xD

rad/src/lib.rs
lrubiorod: agree

rad/src/lib.rs
```diff
@@ -201,11 +208,81 @@ pub async fn run_retrieval_report(
...
+                    RadError::InconsistentSource,
+                )));
+            }
+        }
+    }
+
+    // If all values pass the filters, return the first value (any of them will work, actually)
+    Ok(values
+        .get(0)
+        .expect("At least one value must have been received")
```
lrubiorod, line 282: It only happens if there are zero values, so... it is correct that message?

lrubiorod:
image
Should we show that logs in debug or trace level?

rad/src/lib.rs
girazoki: Just by doing a couple of tests it can be seen that several proxies are not able to access some APIs. Shall we consider that timeout error as a valid input for the filter? In my opinion we should discard them. Of course in that case we should discuss what to do if all cases gave timeout

net/src/client/http/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+
+/// A surf-alike HTTP client that additionally supports proxies (HTTP(S), SOCKS4 and SOCKS5)
+#[derive(Clone, Debug)]
+pub struct WitnetHttpClient {
+    client: Arc<isahc::HttpClient>,
+}
+
+impl WitnetHttpClient {
+    /// Create a new `WitnetHttpClient`
+    pub fn new(proxy: &Option<String>) -> Self {
```
tmpolaczyk, line 19:
```suggestion
    pub fn new(proxy: Option<&str>) -> Self {
```

And on the caller side, you can use `.as_deref()` to convert `Option<String>` into `Option<&str>`.

net/src/client/http/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+            .build()
+            .unwrap(),
+        );
+
+        Self { client }
+    }
+
+    /// Turn this `WitnetHttpClient` into a `surf`-compatible client
+    pub fn as_surf_client(&self) -> Client<Self> {
+        Client::with_client(self.clone())
```
tmpolaczyk, line 36: Move the clone to the caller, so this function can also be used without cloning.

```suggestion
    pub fn as_surf_client(self) -> Client<Self> {
        Client::with_client(self)
```

tmpolaczyk: Ah nevermind, it's just an Arc clone so it doesn't matter

node/src/actors/node.rs
```diff
@@ -53,7 +53,8 @@ pub fn run(config: Arc<Config>, callback: fn()) -> Result<(), failure::Error> {
     SystemRegistry::set(inventory_manager_addr);
 
     // Start RadManager actor
-    let rad_manager_addr = RadManager::default().start();
+    let rad_manager_addr =
+        RadManager::with_proxies(config.connections.retrieval_proxies.clone()).start();
```
tmpolaczyk, line 57: Not sure if setting the proxies here will save them if the actor panics and then restarts. That's why we always use the `starting` method in `actor.rs` to read the config on actor startup.

tmpolaczyk: Well, now it doesn't matter because if any actor panics we always stop the node.

rad/src/lib.rs
```diff
@@ -201,11 +208,81 @@ pub async fn run_retrieval_report(
...
+    // If there was only one retrieved value, there's no actual need to run the tally, as this means
+    // that only one transport was used and therefore the node is not in paranoid mode.
+    if values.len() > 1 {
+        let mut context = ReportContext::from_stage(Stage::Tally(TallyMetaData::default()));
+        let _tally = run_tally_with_context_report(values.clone(), tally, &mut context, settings);
+
+        // If at least one of the retrieved values is marked as an outlier, resolve to the
+        // `InconsistentSource` error.
+        if let Stage::Tally(TallyMetaData { liars, .. }) = context.stage {
+            if !liars.is_empty() {
```
girazoki, line 271:
```suggestion
            if liars.iter().any(|&v| v==true) {
```

girazoki: Also, seems that as soon as there is some specific error (403), the paranoid_retrieval returns the error whitout evaluating the responses. This is, if locally I get a value  and through tor I get a 403, the 403 is returned and committed, I think we should refrain from commiting in that case

aesedepece:
Closed for inactivity. A reboot of this proposal will probably need a huge rebase anyway.

PR #1527, opened 2020-09-09
Store superblocks in storage

### LINKED ISSUE #1548: Store superblocks in storage
The node should store the created superblocks in the storage. The superblocks are needed for some funcionality of the bridge, and the last consolidated superblock is needed to allow new nodes to synchronize.

The superblock votes will not be stored yet.

tmpolaczyk:
Close #1548 

Add JSON-RPC method `getSuperblock` that given a `superblock_index` returns the superblock, and the list of block hashes that pertain to that superblock. But only if the superblock has already seen a majority of votes and is consolidated. Also works when given a `block_epoch` instead of a `superblock_index`: will convert the block epoch to superblock index.

Note that nodes will only store superblocks when they are consolidated, so a new node that synchronizes to the testnet will not store the superblocks that were created when the node was offline.

You can test this method using:

```
echo '{"jsonrpc":"2.0","id":"1","method":"getSuperblock","params":{"superblock_index":0}}' | cargo run -- -c witnet.toml node raw
```

### DIFF
--- a/node/src/actors/json_rpc/json_rpc_methods.rs
+++ b/node/src/actors/json_rpc/json_rpc_methods.rs
@@ -1242,53 +1242,42 @@ pub fn get_superblock(
         Err(e) => return Box::new(futures::failed(e)),
     };
 
-    match params {
+    let fut = match params {
         GetSuperblockBlocksParams::SuperblockIndex(superblock_index) => {
-            let inventory_manager_addr = InventoryManager::from_registry();
-
-            let fut = inventory_manager_addr
-                .send(GetItemSuperblock { superblock_index })
-                .map_err(internal_error)
-                .and_then(|dr_info| match dr_info {
-                    Ok(x) => match serde_json::to_value(&x) {
-                        Ok(x) => futures::finished(x),
-                        Err(e) => {
-                            let err = internal_error_s(e);
-                            futures::failed(err)
-                        }
-                    },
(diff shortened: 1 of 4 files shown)


node/src/actors/messages.rs
```diff
@@ -903,7 +915,7 @@ pub struct BlockNotify {
 ///
 /// As per current consensus algorithm, "consolidated blocks" implies that there exists at least one
 /// superblock in the chain that builds upon the superblock where those blocks were anchored.
-#[derive(Message, Serialize)]
+#[derive(Clone, Debug, PartialEq, Eq, Message, Serialize, Deserialize)]
```
girazoki, line 918: Order pleaze

node/src/actors/json_rpc/json_rpc_methods.rs
```diff
@@ -1516,6 +1589,7 @@ mod tests {
                 "getPublicKey",
                 "getReputation",
                 "getReputationAll",
+                "getSuperblock",
```
lrubiorod, line 1581:
```suggestion
                "getSuperBlock",
```
???

tmpolaczyk: BREAKING CHANGE

lrubiorod: xD

lrubiorod (approved):

PR #1531, opened 2020-09-10
feat(bridge): add timeout to post_to_witnet_more_than_once

### LINKED ISSUE #1293: The bridge does not get out of including state if posting DR fails

tmpolaczyk:
Fix #1293 (if post_to_witnet_more_than_once is set to true)

This solves issues where data requests get stuck in claimed state because they are not included in a witnet block.


tmpolaczyk:
Included in PR #1533

Issue #1532, opened 2020-09-10
Persist chain state during synchronizing


lrubiorod:
Currently, we have to achieve the synced state until persist the chain_state, it produces that if I have some problems during synchronizing we have to start from block zero again. So it could produce that a node could never synch.

A possible solution will be to trust in the majority of last superblock consolidated of our outbounds to persist the chain_state while we are below of this specific superblock

Issue #1526, opened 2020-09-09
Subtract with overflow in current_committee_size_requirement


tmpolaczyk:
After synchronizing a new node to the current testnet, this line causes a subtract with overflow:

https://github.com/witnet/witnet-rust/blob/c86e6b7390902a90c3218a90424bee4541c8822e/node/src/actors/chain_manager/mod.rs#L2084

Used here:

https://github.com/witnet/witnet-rust/blob/c86e6b7390902a90c3218a90424bee4541c8822e/node/src/actors/chain_manager/mod.rs#L1279-L1286

This happened in a custom branch so the first step should be to try to reproduce it on master. Also, it only happens in debug mode, nodes that use the compiled released will silently wrap on overflow, which may result in invalid behavior and forks.

PR #1535, opened 2020-09-10
fix(node): do not restore chain state when processing empty AddBlocks

tmpolaczyk:
This is a workaround to #1532.

That issue is still not solved because stopping the node while synchronizing does not store the chain state.

Explanation: the `Session` actor implicitly use an empty `AddBlocks` message to indicate an error. For example, when a peer disconnects while the node is receiving blocks from them, the `Session` actor will send an empty add blocks.

Before this PR, the `ChainManager` would restore the chain state to the last consolidated state, which is epoch 0 in nodes that are synchronizing the first time. This would make the synchronization process restart, and as the number of blocks increases synchronizing a new node would become impossible. After this PR, the `ChainManager` will simply request a new blocks batch from another peer when a peer fails to send the required blocks.


girazoki (approved):

tmpolaczyk:
Trying to synchronize in debug mode but I keep hitting #1526. Not sure what happens in release mode.

PR #1538, opened 2020-09-11
fix(node): fix bug in synchronization to do not revert in case of fir…

lrubiorod:
…st block fail

### DIFF
--- a/node/src/actors/chain_manager/mod.rs
+++ b/node/src/actors/chain_manager/mod.rs
@@ -1452,15 +1452,20 @@ impl ChainManager {
         let mut num_processed_blocks = 0;
 
         for block in blocks.iter() {
-            num_processed_blocks += 1;
-
             if let Err(e) = self.process_requested_block(ctx, block.clone()) {
                 log::error!("Error processing block: {}", e);
-                self.initialize_from_storage(ctx);
-                log::info!("Restored chain state from storage");
+                if num_processed_blocks > 0 {
+                    // Restore only in case there were several blocks consolidated before
+                    // This is not needed if the error is in the first block because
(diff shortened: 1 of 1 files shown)


node/src/actors/chain_manager/mod.rs
```diff
@@ -1439,13 +1439,17 @@ impl ChainManager {
...
 
-        for block in blocks.iter() {
+        for (i, block) in blocks.iter().enumerate() {
             num_processed_blocks += 1;
 
             if let Err(e) = self.process_requested_block(ctx, block.clone()) {
                 log::error!("Error processing block: {}", e);
-                self.initialize_from_storage(ctx);
-                log::info!("Restored chain state from storage");
+                if i > 0 {
```
tmpolaczyk, line 1447: We already have a counter variable:

```suggestion
                if num_processed_blocks > 1 {
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -1439,13 +1439,17 @@ impl ChainManager {
...
+                    // Restore only in case there were several blocks consolidated before
+                    // This is safe because the state has not been mutated yet
```
tmpolaczyk, line 1449:
```suggestion
                    // This is not needed if the error is in the first block because the state has not been mutated yet
```

tmpolaczyk (approved):

tmpolaczyk:
Note for the future: `process_blocks_batch` returns `(bool, usize)` when it could return `Option<usize>`.

clbartoli (approved):

aesedepece:
This is ready to rebase and merge, right?

PR #1541, opened 2020-09-11
feat(wallet): syncronization + superblocks + blocks indexing

mariocao:
This PR is built on top of @aesedepece PR #1488 as superblock notifications are required.

This PR aims to:

- update synchronization algorithm to be complaint with the new node consensus algorithm
- handle superblock notifications
- support pending and confirmed blocks, transactions and address information
- implement recovery mechanism in case node chain reverts (reorganizations)
  - handling edge-cases of block notifications
  - cleaning pending state if reorganization is detected
- improve block indexing logs (in debug mode)

However, there are still 2 issues missing that were created:
 - #1539 update dr balance movements (if tally txn is indexed)
 - #1540 index txn should take into account that address infos might be in pending state

Closes #1479 
Closes #1480 
Closes #1506
Closes #1545

PR #1545, opened 2020-09-14
Wallet/feat/handle superblock notifications

### LINKED ISSUE #1479: [Wallet] Handle consolidated block(s) notifications
- Accumulate balance movements linked to unconsolidated blocks (e.g. hasmap)
- Update block balance movements to confirmed
- Notify clients

### LINKED ISSUE #1480: [Wallet] Add `confirmed` flag to `Transaction` in `BalanceMovement`
Add `confirmed` flag to `Transaction` in `BalanceMovement` data structure.
Additionally, when indexing blocks and creating new balance movements set `confirmed` flag to `false`.

### LINKED ISSUE #1506: [Wallet] Persist transaccional data if block is confirmed, otherwise accumulate until confirmation

### LINKED ISSUE #1540: [Wallet] Update addresses information considering pending states (after indexing a transaction)
Current `_index_transaction()` logic, retrieves from database the last `address_info` object and updates it accordingly.
However, after latest changes in the wallet now `pending` on-memory `address_info` should also be considered.

Right now it is easy to find a specific `address_info` in the `pending` wallet state.
They are stored in a `HashMap<BlockHash, Vec<AddressInfo>>`.

Despite, there are not many block hashes and less address info items in the `pending` wallet state, it might be desirable an additional structure (e.g. hashmap) in the wallet `State` struct that facilitates finding `addresses` without having to iterate through the `pending_address_infos` block hashes and vectors.

### LINKED ISSUE #1551: [Wallet] Transactions can be added to balance twice when indexing transactions
While testing #1540, I got one transaction indexed only once but added twice to the account balance.

This was most likely caused by the wallet restarting while the transaction was pending confirmation, which may have caused the wallet to resync and process the transaction again.

We'll need to try to reproduce it and provide more context.

aesedepece:
This is a follow-up of #1541, rebasing it on top of master and effectively superseding it. 

Closes #1479
Closes #1480
Closes #1506 
Closes #1540
Closes #1541
Closes #1551


wallet/src/actors/worker/methods.rs
```diff
@@ -621,12 +641,24 @@ impl Worker {
...
                 let get_block_future = self.get_block(id.clone());
                 let block: types::ChainBlock = futures03::executor::block_on(get_block_future)?;
 
-                // Process each block
-                self.handle_block(block.clone(), wallet.clone(), DynamicSink::default())?;
+                // Compute if block should be considered confirmed
+                // Note: blocks confirmed in past superblocks are considered confirmed
+                // FIXME(#1437): best approach would be the node signaling if blocks are confirmed
+                let tip_superblock_index = self.current_epoch()? / superblock_period;
+                let current_superblock_index = epoch / superblock_period;
```
girazoki, line 652: This is a bit missleading, as current refers to the block we are processing inside this loop while tip refers to the current epoch divided with the period. I would change them for:

let current_superblock_index = self.current_epoch()? / superblock_period;
let processed_superblock_index = epoch / superblock_period

Or something similar

wallet/src/actors/worker/methods.rs
```diff
@@ -741,59 +773,152 @@ impl Worker {
...
+            wallet.update_sync_state(new_last_sync, confirmed)?;
+        }
+
+        Ok(())
+    }
+
+    /// Handle superblock notification by confirming the transactions of the consolidated blocks
+    pub fn handle_superblock(
+        &self,
+        block_notification: types::SuperBlockNotification,
```
girazoki, line 852:
```suggestion
        superblock_notification: types::SuperBlockNotification,
```

wallet/src/model.rs
```diff
@@ -23,7 +22,7 @@ pub struct UnlockedWallet {
     pub available_accounts: Vec<u32>,
 }
 
-#[derive(Debug, Serialize, PartialEq)]
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
```
girazoki, line 25: order

wallet/src/repository/wallet/mod.rs
```diff
@@ -1,27 +1,29 @@
...
     model,
     params::Params,
-    types::{self, Hashable as _},
+    types::{self, signature, ExtendedPK, Hash, Hashable as _, RadonError},
 };
 
+use super::*;
+use crate::constants::{EXTERNAL_KEYCHAIN, INTERNAL_KEYCHAIN};
+use crate::model::AddressInfo;
+use std::sync::Arc;
```
girazoki, line 20: fmt

wallet/src/actors/worker/methods.rs
aesedepece: What about this?
```rust
                let node_tip_superblock_index = self.current_epoch()? / superblock_period;
                let local_tip_superblock_index = epoch / superblock_period;
```

wallet/src/repository/wallet/state.rs
```diff
@@ -1,21 +1,43 @@
...
+    pub name: Option<String>,
+    /// Next external index used to derive addresses
     pub next_external_index: u32,
+    /// Next internal index used to derive addresses
     pub next_internal_index: u32,
-    pub available_accounts: Vec<u32>,
-    pub balance: u64,
+    /// List of pending address infos, waiting to be confirmed with a superblock
+    pub pending_addresses_by_block: HashMap<String, Vec<Arc<model::Address>>>,
+    pub pending_addresses_by_path: HashMap<String, Arc<model::Address>>,
```
aesedepece, line 34: Missing docs

mariocao (approved):

wallet/src/actors/worker/methods.rs
girazoki: this is cool

PR #1553, opened 2020-09-14
fix(data structures): default committee size after transitioning from…

girazoki:
This PR fixes the transition from bootstrap committee to ARS based committee. Essentially, as soon as the epoch limit is reached and as long as the boootstrap committee keeps being the last committee, returns the default committee size. I implemented like this because the committee size is always calculated based on the number of votes we had in the previous superblock, which is the bootstrap committee length at the moment of transitioning.

In short, not very convinced about this

 However, this has some caveats:

- If the ARS members are offline at the moment of transitioning, the committee size does not get reduced.
- If after transitioning, for some reason the ARS is identical to the bootstrap committee, and one of them goes offline, the committee size does not get reduced

### DIFF
--- a/node/src/actors/chain_manager/mod.rs
+++ b/node/src/actors/chain_manager/mod.rs
@@ -1251,14 +1251,14 @@ impl ChainManager {
 
                         let chain_info = act.chain_state.chain_info.as_ref().unwrap();
                         let reputation_engine = act.chain_state.reputation_engine.as_ref().unwrap();
+                        let last_superblock_signed_by_bootstrap = last_superblock_signed_by_bootstrap(&chain_info.consensus_constants);
 
                         let reputed_ars_members =
                             // Before reaching the epoch activity_period + collateral_age the bootstrap committee signs the superblock
                             // collateral_age is measured in blocks instead of epochs, but this only means that the period in which
                             // the bootstrap committee signs is at least epoch activity_period + collateral_age
-                            if block_epoch
-                                > chain_info.consensus_constants.collateral_age
-                                    + chain_info.consensus_constants.activity_period
+                            if superblock_index
+                                >= last_superblock_signed_by_bootstrap
                             {
                                 let ars_members = reputation_engine.get_rep_ordered_ars_list();
                                 let reputed = reputed_ars(&ars_members, &reputation_engine);
@@ -1290,6 +1290,7 @@ impl ChainManager {
                         consensus_constants.superblock_committee_decreasing_step,
                         chain_info.highest_superblock_checkpoint.checkpoint,
                         superblock_index,
+                        last_superblock_signed_by_bootstrap
                     );
                     log::debug!("The current signing committee size is {}", committee_size);
 
@@ -2121,6 +2122,10 @@ fn show_sync_progress(
     );
 }
 
+fn last_superblock_signed_by_bootstrap(consensus_constants: &ConsensusConstants) -> u32 {
+    (consensus_constants.collateral_age + consensus_constants.activity_period)
+        / u32::from(consensus_constants.superblock_period)
+}
 // Returns the committee size to be applied given the default committee size, decreasing period
 // and  step, last consolidated epoch and the current checkpoint
 fn current_committee_size_requirement(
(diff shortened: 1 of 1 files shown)


node/src/actors/chain_manager/mod.rs
```diff
@@ -1251,46 +1251,53 @@ impl ChainManager {
...
+                                    let reputed = reputed_ars(&ars_list, &reputation_engine);
+                                    // In case of no reputed nodes, return all active nodes
+                                    if reputed.is_empty() {
+                                        ars_list
+                                    } else {
+                                        reputed
+                                    }
+                                };
+                                let committee_size = {
+                                    if act.chain_state.superblock_state.get_committee() == bootstrap_committee.into_iter().collect::<HashSet<PublicKeyHash>>() {
```
tmpolaczyk, line 1275: This checks if the last superblock has been signed by the bootstrap committee. But wouldn't it be easier to just check the superblock index, and compare it to a variable like `last_superblock_index_that_will_be_signed_by_the_bootstrap_committee`?

data_structures/src/superblock.rs
```diff
@@ -438,6 +438,11 @@ impl SuperBlockState {
         self.votes_mempool.get_received_votes()
     }
 
+    /// Returns the current committee
+    pub fn get_committee(&self) -> HashSet<PublicKeyHash> {
+        self.signing_committee.clone()
```
tmpolaczyk, line 443: You should be able to remove this clone and use a reference instead

```suggestion
    pub fn get_committee(&self) -> &HashSet<PublicKeyHash> {
        self.signing_committee
```

node/src/actors/chain_manager/mod.rs
girazoki: definitely easier, I can modify that

node/src/actors/chain_manager/mod.rs
```diff
@@ -2131,9 +2133,12 @@ fn current_committee_size_requirement(
     decreasing_step: u32,
     last_consolidated_checkpoint: u32,
     current_checkpoint: u32,
+    last_checkpoint_signed_by_bootstrap: u32,
 ) -> u32 {
-    // If the last consolidated superblock is 0, return the default committee size
-    if last_consolidated_checkpoint == 0 {
+    // If the last consolidated superblock or the current checkpoint is below last_checkpoint_signed_by_bootstrap, return the default committee size
+    if current_checkpoint <= last_checkpoint_signed_by_bootstrap
+        || last_consolidated_checkpoint <= last_checkpoint_signed_by_bootstrap
```
tmpolaczyk, line 2140: If we assume that `last_consolidated_checkpoint <= current_checkpoint`, this can be simplified to

```suggestion
    if last_consolidated_checkpoint <= last_checkpoint_signed_by_bootstrap
```

tmpolaczyk: But issue #1526 shows that that assumption is not always true, for some unknown reason.

girazoki: I have not being able to sync with the new testnet from master so I dont think I will be able to reproduce this error. What do we think about it?

node/src/actors/chain_manager/mod.rs
```diff
@@ -1290,6 +1290,8 @@ impl ChainManager {
                         consensus_constants.superblock_committee_decreasing_step,
                         chain_info.highest_superblock_checkpoint.checkpoint,
                         superblock_index,
+                        (chain_info.consensus_constants.collateral_age
+                            + chain_info.consensus_constants.activity_period)/superblock_period
```
tmpolaczyk, line 1294: I would prefer to have this logic defined in a separate function. Something like

```rust
fn last_superblock_signed_by_bootstrap(consensus_constants: &ConsensusConstants) -> u32 {}
```

And reuse the same variable when calculating the `reputed_ars_members`.

tmpolaczyk (approved):

node/src/actors/chain_manager/mod.rs
```diff
@@ -2130,9 +2135,11 @@ fn current_committee_size_requirement(
     decreasing_step: u32,
     last_consolidated_checkpoint: u32,
     current_checkpoint: u32,
+    last_checkpoint_signed_by_bootstrap: u32,
 ) -> u32 {
-    // If the last consolidated superblock is 0, return the default committee size
-    if last_consolidated_checkpoint == 0 {
+    assert!(last_consolidated_checkpoint <= current_checkpoint, "Something went wrong as the last consolidated checkpoint is bigger than our current checkpoint {} > {}", last_consolidated_checkpoint, current_checkpoint);
+    // If the last consolidated superblock or the current checkpoint is below last_checkpoint_signed_by_bootstrap, return the default committee size
+    if last_consolidated_checkpoint <= last_checkpoint_signed_by_bootstrap {
```
clbartoli, line 2142: Maybe we could separate in two cases? Something like
- If `last_consolidated_checkpoint = current_checkpoint`, then return the default committee size
- if `last_consolidated_checkpoint < current_checkpoint`, the return the last committee size, since it should be the same as the bootstrap committee size

girazoki:
Closed by #1557

PR #1560, opened 2020-09-16
fix(chain manager): avoid overflows while selecting committee

### LINKED ISSUE #1526: Subtract with overflow in current_committee_size_requirement

girazoki:
This PR solves the issue observed during synchronization that generated substraction overflows. This happened because the node tried to create a consolidate and candidate superblocks that it already had built before but then lost synchronization because of lack of consensus on the beacons sent by its peers. While constructing the target superblock again after having constructed the candidate, the substaction between these two checkpoints yields negative.

This PR adds functionality to check whether we need to construct the target AND CANDIDATE blocks while syncronizing

Closes #1526

### DIFF
--- a/node/src/actors/chain_manager/handlers.rs
+++ b/node/src/actors/chain_manager/handlers.rs
@@ -479,7 +479,7 @@ impl Handler<AddBlocks> for ChainManager {
 
                                     // Update ARS if there were no blocks right before the epoch during
                                     // which we should construct the target superblock
-                                    let candidate_superblock_epoch = (act.current_epoch.unwrap() / superblock_period) * superblock_period;
+                                    let candidate_superblock_checkpoint = act.current_epoch.unwrap() / superblock_period;
 
                                     // We need to persist blocks in order to be able to construct the
                                     // superblock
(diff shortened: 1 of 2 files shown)


node/src/actors/chain_manager/handlers.rs
```diff
@@ -469,35 +469,41 @@ impl Handler<AddBlocks> for ChainManager {
...
-                                if !batch_succeeded {
-                                    log::error!("Received invalid blocks batch...");
-                                    act.update_state_machine(StateMachine::WaitingConsensus);
-                                    act.sync_waiting_for_add_blocks_since = None;
-
-                                    return actix::fut::err(());
-                                }
-                                log_sync_progress(&sync_target, &remaining_blocks, num_processed_blocks, "SyncWithCandidate(remaining)");
+                                if let Some(candidate_superblock_epoch) = candidate_superblock_epoch {
+                                    // Process remaining blocks
```
tmpolaczyk, line 487: Not sure if this `if` is correct. When `superblock_candidate_is_needed` returns `None` it means that the candidate superblock has already been created. But the remaining blocks are the ones after that superblock, so I think they should always be processed. Because if we already had this blocks we wouldn't have received them.

tmpolaczyk: The only think inside the `if` should be:

    act.move_chain_state_forward(superblock_index);

Because we don't need to save the chain state if we already have this superblock saved.

node/src/actors/chain_manager/handlers.rs
```diff
@@ -494,19 +494,25 @@ impl Handler<AddBlocks> for ChainManager {
...
+                                else{
+                                    actix::fut::Either::B(actix::fut::ok(None))
+                                }
                             })
                             .and_then(move |candidate_superblock_epoch, act, ctx| {
-                                // Store chain state at candidate superblock, but do not persist it yet
-                                let superblock_index = candidate_superblock_epoch / superblock_period;
-                                act.move_chain_state_forward(superblock_index);
-
+                                if let Some(candidate_superblock_epoch) = candidate_superblock_epoch {
```
tmpolaczyk, line 511: Now that I see it, this if can be combined with the previous one, and we don't need to return `candidate_superblock_epoch` anymore.

tmpolaczyk (approved):

PR #1562, opened 2020-09-17
Add config parameter to limit the number of blocks downloaded when synchronizing

### LINKED ISSUE #1279: Synchronization: allow to configure batch block size
Currently there is a constant in the code that sets the maximum blocks number to be sent during synchronization process:

```
pub const MAX_BLOCKS_SYNC: usize = 500;
```

It cannot be changed in `witnet.toml`, and also changing it does only affect the blocks sent by this node, not the received blocks; meaning that there is no way for a node operator to set a smaller batch size. Nodes with slower storage or internet connection may want to lower the batch size, although the current limit of 500 does not seem to cause any major problems.

tmpolaczyk:
Close #1279

Built on top of #1577 

Adds a configurable limit to the received blocks batch size, used when synchronizing.

```
[connections]
requested_blocks_batch_limit = 50
```

Note that values higher than 500 will work as 500, because there is a hardcoded limit to the sent blocks batch size:

```rust
/// Maximum blocks number to be sent during synchronization process
pub const MAX_BLOCKS_SYNC: usize = 500;
```

Also, setting `requested_blocks_batch_limit = 1` will result in an infinite loop and the node not syncing, because of #1580 

Setting `requested_blocks_batch_limit = 2` does also not work because the node stops requesting new blocks after reaching the target superblock.

So setting values smaller than 2 times the superblock period (2 * 10 in the current testnet) are silently set to that value (2 * 10).

### DIFF
--- a/node/src/actors/session/handlers.rs
+++ b/node/src/actors/session/handlers.rs
@@ -1,4 +1,4 @@
-use std::{cmp::Ordering, io::Error, net::SocketAddr};
+use std::{cmp::Ordering, convert::TryFrom, io::Error, net::SocketAddr};
 
 use actix::{
     io::WriteHandler, ActorContext, ActorFuture, Context, ContextFutureSpawner, Handler,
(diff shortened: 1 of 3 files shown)


tmpolaczyk:
This PR would be easier if `Session` had access to `Config`. I think it's time to add an `Arc<Config>` everywhere.

aesedepece:
> This PR would be easier if `Session` had access to `Config`. I think it's time to add an `Arc<Config>` everywhere.

Can't agree more.

tmpolaczyk:
Opened #1577 to add an `Arc<Config>` to the `Session` actor

node/src/actors/session/handlers.rs
```diff
@@ -681,19 +681,32 @@ fn inventory_process_superblock(
...
+    let mut limit = usize::try_from(session.config.connections.requested_blocks_batch_limit)
+        .unwrap_or(usize::MAX);
+    if limit == 0 {
+        // unlimited
+        limit = usize::MAX;
+    }
+
+    // Small limits break the synchronization, so force a minimum value of 2 times the superblock
+    // period
+    let min_limit = 2 * usize::from(session.config.consensus_constants.superblock_period);
```
tmpolaczyk, line 697: After testing a bit I decided to set the minimum limit to 2 times the superblock period. Now any values smaller than that are silently overwritten.

node/src/actors/session/handlers.rs
```diff
@@ -681,19 +681,32 @@ fn inventory_process_superblock(
...
+    if limit < min_limit {
```
lrubiorod, line 698: should use a match to have a more fancy code to specify minimum and maximum?

node/src/actors/session/handlers.rs
tmpolaczyk: The issue with smaller batch size is that the `AddBlocks` handler does not request additional blocks after reaching the target superblock.

lrubiorod (approved):

clbartoli (approved):

PR #1565, opened 2020-09-17
fix(chain manager): compare superblock votes against current index

### LINKED ISSUE #1564: Superblock vote validation should take into account the current superblock epoch
When a superblock vote is received, we check if it's valid by comparing its index to `self.current_superblock_beacon.checkpoint`. This is wrong because `self.current_superblock_beacon.checkpoint` is not always the current superblock, it is the last created superblock, and that difference matters if there is a reorganization.

https://github.com/witnet/witnet-rust/blob/dd9ab39d7bebf43f1e128059b032bb61e2e6b1a3/data_structures/src/superblock.rs#L299

Instead, we should calculate the current superblock checkpoint and pass it as an argument to the `is_valid` function.

girazoki:
Close #1564

### DIFF
--- a/data_structures/src/superblock.rs
+++ b/data_structures/src/superblock.rs
@@ -262,7 +262,11 @@ impl SuperBlockState {
 
     /// Add a vote sent by another peer.
     /// This method assumes that the signatures are valid, they must be checked by the caller.
-    pub fn add_vote(&mut self, sbv: &SuperBlockVote) -> AddSuperBlockVote {
+    pub fn add_vote(
+        &mut self,
+        sbv: &SuperBlockVote,
+        current_superblock_index: u32,
+    ) -> AddSuperBlockVote {
         let r = if self.votes_mempool.contains(sbv) {
             // Already processed before
             AddSuperBlockVote::AlreadySeen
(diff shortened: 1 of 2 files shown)


node/src/actors/chain_manager/mod.rs
```diff
@@ -827,8 +837,10 @@ impl ChainManager {
             // of the ARS, even if the superblock hash is different from our local superblock hash.
             // If the superblock index is different from the current one we cannot check ARS membership,
             // so we broadcast it if the index is within an acceptable range (not too old).
-            let should_broadcast = match act.chain_state.superblock_state.add_vote(&superblock_vote)
-            {
+            let should_broadcast = match act.chain_state.superblock_state.add_vote(
+                &superblock_vote,
+                act.current_epoch.unwrap() / superblock_period,
```
tmpolaczyk, line 842: `act.current_epoch` can be `None` if the current epoch is below epoch 0. I think here we can safely set the epoch to 0 in that case:

```suggestion
                    act.current_epoch.unwrap_or(0) / superblock_period,
```

data_structures/src/superblock.rs
```diff
@@ -1548,27 +1641,96 @@ mod tests {
...
+            genesis_hash,
+            &alt_keys,
+            None,
+        );
+
+        let _sb2 = sbs.build_superblock(
+            &block_headers,
+            ars_identities,
+            2,
+            1,
```
tmpolaczyk, line 1725: This is the superblock index, it should be the same as the vote index because that used to be marked as "NotInSigningCommittee"

```suggestion
            5,
```

tmpolaczyk: Also, you could add this assert:

```rust
         assert_eq!(sbs.add_vote(&v0, 5), AddSuperBlockVote::NotInSigningCommittee); 
```

lrubiorod (approved):

tmpolaczyk (approved):

PR #1572, opened 2020-09-21
fix(node): broadcast candidates in WaitingConsensus and Synchronizing…

lrubiorod:
… states

### DIFF
--- a/node/src/actors/chain_manager/mod.rs
+++ b/node/src/actors/chain_manager/mod.rs
@@ -422,7 +422,10 @@ impl ChainManager {
                 if self.sm_state == StateMachine::WaitingConsensus
                     || self.sm_state == StateMachine::Synchronizing
                 {
-                    self.candidates.insert(hash_block, block);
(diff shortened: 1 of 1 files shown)


node/src/actors/chain_manager/mod.rs
```diff
@@ -422,7 +422,8 @@ impl ChainManager {
                 if self.sm_state == StateMachine::WaitingConsensus
                     || self.sm_state == StateMachine::Synchronizing
                 {
-                    self.candidates.insert(hash_block, block);
+                    self.candidates.insert(hash_block, block.clone());
+                    self.broadcast_item(InventoryItem::Block(block));
```
tmpolaczyk, line 426: Could add a comment:
```suggestion
                    // If the node is not synced, broadcast candidates without validating them
                    self.broadcast_item(InventoryItem::Block(block));
```

tmpolaczyk (approved):

PR #1577, opened 2020-09-22
refactor(node): store Arc<Config> in Session actor

tmpolaczyk:


aesedepece:
Next step would probably be removing the config manager altogether. I wonder whether we can simply read the config as we start the node and simply send `Arc<Config>` to every other actor as we create them, or at least to every non-supervised one for which we made the assumption that they live for as long as the node is running.

aesedepece (approved):

mariocao (approved):

PR #1581, opened 2020-09-22
get_superblock_votes return valid votes

### LINKED ISSUE #1563: GetSuperBlockVotes message returns all the votes, including invalid ones
The `GetSuperBlockVotes` message is used to obtain the votes for the current superblock that should be sent to new peers. But it currently sends all the seen votes, even invalid ones:

https://github.com/witnet/witnet-rust/blob/dd9ab39d7bebf43f1e128059b032bb61e2e6b1a3/data_structures/src/superblock.rs#L453-L455

To fix this, we can create a function similar to `get_valid_votes` but that new function should return a `HashSet<SuperBlockVote>` or a `Vec<SuperBlockVote>`:

https://github.com/witnet/witnet-rust/blob/dd9ab39d7bebf43f1e128059b032bb61e2e6b1a3/data_structures/src/superblock.rs#L152-L154

Then we can simply do

```rust
self.votes_mempool.get_valid_votes() 
```

And add tests to check that `get_current_superblock_votes` does not return votes from members not in the signing committee.

clbartoli:
Before this PR, the function `get_current_superblock_votes` returned all the votes, including the invalid ones.
Now the function returns only the valid vote.
The PR adds a test to the function.

Closes #1563

### DIFF
--- a/data_structures/src/superblock.rs
+++ b/data_structures/src/superblock.rs
@@ -144,11 +144,6 @@ impl SuperBlockVotesMempool {
         }
     }
 
-    fn get_received_votes(&self) -> HashSet<SuperBlockVote> {
-        self.received_votes.clone()
-    }
-
-    #[cfg(test)]
     fn get_valid_votes(&self) -> HashMap<Hash, Vec<SuperBlockVote>> {
         self.votes_on_each_superblock.clone()
     }
@@ -458,9 +453,14 @@ impl SuperBlockState {
         self.current_superblock_beacon
     }
 
-    /// Returns the current superblock votes.
+    /// Returns the current valid superblock votes.
     pub fn get_current_superblock_votes(&self) -> HashSet<SuperBlockVote> {
-        self.votes_mempool.get_received_votes()
+        self.votes_mempool
+            .get_valid_votes()
+            .values()
+            .flatten()
+            .cloned()
+            .collect()
     }
 
     #[allow(clippy::cast_possible_truncation)]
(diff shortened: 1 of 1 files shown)


data_structures/src/superblock.rs
```diff
@@ -1751,6 +1751,91 @@ mod tests {
...
+        v3.secp256k1_signature.public_key = p2;
+        assert_eq!(sbs.add_vote(&v3, 1), AddSuperBlockVote::ValidWithSameHash);
+        // If we change the index of the vote to x+1 it should be set as MaybeValid
+        let mut v4 = SuperBlockVote::new_unsigned(sb2_hash, 2);
+        v4.secp256k1_signature.public_key = p4;
+        assert_eq!(sbs.add_vote(&v4, 1), AddSuperBlockVote::MaybeValid);
+
+        // The function get_current_superblock_votes should return only the vote v3
+        // and exclude the rest of the votes
+        assert_eq!(sbs.get_current_superblock_votes().contains(&v1), false);
```
tmpolaczyk, line 1833: In some other tests there is an example on how to use `assert_eq` with a `HashSet`, something like:

```suggestion
        assert_eq!(sbs.get_current_superblock_votes(), vec![v3].into_iter().collect());
```

data_structures/src/superblock.rs
```diff
@@ -458,9 +453,14 @@ impl SuperBlockState {
         self.current_superblock_beacon
     }
 
-    /// Returns the current superblock votes.
+    /// Returns the current  valid superblock votes.
```
girazoki, line 456:
```suggestion
    /// Returns the current valid superblock votes.
```

tmpolaczyk (approved):

girazoki (approved):

PR #1582, opened 2020-09-22
feat(node): delay commit and reveal processing

### LINKED ISSUE #1569: Reveal transactions may be ignored if received before the epoch notification
We can sometimes see this warning in the logs:

```
[2020-09-18T13:27:45Z WARN  witnet_node::actors::chain_manager] Error when validating transaction 77dd262e5576ad71531f9dc56ef5e0eac7c750a7a015995d49c69d3f759af9f6: Received a reveal and Data Request is not in Reveal stage
[2020-09-18T13:27:45Z DEBUG witnet_node::actors::epoch_manager] Current epoch Ok(57). Last checked epoch Some(56)
[2020-09-18T13:27:45Z INFO  witnet_node::actors::epoch_manager] [Checkpoints] We are now in epoch #57
```

This is a clear race condition, and reveals are cleared on every epoch so it is possible that the received reveals are not included in a block.

The method `TransactionsPool::insert` should have some logic to detect transactions that are not valid when received, but may be valid on the next epoch, and process that transactions after consolidating a new block.

lrubiorod:
Close #1569

### DIFF
--- a/node/src/actors/chain_manager/handlers.rs
+++ b/node/src/actors/chain_manager/handlers.rs
@@ -210,6 +210,12 @@ impl Handler<EpochNotification<EveryEpochPayload>> for ChainManager {
             }
         }
 
+        // After block consolidation, commits and reveals that arrive in an incorrect moment
+        // are processed now
+        for transaction in self.temp_commits_and_reveals.drain(..) {
+            ctx.notify(AddTransaction { transaction });
+        }
+
         self.peers_beacons_received = false;
     }
 }
--- a/node/src/actors/chain_manager/mod.rs
+++ b/node/src/actors/chain_manager/mod.rs
@@ -223,6 +223,8 @@ pub struct ChainManager {
     external_percentage: u8,
     /// List of superblock votes received while we are synchronizing
     temp_superblock_votes: Vec<SuperBlockVote>,
+    /// Commits and reveals to process later
+    temp_commits_and_reveals: Vec<Transaction>,
 }
 
 /// Wrapper around a block candidate that contains additional metadata regarding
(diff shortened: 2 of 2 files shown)


node/src/actors/chain_manager/mod.rs
```diff
@@ -956,15 +958,19 @@ impl ChainManager {
...
-
-                if timestamp_now > timestamp_mining {
-                    let e = ChainManagerError::TooLateToCommit;
-                    log::debug!("{}", e);
-                    return Box::new(actix::fut::err(e.into()));
+            let timestamp_mining = epoch_constants
+                .block_mining_timestamp(current_epoch)
+                .unwrap();
+
+            if timestamp_now > timestamp_mining {
```
tmpolaczyk, line 965: Race condition: this can happen before the mining starts, is that a problem?

lrubiorod: not really, worst case a commit that arrive in the limit of block creation that could be included is not

node/src/actors/chain_manager/handlers.rs
```diff
@@ -210,6 +210,15 @@ impl Handler<EpochNotification<EveryEpochPayload>> for ChainManager {
             }
         }
 
+        // After block consolidation, commits and reveals that arrive in an incorrect moment
+        // are processed now
+        let timestamp_now = get_timestamp();
+        let delayed_transactions: Vec<Transaction> =
+            self.temp_commits_and_reveals.drain(..).collect();
+        for transaction in delayed_transactions {
+            let _aux = self.add_transaction(AddTransaction { transaction }, timestamp_now);
```
girazoki, line 219: is _aux needed?

lrubiorod: clippy requirements

tmpolaczyk: `add_transaction` returns a future, so it will not be executed... Try `ctx.notify(AddTransaction` instead

lrubiorod: done!

node/src/actors/chain_manager/mod.rs
```diff
@@ -956,15 +958,19 @@ impl ChainManager {
...
+                if let Transaction::Commit(_) = &msg.transaction {
```
tmpolaczyk, line 966: Does this compile?

```suggestion
                if let Transaction::Commit(_) | Transaction::Reveal(_) = &msg.transaction {
```

tmpolaczyk (approved):

girazoki (approved):

PR #1583, opened 2020-09-23
WIP store utxo set in the database and not in ram

### LINKED ISSUE #1275: Keep UTXO set in storage

tmpolaczyk:
This PR stores consolidated entries of the UTXO set in the storage, and the non-consolidated entries in RAM. Consolidated means included in a consolidated superblock.

Close #1275

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4638,6 +4638,7 @@ name = "witnet_data_structures"
 version = "0.3.2"
 dependencies = [
  "bech32",
+ "bincode",
  "bls-signatures-rs",
  "byteorder",
  "cbor-codec",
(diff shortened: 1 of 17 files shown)


tmpolaczyk:
Unresolved problem: migrations. Currently the node stores the chain state using the magic number from consensus constants as the key. In this PR, UTXOs are stored without any magic number. This means that the storage cannot be reused by the same node on different testnets, because the old UTXOs will still be marked as valid. So we need to either add the magic number to the UTXO key, as `format!("UTXO-{}-{}", magic, output_pointer)`, or find some way to detect if the environment has changed and delete all the existing UTXOs from the storage (this should now be possible thanks to the iteration support in storage).

tmpolaczyk:
Continued in #2159

PR #1584, opened 2020-09-24
fix(storage): fix delete operation on encrypted backend

tmpolaczyk:
This only affects the wallet, and we never tried to delete anything there, so this issue shouldn't have affected us


mariocao (approved):
Good catch!  :fireworks: 
It seems nobody was using the `delete` function! :rofl:

girazoki (approved):

Issue #1075, opened 2020-03-12
Make Config serializable and implement ShowConfig command


lrubiorod:
https://github.com/witnet/witnet-rust/blob/master/src/cli/node/with_node.rs#L54

tmpolaczyk:
I tried to implement this but got an error:

    Error: values must be emitted before tables

I think the solution is explained here:

https://github.com/alexcrichton/toml-rs/blob/7a43ac9e9c28afec8414f6d924024f5d33fe8fa2/src/ser.rs#L7

twjang:
Hello! I implemented this feature and created a PR. Could you review https://github.com/witnet/witnet-rust/pull/1591 ?

PR #1591, opened 2020-09-24
feat: implemented 'witnet node config' command

twjang:
Hello!

I implemented show-config command. I made the Config struct be able to be converted into PartialConfig, so that it can be serialized into toml format.

The related issue is #1075

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -47,6 +47,7 @@ use crate::{
     dirs,
 };
 use partial_struct::PartialStruct;
+use std::convert::TryFrom;
 use witnet_crypto::hash::HashFunction;
 use witnet_data_structures::chain::{ConsensusConstants, Environment, PartialConsensusConstants};
 use witnet_protected::{Protected, ProtectedString};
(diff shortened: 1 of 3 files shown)


config/src/config.rs
```diff
@@ -303,6 +360,13 @@ pub struct Storage {
     pub master_key_import_path: Option<PathBuf>,
 }
 
+fn to_protected_string<S>(_val: &Option<Protected>, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: serde::Serializer,
+{
+    serializer.serialize_str("**** REDACTED ****")
```
tmpolaczyk, line 367: This should check if the password is `None`, because in that case we must keep it as `None`

twjang: I modified this part as you suggested in the latest commit 2a6fc17. Could you review new code? Thanks in advance.

config/src/config.rs
```diff
@@ -245,6 +278,17 @@ pub struct Connections {
     pub reject_sybil_inbounds: bool,
 }
 
+fn to_millis<S>(val: &Option<Duration>, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: serde::Serializer,
+{
+    if let Some(duration) = val {
+        serializer.serialize_u64(u64::try_from(duration.as_millis()).unwrap())
```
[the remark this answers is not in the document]
twjang: I changed `serde::de::Error::custom` to `serde::ser::Error::custom` to fix compile error in the latest commit.

tmpolaczyk:
Ok, good! Before merging, can you squash all the commits into one? You can do that by running `git rebase -i upstream/master` and then replace "pick" with "f" for all lines except the first one.

Thank you!

tmpolaczyk (approved):

PR #1593, opened 2020-09-25
fix(validations): modify consensus percentage to use commits number i…

### LINKED ISSUE #1307: Ensure min_consensus_percentage is achieved
Ensure that  min_consensus_percentage is calculated with the commits number.

Currently I think is calculated with reveals number so... it implies if there are too much non revealers, the result could not achieve the min_consensus_percentage specified for the witnesses required

lrubiorod:
…nstead of reveals

Close #1307


tmpolaczyk:
Can you add a test that returns `InsufficientConsensus` now but would have been a valid result before this change? Something like 4 witnesses but only 1 reveal.

tmpolaczyk (approved):

mariocao (approved):

PR #1594, opened 2020-09-25
feat(wallet): create local transaction movements

mariocao:


wallet/src/actors/app/methods.rs
```diff
@@ -634,7 +636,17 @@ impl App {
...
                 .get_wallet_by_session_and_id(&session_id, &wallet_id),
         )
-        .and_then(move |_wallet, slf: &mut Self, _| slf.send_inventory_transaction(transaction));
+        .and_then(move |wallet, slf: &mut Self, _| {
+            slf.send_inventory_transaction(transaction.clone())
+                .and_then(move |value, _slf, _ctx| {
+                    let _ = wallet.add_local_movement(&model::ExtendedTransaction {
+                        transaction,
+                        metadata: None,
+                    });
```
tmpolaczyk, line 645: This is ignoring any errors returned by `add_local_movement`. Try this to return the errors:
```suggestion
                    if let Some(err) = wallet.add_local_movement(&model::ExtendedTransaction {
                        transaction,
                        metadata: None,
                    }) {
                        return actix::fut::err(err);
                    }
```

wallet/src/repository/wallet/mod.rs
```diff
@@ -313,23 +319,22 @@ where
         // Query database `transaction_next_id` to compute total amount of transactions
         let db_total = self
             .db
-            .get_or_default::<_, u32>(&keys::transaction_next_id(account))?
-            .saturating_sub(1);
+            .get_or_default::<_, u32>(&keys::transaction_next_id(account))?;
```
tmpolaczyk, line 322: Why remove the `.saturating_sub(1)`?

wallet/src/model.rs
```diff
@@ -248,7 +248,7 @@ pub struct KeyBalance {
     pub amount: u64,
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[derive(Debug, Default, Clone, Serialize, Deserialize, PartialEq)]
```
girazoki, line 251: order

wallet/src/repository/wallet/mod.rs
```diff
@@ -1061,12 +1075,44 @@ where
             (input_amount - output_amount, model::MovementType::Negative)
         };
 
-        Ok(AccountMutation {
+        // If UTXO set has not changed, then there is no balance movement derived from the transaction being processed
+        if utxo_inserts.is_empty() && utxo_removals.is_empty() {
+            return Ok(None);
```
tmpolaczyk, line 1080: What kinds of transactions can enter this branch?

wallet/src/repository/wallet/mod.rs
```diff
@@ -302,8 +304,12 @@ where
         let account = state.account;
 
         // Total amount of state and db transactions
-        let total = state.transaction_next_id;
-        let mut transactions = Vec::with_capacity(total as usize);
+        let total = state.transaction_next_id + u32::try_from(state.local_movements.len()).unwrap();
+        let mut transactions: Vec<BalanceMovement> = Vec::with_capacity(100);
```
girazoki, line 308: why 100?

wallet/src/repository/wallet/mod.rs
```diff
@@ -1061,12 +1075,44 @@ where
             (input_amount - output_amount, model::MovementType::Negative)
         };
 
-        Ok(AccountMutation {
+        // If UTXO set has not changed, then there is no balance movement derived from the transaction being processed
+        if utxo_inserts.is_empty() && utxo_removals.is_empty() {
```
girazoki, line 1079: We can do this a few lines up to avoid a couple of substractions right?

wallet/src/repository/wallet/mod.rs
lrubiorod: It was by the error in the indexes of the wallet

wallet/src/repository/wallet/mod.rs
mariocao: The normal ones, the ones the wallet is indexing :)

mariocao: Is a good question. The wallet is suppose to filter transactions beforehand. It does not harm anyway! :rofl:

wallet/src/repository/wallet/mod.rs
```diff
@@ -6,14 +6,16 @@ use std::{
...
+use witnet_data_structures::{
+    chain::{CheckpointBeacon, Environment, Epoch, EpochConstants, PublicKeyHash},
+    get_environment,
 };
-use witnet_data_structures::get_environment;
+use witnet_util::timestamp::get_timestamp;
 
 use crate::constants::{EXTERNAL_KEYCHAIN, INTERNAL_KEYCHAIN};
-use crate::model::AddressInfo;
+use crate::model::{AddressInfo, BalanceMovement};
```
lrubiorod, line 18: fmt

wallet/src/repository/wallet/mod.rs
tmpolaczyk: Then can we add tests so it doesn't happen again? :P

wallet/src/repository/wallet/mod.rs
```diff
@@ -302,8 +304,12 @@ where
         let account = state.account;
 
         // Total amount of state and db transactions
-        let total = state.transaction_next_id;
-        let mut transactions = Vec::with_capacity(total as usize);
+        let total = state.transaction_next_id + u32::try_from(state.local_movements.len()).unwrap();
+        let mut transactions: Vec<BalanceMovement> = Vec::with_capacity(total as usize);
```
tmpolaczyk, line 308: It looks like total should be `usize`:

```suggestion
        let total = usize::try_from(state.transaction_next_id).unwrap() + state.local_movements.len();
        let mut transactions: Vec<BalanceMovement> = Vec::with_capacity(total);
```

lrubiorod, line 308: should be easier cast transaction_next_id to usize instead of local_movements to u32 and later to usize?

girazoki (approved):

wallet/src/actors/app/methods.rs
```diff
@@ -624,17 +624,37 @@ impl App {
...
 
+    /// Send a transaction to the node as inventory item broadcast
+    /// and add a local pending balance movement to the wallet state.
     pub fn send_transaction(
         &self,
         session_id: &types::SessionId,
         wallet_id: &str,
         transaction: types::Transaction,
-    ) -> ResponseActFuture<serde_json::Value> {
+    ) -> ResponseActFuture<SendTransactionResponse> {
```
tmpolaczyk, line 634: Does this change break the public API? (sheikah)

wallet/src/repository/wallet/mod.rs
mariocao: image

:rofl:

wallet/src/repository/wallet/mod.rs
```diff
@@ -313,23 +319,22 @@ where
         // Query database `transaction_next_id` to compute total amount of transactions
         let db_total = self
             .db
-            .get_or_default::<_, u32>(&keys::transaction_next_id(account))?
-            .saturating_sub(1);
+            .get_or_default::<_, u32>(&keys::transaction_next_id(account))?;
 
         if db_total > 0 {
```
tmpolaczyk, line 324:
```suggestion
```

You could remove this if because when `db_total == 0` the only possible range is `0..0`, and that's a valid range with 0 elements.

wallet/src/repository/wallet/mod.rs
tmpolaczyk: Fine, I'll do it myself

https://github.com/tmpolaczyk/witnet-rust/tree/wallet-get-transactions-tests

wallet/src/actors/app/methods.rs
mariocao: Yes, but I think nobody was using it...the response was the json rpc value of sendinng an InventoryItem

wallet/src/repository/wallet/mod.rs
mariocao: It used later as a `Transaction` field of type `u32`

wallet/src/actors/app/methods.rs
tmpolaczyk: Yes, the response was the transaction hash. Now it's an object that contains the transaction hash and a `BalanceMovement`, which is some data structure that's internal to the wallet. I think it would be better to keep this method as before. Can we just map the result like `.map(|result| result.jsonrpc_result)` ?

mariocao: They can use the `BalanceMovement` object to draw a new pending transaction in their transaction list. :smiley:

tmpolaczyk: Ok, then leave it like this.

tmpolaczyk (approved):

Issue #792, opened 2019-08-30
Unify DataRequestReport and DataRequestInfo


lrubiorod:
We have 2 very similar structures, `DataRequestInfo` and `DataRequestReport`. It could be better to only have one of them

aesedepece:
Can you illustrate how similar are they?

drcpu-github:
Volunteered for this refactor via Trello.

I took a look at the code and I think it makes most sense to completely remove DataRequestReport and replace it everywhere with DataRequestInfo as the latter is basically a superset of fields (with some extra indexing etc). Any other remarks and / or things I should look out for?

drcpu-github:
@lrubiorod Any insights on above comment?

PR #1601, opened 2020-09-29
refactor: remove DataRequestReport in favor of DataRequestInfo

drcpu-github:
Replaced DataRequestReport with DataRequestInfo everywhere since the latter is basically an encompassing struct containing all data used by DataRequestReport. Clippy, FMT and all tests (cargo test --all --verbose) pass or are ignored. Let me know any comments you have.

This PR tackles issue #792.

### DIFF
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -2390,52 +2390,6 @@ pub enum InventoryItem {
     SuperBlock(SuperBlock),
 }
 
-/// Data request report to be persisted into Storage and
-/// using as index the Data Request OutputPointer
-// FIXME (#792): Review if this struct is needed
-// It is not needed, we just need to store the transaction hash for all the commits, reveals, and
-// tally. All the information can then be retrieved from the database. The data request transaction
-// hash is used as the key.
-#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
-pub struct DataRequestReport {
-    /// List of commitment output pointers to resolve the data request
-    pub commits: Vec<CommitTransaction>,
-    /// List of reveal output pointers to the commitments (contains the data request result of the witnet)
-    pub reveals: Vec<RevealTransaction>,
-    /// Tally output pointer (contains final result)
-    pub tally: TallyTransaction,
-    /// Hash of the block with the DataRequestTransaction
-    pub block_hash_dr_tx: Hash,
-    /// Hash of the block with the TallyTransaction
-    pub block_hash_tally_tx: Hash,
-    /// Current commit round starting from 1
-    pub current_commit_round: u16,
-    /// Current reveal round starting from 1
-    pub current_reveal_round: u16,
-}
-
-impl TryFrom<DataRequestInfo> for DataRequestReport {
-    type Error = failure::Error;
-
-    fn try_from(x: DataRequestInfo) -> Result<Self, failure::Error> {
-        if let (Some(tally), Some(block_hash_dr_tx), Some(block_hash_tally_tx)) =
-            (x.tally, x.block_hash_dr_tx, x.block_hash_tally_tx)
-        {
-            Ok(DataRequestReport {
-                commits: x.commits.values().cloned().collect(),
-                reveals: x.reveals.values().cloned().collect(),
-                tally,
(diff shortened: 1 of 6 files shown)


data_structures/src/chain.rs
```diff
@@ -2566,15 +2501,12 @@ impl DataRequestState {
         mut self,
         tally: TallyTransaction,
         block_hash_tally_tx: &Hash,
-    ) -> Result<DataRequestReport, failure::Error> {
+    ) -> Result<DataRequestInfo, failure::Error> {
         if let DataRequestStage::TALLY = self.stage {
             self.info.tally = Some(tally);
             self.info.block_hash_tally_tx = Some(*block_hash_tally_tx);
```
tmpolaczyk, line 2507: `data_request_into.current_stage` should be set to `None` when the data request is finished, now it stays at "TALLY".

```suggestion
            self.info.block_hash_tally_tx = Some(*block_hash_tally_tx);
            self.info.stage = None;
```

node/src/actors/chain_manager/handlers.rs
```diff
@@ -1270,9 +1270,9 @@ impl Handler<GetDataRequestReport> for ChainManager {
         } else {
             let dr_pointer_string = format!("DR-REPORT-{}", dr_pointer);
             // Otherwise, try to get it from storage
-            let fut = storage_mngr::get::<_, DataRequestReport>(&dr_pointer_string).and_then(
-                move |dr_report| match dr_report {
-                    Some(x) => futures::finished(DataRequestInfo::from(x)),
+            let fut = storage_mngr::get::<_, DataRequestInfo>(&dr_pointer_string).and_then(
```
tmpolaczyk, line 1273: Changing the type of values from the storage is a breaking change, so any nodes that update to this version will have to delete the storage and synchronize again. We technically could write migrations that convert old entries to the new format, but let's just break it.

node/src/actors/json_rpc/json_rpc_methods.rs
```diff
@@ -65,7 +65,7 @@ pub fn jsonrpc_io_handler(
     //io.add_method("getOutput", |params: Params| get_output(params.parse()));
     io.add_method("syncStatus", |_params: Params| status());
     io.add_method("dataRequestReport", |params: Params| {
-        data_request_report(params.parse())
+        data_request_info(params.parse())
```
tmpolaczyk, line 68: No need to rename this method

src/cli/node/json_rpc_client.rs
```diff
@@ -776,7 +776,7 @@ impl fmt::Display for DataRequestTransactionInfo {
     }
 }
 
-pub fn data_request_report(
+pub fn data_request_info(
```
tmpolaczyk, line 779: This also doesn't need to be renamed

node/src/actors/json_rpc/json_rpc_methods.rs
drcpu-github: Agreed. I wanted to make it clear that it changed, but that may not have been the best approach.

data_structures/src/chain.rs
drcpu-github: I assume you mean setting `self.info.current_stage = None`

node/src/actors/chain_manager/handlers.rs
drcpu-github: I did remove the TryFrom conversion to convert DataRequestReport into DataRequestInfo as that was not needed anymore if I replaced it completely. But, yes, I'd say that if this is merged into master, it is for a completely new release (be it a new testnet or mainnet), so people will resynchronize anyway.

node/src/actors/json_rpc/json_rpc_methods.rs
tmpolaczyk: Actually, since the output of the command is exactly the same you could say that it hasn't changed, which is great for users.

drcpu-github:
Applied the required changes. I will squash the commits (and possibly rebase with master) once it has been approved for merging, just left them separately for now for easier comparison.

tmpolaczyk (approved):
Ok, thank you!

Issue #1145, opened 2020-04-02
[witnet_rad] Parse integer should ignore line breaks


mariocao:
## Summary

Parsing an Integer fails if the return result of an URL includes a line break, i.e. the result contains one extra byte that wasn't expected.

## Example of failing code

Data request source:
```
const source = new Witnet.Source("https://www.random.org/integers/?num=1&min=1&max=100&format=plain&rnd=new&col=1&base=10").asInteger()
```

Data request bytecode:

```
0a800108f0e597f405125e125768747470733a2f2f7777772e72616e646f6d2e6f72672f696e7465676572732f3f6e756d3d31266d696e3d31266d61783d31303026666f726d61743d706c61696e26726e643d6e657726636f6c3d3126626173653d31301a038118731a0d0a0908051205fa3fc00000100322090a0508051201011003100a18042001280130013801400248055046
```

## Error

After doing a `send-request --run --hex` with the node as:
```
./witnet node send-request --run --hex 0a800108e7cc97f405125e125768747470733a2f2f7777772e72616e646f6d2e6f72672f696e7465676572732f3f6e756d3d31266d696e3d31266d61783d31303026666f726d61743d706c61696e26726e643d6e657726636f6c3d3126626173653d31301a038118731a0d0a0908051205fa3fc00000100322090a0508051201011003100a18042001280130013801400248055046
```

I get the following error:

```
[2020-04-02T14:25:20Z INFO  witnet::cli::node::json_rpc_client] Running retrieval for https://www.random.org/integers/?num=1&min=1&max=100&format=plain&rnd=new&col=1&base=10
[2020-04-02T14:25:20Z WARN  witnet_rad::types] `RadError` cannot be converted to `RadonError` but it should, because it is needed for the tally result. Message: None. Inner: `Some(ParseInt { message: "invalid digit found in string" })`
[2020-04-02T14:25:20Z DEBUG witnet_rad] Result for URL https://www.random.org/integers/?num=1&min=1&max=100&format=plain&rnd=new&col=1&base=10: Ok(RadonReport { metadata: Retrieval, result: RadonError(RadonError { inner: UnhandledIntercept { inner: Some(ParseInt { message: "invalid digit found in string" }), message: None } }), running_time: 44.909µs })
```

tmpolaczyk:
A simple workaround would be to `String.replace("\n", "")`

PR #1605, opened 2020-09-29
feat(log): show current commit stage and current reveal stage

drcpu-github:
This PR (should) solve(s) issue #1005.

I split logging format string for COMMIT and REVEAL (taking the appropriate current round) on the one hand and TALLY on the other hand. Issue #1005 hints to use `extra_commit_rounds`, but that parameter does not exist anymore, so instead I took the `extra_rounds` parameter from the DataRequestPool struct.

I wanted to test whether the logging actually prints the correct messages, but I have not minted any blocks yet, so I am too poor to launch data requests and no one else is launching any either.

PR #1606, opened 2020-09-30
feat(witnet_rad): ignore line breaks at the end of a float, int or bool when parsing from string

drcpu-github:
Implemented ignoring a line break when parsing float / int or bool tackling issue #1145 

I chose to truncate the string with one character if necessary rather than "blindly" replacing newlines because that is more strict. Even though I cannot think of anything exploiting replacing all newlines, I'd say stricter is always better.

Even though the issue only mentions integer parsing, the same goes for floats and bools, so I also added the code there. I guess I could abstract the truncating code to a separate function but since there are only three occurences, that would be pretty much the same overhead in terms of lines of code. Would you still prefer putting it into a separate function (e.g., for maintainability)?

### DIFF
--- a/rad/src/operators/string.rs
+++ b/rad/src/operators/string.rs
@@ -35,20 +35,31 @@ pub fn parse_json_array(input: &RadonString) -> Result<RadonArray, RadError> {
     item.try_into()
 }
 
+pub fn radon_trim(input: &RadonString) -> String {
+    if input.value().ends_with('\n') {
+        input.value()[..input.value().len() - 1].to_string()
+    } else {
+        input.value()
+    }
+}
+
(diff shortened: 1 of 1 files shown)


tmpolaczyk:
> Would you still prefer putting it into a separate function (e.g., for maintainability)?

Yes please! This way we can easily change it in the future if for example we decide to also strip trailing spaces.

```rust
let str_value = radon_trim(input.value());
```

drcpu-github:
Because you asked so nicely, I made a separate function out of it. :wink:

I will squash the commits and rebase master into it once it has been approved (and PR #1605 is merged into master).

lrubiorod (approved):

Issue #1133, opened 2020-03-30
Error when parsing floats with commas


girazoki:
Source: [https://www.chards.co.uk/data/update-prices?full=false&currency=eur]

Script:
```js
const chards = new Witnet.Source("https://www.chards.co.uk/data/update-prices?full=false&currency=eur")
  .parseMapJSON()
  .getMap("gold")
  .getFloat("gram")
  .multiply(1000)
  .round()
```

Error:
> `RadError` cannot be converted to `RadonError` but it should, because it is needed for the tally result. Message: None. Inner: `Some(ParseFloat { message: "invalid float literal" })`

tmpolaczyk:
That's not supported, a simple workaround would be to `String.replace(",", "")`

aesedepece:
> That's not supported, a simple workaround would be to `String.replace(",", "")`

If we only had a `.replace` RADON method...

drcpu-github:
@aesedepece Would it make sense to have a getFloat RADON function that allows to specify thousand and decimal separators for greater parsing flexibility? Example `getFloat("gram", thousand=" ", decimal=",")` for a number `111 111,111`.

Tommytrg:
Could we create a WIP for this? @tmpolaczyk @lrubiorod

tmpolaczyk:
We could

aesedepece:
Yes, we could

aesedepece:
This was solved by #2235, now about to be released inside 1.6 as per [WIP-0024](https://github.com/witnet/WIPs/blob/master/wip-0024.md).

PR #1610, opened 2020-10-01
feat(witnet_rad): Allow specification of thousand and decimal separators in get_integer and get_float

drcpu-github:
This is more of a proposal on how to solve issue #1133, rather than a finalized PR.

What I would do is allow a data requester to specify the format of the numeric type he wants to fetch by defining the thousand and decimal separator. This would allow fetching data from "weird" API's that follow, e.g., the European convention of adding spaces or dots between groups of thousand multiples or a comma as a decimal separator.

The getFloat or getInt functions are (apparently) extensible by default to allow extra arguments, so the data request from issue #1133 could be extended as follows:
```
import * as Witnet from "witnet-requests"

const random = new Witnet.Source("https://www.chards.co.uk/data/update-prices?full=false&currency=eur")
    .parseJSONMap()
    .getMap("gold")
    .getFloat("kilogram", ",", ".") // First argument is the map key, second the thousand separator and third the decimal separator
    .multiply(1000)
    .round()

const aggregator = new Witnet.Aggregator({
    filters: [],
    reducer: Witnet.Types.REDUCERS.averageMean
})

const tally = new Witnet.Tally({
    filters: [],
    reducer: Witnet.Types.REDUCERS.averageMean
})

const request = new Witnet.Request()
    .addSource(random) // Use source 1
    .setAggregator(aggregator) // Set the aggregation script
    .setTally(tally) // Set the tally script
    .setQuorum(4, 70) // Set witness count
    .setFees(10, 1) // Set economic incentives
    .schedule(0) // Make this request immediately solvable
// Do not forget to export the request object
export { request as default }
```

This yields following hex string:
`0a7a08a199d2fb05126a124368747470733a2f2f7777772e6368617264732e636f2e756b2f646174612f7570646174652d7072696365733f66756c6c3d66616c73652663757272656e63793d6575721a2385187782186664676f6c64841864686b696c6f6772616d612c612e8218571903e8185b1a02100322021003100a180420012846308094ebdc03`

Losing the separator definitions and replace kilogram by gram (for a valid float value), yields following hex string:
`0a720896efd6fb051262124368747470733a2f2f7777772e6368617264732e636f2e756b2f646174612f7570646174652d7072696365733f66756c6c3d66616c73652663757272656e63793d6575721a1b85187782186664676f6c64821864646772616d8218571903e8185b1a02100322021003100a180420012846308094ebdc03`

Both of these still work, but the extra separator arguments allow for greater flexibility on the data requester side in dealing with "weird" API data.

I am obviously open to any suggestions (also, there is a significant possibility I sinned against more than one Rust conversion convention).

### DIFF
--- a/rad/src/operators/map.rs
+++ b/rad/src/operators/map.rs
@@ -9,6 +9,22 @@ use crate::{
     },
 };
 
+pub fn replace_separators(
+    value: RadonTypes,
+    thousand_separator: serde_cbor::Value,
+    decimal_separator: serde_cbor::Value,
+) -> RadonTypes {
+    let rad_str_value: RadonString = value.try_into().unwrap();
+    let thousand = from_value::<String>(thousand_separator).unwrap_or_else(|_| "".to_string());
+    let decimal = from_value::<String>(decimal_separator).unwrap_or_else(|_| ".".to_string());
+    RadonTypes::from(RadonString::from(
+        rad_str_value
+            .value()
+            .replace(&thousand, "")
+            .replace(&decimal, "."),
+    ))
+}
+
 pub fn get(input: &RadonMap, args: &[Value]) -> Result<RadonTypes, RadError> {
     let wrong_args = || RadError::WrongArguments {
         input_type: RadonMap::radon_type_name(),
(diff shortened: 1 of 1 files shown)


tmpolaczyk:
Now this is only implemented for `RadonMap::getFloat` and `RadonMap::getInteger`, for consistency we would also need to implement it for `RadonArray`, and also for `String::asInteger` and `String::asFloat`.

But let's discuss first. I'm in favor of simply adding a `RadonString::replace` operator and letting users implement their custom conversions. Because for example, how do you convert strings like `$2.123` or `412,33 €`? So for me the initial example would be written as:

```
const random = new Witnet.Source("https://www.chards.co.uk/data/update-prices?full=false&currency=eur")
    .parseJSONMap()
    .getMap("gold")
    .replace(",", "")
    .getFloat("kilogram")
    .multiply(1000)
    .round()
```

drcpu-github:
I was not convinced @aesedepece remark in the initial issue was a serious suggestion. I chose to implement it like this because it allows for flexibility in defining the number format without giving the data requester control over the actual replacement. If you want to support arbitrary conversions, then yes, you need a replace Radon operator.

I am wondering if this can be misused? Could you, for example, generate an idiotically large data request by using the replace operator to generate a very large string?

tmpolaczyk:
@drcpu-github string replace can be misused, that's probably the reason why it doesn't exist. For example:

```
    .replace("", "AAAA")
    .replace("", "AAAA")
    .replace("", "AAAA")
    .replace("", "AAAA")
```

Will turn an empty string into 624 "A"s

drcpu-github:
Yes, that is basically what I meant. So do we want to implement a generic Radon replace operator or is it better to stay with a more restricted version? Obviously, we could also implement a replace operator that puts restrictions on the replacements (e.g., the first string needs to be at least x characters and the second can be at most y characters)?

aesedepece:
Superseded by #1784, in a backwards compatible way.

PR #1612, opened 2020-10-01
feat(reputation): update active identities acording to block holes

lrubiorod:


tmpolaczyk:
https://github.com/witnet/witnet-rust/blob/1fa31af5125b98bdea5faa955c9244be588db419/reputation/src/ars.rs#L169

This is the `update_empty` function that we used before. Maybe we can add it back?

reputation/src/ars.rs
```diff
@@ -114,12 +114,6 @@ where
         M: IntoIterator<Item = K>,
     {
         let identities: HashSet<K> = identities.into_iter().collect();
-        // Push activity with no entries is not allowed
-        // Note: at least the node's miner activity should be pushed
-        assert!(
```
tmpolaczyk, line 119: Maybe we can move this assert to the `update` function?

lrubiorod:
> https://github.com/witnet/witnet-rust/blob/1fa31af5125b98bdea5faa955c9244be588db419/reputation/src/ars.rs#L169
> 
> This is the `update_empty` function that we used before. Maybe we can add it back?

In the past, we use in different places. Now it is only used in one place, so... I think it is not necessary

reputation/src/ars.rs
lrubiorod: identities is a generic parameter so it hasnt "is_empty()" function

tmpolaczyk (approved):

mariocao (approved):

PR #1613, opened 2020-10-01
feat(update tally balance movements)

girazoki:
This PR udpates the wallet DR transacions with the metadata returned by tally.

### DIFF
--- a/wallet/src/repository/wallet/mod.rs
+++ b/wallet/src/repository/wallet/mod.rs
@@ -554,7 +554,9 @@ where
                     .db
                     .get::<_, model::Path>(&keys::transactions_index(tally.dr_pointer.as_ref()))
                     .is_ok()
-                    || state.pending_dr_movements.contains_key(&tally.dr_pointer)
+                    || state
+                        .pending_dr_movements
+                        .contains_key(&tally.dr_pointer.to_string())
                 {
                     filtered_txns.push(txn.clone());
                     continue;
@@ -604,7 +606,7 @@ where
                     Ok(Some((balance_movement, mut new_addresses))) => {
                         if let types::Transaction::DataRequest(dr_tx) = &txn.transaction {
                             dr_balance_movements.insert(
-                                dr_tx.hash(),
+                                dr_tx.hash().to_string(),
                                 (block_info.block_hash, block_balance_movements.len()),
                             );
                         }
@@ -623,8 +625,10 @@ where
             }
             if let types::Transaction::Tally(tally) = &txn.transaction {
                 // The DR transaction is in pending state
-                if let Some((pending_block_hash, index)) =
-                    state.pending_dr_movements.get(&tally.dr_pointer).cloned()
+                if let Some((pending_block_hash, index)) = state
+                    .pending_dr_movements
+                    .get(&tally.dr_pointer.to_string())
+                    .cloned()
                 {
                     let dr_movement = state
                         .pending_movements
@@ -636,7 +640,7 @@ where
                             let mut updated_dr_movement = dr_movement.clone();
                             updated_dr_movement.transaction.data = build_updated_dr_transaction_data(dr_data, tally, &txn.metadata)?;
                             state.pending_movements.get_mut(&pending_block_hash.to_string()).unwrap()[index] = updated_dr_movement;
-                            state.pending_dr_movements.remove(&tally.dr_pointer);
+                            state.pending_dr_movements.remove(&tally.dr_pointer.to_string());
                         }
                         _ => log::warn!("data request tally update failed because wrong transaction type (txn: {})", tally.dr_pointer),
                     }
@@ -660,7 +664,6 @@ where
                             dr_movement_to_update.transaction.data = build_updated_dr_transaction_data(dr_data, tally, &txn.metadata)?;
                             dr_movement_to_update.db_key = txn_id;
                             db_movements_to_update.push(dr_movement_to_update);
-                            state.pending_dr_movements.remove(&tally.dr_pointer);
                         }
                         _ => log::warn!("data request tally update failed because wrong transaction type (txn: {})", tally.dr_pointer),
                     }
@@ -698,7 +701,7 @@ where
             balance_movements_to_persist.extend_from_slice(&db_movements_to_update);
 
             self._persist_block_txns(
-                balance_movements_to_persist,
+                balance_movements_to_persist.clone(),
                 addresses,
                 state.transaction_next_id,
                 state.next_external_index,
@@ -706,7 +709,13 @@ where
                 state.utxo_set.clone(),
                 &state.balance.unconfirmed,
                 block_info,
-            )?
+            )?;
+
+            // Update pending DR movements if they were persisted
+            // balance_movements_to_persist.
+            balance_movements_to_persist.iter().for_each(|x| {
+                state.pending_dr_movements.remove(&x.transaction.hash);
+            });
         } else {
             for address in &addresses {
                 let path = address.path.clone();
--- a/wallet/src/repository/wallet/mod.rs
+++ b/wallet/src/repository/wallet/mod.rs
@@ -179,6 +179,7 @@ where
             pending_addresses_by_block: Default::default(),
             pending_addresses_by_path: Default::default(),
             pending_blocks: Default::default(),
+            pending_dr_movements: Default::default(),
         });
 
         Ok(Self {
@@ -495,6 +496,38 @@ where
                 _ => continue,
             };
 
+            // Check if tally txn corresponds to a wallet sent data request
+            if let types::Transaction::Tally(tally) = &txn {
+                if self
+                    .db
+                    .get::<_, model::Path>(&keys::transactions_index(tally.dr_pointer.as_ref()))
+                    .is_ok()
+                {
+                    filtered_txns.push(txn.clone());
+                    continue;
+                }
+                else if let Some(_dr_tx) = state.pending_dr_movements.get(&tally.dr_pointer) {
+                        filtered_txns.push(txn.clone());
+                        continue;
+                }
+                /*else {
+                    let push_needed = state.pending_blocks.iter().any(|&(block_hash, _)| {
+                        if let Some(balance_m{ovements) = state.pending_movements.get(block_hash) {
+                            balance_movements.iter().any(|&balance_movement| {
+                                balance_movement.transaction.hash == tally.dr_pointer.to_string()
+                            })
+                        }
+                        else {
+                            false
+                        }
+                    });
+                    if push_needed {
+                        filtered_txns.push(txn.clone());
+                        continue;
+                    }
+                }*/
+            }
+
             let check_db = |output: &types::VttOutput| {
                 self.db
                     .get::<_, model::Path>(&keys::pkh(&output.pkh))
@@ -523,6 +556,7 @@ where
         let mut state = self.state.write()?;
         let mut addresses = Vec::new();
         let mut balance_movements = Vec::new();
+        let mut dr_balance_movements = HashMap::new();
 
         // Index all transactions
         for txn in txns {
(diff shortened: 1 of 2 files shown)


wallet/src/repository/wallet/mod.rs
```diff
@@ -372,8 +364,24 @@ where
...
         }
 
+        // Append balance movements of pending blocks
+        state.pending_movements.values().for_each(|x| {
+            keyed_transactions.extend(x.iter().map(|movement| (movement.db_key, movement.clone())));
+        });
+
+        let total_txns = keyed_transactions.len();
+        // Append local pending balance movements (not yet included in blocks)
+        let local_movements = state.local_movements.values().cloned().collect::<Vec<_>>();
```
tmpolaczyk, line 374: If you are cloning the `movement` later, you don't need to clone it here. You don't even need to collect it because you are using `local_movements` as an iterator:

```suggestion
        let local_movements = state.local_movements.values();
```

wallet/src/repository/wallet/mod.rs
```diff
@@ -372,8 +364,24 @@ where
...
+            (
+                u32::try_from(total_txns + index)
+                    .expect("Transaction length should be convertible to u32"),
+                movement.clone(),
+            )
+        }));
+
         Ok(model::Transactions {
-            transactions,
+            transactions: keyed_transactions.values().cloned().rev().collect(),
```
tmpolaczyk, line 384: Here you don't need to clone the elements, you should be able to consume `keyed_transactions` using `into_iter`:

```suggestion
            transactions: keyed_transactions.into_iter().map(|(k, v)| v).rev().collect(),
```

wallet/src/repository/wallet/mod.rs
```diff
@@ -496,6 +504,24 @@ where
                 _ => continue,
             };
 
+            // Check if tally txn corresponds to a wallet sent data request
+            if let types::Transaction::Tally(tally) = &txn {
+                // There is a DR transaction persisted in database whose tally was found
+                if self
+                    .db
+                    .get::<_, model::Path>(&keys::transactions_index(tally.dr_pointer.as_ref()))
+                    .is_ok()
```
tmpolaczyk, line 513: I think `self.db.get` returns `Result<Option<T>>`, so it will always be ok. To check if some item exists in the database you should check if the result is `Ok(Some(_))`.

tmpolaczyk: Or maybe there is a `.contains`

wallet/src/repository/wallet/mod.rs
```diff
@@ -328,16 +329,7 @@ where
...
-
-        // Prepend local pending balance movements (not yet included in blocks)
-        let local_movements = state.local_movements.values().cloned().collect::<Vec<_>>();
-        transactions.extend_from_slice(&local_movements);
-
-        // Prepend balance movements of pending blocks
-        state.pending_movements.values().for_each(|x| {
-            transactions.extend_from_slice(x);
-        });
+        let mut keyed_transactions: BTreeMap<u32, model::BalanceMovement> = BTreeMap::new();
```
tmpolaczyk, line 332: This is a `BTreeMap` because it needs to be sorted, why does it need to be sorted?

mariocao: It is a nice to have, as in that ways clients will retrieve all transactions having some order.

wallet/src/repository/wallet/mod.rs
mariocao: I think there is no `contains`, but it should be really easy to implement. :)

wallet/src/repository/wallet/mod.rs
mariocao: The clone wars! :rofl:

wallet/src/repository/wallet/mod.rs
tmpolaczyk: Not today

wallet/src/repository/wallet/mod.rs
```diff
@@ -372,8 +364,24 @@ where
             }
         }
 
+        // Append balance movements of pending blocks
+        state.pending_movements.values().for_each(|x| {
+            keyed_transactions.extend(x.iter().map(|movement| (movement.db_key, movement.clone())));
```
tmpolaczyk, line 369: The `offset` and `limit` arguments are ignored here, should we check if `movement.db_key` is inside the `range` that the user asked for?

tmpolaczyk: Although this is complicated because of `local_movements`, which are always added to the end of the list. So do we would need to subtract `state.local_movements.len()` from the limit when fetching from the database?

wallet/src/repository/wallet/mod.rs
```diff
@@ -496,6 +504,24 @@ where
...
+                if self
+                    .db
+                    .get::<_, model::Path>(&keys::transactions_index(tally.dr_pointer.as_ref()))
+                    .is_ok()
+                {
+                    filtered_txns.push(txn.clone());
+                    continue;
+                }
+                // There is a DR transaction in pending state whose tally was found
+                else if let Some(_dr_tx) = state.pending_dr_movements.get(&tally.dr_pointer) {
```
tmpolaczyk, line 519: But here we do have a contains:

```suggestion
                else if state.pending_dr_movements.contains(&tally.dr_pointer) {
```

wallet/src/repository/wallet/state.rs
```diff
@@ -51,6 +51,8 @@ pub struct State {
     pub pending_addresses_by_path: HashMap<String, Arc<model::Address>>,
     /// List of pending blocks with state snapshots waiting to be confirmed
     pub pending_blocks: HashMap<String, StateSnapshot>,
+    /// List of pending dr movements, waiting to be confirmed with a superblock
```
tmpolaczyk, line 54: Can you give names to the types? This is a map of (dr_pointer) to (pending_block_hash, index). The index can be used to index pending_movements, which is a map of (pending_block_hash) to (Vec<BalanceMovement).

wallet/src/repository/wallet/mod.rs
```diff
@@ -372,8 +364,24 @@ where
             }
         }
 
+        // Append balance movements of pending blocks
+        state.pending_movements.values().for_each(|x| {
+            keyed_transactions.extend(x.iter().map(|movement| (movement.db_key, movement.clone())));
+        });
+
+        let total_txns = keyed_transactions.len();
```
tmpolaczyk, line 372: Careful, this is number of transactions in the `BTreeMap`, I think you want the index of the highest transaction. Try something like this:

```suggestion
        // Get the highest index
        let last_index = keyed_transactions.values().rev().next().unwrap_or(0);
```

And later when you do

```
total_txns + index
```

It should be

```
total_txns + index + 1
```

To test this we need a `keyed_transactions` where the first key is not 0.

wallet/src/repository/wallet/state.rs
girazoki: of course

wallet/src/repository/wallet/mod.rs
girazoki: we do not have a contains

wallet/src/repository/wallet/mod.rs
```diff
@@ -328,16 +329,7 @@ where
 
         // Total amount of state and db transactions
         let total = state.transaction_next_id + u32::try_from(state.local_movements.len()).unwrap();
```
tmpolaczyk, line 331: What's the difference between `total` and `db_total`? `db_total` is obtained from the database, but shouldn't that be equal to `state.transaction_next_id`?

mariocao: exactly! :)

wallet/src/repository/wallet/mod.rs
```diff
@@ -327,52 +332,103 @@ where
...
-
-        // Prepend local pending balance movements (not yet included in blocks)
-        let local_movements = state.local_movements.values().cloned().collect::<Vec<_>>();
-        transactions.extend_from_slice(&local_movements);
-
-        // Prepend balance movements of pending blocks
-        state.pending_movements.values().for_each(|x| {
-            transactions.extend_from_slice(x);
-        });
+        let mut keyed_transactions: Vec<model::BalanceMovement> = Vec::new();
```
tmpolaczyk, line 335:
```suggestion
        let mut transactions: Vec<model::BalanceMovement> = Vec::new();
```

wallet/src/repository/wallet/mod.rs
```diff
@@ -327,52 +332,103 @@ where
...
             .get_or_default::<_, u32>(&keys::transaction_next_id(account))?;
 
-        if db_total > 0 {
-            let end = db_total.saturating_sub(offset);
-            let start = end.saturating_sub(limit);
-            let range = start..end;
+        // get number of non-repated pending movements.
+        let pending_length = {
+            let mut glob_acc: usize = 0;
+            state.pending_movements.values().for_each(|x| {
```
tmpolaczyk, line 345: This looks like a `fold`

wallet/src/repository/wallet/mod.rs
```diff
@@ -327,52 +332,103 @@ where
...
-                            index,
-                            e
-                        );
+        // Append local movements if any
+        if let Some(range_local) = range_local {
+            // Append local pending balance movements (not yet included in blocks)
+            let mut local_movements: Vec<model::BalanceMovement> =
+                state.local_movements.values().cloned().collect();
+            local_movements.sort_by(|a, b| a.db_key.cmp(&b.db_key));
+            keyed_transactions.extend_from_slice(&local_movements[range_local]);
```
tmpolaczyk, line 367: In this block of code we are cloning the `model::BalanceMovement` two times: in `.cloned()` and also in `extend_from_slice`. Not sure what's the best way to avoid that, but the simplest one to use `drain` to move the elements from one vector into the other one:

```suggestion
            keyed_transactions.extend(local_movements.drain(range_local));
```

wallet/src/repository/wallet/mod.rs
```diff
@@ -495,6 +551,23 @@ where
...
+            if let types::Transaction::Tally(tally) = &txn {
+                // There is a DR transaction persisted in database whose tally was found or
+                // there is a DR transaction in pending state whose tally was found
+                if self
+                    .db
+                    .get::<_, u32>(&keys::transactions_index(tally.dr_pointer.as_ref()))
+                    .is_ok()
+                    || state
+                        .pending_dr_movements
+                        .contains_key(&tally.dr_pointer.to_string())
```
tmpolaczyk, line 564: Does it make sense to change the order of the condition here? So we first check if the transaction is in memory, and only check the database if it's not in memory.

wallet/src/repository/wallet/mod.rs
```diff
@@ -547,6 +628,63 @@ where
...
+                    );
+                    let dr_movement = state
+                        .pending_movements
+                        .get(&pending_block_hash.to_string())
+                        .unwrap()[index]
+                        .clone();
+
+                    match &dr_movement.transaction.data {
+                        model::TransactionData::DataRequest(dr_data) => {
+                            let mut updated_dr_movement = dr_movement.clone();
```
tmpolaczyk, line 650:
```suggestion
                            let mut updated_dr_movement = dr_movement;
```

wallet/src/repository/wallet/mod.rs
```diff
@@ -547,6 +628,63 @@ where
...
+                                state.account,
+                                txn_id,
+                            ))
+                            .map(|dr_movement| (dr_movement, txn_id))
+                    })
+                {
+                    log::debug!("Found a tally for data request {:?} that was in DB", txn_id);
+                    match &dr_movement.transaction.data {
+                        model::TransactionData::DataRequest(dr_data) => {
+                            let mut dr_movement_to_update = dr_movement.clone();
```
tmpolaczyk, line 674:
```suggestion
                            let mut dr_movement_to_update = dr_movement;
```

tmpolaczyk (approved):

mariocao (approved):

PR #1618, opened 2020-10-07
feat(superblock): remove reputed ars in signing committee

lrubiorod:


tmpolaczyk (approved):

tmpolaczyk:
The concern is that it may be possible to have 1 node with reputation and 100 nodes without reputation. In that case, before this PR only that 1 node would be able to vote. So if that node goes offline, the network cannot recover.

PR #1619, opened 2020-10-07
feat(wallet): make the wallet be 1 address ahead

### LINKED ISSUE #1260: Wallet should look for addresses when indexing blocks for the first time
When using the sheikah feature "import wallet from mnemonics", the wallet does not index old transactions. This is because there are no addresses generated yet.

A simple fix would be to generate a fixed amount of addresses (100 internal and 100 external, for example) before starting the indexing process.

Another option would be to add a method to manually re-index the block chain.

mariocao:
The wallet will always be ahead 1 address in their address list. It was an "easy" way to allow re-importing and re-indexing transactions targeting the wallet while re-importing a wallet into a client (Sheikah).

The main disadvantage of this "naive" approach is that if there are holes (addresses not being used) in the address list, then the wallet will stop generating them while syncing.

A proposal would be that while syncing, the wallet could generate address batches (with a significant amount of address, i.e. 100) and index all block transactions checking if they affect the current wallet's account.

Closes #1260

### DIFF
--- a/wallet/src/actors/worker/handlers/gen_address.rs
+++ b/wallet/src/actors/worker/handlers/gen_address.rs
@@ -1,6 +1,7 @@
 use actix::prelude::*;
 
 use crate::actors::worker;
+use crate::actors::worker::Error::AddressGeneration;
 use crate::{model, types};
 
 pub struct GenAddress(pub types::SessionWallet, pub bool, pub Option<String>);
@@ -17,8 +18,14 @@ impl Handler<GenAddress> for worker::Worker {
         GenAddress(wallet, external, label): GenAddress,
         _ctx: &mut Self::Context,
     ) -> Self::Result {
-        self.gen_address(&wallet, external, label).map(|address| {
-            (*address.expect("Address cannot be generated if wallet was never unlocked")).clone()
-        })
+        self.gen_address(&wallet, external, label)
+            .and_then(|addr_opt| {
+                addr_opt.ok_or_else(|| {
+                    AddressGeneration(
+                        "Address cannot be generated if wallet was never unlocked".to_string(),
+                    )
+                })
+            })
+            .map(|addr| (*addr).clone())
     }
 }
--- a/wallet/src/actors/worker/handlers/gen_address.rs
+++ b/wallet/src/actors/worker/handlers/gen_address.rs
@@ -17,7 +17,8 @@ impl Handler<GenAddress> for worker::Worker {
         GenAddress(wallet, external, label): GenAddress,
         _ctx: &mut Self::Context,
     ) -> Self::Result {
-        self.gen_address(&wallet, external, label)
-            .map(|address| (*address).clone())
+        self.gen_address(&wallet, external, label).map(|address| {
+            (*address.expect("Address cannot be generated if wallet was never unlocked")).clone()
+        })
     }
 }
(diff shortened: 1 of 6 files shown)


wallet/src/actors/worker/handlers/gen_address.rs
```diff
@@ -17,7 +17,8 @@ impl Handler<GenAddress> for worker::Worker {
         GenAddress(wallet, external, label): GenAddress,
         _ctx: &mut Self::Context,
     ) -> Self::Result {
-        self.gen_address(&wallet, external, label)
-            .map(|address| (*address).clone())
+        self.gen_address(&wallet, external, label).map(|address| {
+            (*address.expect("Address cannot be generated if wallet was never unlocked")).clone()
```
tmpolaczyk, line 21: Why `expect` in a function that returns a result? We should create a new error for this case and use `address?.clone()`.

wallet/src/repository/wallet/mod.rs
```diff
@@ -267,32 +307,33 @@ where
...
-            address,
-            path,
-            info,
-            index,
-            account,
-            keychain,
-            pkh,
-        };
+        // Return previous address index (always 1 address ahead)
+        let address = self._get_address(state, account, keychain, index).ok();
```
tmpolaczyk, line 317: This ignores any errors returned my `_get_address`. If `address` should be `None` in some special cases, handle that case specially. For example, if `_get_address` returns `None` when the index is 0:

```suggestion
        let address = if index != 0 { Some(self._get_address(state, account, keychain, index)?) } else { None };
```

wallet/src/repository/wallet/mod.rs
```diff
@@ -1026,7 +1053,9 @@ where
                 let change_pkh = if let Some(pkh) = first_pkh {
                     pkh
                 } else {
-                    self._gen_internal_address(state, None)?.pkh
+                    self._gen_internal_address(state, None)?
+                        .expect("Address cannot be generated if wallet was never unlocked")
```
tmpolaczyk, line 1057: `expect` in a function that returns a `Result`

wallet/src/repository/wallet/mod.rs
```diff
@@ -806,14 +843,6 @@ where
...
         batch.put(keys::account_utxo_set(account).into_bytes(), utxo_set)?;
         batch.put(keys::account_balance(account).into_bytes(), balance)?;
-        batch.put(
-            keys::account_next_index(account, constants::EXTERNAL_KEYCHAIN),
-            next_external_address_index,
-        )?;
-        batch.put(
-            keys::account_next_index(account, constants::INTERNAL_KEYCHAIN),
-            next_internal_address_index,
-        )?;
```
tmpolaczyk, line 816: Why can this be removed?

mariocao: Because addresses indices are persisted while generating addresses. No need to do it twice. :P

tmpolaczyk: Ok, perfect

wallet/src/repository/wallet/mod.rs
tmpolaczyk: Or even better, make `_get_address` return an `Option`

wallet/src/actors/worker/handlers/gen_address.rs
mariocao: We cannot use `?` inside a closure... how do you propose to change it?

tmpolaczyk: Either rewrite the map as a match, or in this case you can use `and_then` instead of `map`.

PR #1621, opened 2020-10-08
feat(node): include halving and block reward in the consensus constants

lrubiorod:


validations/src/validations.rs
```diff
@@ -2090,11 +2099,11 @@ pub fn block_reward(epoch: Epoch) -> u64 {
...
 
 /// Calculate the total amount of wits that will be rewarded to miners.
-pub fn total_block_reward() -> u64 {
+pub fn total_block_reward(initial_block_reward: u64, halving_period: u32) -> u64 {
     let mut total_reward = 0;
-    let mut base_reward = INITIAL_BLOCK_REWARD;
+    let mut base_reward = initial_block_reward;
     while base_reward != 0 {
-        total_reward += base_reward * u64::from(HALVING_PERIOD);
+        total_reward += base_reward * u64::from(halving_period);
```
tmpolaczyk, line 2106: We should make sure that this function panics on overflow, just to be safe.
```suggestion
        total_reward = total_reward.checked_add(base_reward.checked_mul(u64::from(halving_period)).expect("overflow")).expect("overflow");
```

validations/src/validations.rs
```diff
@@ -2430,20 +2442,25 @@ mod tests {
...
-        assert_eq!(block_reward(3_500_000), (62.5 * wit as f64).floor() as u64);
-        assert_eq!(block_reward(1_750_000 * 36), 3);
-        assert_eq!(block_reward(1_750_000 * 37), 1);
-        assert_eq!(block_reward(1_750_000 * 38), 0);
-        assert_eq!(block_reward(1_750_000 * 63), 0);
-        assert_eq!(block_reward(1_750_000 * 64), 0);
-        assert_eq!(block_reward(1_750_000 * 65), 0);
-        assert_eq!(block_reward(1_750_000 * 100), 0);
+        let ibr = INITIAL_BLOCK_REWARD;
+        let hp = HALVING_PERIOD;
```
tmpolaczyk, line 2446: You can use a closure to avoid passing the same arguments all the time:

```suggestion
        let reward = |epoch| block_reward(epoch, INITIAL_BLOCK_REWARD, HALVING_PERIOD);
```

config/src/defaults.rs
```diff
@@ -127,6 +127,18 @@ pub trait Defaults {
...
+    /// Default initial block reward
+    fn consensus_constants_initial_block_reward(&self) -> u64 {
+        // 250 wits
+        250 * 1_000_000_000
+    }
+
+    /// Default halving period
+    fn consensus_constants_halving_period(&self) -> u32 {
+        // 3.5M epochs * (45 secs/epoch) ~> 5 years
+        3_500_000
```
tmpolaczyk, line 139: This will be the first timestamp with halved reward:

1760166000 // 3500000 * 45 + 1602666000
10/11/2025 @ 7:00am (UTC)

tmpolaczyk (approved):

PR #1623, opened 2020-10-09
feat(config): update mainnet consensus constants

lrubiorod:
Respect to initial difficulty, due to after initial difficulty period, it will return to the calculate difficulty and the activity period is 2000, use a initial difficulty bigger than this number only delays a possible scale problem. If we can not manage a really big quantity of nodes, it is better to crash at the beginning instead of 25 hours later, when some miners have block rewards


drcpu-github:
I'm probably missing something but why does it make sense to increase the period with a fixed initial difficulty? The longer that period, the more likely it becomes that the number of active nodes is significantly bigger than the difficulty and would that not lead to more scaling problems (or chain splits if there are way too many blocks proposed)?

tmpolaczyk:
@drcpu-github when the initial difficulty period finishes, the mining difficulty will depend on the size of the ARS. The problem is that during the first blocks the only way to enter into the ARS is to mine a block, because there is no collateral for data requests. So when the difficulty period ends, there will be at most as many identities as unique miners.

There were proposals to count the candidate blocks and use them to update the difficulty, but it was too complex and we decided to just add the initial_difficulty option to fix the bootstrapping.

drcpu-github:
Sure, I know those mechanics, but I'm talking about whether lenghtening this initial difficulty period does not cause problems in itself too:

A couple of testnets ago, there was a problem with not having an initial difficulty (or it was too low?). Way too many blocks were being proposed and the network basically split and crashed. If you lengthen the initial difficulty period, the odds increase that more nodes come online causing this initial difficulty to be too low and introducing the same problem again, no?

However, let me see if I understand you correctly. If the initial difficulty period is too short, the ARS will be too small and difficulty will drop too much, basically leading to the same problem, correct? If both of these things are true, then it is just a difficult balancing problem, I guess.

By the way, I thought the difficulty algorithm was decoupled from simply being equal to the ARS size starting at testnet 0.9.3?

lrubiorod:
There are two difficults, a initial difficulty fixed to avoid what happened in one of the testnets, and another dynamic related to the ARS size. We use the initial difficulty to keep safe the network until the ARS lenght would be big enough to ensure a proper difficulty. As @tmpolaczyk says, in a environment without data requests, the only active identities comes from miners, and the activty_period is 2000, so it would be the maximum number for the ARS.

So putting a difficulty bigger than this number only delays the problem in case it would exist. Because after the initial_diffcult_period, it would be reduced to 2000 (technicaly 2000/7, but it is another story :P)

witnet.toml
```diff
@@ -10,40 +10,40 @@ public_addr = "0.0.0.0:21337"
...
+        "167.86.112.22:21337",
         "167.86.112.23:21337",
-        "167.172.196.106:21337",
-        "172.104.233.132:21337",
-        "172.105.68.149:21337",
-        "173.212.245.133:21337",
-        "178.62.85.63:21337",
+        "167.99.154.252:21337",
+        "173.249.3.178:21337",
+        "173.249.8.65:213377",
```
drcpu-github, line 37: This is a typo. :)

Also, that VPS will run the superblock bootstrap committee node, not sure if you want to have it do both since I have seen nodes crash because of too many incoming connections.

lrubiorod: Thanks!! I would recommend you to run the superblock bootstrap committee node in a machine that you know is not going to crash, for example in a machine with no inbounds, and respect to that IP, if you detect problems with inbound connection you can add an inbound_limit in the witnet.toml file

aesedepece:
`5f5f5f2f6f726967696e616c2f3030312f3738392f3432382f6130312e676966`

aesedepece (approved):

girazoki (approved):

config/src/defaults.rs
```diff
@@ -145,16 +145,15 @@ pub trait Defaults {
 
     /// Default Hash value for the auxiliary bootstrap block
     fn consensus_constants_bootstrap_hash(&self) -> Hash {
-        // FIXME(#1615) Decide an appropriate default value for Mainnet
-        "00000000000000000000000000000000000000007769746e65742d302e392e33"
+        // Brrr
```
mariocao, line 148: :rofl:

aesedepece (approved):
# EPIC

mariocao (approved):
```
______                       _ 
| ___ \                     | |
| |_/ / ___   ___  _ __ ___ | |
| ___ \/ _ \ / _ \| '_ ` _ \| |
| |_/ / (_) | (_) | | | | | |_|
\____/ \___/ \___/|_| |_| |_(_)
```

PR #1624, opened 2020-10-09
feat(wallet): add sync transient addresses for support wallet imports

mariocao:
Improves and closes #1619

### DIFF
--- a/wallet/src/repository/wallet/mod.rs
+++ b/wallet/src/repository/wallet/mod.rs
@@ -50,6 +50,75 @@ impl<T> Wallet<T>
 where
     T: Database,
 {
+    /// Generate transient addresses for synchronization purposes
+    /// This function only creates and inserts addreses
+    pub fn generate_transient_addresses(
+        &self,
+        external_addresses: u16,
+        internal_addresses: u16,
+    ) -> Result<()> {
+        let mut state = self.state.write()?;
+
+        self._generate_transient_addresses(&mut state, external_addresses, internal_addresses)
+    }
+
+    /// Non-locking version of `generate_transient_addresses`
+    pub fn _generate_transient_addresses(
+        &self,
+        state: &mut State,
+        external_addresses: u16,
+        internal_addresses: u16,
+    ) -> Result<()> {
+        // Compute ranges for external and internal addresses
+        let external_range =
+            state.next_external_index..state.next_external_index + u32::from(external_addresses);
+        let internal_range =
+            state.next_internal_index..state.next_internal_index + u32::from(internal_addresses);
+
+        // Generate external addresses
+        for index in external_range {
+            let account = state.account;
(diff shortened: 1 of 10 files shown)


wallet/src/repository/wallet/mod.rs
```diff
@@ -564,16 +631,24 @@ where
...
 
-            let check_db = |output: &types::VttOutput| {
+            let check_db_and_transient = |output: &types::VttOutput| {
                 self.db
                     .get::<_, model::Path>(&keys::pkh(&output.pkh))
                     .is_ok()
+                    || state
+                        .transient_external_addresses
+                        .get(&output.pkh)
+                        .is_some()
```
tmpolaczyk, line 641: This is .contains()

wallet/src/repository/wallet/mod.rs
```diff
@@ -50,6 +50,75 @@ impl<T> Wallet<T>
...
+            state.next_internal_index..state.next_internal_index + u32::from(internal_addresses);
+
+        // Generate external addresses
+        for index in external_range {
+            let account = state.account;
+            let keychain = constants::EXTERNAL_KEYCHAIN;
+            let parent_key = &state.keychains[keychain as usize].clone();
+
+            let (address, _) =
+                self.gen_address(None, parent_key, account, keychain, index, false)?;
```
tmpolaczyk, line 85: This will trigger a deadlock

```suggestion
                self._gen_address(None, parent_key, account, keychain, index, false)?;
```

wallet/src/repository/wallet/mod.rs
```diff
@@ -1321,22 +1440,11 @@ where
...
         let (address, next_index) =
-            self.gen_address(label, parent_key, account, keychain, index)?;
-
-        let path = model::Path {
-            account,
-            keychain,
-            index,
-        }
-        .to_string();
+            self.gen_address(label, &parent_key, account, keychain, index, true)?;
```
tmpolaczyk, line 1446: Deadlock? Even though the code is the same as before :thinking: 

```suggestion
            self._gen_address(label, &parent_key, account, keychain, index, true)?;
```

wallet/src/repository/wallet/mod.rs
mariocao: `_gen_address` does not exist :) Maybe in the past

wallet/src/repository/wallet/mod.rs
mariocao: true :smile:

wallet/src/repository/wallet/mod.rs
tmpolaczyk: Ah sorry, this is not the `gen_address` that uses the state mutex, it's a different function with the same name

tmpolaczyk (approved):

PR #1627, opened 2020-10-12
feat(wallet): add support for generating addresses by transaction outputs

mariocao:
Add support for generating addresses by transaction outputs (instead of just transactions).

Before, the synchronization address generation batch only applied to single transactions and not to all outputs, i.e. a transactions with 1000 outputs could never work with the previous implementation.

Now, if a single transaction has several outputs and their index difference does not exceed the `sync_address_batch_length`, the current implementation will allow a successful synchronization.

### DIFF
--- a/wallet/src/repository/wallet/mod.rs
+++ b/wallet/src/repository/wallet/mod.rs
@@ -1291,6 +1291,7 @@ where
         Ok(None)
     }
 
+    // Returns the account mutation in terms of changes to the UTXO set and balance
     fn _get_account_mutation(
         &self,
         state: &State,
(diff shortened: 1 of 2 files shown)


wallet/src/repository/wallet/mod.rs
```diff
@@ -1353,10 +1368,6 @@ where
                         );
                     }
                 }
-            } else if state.transient_external_addresses.contains_key(&output.pkh) {
-                own_outputs.insert(output.pkh, model::OutputType::External);
-            } else if state.transient_internal_addresses.contains_key(&output.pkh) {
-                own_outputs.insert(output.pkh, model::OutputType::Internal);
```
tmpolaczyk, line 1359: Is this not needed anymore?

wallet/src/repository/wallet/mod.rs
```diff
@@ -1292,6 +1252,79 @@ where
...
 
+    // During wallet synchronization, generate external and internal addresses
+    // if transaction outputs are pointing to transient addresses
+    fn _sync_address_generation(
+        &self,
+        state: &mut State,
+        outputs: &[types::VttOutput],
+    ) -> Result<()> {
+        loop {
+            let (new_external_index, new_internal_index) = outputs.iter().fold(
```
tmpolaczyk, line 1263: Is it possible to move the fold outside of the loop, or would that break everything?

wallet/src/repository/wallet/mod.rs
mariocao: Nope, because the external and internal addresses are generated (as normal ones) before getting there.

wallet/src/repository/wallet/mod.rs
mariocao: I think yes. For sure, there is a better way to do that address generation, but we valued simplicity over performance (giving the fact the little time we have).

tmpolaczyk (approved):

PR #1638, opened 2020-10-15
feat(wallet): add getInternal addresses in jsonrpc

### LINKED ISSUE #1634: [Wallet] Add JsonRPC method to retrieve internal addresses
Currently, the `get_addresses` method only returns only the external addresses.

A possible implementation might be to include an optional field called `external` to that method.

girazoki:
Closes #1634

### DIFF
--- a/wallet/src/actors/worker/handlers/get_addresses.rs
+++ b/wallet/src/actors/worker/handlers/get_addresses.rs
@@ -9,6 +9,8 @@ pub struct GetAddresses(
     pub u32,
     /// Limit
     pub u32,
+    /// External
+    pub bool,
 );
 
 impl Message for GetAddresses {
@@ -20,9 +22,9 @@ impl Handler<GetAddresses> for worker::Worker {
 
     fn handle(
         &mut self,
-        GetAddresses(wallet, offset, limit): GetAddresses,
+        GetAddresses(wallet, offset, limit, external): GetAddresses,
         _ctx: &mut Self::Context,
     ) -> Self::Result {
-        self.addresses(&wallet, offset, limit)
+        self.addresses(&wallet, offset, limit, external)
     }
 }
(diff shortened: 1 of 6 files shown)


wallet/src/actors/worker/handlers/get_addresses.rs
```diff
@@ -9,6 +9,8 @@ pub struct GetAddresses(
     pub u32,
     /// Limit
     pub u32,
+    /// External
+    pub bool,
```
tmpolaczyk, line 13: Refactor this into a struct with name fields please

```suggestion
    /// External
    pub external: bool,
```

mariocao: Yep, that's a known a easy refactor! :)

wallet/src/repository/wallet/mod.rs
```diff
@@ -373,7 +373,34 @@ where
...
+        for index in range.rev() {
+            let address = self.get_address(account, keychain, index)?;
+            addresses.push((*address).clone());
+        }
+
+        Ok(model::Addresses { addresses, total })
+    }
+
+    /// Return a list of the generated internal addresses that.
+    pub fn internal_addresses(&self, offset: u32, limit: u32) -> Result<model::Addresses> {
```
mariocao, line 391: I think this function can be refactored together with `external_addresses`.

Maybe creating a generic non-public `addresses` function with the `limit`, `offset` and `keychain` as argument.

mariocao (changes requested):

wallet/src/actors/worker/handlers/get_addresses.rs
girazoki: this is a common issue with all handlers. I would change them all in the same issue, not on this one obviously

tmpolaczyk (approved):

mariocao (approved):

PR #1639, opened 2020-10-15
fix(wallet): create wallet twice should return an error

mariocao:
If a user tries to create twice a wallet with the same seed phrase, the wallet will return a `409 Conflict` error message.

### DIFF
--- a/wallet/src/actors/worker/methods.rs
+++ b/wallet/src/actors/worker/methods.rs
@@ -78,6 +78,16 @@ impl Worker {
             self.params.id_hash_iterations,
         );
 
+        // If wallet ID already exists, return Err
+        if self
+            .wallets
+            .infos()?
+            .into_iter()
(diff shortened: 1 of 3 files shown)


wallet/src/actors/worker/methods.rs
```diff
@@ -78,6 +78,16 @@ impl Worker {
...
 
+        // If wallet ID already exists, return Err
+        if self
+            .wallets
+            .infos()?
+            .into_iter()
+            .any(|wallet| wallet.id == id)
+        {
+            return Err(Error::WalletAlreadyExists(id));
+        };
```
tmpolaczyk, line 89: We can also check if a wallet with this id already exists by looking if the database contains some key. But since we already keep the wallet ids in memory, and we don't expect to have thousands of wallets, iterating is fine.

tmpolaczyk (approved):

PR #1641, opened 2020-10-16
refactor(wallet): use type safe database api

tmpolaczyk:
This PR adds types to the keys used in the wallet database.

This way one does not need to specify the type when using `db.get()`, and also it's impossible to deserialize a value as a wrong type. All the keys and types should be compatible with the ones in master, but we should double check that this does not break anything.

### DIFF
--- a/wallet/src/repository/wallet/tests/mod.rs
+++ b/wallet/src/repository/wallet/tests/mod.rs
@@ -1,6 +1,7 @@
 use std::{collections::HashMap, iter::FromIterator as _, mem};
 
 use super::*;
+use crate::db::HashMapDb;
 use crate::repository::wallet::tests::factories::vtt_from_body;
 use crate::types::Hashable;
 use crate::*;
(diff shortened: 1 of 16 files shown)


wallet/src/repository/wallet/tests/mod.rs
```diff
@@ -577,10 +552,16 @@ fn test_create_vtt_does_not_spend_utxos() {
 
     assert_eq!(1, wallet.balance().unwrap().confirmed.available);
     assert_eq!(0, wallet.balance().unwrap().confirmed.locked);
-    assert_eq!(1, db.get::<_, u64>(&keys::account_balance(0)).unwrap());
```
tmpolaczyk, line 580: This was the only type mismatch between `db.put` and `db.get`, and this test did not fail because bincode serializes structs by concatenating all the fields, and the first field was a `u64` with value 1.

aesedepece (approved):
This is super cool :muscle:

PR #1644, opened 2020-10-17
chore: updated directories 2.0.2 to directories-next 1.0.2

drcpu-github:
Solves issue #1642.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,6 @@ travis-ci = { repository = "https://github.com/witnet/witnet-rust", branch = "ma
 ansi_term = "0.12.1"
 bytecount = "0.6.0"
 ctrlc = "3.1.3"
-directories = "2.0.2"
 env_logger = "0.7.1"
 failure = "0.1.8"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -26,7 +26,7 @@ travis-ci = { repository = "https://github.com/witnet/witnet-rust", branch = "ma
 ansi_term = "0.12.1"
 bytecount = "0.6.0"
 ctrlc = "3.1.3"
-directories = "2.0.2"
+directories-next = "1.0.2"
```
tmpolaczyk, line 29: Actually, this line can be removed

```suggestion
```

drcpu-github: Yes, I was not sure why that dependency was included there since it wasn't used, but I did not want to break anything.

drcpu-github:
Added suggestions and rebased master.

tmpolaczyk (approved):

Cargo.toml
tmpolaczyk: Yeah, ideally we would have some automated check that would detect unused dependencies, but we haven't found any.

PR #1663, opened 2020-10-20
feat(jsonrpc): add "confirmed" field to getBlock method

### LINKED ISSUE #1437: Superblock Sync: discard persisted and non-consolidated blocks
Currently when the chain reorganizes we do not discard the non-consolidated blocks, leading to a client ( and the wallet) being able to still query them

tmpolaczyk:
This allows to know if a block has been consolidated by a majority of the superblock committee.

The implementation checks that the block epoch is below the last confirmed block according to the superblock beacon, and that the block hash exists at the expected position in the block chain.

This method is used in the wallet synchronization.

Close #1437

### DIFF
--- a/node/src/actors/chain_manager/handlers.rs
+++ b/node/src/actors/chain_manager/handlers.rs
@@ -26,8 +26,9 @@ use crate::{
             AddTransaction, Broadcast, BuildDrt, BuildVtt, EpochNotification, GetBalance,
             GetBlocksEpochRange, GetDataRequestInfo, GetHighestCheckpointBeacon,
             GetMemoryTransaction, GetMempool, GetMempoolResult, GetNodeStats, GetReputation,
-            GetReputationResult, GetState, GetSuperBlockVotes, GetUtxoInfo, PeersBeacons,
-            ReputationStats, SendLastBeacon, SessionUnitResult, SetLastBeacon, TryMineBlock,
+            GetReputationResult, GetState, GetSuperBlockVotes, GetUtxoInfo, IsConfirmedBlock,
+            PeersBeacons, ReputationStats, SendLastBeacon, SessionUnitResult, SetLastBeacon,
+            TryMineBlock,
         },
         sessions_manager::SessionsManager,
     },
@@ -388,9 +389,6 @@ impl Handler<AddBlocks> for ChainManager {
                         }
 
                         // Persist blocks batch when target not reached
-                        // FIXME(#1437): this blocks should be marked as "pending" because we don't
-                        // know if they are valid yet. We can mark them as valid after verifying
-                        // that the top blocks pertain to the target superblock.
                         self.persist_blocks_batch(ctx, blocks);
                         let to_be_stored =
                             self.chain_state.data_request_pool.finished_data_requests();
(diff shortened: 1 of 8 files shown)


tmpolaczyk:
This PR currently does not work for blocks received when synchronizing, because when synchronizing the node does not create superblocks aside from the target superblock.

node/src/actors/chain_manager/handlers.rs
```diff
@@ -1497,6 +1498,37 @@ impl Handler<AddSuperBlock> for ChainManager {
...
+            .chain_state
+            .chain_info
+            .as_ref()
+            .ok_or(ChainManagerError::ChainNotReady)?
+            .consensus_constants
+            .superblock_period;
+        let superblock_beacon = self.get_superblock_beacon();
+        // Superblock 1 confirms blocks 0..=9
+        let last_confirmed_block =
+            (superblock_beacon.checkpoint * u32::from(superblock_period)).saturating_sub(1);
```
tmpolaczyk, line 1515: This is the kind of logic that I would like to extract to a dedicated function

node/src/actors/json_rpc/json_rpc_methods.rs
```diff
@@ -588,15 +591,44 @@ pub fn get_block(hash: Result<(Hash,), jsonrpc_core::Error>) -> JsonRpcResultAsy
...
+                            block_epoch,
+                        })
+                        .then(|res| match res {
+                            Ok(Ok(confirmed)) => {
+                                if confirmed {
+                                    // Append {"confirmed":true} to the result
+                                    value
+                                        .as_object_mut()
+                                        .expect("The result of getBlock should be an object")
+                                        .insert("confirmed".to_string(), Value::Bool(true));
```
girazoki, line 608: Is there a way we return a Value::Bool(false)?

girazoki: I mean, if there is any case for that. I dont see it and seems we should right?

tmpolaczyk: I assumed that if the "confirmed" key does not exist it is the same as setting it to false

girazoki: Ok, a bit confusing  but I understand it is to save some bytes in some cases. Although given the majority of blocks will have the true field maybe it should be the other way around haha

girazoki: lets leave it like that though, it might be more confusing otherwise

mariocao: I agree with @girazoki :smile:

mariocao (changes requested):

node/src/actors/json_rpc/json_rpc_methods.rs
tmpolaczyk: Ok, changed. Now all blocks should have a "confirmed" field.

wallet/src/actors/worker/methods.rs
```diff
@@ -777,7 +759,18 @@ impl Worker {
...
-                serde_json::from_value::<types::ChainBlock>(json).map_err(node_error)
+                // Set confirmed to true if the result contains {"confirmed": true}
+                let mut confirmed = false;
+                if let Some(obj) = json.as_object() {
+                    if let Some(c) = obj.get("confirmed") {
+                        if let Some(true) = c.as_bool() {
+                            confirmed = true;
+                        }
+                    }
+                }
```
mariocao, line 770: After you last changes, block will always have a `confirmed` field, don't they?

tmpolaczyk: This will work as well, but should I rewrite it to return an error if there is no `confirmed` field?

tmpolaczyk: This way the wallet could detect if it connects to a node running an old version.

girazoki (approved):

mariocao (approved):

node/src/actors/messages.rs
```diff
@@ -345,6 +345,19 @@ pub struct AddSuperBlock {
     pub superblock: SuperBlock,
 }
 
+/// Returns true if the provided block hash is the consolidated block for the provided epoch, and
```
aesedepece, line 348: How does a data structure return? :stuck_out_tongue:

wallet/src/actors/worker/methods.rs
```diff
@@ -703,16 +692,9 @@ impl Worker {
...
                 let get_block_future = self.get_block(id.clone());
-                let block: types::ChainBlock = futures03::executor::block_on(get_block_future)?;
-
-                // Compute if block should be considered confirmed
-                // Note: blocks confirmed in past superblocks are considered confirmed
-                // FIXME(#1437): best approach would be the node signaling if blocks are confirmed
-                let node_tip_superblock_index = self.current_epoch()? / superblock_period;
-                let local_tip_superblock_index = epoch / superblock_period;
-                let confirmed = node_tip_superblock_index - local_tip_superblock_index > 1;
+                let (block, confirmed) = futures03::executor::block_on(get_block_future)?;
```
aesedepece, line 697: Won't this degrade the performance of the synchronization progress significantly? If we get the blocks list in batches, maybe we could also ask for their confirmation status in batches. Or maybe offer that information upfront in the blocks list.

tmpolaczyk: I don't know about performance, but yes, we could expose a `getLastConfirmedBlockEpoch` method in the JSON-RPC interface, and set all blocks before than to confirmed.

aesedepece: So I guess that we can settle with this PR for now and do what you just mentioned in a subsequent update, when reducing synchronization times start to be more pressing as the chain grows.

PR #1669, opened 2020-10-21
Cache uncompressed bn keys

### LINKED ISSUE #1668: Calculating the hash of BN256 public keys is very expensive
This is the code that calculates the hash of a BN256 public key, used in the function build_superblock to calculate the ARS hash:

```rust
impl Hashable for Bn256PublicKey {
    fn hash(&self) -> Hash {
        calculate_sha256(&self.to_uncompressed().unwrap()).into()
    }
}
```

`to_uncompressed` is an expensive operation, so we should avoid calling it here.

Possible solutions

* Cache the hash. The operation will still be expensive but only the first time a public key is used.
* Cache the uncompressed key. Same as above, but we can reuse the uncompressed key in other places.
* Store the uncompressed key. We can assess whether we need the compressed key or the uncompressed key, or even both, and calculate them ahead of time.
* Change the hash function to use compressed keys (BREAKING CHANGE). This depends on the bridge, and what is cheaper in ethereum.

tmpolaczyk:
Close #1668

When running a node in debug mode, the superblock creation step used to take more than 60 seconds. With this PR, it takes less than 1 second. In release mode the difference may not be as noticeable because the node is already fast enough (unless someone is running a node on slow hardware), but with this PR we can support bigger ARS with more identities.

The size of an uncompressed key is 128 bytes, so this PR will increase the memory usage by 128 bytes times the size of the ARS. For an ARS with 2000 identities, this is about 250 KiB.

### DIFF
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -37,8 +37,8 @@ use crate::{
     proto::{schema::witnet, ProtobufConvert},
     superblock::SuperBlockState,
     transaction::{
-        CommitTransaction, DRTransaction, DRTransactionBody, MintTransaction, RevealTransaction,
-        TallyTransaction, Transaction, TxInclusionProof, VTTransaction,
+        CommitTransaction, DRTransaction, DRTransactionBody, Memoized, MintTransaction,
+        RevealTransaction, TallyTransaction, Transaction, TxInclusionProof, VTTransaction,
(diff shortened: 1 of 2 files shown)


lrubiorod (approved):

mariocao (approved):
Great PR! :+1:

Issue #1603, opened 2020-09-29
Transactions with fee 0 should not be valid


lrubiorod:

tmpolaczyk:
They are valid in bitcoin, however bitcoin nodes do not relay transactions with small fees, but "small" is configurable.

lrubiorod:
Implemented a configurable solution in #1671

PR #1671, opened 2020-10-21
feat(mining): allow miners to set a minimum required fee for including a VTT in a block

drcpu-github:
This is a proposal to limit spamming the network with free VTT's (like we saw today). It could also be extended to DR's, but I did not include that yet because I'm not sure that is something we ultimately want.

This PR would allow a node operator to set a minimum transaction fee he wants to receive before including a VTT into a block whenever he's eligible. I would set this to 1 nanoWIT (though I did not set that as default), but obviously everyone can choose a setting.

Let me know what you think about this and if this unexpectedly breaks something. For example, I am not sure if this would let the VTT pool grow unlimited in the case of spam (I thought there was some sort of periodic cleanup).

### DIFF
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -1517,6 +1517,9 @@ pub struct TransactionsPool {
     // Values greater than 1 mean keep more value transfer transactions than data request
     // transactions.
     vt_to_dr_factor: f64,
+    // Minimum fee required to include a VTT into a block. We check for this fee in the
+    // TransactionPool so we can choose not to insert a transaction we will not mine anyway.
+    minimum_vtt_fee: u64,
 }
 
 impl Default for TransactionsPool {
@@ -1538,6 +1541,8 @@ impl Default for TransactionsPool {
             weight_limit: u64::MAX,
             // Try to keep the same amount of value transfer weight and data request weight
             vt_to_dr_factor: 1.0,
+            // Default is to include all transactions into the pool and blocks
+            minimum_vtt_fee: 0,
         }
     }
 }
@@ -1574,6 +1579,15 @@ impl TransactionsPool {
         self.remove_transactions_for_size_limit()
     }
 
+    /// As a miner, require a minimum flat fee to include a VTT into the `TransactionsPool` and
+    /// blocks.
+    ///
+    /// This fee is irrespective of the complexity of the VTT so this setting is as uncomplicated
+    /// as possible.
+    pub fn set_minimum_vtt_fee(&mut self, minimum_vtt_fee: u64) {
+        self.minimum_vtt_fee = minimum_vtt_fee;
+    }
+
     /// Returns `true` if the pool contains no transactions.
     ///
     /// # Examples:
@@ -1608,6 +1622,7 @@ impl TransactionsPool {
             total_dr_weight,
             weight_limit: _,
             vt_to_dr_factor: _,
+            minimum_vtt_fee: _,
         } = self;
 
         vt_transactions.clear();
@@ -2084,6 +2099,9 @@ impl TransactionsPool {
     /// Due to the size limit, inserting a new transaction may result in some older ones being
     /// removed. This method returns a list of the removed transactions.
     ///
+    /// If the transaction is not added due to configurable constraints, the method returns a
+    /// list containing only the transaction that was not added
+    ///
     /// # Examples:
     ///
     /// ```
@@ -2104,17 +2122,21 @@ impl TransactionsPool {
                 let weight = f64::from(vt_tx.weight());
                 let priority = OrderedFloat(fee as f64 / weight);
 
-                self.total_vt_weight += u64::from(vt_tx.weight());
+                if fee < self.minimum_vtt_fee {
+                    return vec![Transaction::ValueTransfer(vt_tx)];
+                } else {
+                    self.total_vt_weight += u64::from(vt_tx.weight());
 
-                for input in &vt_tx.body.inputs {
-                    self.output_pointer_map
-                        .entry(input.output_pointer.clone())
-                        .or_insert_with(Vec::new)
-                        .push(vt_tx.hash());
+                    for input in &vt_tx.body.inputs {
+                        self.output_pointer_map
+                            .entry(input.output_pointer.clone())
+                            .or_insert_with(Vec::new)
+                            .push(vt_tx.hash());
+                    }
+
+                    self.vt_transactions.insert(key, (priority, vt_tx));
+                    self.sorted_vt_index.insert((priority, key));
                 }
-
-                self.vt_transactions.insert(key, (priority, vt_tx));
-                self.sorted_vt_index.insert((priority, key));
             }
             Transaction::DataRequest(dr_tx) => {
                 let weight = f64::from(dr_tx.weight());
@@ -4295,6 +4317,27 @@ mod tests {
             // Assert there are exactly 10 removed VTTs
             assert_eq!(removed_vt_count, 10);
         }
+    }
+
+    #[test]
+    fn transactions_pool_minimum_vtt_fee() {
+        let input = Input::default();
+        let vt1 = Transaction::ValueTransfer(VTTransaction::new(
+            VTTransactionBody::new(vec![input], vec![ValueTransferOutput::default()]),
+            vec![],
+        ));
+
+        let mut transactions_pool = TransactionsPool::default();
+        transactions_pool.set_minimum_vtt_fee(1);
+
+        // Inserting a transaction with a fee lower than `minimum_vtt_fee` results in the transaction not being added
+        let removed = transactions_pool.insert(vt1.clone(), 0);
+        assert_eq!(removed, vec![vt1.clone()]);
+        assert!(!transactions_pool.contains(&vt1).unwrap());
+        // Inserting a transaction with a fee higher than `minimum_vtt_fee` results in the transaction being added
+        let removed = transactions_pool.insert(vt1.clone(), 1);
+        assert_eq!(removed, vec![]);
+        assert!(transactions_pool.contains(&vt1).unwrap());
     }
 
     #[test]
--- a/node/src/actors/chain_manager/mod.rs
+++ b/node/src/actors/chain_manager/mod.rs
@@ -2351,7 +2351,7 @@ pub fn log_removed_transactions(removed_transactions: &[Transaction], inserted_t
 
         if removed_the_one_we_just_inserted {
             log::trace!(
-                "Transaction {} was not added because the TransactionsPool is full",
+                "Transaction {} was not added because the fee was too low",
                 inserted_tx_hash
             );
         } else {
@@ -2359,7 +2359,7 @@ pub fn log_removed_transactions(removed_transactions: &[Transaction], inserted_t
         }
 
         log::debug!(
-            "TransactionsPool is full! Removed the following transactions: {:?}",
+            "Removed the following transactions: {:?}",
             removed_tx_hashes
         );
     }
--- a/witnet.toml
+++ b/witnet.toml
@@ -91,6 +91,8 @@ genesis_path = ".witnet/config/genesis_block.json"
 # the percentage of the block rewards that will be assigned to `mint_external_address` (50% by default)
 #mint_external_address = "twit1jqgf4rxjrgas3kdhj3t4cr3mg3n33m8zw0aglr"
 #mint_external_percentage = 50
+# Set a minimum fee you require before your node includes a value transfer transaction into a block
+minimum_vtt_fee_nanowits = 1
 
 [log]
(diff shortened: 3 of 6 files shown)


tmpolaczyk:
This would work, but the transactions would still flood the network. A way to reduce spamming is to use this minimum fee when validating transactions, and stop broadcasting transactions with small fees. See issue #1603.

tmpolaczyk:
https://github.com/witnet/witnet-rust/blob/e45f186731787e5d1c9d9b1e56c6da2522c9057c/validations/src/validations.rs#L1833-L1835

Maybe we can modify this function to accept a `minimum_vtt_fee` argument, and return a validation error if the fee is smaller than that. This way the node would not broadcast the transaction, and also not include it in the mempool.

drcpu-github:
Yes, this is a miner-centric update. I did realize it does not prevent spamming itself, but as long as they are not included in blocks, they don't bloat the chain.

Does every node call validate_new_transaction when it sees one or only the node that initializes the transaction? The former would prevent 0 fee transactions to propagate through the network indeed. In the latter case, a malicious actor can just modify that function to remove the `minimum_vtt_fee` allowing for spamming of 0 fee transactions again.

tmpolaczyk:
Ah ok, miners need to pay for storage space so it makes sense to set a minimum fee.

> Does every node call validate_new_transaction when it sees one or only the node that initializes the transaction?

It should call validate_new_transaction when it receives a new transaction, before broadcasting it to the network. And also when creating a new transaction, so the error that we display should be user-friendly.

tmpolaczyk:
Maybe it's easier to implement here instead:

https://github.com/witnet/witnet-rust/blob/4cadef76ab32c03850936c1d61e29feff4cf49dc/node/src/actors/chain_manager/mod.rs#L1013

Add something like

```rust
if fee < minimum { return actix::fut::err(e) };
```

drcpu-github:
I do think we should differentiate between these two things though. My PR focuses on a miner's view (and is economic in nature). Setting a minimum fee in either `add_transaction` or `validate_new_transaction` is more of a network-centric update.

Disallowing spam of 0-cost transactions may make sense, but do we want to make a network-centric update to be something configurable? I think it makes little sense to allow a miner to choose whether to broadcast / create a transaction based on the fee?

tmpolaczyk:
Ok, let's focus on storage space in this PR and leave the networking aspects out for a future issue.

The only problem I see in this PR is that transactions with fee 0 will still be inserted into the transactions pool when the minimum fee is greater than 0, and that's not needed. But we can ignore this problem because: 1) the transactions pool has a size limit and when it is full the transactions with lower fee are removed first, and 2) as soon as these transactions become invalid or are included by another miner, they will be removed from the transactions pool, so they shouldn't stay in memory forever.

The minimum fee should probably be set in nanowits / weight unit so that bigger transactions need bigger fee. But this adds some confusion because our users have never seen fees represented using that unit, and we would need to add some command like `node send --fee-per-weight-unit=100` to replace `node send --fee=1`. Also, there is no way to see the fee using getTransaction, or even in the block explorer. So I guess we can also ignore this problem for now and set the minimum fee in nanowits.

witnet.toml
```diff
@@ -91,6 +91,8 @@ genesis_path = ".witnet/config/genesis_block.json"
 # the percentage of the block rewards that will be assigned to `mint_external_address` (50% by default)
 #mint_external_address = "twit1jqgf4rxjrgas3kdhj3t4cr3mg3n33m8zw0aglr"
 #mint_external_percentage = 50
+# Set a minimum fee you require before your node includes a value transfer transaction into a block
+minimum_vtt_fee = 1
```
tmpolaczyk, line 95: This will set this value to 1 for people who use the default witnet.toml
Comment this line to use the default value, 0.

```suggestion
#minimum_vtt_fee = 1
```

drcpu-github: I set it to 1 because (currently) most node operators are not quite the type of people that are familiar enough with node operation to modify that type of parameters (even though it would still make sense for them to request a minimum fee of 1 nanoWit). However, if you still think it is better to set it to the default value of 0, I'm fine with that.

drcpu-github:
> The only problem I see in this PR is that transactions with fee 0 will still be inserted into the transactions pool when the minimum fee is greater than 0, and that's not needed. But we can ignore this problem because: 1) the transactions pool has a size limit and when it is full the transactions with lower fee are removed first, and 2) as soon as these transactions become invalid or are included by another miner, they will be removed from the transactions pool, so they shouldn't stay in memory forever.

That was basically my reasoning in implementing this PR, so it's good that that is confirmed.

> The minimum fee should probably be set in nanowits / weight unit so that bigger transactions need bigger fee. But this adds some confusion because our users have never seen fees represented using that unit, and we would need to add some command like `node send --fee-per-weight-unit=100` to replace `node send --fee=1`. Also, there is no way to see the fee using getTransaction, or even in the block explorer. So I guess we can also ignore this problem for now and set the minimum fee in nanowits.

Agreed that it makes sense to have a fee-per-weight unit for more complex VTT's but also that it is quite a complex thing for most node operators to define.

drcpu-github:
On a related note, do I extend this PR to include the same minimum fee option for a DR or not?

tmpolaczyk:
> On a related note, do I extend this PR to include the same minimum fee option for a DR or not?

Not needed

witnet.toml
tmpolaczyk: Well, it's easy to approve PRs that add extra features and don't change existing ones, but if some feature changes the default behavior of a node then those changes need to be carefully discussed.

I would prefer to keep the value of 0 because I don't see transactions with 0 fee as a big problem, but let's wait for a third opinion.

drcpu-github: I agree, that's why I'm taking the time to discuss this. While it is not a breaking change in the codebase, the impact on the economics of the network are definitely not neglible. I'm not in a rush to get this merged, so yes, let's wait for more opinions.   😉

node/src/actors/chain_manager/mining.rs
```diff
@@ -739,6 +742,12 @@ fn build_block(
             }
         };
 
+        // The value transfer transaction are sorted from highest to lowest fee, so as
+        // soon as the transaction fee is below our minimum fee, we can just stop searching
+        if transaction_fee < minimum_vtt_fee {
+            break;
```
lrubiorod, line 748: Transactions are sorted by fee/weight, so could happen that a transaction with a fee lower than the minimum_vtt_fee was before than other with a bigger fee but much bigger weight

witnet.toml
lrubiorod: I agree that at least 1 nanowit is need for a miner to earn a really minimum profit by include the transaction in the block

node/src/actors/chain_manager/mining.rs
drcpu-github: Okay, I missed that when going over the code. 

That leaves two options:
1) I still handle `minimum_vtt_fee` as a flat fee and just continue the iteration.
2) I do implement it as a fee / weight as suggested by @tmpolaczyk.

If I go for case 2, I would abstract the `fee-per-weight` away from the miner and say that if you set `minimum_vtt_fee` to 1, that means you want 1 nanoWIT for the most basic VTT (and more for more complex transactions). The most basic VTT has one input one output with a weight of `1*133 + 1*36*10=493`. So in code that would be something like (with proper rounding):
```
if transaction_fee / (weight / 493) < minimum_vtt_fee {
            break;
}
```

What do you think?

tmpolaczyk: There is already a `min_vt_weight` variable there, you can use that instead of 493. If minimum_vtt_fee is set in nanowits, then the code should use floating point division to avoid problems with integer division:

```
if f64::from(transaction_fee) / f64::from(weight) < f64::from(minimum_vtt_fee) / f64::from(min_vt_weight) {
            break;
}
```

If this works then go with option 2, otherwise just leave it like it was before and add a warning that this feature is experimental and the node may skip some transactions with higher fee.

drcpu-github:
I implemented a "normalized" fee per weight `minimum_vtt_fee` and added a (big) test to check this works well (spoiler alert, it does). I had to convert using `as f64` because precise conversion from u64 to f64 using from or try_from is not implemented due to possible loss of precision.

I left the default `mining_minimum_vtt_fee` at 0 because I believe that deleting the setting in witnet.toml (default at 1) should result in all VTT's being accepted.

witnet.toml
aesedepece, line 95: Let me suggest that all configuration file entries should clearly state their unit, in this case, `nanowit`:
```suggestion
minimum_vtt_fee_nanowits = 1
```

drcpu-github: Done!

drcpu-github:
After some thinking I decided to not insert the VTT transaction into the transactions pool if its fee is too low, which is (partially) what @tmpolaczyk suggested in the beginning. It does not make too much sense to include it into the pool if you are not planning to include it in a block. I do still broadcast the transaction because I don't think a miner should censor a transaction simply because he does not want to mine a block with it.

node/src/actors/chain_manager/mod.rs
```diff
@@ -1932,7 +1932,9 @@ fn update_pools(
     }
 
     for vt_tx in &block.txns.value_transfer_txns {
-        transactions_pool.vt_remove(&vt_tx.hash());
+        if transactions_pool.vt_contains(&vt_tx.hash()) {
+            transactions_pool.vt_remove(&vt_tx.hash());
```
lrubiorod, line 1936: vt_removes also tries to remove the transactions that could have the same inputs that the consolidated transaction, so it would have to be applied without the contains, because it could happen that you can not have the transaction in your mempool but you would have another transactions with same inputs...

But, to enforce that, we should include the transaction in vt_remove instead of the hash. I could do that in other PR

data_structures/src/chain.rs
```diff
@@ -2101,22 +2119,29 @@ impl TransactionsPool {
...
+        let tx_not_added = transaction.clone();
+
         match transaction {
             Transaction::ValueTransfer(vt_tx) => {
                 let weight = f64::from(vt_tx.weight());
                 let priority = OrderedFloat(fee as f64 / weight);
 
-                self.total_vt_weight += u64::from(vt_tx.weight());
+                if fee < self.minimum_vtt_fee {
+                    added_transaction = false;
```
tmpolaczyk, line 2131: I think you can do an early return here to avoid the transaction clone and the `added_transaction` flag

```suggestion
                    // Do not insert the transaction if the fee is lower than the minimum fee
                    return vec![transaction];
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -2336,15 +2338,15 @@ pub fn log_removed_transactions(removed_transactions: &[Transaction], inserted_t
 
         if removed_the_one_we_just_inserted {
             log::trace!(
-                "Transaction {} was not added because the TransactionsPool is full",
+                "Transaction {} was not added because the fee was too low",
```
tmpolaczyk, line 2341: This is tricky, because now there are 2 reasons why a transaction is not inserted.

```suggestion
                "Transaction {} was not added",
```

node/src/actors/chain_manager/mod.rs
tmpolaczyk: So, currently, in master, this transactions could stay in the mempool forever? Open an issue

node/src/actors/chain_manager/mod.rs
drcpu-github: I changed the comment to `the fee was too low` because this is essentially also what happens when you do not add the new transaction because of the size limitations to the transaction pool. If the fee of said transaction was high enough, it would've kicked out another transaction instead of not adding the new one. Or do I misunderstand something?

tmpolaczyk: True, good idea

node/src/actors/chain_manager/mod.rs
lrubiorod: merged a PR :P

drcpu-github: I'll have a look at this and merge the newest master with my PR.

tmpolaczyk: Basically we realized that we must call `vt_remove` even if the transactions are not in the transactions pool, and that was a potential issue. You simply need to revert this line, and call `vt_remove` without checking `vt_contains`.

drcpu-github: Yes, I understood that from the discussion. 😉

Only one question, will vt_remove_inner no crash when the initial tx_key was not present? In other words will `self.vt_transactions.remove(key)` graciously stop or crash and burn?

lrubiorod (approved):

aesedepece (approved):

PR #1672, opened 2020-10-22
feat(wallet): allow re-create wallet with same seed phrase

### LINKED ISSUE #1631: [Wallet] Return error creating a wallet with already used mnemonics
When importing a wallet in sheikah, it creates the wallet and then tries to unlock it. However, if the user is reusing mnemonics an already created wallet, it will not return an error and will fail if the user introduces a different password.

Because of that, it would be useful for Sheikah receive an error when the wallet is already created.

### LINKED ISSUE #1673: [Wallet] Overwrite already created wallet with same seed phrase
The wallet should support creating a wallet whose seed phrase has already been used for creating another wallet account by specifying a flag (e.g. `overwrite`) to true.

Some relevant context for solving this issue:
- `wallet_id` is deterministically computed from seed source
- Database persistence uses wallet id as `prefix` (what happens to existing stored values if password is changed)
- Database `delete` method is not (yet?) available in the wallet

mariocao:
This PR:
- allow re-create wallet with same seed phrase using the flag `overwrite` in `create_wallet`
- return `wallet_id` in `validate_mnemonics`

Closes  #1639 (previous PR)
Closes #1673 
Closes #1631

### DIFF
--- a/wallet/src/actors/app/handlers/validate_mnemonics.rs
+++ b/wallet/src/actors/app/handlers/validate_mnemonics.rs
@@ -1,8 +1,6 @@
 use actix::prelude::*;
 use serde::{Deserialize, Serialize};
 
-use witnet_crypto::mnemonic::Mnemonic;
-
 use crate::actors::app;
 use crate::types;
 
@@ -14,7 +12,8 @@ pub struct ValidateMnemonicsRequest {
 
 #[derive(Debug, Serialize, Deserialize)]
 pub struct ValidateMnemonicsResponse {
-    pub valid: bool,
+    pub exist: bool,
+    pub wallet_id: String,
 }
 
 impl Message for ValidateMnemonicsRequest {
@@ -25,39 +24,8 @@ impl Handler<ValidateMnemonicsRequest> for app::App {
     type Result = app::ResponseActFuture<ValidateMnemonicsResponse>;
 
     fn handle(&mut self, req: ValidateMnemonicsRequest, _ctx: &mut Self::Context) -> Self::Result {
-        let validated_params = validate(req).is_ok();
-
-        let f = fut::result(Ok(ValidateMnemonicsResponse {
-            valid: validated_params,
-        }));
+        let f = self.validate_seed(req.seed_source, req.seed_data);
 
         Box::new(f)
-    }
-}
-
-struct Validated {
-    pub seed_source: types::SeedSource,
-}
-
-/// Validate `ValidateMnemonicsRequest`.
-///
-/// To be valid it must pass these checks:
-/// - seed_sources has to be `mnemonics | xprv`
-fn validate(req: ValidateMnemonicsRequest) -> Result<Validated, app::ValidationErrors> {
-    let seed_data = req.seed_data;
-
-    match req.seed_source.as_ref() {
-        "xprv" => Ok(Validated {
-            seed_source: types::SeedSource::Xprv(seed_data),
-        }),
-        "mnemonics" => Mnemonic::from_phrase(seed_data)
-            .map_err(|err| app::field_error("seed_data", format!("{}", err)))
-            .map(|seed_data| Validated {
-                seed_source: types::SeedSource::Mnemonics(seed_data),
-            }),
-        _ => Err(app::field_error(
-            "seed_source",
-            "Seed source has to be mnemonics|xprv.",
-        )),
     }
 }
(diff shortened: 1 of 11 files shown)


tmpolaczyk:
Does this supersede #1639?

mariocao:
> Does this supersede #1639?

Partially yes

tmpolaczyk:
> Partially yes

Then partially close that PR :stuck_out_tongue_closed_eyes:

wallet/src/actors/app/handlers/validate_mnemonics.rs
```diff
@@ -14,7 +12,8 @@ pub struct ValidateMnemonicsRequest {
 
 #[derive(Debug, Serialize, Deserialize)]
 pub struct ValidateMnemonicsResponse {
-    pub valid: bool,
+    pub exist: bool,
+    pub wallet_id: String,
```
tmpolaczyk, line 16: Breaking change. What's the equivalent PR in sheikah?

wallet/src/actors/worker/handlers/check_wallet_seed.rs
```diff
@@ -0,0 +1,25 @@
+use actix::prelude::*;
+
+use crate::actors::worker;
+use crate::types;
```
lrubiorod, line 4: fmt :P

wallet/src/actors/app/handlers/validate_mnemonics.rs
mariocao: I think they were not using `validate_mnemonics` until now, so they shouldn't have any problem. :smile:

tmpolaczyk: I found it here, it is used when importing mnemonics:

https://github.com/witnet/sheikah/blob/ff65864a464591ef3275b95caac82465fb867137/src/store/wallet.js#L773

wallet/src/actors/worker/methods.rs
```diff
@@ -78,6 +79,17 @@ impl Worker {
             self.params.id_hash_iterations,
         );
 
+        // Return error if `overwrite=false` and wallet already exists
+        if self
+            .wallets
+            .infos()?
+            .into_iter()
+            .any(|wallet| wallet.id == id)
+            && !overwrite
```
tmpolaczyk, line 88: Flip this condition, `if !overwrite && ....`

wallet/src/repository/wallets/mod.rs
```diff
@@ -106,11 +106,9 @@ impl<T: Database> Wallets<T> {
         let wallet_id = id.to_string();
         let lock = self.wallets_mutex.lock()?;
         let mut ids: Vec<String> = self.db.get_or_default(&keys::wallet_ids())?;
-        if !ids.contains(&wallet_id) {
```
tmpolaczyk, line 109: Maybe this was used to avoid having duplicate keys in `ids`?

Check that ids does not grow every time we re-create the same wallet:

```
ids: []
// Create wallet 111
ids: [111]
// Re-create wallet 111
ids: [111, 111]
// Re-create wallet 111
ids: [111, 111, 111]
```

mariocao: Ups...agree, the problem was the write batch being inside that condition!

wallet/src/actors/worker/handlers/create_wallet.rs
```diff
@@ -8,8 +8,12 @@ pub struct CreateWallet(
     pub Option<String>,
     /// Wallet caption
     pub Option<String>,
+    /// Wallet user-defined password
     pub types::Password,
+    /// Seed data (mnemonics or xpriv)
```
lrubiorod, line 13:
```suggestion
    /// Seed data (mnemonics or xprv)
```

wallet/src/actors/app/methods.rs
```diff
@@ -727,4 +734,38 @@ impl App {
...
+                "seed_source",
+                "Seed source has to be mnemonics|xprv.",
+            ))),
+        })
+        // Check if seed was already used in wallet
+        .and_then(|seed, slf: &mut Self, _| {
+            slf.params
+                .worker
+                .send(worker::CheckWalletSeedRequest(seed))
+                .flatten()
```
girazoki, line 763: beautiful

girazoki (approved):

lrubiorod (approved):

PR #1675, opened 2020-10-22
feat(wallet): support clearing chain data and resyncing

### LINKED ISSUE #1674: [Wallet] Allow a re-sync method in the wallet
The wallet should allow to re-sync the entire chain just in case the stored chain has some errors

aesedepece:
Also: add `delete` method in `Database` trait.

Fix #1674

### DIFF
--- a/wallet/src/repository/wallet/mod.rs
+++ b/wallet/src/repository/wallet/mod.rs
@@ -1666,9 +1666,7 @@ where
         let mut state = self.state.write()?;
 
         // Prevent chain data from being cleared if a sync or resync process is in progress
-        if state.transient_internal_addresses.is_empty()
-            && state.transient_external_addresses.is_empty()
-        {
+        if !state.is_syncing() {
             state.clear_chain_data(&self.params.genesis_prev_hash);
 
             Ok(true)
--- a/wallet/src/repository/wallet/state.rs
+++ b/wallet/src/repository/wallet/state.rs
@@ -103,4 +103,16 @@ impl State {
         self.transient_internal_addresses.clear();
         self.transient_external_addresses.clear();
     }
+
+    /// Tell whether the wallet is undergoing a synchronization.
+    ///
+    /// Currently, this function derives the information from the presence or absence of transient
+    /// addresses, i.e. addresses that are temporarily generated during the synchronization process.
+    ///
+    /// TODO: refactor all synchronization-related fields into a `SynchronizationState` structure
+    ///  so that the number of fields in `State` does not keep growing, and take that as a chance
+    ///  to add a private `is_syncing` field for which this would act as a getter.
+    pub fn is_syncing(&self) -> bool {
+        self.transient_internal_addresses.is_empty() && self.transient_external_addresses.is_empty()
+    }
 }


wallet/src/repository/wallet/mod.rs
```diff
@@ -1645,6 +1645,20 @@ where
...
+    /// - Synchronization status
+    /// - Balances
+    /// - Movements
+    /// - Addresses and their metadata
+    pub fn clear_chain_data(&self) -> Result<()> {
+        let mut state = self.state.write()?;
+        state.clear_chain_data(&self.params.genesis_prev_hash);
+
+        Ok(())
+    }
```
aesedepece, line 1661: This is were theoretically we could actually delete data from the storage.

wallet/src/db/encrypted/mod.rs
```diff
@@ -72,6 +72,15 @@ impl Database for EncryptedDb {
...
 
+    fn delete<K>(&self, key: &K) -> Result<()>
+    where
+        K: AsRef<[u8]>,
+    {
+        self.as_ref().delete(key)?;
+
+        Ok(())
+    }
+
```
mariocao, line 83: This key is not correct! :) Maybe we should take this delete feature from this PR as it is not used.

mariocao (changes requested):

mariocao:
As you mentioned to me, re-syncing several times may lead to some problems.
We should about this by avoiding re-syncing while syncing using this condition:
```
        // Exit if not syncing
        if state.transient_internal_addresses.is_empty()
            && state.transient_external_addresses.is_empty()
```

wallet/src/repository/wallet/mod.rs
```diff
@@ -1645,6 +1645,32 @@ where
...
+    ///
+    /// In order to prevent data race conditions, resyncing is not allowed while a sync or resync
+    /// process is already in progress. Accordingly, this function returns whether chain data has
+    /// been cleared or not.
+    pub fn clear_chain_data(&self) -> Result<bool> {
+        let mut state = self.state.write()?;
+
+        // Prevent chain data from being cleared if a sync or resync process is in progress
+        if state.transient_internal_addresses.is_empty()
+            && state.transient_external_addresses.is_empty()
```
tmpolaczyk, line 1665: I would prefer using a dedicated "syncing" flag instead of reusing other fields.

wallet/src/actors/worker/methods.rs
```diff
@@ -994,4 +994,28 @@ impl Worker {
...
+    /// In order to prevent data race conditions, resyncing is not allowed while a sync or resync
+    /// process is already in progress. Accordingly, this function returns whether chain data has
+    /// been cleared or not.
+    pub fn clear_chain_data_and_resync(
+        &self,
+        wallet_id: &str,
+        wallet: types::SessionWallet,
+        sink: DynamicSink,
+    ) -> Result<bool> {
+        // Only trigger sync of chain data has been cleared
```
girazoki, line 1015:
```suggestion
        // Only trigger sync of chain after data has been cleared
```

wallet/src/repository/wallet/mod.rs
```diff
@@ -1645,6 +1635,30 @@ where
...
+    /// - Addresses and their metadata
+    ///
+    /// In order to prevent data race conditions, resyncing is not allowed while a sync or resync
+    /// process is already in progress. Accordingly, this function returns whether chain data has
+    /// been cleared or not.
+    pub fn clear_chain_data(&self) -> Result<bool> {
+        let mut state = self.state.write()?;
+
+        // Prevent chain data from being cleared if a sync or resync process is in progress
+        if state.is_syncing() {
```
girazoki, line 1654: If I understand correctly, if is_syncing is true we clear data and issue a re-sync. I would change the naming as this is a bit confusing to me

girazoki (approved):

PR #1676, opened 2020-10-23
feat(node): persist chain state during synchronization

### LINKED ISSUE #1532: Persist chain state during synchronizing

lrubiorod:
Close #1532

### DIFF
--- a/node/src/actors/chain_manager/handlers.rs
+++ b/node/src/actors/chain_manager/handlers.rs
@@ -399,6 +399,9 @@ impl Handler<AddBlocks> for ChainManager {
                         log::debug!("TargetNotReached: superblock target #{} not reached, requesting more blocks. ({} processed blocks)",
                             sync_target.superblock.checkpoint, num_processed_blocks);
                         self.request_blocks_batch(ctx);
+
+                        // Copy current chain state into previous chain state, and persist it
+                        self.persist_chain_state(None, ctx);
                     }
                     // SyncWithoutCandidate:
                     // 1. process blocks
--- a/node/src/actors/chain_manager/mod.rs
+++ b/node/src/actors/chain_manager/mod.rs
@@ -251,12 +251,20 @@ impl SystemService for ChainManager {}
 /// Auxiliary methods for ChainManager actor
 impl ChainManager {
     /// Persist previous chain state into storage
-    fn persist_chain_state(&mut self, superblock_index: u32, ctx: &mut Context<Self>) {
-        let previous_chain_state = self.chain_state_snapshot.restore(superblock_index);
-        if previous_chain_state.is_none() {
-            return;
-        }
-        let previous_chain_state = previous_chain_state.unwrap();
+    /// None case: persist current chain state into storage (during synchronization)
+    fn persist_chain_state(&mut self, superblock_index: Option<u32>, ctx: &mut Context<Self>) {
+        let previous_chain_state = if let Some(superblock_index) = superblock_index {
+            let chain_state_snapshot = self.chain_state_snapshot.restore(superblock_index);
+
+            if chain_state_snapshot.is_none() {
+                return;
+            }
+
+            chain_state_snapshot.unwrap()
+        } else {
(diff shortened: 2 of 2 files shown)


node/src/actors/chain_manager/handlers.rs
```diff
@@ -399,6 +399,10 @@ impl Handler<AddBlocks> for ChainManager {
                         log::debug!("TargetNotReached: superblock target #{} not reached, requesting more blocks. ({} processed blocks)",
                             sync_target.superblock.checkpoint, num_processed_blocks);
                         self.request_blocks_batch(ctx);
+
+                        // Copy current chain state into previous chain state, and persist it
+                        self.move_chain_state_forward(0);
+                        self.persist_chain_state(0, ctx);
```
tmpolaczyk, line 405: This `0` should be the superblock index, but here we do not use the superblock index, we use the sync blocks batch. How about we change it into an option, and use `None` here but `Some(_)` everywhere else?

This way in theory we can distinguish if the chain state is valid because it is a consolidated block, or it may be invalid because we received it while syncing.

tmpolaczyk:
So the main problem with this PR is that a malicious peer can send a forked blocks batch, and the syncing node will never detect it, it will get stuck in synchronizing forever. Before this PR this could be fixed by restarting the node, because that restarts the synchronization.

lrubiorod:
> So the main problem with this PR is that a malicious peer can send a forked blocks batch, and the syncing node will never detect it, it will get stuck in synchronizing forever. Before this PR this could be fixed by restarting the node, because that restarts the synchronization.

I think at the beginning we should assume that we have to trust in our outbound peers. Optimistic Rollup!! The problem to do not merge a solution to this, is when the blockchain would be too big, to have a problem and start from zero could be super inefficient

Before this PR, the same malicious peer could send a forked batch and force to start again, and he can repeat it to get the other peer stuck

lrubiorod:
> Do we need to add a `Resync` method to the node as well? laughing
Meybe in a new issue :P

girazoki:
In any case deleting the storage would work, and I do not see that as something very difficult to do in such a situation

girazoki (approved):

node/src/actors/chain_manager/mod.rs
```diff
@@ -273,14 +280,16 @@ impl ChainManager {
...
+            log::debug!(
+                "Persisting chain state for superblock #{} with chain beacon {:?} and super beacon {:?}",
+                superblock_index,
+                chain_beacon,
+                superblock_beacon
+            );
 
-        assert_eq!(superblock_beacon.checkpoint, superblock_index);
+            assert_eq!(superblock_beacon.checkpoint, superblock_index);
+        }
```
tmpolaczyk, line 292: Add a different log when superblock_index is None? "Persisting chain state during synchronization, chain beacon: {:?}"

node/src/actors/chain_manager/handlers.rs
```diff
@@ -399,6 +399,10 @@ impl Handler<AddBlocks> for ChainManager {
                         log::debug!("TargetNotReached: superblock target #{} not reached, requesting more blocks. ({} processed blocks)",
                             sync_target.superblock.checkpoint, num_processed_blocks);
                         self.request_blocks_batch(ctx);
+
+                        // Copy current chain state into previous chain state, and persist it
+                        self.move_chain_state_forward(None);
```
tmpolaczyk, line 404: `self.move_chain_state_forward(None)` doesn't do anything, I think we can remove this line and revert the option back to a u32. Same with `.take()`. But we need to document that `self.persist_chain_state(None, ctx)` always persists the current chain state.


```suggestion
```

tmpolaczyk (approved):

PR #1677, opened 2020-10-23
Create Node status notification

### LINKED ISSUE #1662: Create node status event
The node should send an event when its status changes. This way Sheikah will be able to update the wallet's node synchronization

clbartoli:
This PR creates a notification in the node every time its status changes. This way the wallet will inform Sheikah, and Sheikahk will update the node status.

Closes #1662


node/src/actors/messages.rs
```diff
@@ -481,6 +481,8 @@ pub enum StoreInventoryItem {
     Transaction(Hash, PointerToBlock),
     /// Superblocks are stored as the list of block hashes
     Superblock(SuperBlockNotify),
+    /// States of the node
+    StateMachine(ChangeNodeStatusNotify),
```
tmpolaczyk, line 485: This does not belong here

```suggestion
```

node/src/actors/session/handlers.rs
```diff
@@ -690,7 +690,7 @@ fn inventory_process_transaction(
     transaction: Transaction,
 ) {
     // Get ChainManager address
-    let chain_manager_addr = ChainManager::from_registry();
+    let chain_manager_handladdr = ChainManager::from_registry();
```
tmpolaczyk, line 693:
```suggestion
    let chain_manager_addr = ChainManager::from_registry();
```

wallet/src/actors/worker/methods.rs
```diff
@@ -963,6 +963,26 @@ impl Worker {
         Ok(())
     }
 
+    pub fn handle_change_node_status(
+        &self,
+        status: types::StateMachine,
+        _confirmed: bool,
```
tmpolaczyk, line 969: "confirmed" refers to blocks, it doesn't apply here

```suggestion
```

wallet/src/actors/worker/methods.rs
```diff
@@ -963,6 +963,26 @@ impl Worker {
...
+        wallet: types::SessionWallet,
+        sink: types::DynamicSink,
+    ) -> Result<()> {
+        let changed_node_status = status;
+        log::debug!(
+            "The node has chaged its status into {:?}",
+            changed_node_status
```
tmpolaczyk, line 976: I think `changed_node_status` variable is not needed?

```suggestion
            status
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -1466,7 +1446,7 @@ impl ChainManager {
                 next_state
             ),
         }
-
+        self.notify_change_node_status(next_state);
```
tmpolaczyk, line 1449: This will be executed even if the state does not change, we can check that `old != new` to avoid sending redundant notifications, or we can leave it like this if it's not a problem.

node/src/actors/chain_manager/mod.rs
```diff
@@ -1667,6 +1647,12 @@ impl ChainManager {
             JsonRpcServer::from_registry().do_send(superblock_notify);
         }
     }
+
+    /// Let JSON-RPC clients know that if the nodes changes its status
```
aesedepece, line 1651:
```suggestion
    /// Let JSON-RPC clients know when the node changes its status
```

node/src/actors/messages.rs
```diff
@@ -959,3 +959,14 @@ pub struct SuperBlockNotify {
     /// The hashes of the blocks that we are signaling as consolidated.
     pub consolidated_block_hashes: Vec<Hash>,
 }
+
+/// Notification signaling that the node's state has changed.
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub struct ChangeNodeStatusNotify {
```
aesedepece, line 965:
```suggestion
pub struct NodeStatusNotify {
```

node/src/actors/messages.rs
```diff
@@ -959,3 +959,14 @@ pub struct SuperBlockNotify {
...
+    /// The node status.
+    pub node_status: StateMachine,
```
tmpolaczyk, line 967: I think a client only needs to know if the node is synced or not, so what do you think about changing this into a simple flag? This way we would avoid moving `StateMachine` to `chain.rs`

```suggestion
    pub node_synced: bool,
```

wallet/src/actors/worker/handlers/handle_change_node_status.rs
```diff
@@ -0,0 +1,26 @@
+use actix::{Handler, Message};
+
+use crate::actors::worker;
+use crate::types;
+
+pub struct HandleChangeNodeStatusRequest {
```
aesedepece, line 6:
```suggestion
pub struct NodeStatusRequest {
```

wallet/src/actors/worker/methods.rs
```diff
@@ -978,6 +978,21 @@ impl Worker {
         Ok(())
     }
 
+    pub fn handle_change_node_status(
```
aesedepece, line 981:
```suggestion
    pub fn handle_change_node_status(
```

wallet/src/types.rs
```diff
@@ -234,6 +234,8 @@ pub enum Event {
     BlocksOrphan(Vec<String>),
     /// A new movement (transaction) affecting balance.
     Movement(model::BalanceMovement),
+    /// Node status has changed
+    NodeStatusChanged(StateMachine),
```
aesedepece, line 238:
```suggestion
    NodeStatus(StateMachine),
```

aesedepece (changes requested):

node/src/actors/chain_manager/mod.rs
aesedepece: I guess that redundant notifications don't do any harm, and we could speculate that maybe they could be even helpful in some cases. I'm for leaving it in here and just wait and see if they are too spammy.

node/src/actors/chain_manager/mod.rs
```diff
@@ -1680,6 +1660,12 @@ impl ChainManager {
             JsonRpcServer::from_registry().do_send(superblock_notify);
         }
     }
+
+    /// Let JSON-RPC clients know when the node changes its status
+    fn notify_change_node_status(&mut self, node_status: StateMachine) {
```
aesedepece, line 1665:
```suggestion
    fn notify_node_status(&mut self, node_status: StateMachine) {
```

node/src/actors/messages.rs
aesedepece: I kind of like more serving raw states as they give clients the chance to offer more insightful and helpful information about what's up with the node they are using in the back-end.

node/src/actors/json_rpc/server.rs
```diff
@@ -237,3 +233,35 @@ impl Handler<SuperBlockNotify> for JsonRpcServer {
...
+impl Handler<NodeStatusNotify> for JsonRpcServer {
+    type Result = ();
+
+    fn handle(&mut self, msg: NodeStatusNotify, ctx: &mut Self::Context) -> Self::Result {
+        if let Ok(subs) = self.subscriptions.lock() {
+            let empty_map = HashMap::new();
+            for (subscription, (sink, _subscription_params)) in
+                subs.get("status").unwrap_or(&empty_map)
+            {
+                log::debug!("Sending node status notification");
```
aesedepece, line 246:
```suggestion
                log::debug!("Sending node status notification ({})", msg.node_status);
```

wallet/src/actors/app/methods.rs
```diff
@@ -574,6 +575,26 @@ impl App {
         Ok(())
     }
 
+    /// Handle node status notifications received from a Witnet node.
+    pub fn handle_node_status_notification(&mut self, value: types::Json) -> Result<()> {
+        let status = serde_json::from_value::<types::StateMachine>(value).map_err(node_error)?;
```
aesedepece, line 580: Feel free to use this syntax if it looks better to you. 
```suggestion
        let status: StateMachine = serde_json::from_value(value).map_err(node_error)?;
```

wallet/src/actors/app/methods.rs
```diff
@@ -574,6 +575,26 @@ impl App {
...
+            .state
+            .wallets
+            .iter()
+            .map(|(_, wallet)| wallet.clone())
+            .collect();
+
+        for wallet in wallets.iter() {
+            let sink = self.state.get_sink(&wallet.session_id);
+            self.handle_change_status_in_worker(&status, &wallet, sink.clone());
+        }
```
aesedepece, line 593: I think you don't even need to `.iter()` in `for` statements, and in this case I think we can even iterate over a slice.
```suggestion
        for wallet in &wallets {
            let sink = self.state.get_sink(wallet.session_id);
            self.handle_change_status_in_worker(&status, wallet, sink.clone());
        }
```

wallet/src/actors/app/methods.rs
```diff
@@ -607,6 +628,24 @@ impl App {
...
+    /// server thread, so as not to lock the rest of the application.
+    pub fn handle_change_status_in_worker(
+        &self,
+        status: &types::StateMachine,
+        wallet: &types::SessionWallet,
+        sink: types::DynamicSink,
+    ) {
+        // NOTE: Possible enhancement.
+        // Maybe is a good idea to use a shared reference Arc instead of cloning `block` altogether,
+        // moreover when this method is called iteratively by `handle_block_notification`.
```
aesedepece, line 641: This method has nothing to do with `handle_block_notification` :wink: 
```suggestion
        // Maybe is a good idea to use a shared reference Arc instead of cloning `block` altogether
```

wallet/src/actors/worker/methods.rs
```diff
@@ -1007,6 +1007,21 @@ impl Worker {
         Ok(())
     }
 
+    pub fn handle_node_status(
+        &self,
+        status: types::StateMachine,
+        wallet: types::SessionWallet,
+        sink: types::DynamicSink,
+    ) -> Result<()> {
+        log::debug!("The node has chaged its status into {:?}", status);
```
aesedepece, line 1016:
```suggestion
        log::debug!("The node has changed its status into {:?}", status);
```

data_structures/src/chain.rs
```diff
@@ -67,6 +67,27 @@ pub struct ChainInfo {
     pub highest_vrf_output: CheckpointVRF,
 }
 
+/// State Machine
```
aesedepece, line 70:
```suggestion
/// State machine for the synchronization status of a Witnet node
```

aesedepece (changes requested):

data_structures/src/chain.rs
```diff
@@ -67,6 +67,27 @@ pub struct ChainInfo {
...
+#[derive(Copy, Clone, Deserialize, Debug, Eq, PartialEq, Serialize)]
+pub enum StateMachine {
+    /// First state, ChainManager is waiting to consensus between its peers
```
mariocao, line 73:
```suggestion
    /// First state, ChainManager is waiting for reaching consensus between its peers
```

mariocao (commented):
Great PR!!! :fireworks:

tmpolaczyk (approved):

aesedepece (approved):

PR #1678, opened 2020-10-23
Update wallet infos

girazoki:
This PR modfies how wallet infos is retrieved. The caption field is renamed to description, and wallet_infos only returns the name. The description is only stored in the private encrypted section of the wallet DB.

### DIFF
--- a/wallet/src/actors/app/handlers/create_wallet.rs
+++ b/wallet/src/actors/app/handlers/create_wallet.rs
@@ -9,7 +9,7 @@ use crate::types;
 #[derive(Debug, Serialize, Deserialize)]
 pub struct CreateWalletRequest {
     name: Option<String>,
-    caption: Option<String>,
+    description: Option<String>,
     password: types::Password,
     seed_source: String,
     seed_data: types::Password,
@@ -36,7 +36,7 @@ impl Handler<CreateWalletRequest> for app::App {
                 params.password,
                 params.seed_source,
                 params.name,
-                params.caption,
+                params.description,
                 params.overwrite,
             )
             .map(|wallet_id| CreateWalletResponse { wallet_id })
@@ -48,7 +48,7 @@ impl Handler<CreateWalletRequest> for app::App {
 }
 
 struct Validated {
-    pub caption: Option<String>,
+    pub description: Option<String>,
     pub name: Option<String>,
     pub overwrite: bool,
     pub password: types::Password,
@@ -62,7 +62,7 @@ struct Validated {
 /// - seed_sources has to be `mnemonics | xprv`
 fn validate(req: CreateWalletRequest) -> Result<Validated, app::ValidationErrors> {
     let name = req.name;
-    let caption = req.caption;
+    let description = req.description;
     let seed_data = req.seed_data;
     let source = match req.seed_source.as_ref() {
         "xprv" => Ok(types::SeedSource::Xprv(seed_data)),
@@ -85,7 +85,7 @@ fn validate(req: CreateWalletRequest) -> Result<Validated, app::ValidationErrors
     let overwrite = req.overwrite.unwrap_or(false);
 
     app::combine_field_errors(source, password, move |seed_source, password| Validated {
-        caption,
+        description,
         name,
         overwrite,
         password,
--- a/wallet/src/actors/worker/handlers/create_wallet.rs
+++ b/wallet/src/actors/worker/handlers/create_wallet.rs
@@ -6,7 +6,7 @@ use crate::types;
 pub struct CreateWallet(
     /// Wallet name
     pub Option<String>,
-    /// Wallet deescription
+    /// Wallet description
     pub Option<String>,
     /// Wallet user-defined password
     pub types::Password,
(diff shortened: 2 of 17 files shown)


wallet/src/repository/keys.rs
```diff
@@ -12,10 +12,10 @@ pub fn wallet_name() -> &'static str {
     "name"
 }
 
-/// A wallet's caption.
+/// A wallet's description.
 #[inline]
-pub fn wallet_caption() -> &'static str {
-    "caption"
+pub fn wallet_description() -> &'static str {
+    "description"
```
tmpolaczyk, line 18: Any changes to the database keys are BREAKING CHANGE

tmpolaczyk: In this case it's not a breaking change because `wallet_description` was always read using `get_opt`, so nothing breaks if it does not exist, and the old `wallet_caption` was not used, so nothing breaks if we delete it.

girazoki: Yep

wallet/src/actors/app/handlers/create_wallet.rs
```diff
@@ -9,7 +9,7 @@ use crate::types;
 #[derive(Debug, Serialize, Deserialize)]
 pub struct CreateWalletRequest {
     name: Option<String>,
-    caption: Option<String>,
+    description: Option<String>,
```
tmpolaczyk, line 12: As always, we must remember to change this in sheikah.

wallet/src/actors/worker/handlers/create_wallet.rs
```diff
@@ -6,7 +6,7 @@ use crate::types;
 pub struct CreateWallet(
     /// Wallet name
     pub Option<String>,
-    /// Wallet caption
+    /// Wallet deescription
```
mariocao, line 9:
```suggestion
    /// Wallet description
```

wallet/src/actors/worker/methods.rs
```diff
@@ -149,21 +149,16 @@ impl Worker {
...
 
     /// Update the wallet information in the infos database.
-    pub fn update_wallet_info(
-        &self,
-        wallet_id: &str,
-        name: Option<String>,
-        caption: Option<String>,
-    ) -> Result<()> {
-        self.wallets.update_info(wallet_id, name, caption)?;
+    pub fn update_wallet_info(&self, wallet_id: &str, name: Option<String>) -> Result<()> {
```
mariocao, line 160: I would make `update_info` require a `session_id` so that it could only be changed by an authenticated client.
Additionally, I would allow the client to change the description too (as optional param).

mariocao (changes requested):

wallet/src/actors/worker/methods.rs
mariocao: As we discussed, I was wrong! :smile:

mariocao (approved):

wallet/src/repository/wallets/mod.rs
```diff
@@ -73,19 +66,22 @@ impl<T: Database> Wallets<T> {
...
             salt,
             account,
         } = wallet_data;
         let mut wbatch = wallet_db.batch();
 
-        if let Some(name) = name {
+        // We first write name and description into private wallet DB
+        if let Some(name) = name.clone() {
             wbatch.put(keys::wallet_name(), name)?;
         }
```
aesedepece, line 79: Maybe something like this could work?
```suggestion
        if let &Some(name) = name.as_ref() {
            wbatch.put(keys::wallet_name(), name.clone())?;
        }
```

aesedepece: Or maybe we can do `batch.put` directly here as well and avoid having this `if` twice?

wallet/src/repository/wallets/mod.rs
```diff
@@ -116,9 +117,10 @@ impl<T: Database> Wallets<T> {
         Ok(())
     }
 
+    /// Gets a wallet's salt and IV based on its provided ID
```
aesedepece, line 120:
```suggestion
    /// Get a wallet's salt and IV based on its provided ID
```

aesedepece (changes requested):

wallet/src/repository/wallets/mod.rs
```diff
@@ -116,9 +117,10 @@ impl<T: Database> Wallets<T> {
...
     pub fn wallet_salt_and_iv(&self, id: &str) -> Result<(Vec<u8>, Vec<u8>)> {
-        let salt = self.db.get(&keys::wallet_id_salt(id))?;
-        let iv = self.db.get(&keys::wallet_id_iv(id))?;
+        let salt = self.db.get(&keys::wallet_id_salt(&id))?;
+        let iv = self.db.get(&keys::wallet_id_iv(&id))?;
```
tmpolaczyk, line 123: This shouldn't be needed because `id` is already a reference

```suggestion
        let salt = self.db.get(&keys::wallet_id_salt(id))?;
        let iv = self.db.get(&keys::wallet_id_iv(id))?;
```

tmpolaczyk (approved):

wallet/src/repository/wallets/mod.rs
girazoki: I liked the second option and already pushed it

wallet/src/actors/worker/handlers/create_wallet.rs
aesedepece: Ping

girazoki: done

aesedepece (approved):

wallet/src/actors/app/handlers/create_wallet.rs
aesedepece: This is being worked by @Tommytrg  in https://github.com/witnet/sheikah/issues/1486 and https://github.com/witnet/sheikah/issues/1476

PR #1679, opened 2020-10-23
fix(wallet): update address info only if mint or vtt

mariocao:
This PR fixes the update of address payments.

Previously, if Tally or Data Request transactions were returning some change or refunds to the originating address, the `received payments` field of an address was updated wrongly with that "re-funded" amount.

### DIFF
--- a/wallet/src/repository/wallet/mod.rs
+++ b/wallet/src/repository/wallet/mod.rs
@@ -1179,50 +1179,57 @@ where
             .checked_add(1)
             .ok_or_else(|| Error::TransactionIdOverflow)?;
 
-        // Update addresses and their information if there were payments (new UTXOs)
-        let mut addresses = vec![];
+        // Update addresses (externals/internals) and their information if there were payments (new UTXOs)
+        //
+        // - Data Request and Tally transactions are ignored as they only contain refunds to data request
+        // creators. By protocol the tally output can only be set to the first used input of the DR.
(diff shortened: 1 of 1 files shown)


girazoki (approved):

wallet/src/repository/wallet/mod.rs
```diff
@@ -1174,50 +1174,62 @@ where
...
+                        info: model::AddressInfo {
+                            db_key: keys::address_info(path.account, path.keychain, path.index),
+                            label: info.label.clone(),
+                            received_payments,
+                            received_amount: info.received_amount + key_balance.amount,
+                            first_payment_date,
+                            last_payment_date: Some(current_timestamp),
+                        },
+                        pkh: old_address.pkh,
+                    };
```
aesedepece, line 1218: Is [struct update syntax](https://doc.rust-lang.org/stable/book/ch05-01-defining-structs.html#creating-instances-from-other-instances-with-struct-update-syntax) possible here?
```suggestion
                    let updated_address = model::Address {
                        info: model::AddressInfo {
                            db_key: keys::address_info(path.account, path.keychain, path.index),
                            label: info.label.clone(),
                            received_payments,
                            received_amount: info.received_amount + key_balance.amount,
                            first_payment_date,
                            last_payment_date: Some(current_timestamp),
                        },
                        ..old_address
                    };
```

aesedepece (changes requested):

aesedepece (approved):

PR #1683, opened 2020-10-26
feat(transactions_pool): enforce same input transactions cleaning

lrubiorod:


aesedepece:
May I understand that this is a follow-up to #1671?

tmpolaczyk:
> May I understand that this is a follow-up to #1671?

No, this is an issue that existed before but we found it when reviewing #1671

tmpolaczyk:
Add tests: copy `transactions_pool_remove_all_transactions_with_same_output_pointer` into a new test and check all the combinations like this:

```rust
        let mut transactions_pool = TransactionsPool::default();
        // Do not insert vt1
        //transactions_pool.insert(vt1.clone(), 1);
        transactions_pool.insert(vt2.clone(), 1);
        // But remove vt1, maybe it was included in a block before the node received it
        let t = transactions_pool.vt_remove(&vt_1);
        assert_eq!(t, None);
        assert!(!transactions_pool.contains(&vt1).unwrap());
        assert!(!transactions_pool.contains(&vt2).unwrap());
```

tmpolaczyk (approved):

mariocao (approved):

PR #1686, opened 2020-10-27
fix: allow resynchronizing while synchronizing

aesedepece:


wallet/src/repository/wallet/mod.rs
```diff
@@ -1668,14 +1668,22 @@ where
...
             state.clear_chain_data(&self.params.genesis_prev_hash);
 
             Ok(true)
         } else {
             Ok(false)
         }
     }
+
+    /// Run a predicate on the state of a wallet in a thread safe manner, thanks to a write lock.
+    pub fn apply_on_state<P, O>(&self, predicate: P) -> Result<O>
```
tmpolaczyk, line 1681: This is a bad idea because this function can only succeed if the write lock was not held, and there are many functions in our code that take a write lock at the top of the function. This is why there are function names that start with `_`. I would prefer to be explicit about the lock, and therefore remove this function.

wallet/src/repository/wallet/state.rs
```diff
@@ -103,17 +109,49 @@ impl State {
...
+/// The synchronization state and information for a wallet.
+#[derive(Debug)]
+pub struct SynchronizationStatus {
+    is_syncing: bool,
+    is_resyncing: bool,
+}
+
+impl SynchronizationStatus {
+    /// Tell whether a wallet is syncing.
+    #[allow(dead_code)]
```
tmpolaczyk, line 123:
```suggestion
```

wallet/src/repository/wallet/state.rs
```diff
@@ -103,17 +109,49 @@ impl State {
...
+    pub fn set_syncing(&mut self, new_status: bool) -> bool {
+        let old_status = self.is_syncing;
+        self.is_syncing = new_status;
+
+        old_status
+    }
+
+    /// Set the resynchronization status. Returns the old status.
+    pub fn set_resyncing(&mut self, new_status: bool) -> bool {
+        let old_status = self.is_syncing;
```
tmpolaczyk, line 143: Copy-paste error?

```suggestion
        let old_status = self.is_resyncing;
```

wallet/src/repository/wallet/mod.rs
```diff
@@ -1668,14 +1668,22 @@ where
         let mut state = self.state.write()?;
 
         // Prevent chain data from being cleared if a sync or resync process is in progress
-        if !state.is_syncing() {
+        if !state.synchronization.is_resyncing() {
```
tmpolaczyk, line 1671: The comment does not match the code

wallet/src/repository/wallet/state.rs
aesedepece: Yup hehehe

wallet/src/actors/worker/methods.rs
```diff
@@ -615,6 +615,8 @@ impl Worker {
             self.params.sync_address_batch_length,
             self.params.sync_address_batch_length,
         )?;
+        // Flag that a synchronization is in progress
+        wallet.apply_on_state(|mut state| state.synchronization.set_syncing(true))?;
```
tmpolaczyk, line 619: Shouldn't this check if that flag is already set to true, and in that case return some error?

wallet/src/actors/worker/methods.rs
```diff
@@ -1063,7 +1067,13 @@ impl Worker {
     ) -> Result<bool> {
         // Only trigger sync of chain if data has been cleared
         match wallet.clear_chain_data()? {
-            true => self.sync(wallet_id, wallet, sink, true).map(|_| true),
+            true => {
+                wallet.apply_on_state(|mut state| state.synchronization.set_resyncing(true))?;
```
tmpolaczyk, line 1071: Same as above, shouldn't this check if that flag is already set to true, and in that case return some error?

wallet/src/repository/wallet/state.rs
aesedepece: Are you suggesting removing or commenting out the function?

wallet/src/actors/worker/methods.rs
aesedepece: I didn't dare to change any existing behavior. I'm setting the `is_syncing` flag just the sake of completion, but as you saw it's not being used anywhere :shrug:

wallet/src/repository/wallet/state.rs
tmpolaczyk: Removing, public functions cannot be dead code. If clippy complains when you remove it may be because it's not actually public, it may be inside a private module or something. In that case you can keep the allow(dead_code).

wallet/src/actors/worker/methods.rs
aesedepece: That's checked inside `wallet.clear_chain_data`. What do you think about moving that guard outside so that `clear_chain_data` doesn't need to care about the synchronization status?

tmpolaczyk: I don't know, it sounds like `clear_chain_data` should return an error if the wallet is syncing, or cancel the synchronization

wallet/src/actors/worker/methods.rs
tmpolaczyk: Also, if this function returns using a `?` on some error, this flag will never be set to `false`.

wallet/src/repository/wallet/mod.rs
tmpolaczyk: I see that it's useful, so we could rename it to something that mentions a write lock. `apply_on_state_write_lock` for example

wallet/src/repository/wallet/state.rs
aesedepece: I marked it as dead code only because I believe we may use it in the future, but just not yet. My take is that `impl` for data structures are let's say "library code", so, to the greatest extent possible, their API shouldn't be conditioned by usage.

wallet/src/actors/worker/methods.rs
aesedepece: We can't do that. At least for the time being. Because that's exactly the issue we're having right now. `clear_chain_data` is blocking any resynchronization for wallets that are forked.

wallet/src/repository/wallet/mod.rs
aesedepece: I get your point about the write lock. This is sort of dangerous because it can lead to code that looks right at compile time but then causes a deadlock. However, every each of the functions that start with `_` have their public counterparts that acquire a lock in exactly this same way. I don't see why this would be worse than those.

wallet/src/actors/worker/methods.rs
```diff
@@ -1062,9 +1062,16 @@ impl Worker {
         sink: DynamicSink,
     ) -> Result<bool> {
         // Only trigger sync of chain if data has been cleared
```
tmpolaczyk, line 1064:
```suggestion
        // Only trigger sync of chain if the wallet is not already resyncing
```
The function documentation is also outdated. What would happen if the resync is started while a normal sync is already in place?

tmpolaczyk (approved):

PR #1694, opened 2020-10-28
Refactor wallet and connect wallet with build_vtt

### LINKED ISSUE #1633: [Wallet] Implement a "smarter" `create_transaction_components`
Currently the `create_transaction_components` function just takes UTXOs by order of insertion.
Thus, it is not an optimal algorithm for creating UTXO and it does not follow any strategy.

Same sample strategies might be:
- minimizing transaction weight 
- decrease "dust" (small UTXOs)

lrubiorod:
Close #1633


wallet/src/model.rs
```diff
@@ -218,44 +219,6 @@ pub struct Transactions {
     pub total: u32,
 }
 
-#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, Hash)]
-pub struct OutPtr {
```
tmpolaczyk, line 222: Removing this may be a breaking change, depending on the layout of the new type in the database

data_structures/src/chain.rs
```diff
@@ -2648,15 +2648,12 @@ pub struct NodeStats {
 pub struct ChainState {
     /// Blockchain information data structure
     pub chain_info: Option<ChainInfo>,
-    /// Unspent Outputs Pool
-    pub unspent_outputs_pool: UnspentOutputsPool,
+    /// Unspent Outputs Pool (general and own)
+    pub node_unspent_outputs_pool: NodeUtxos,
```
tmpolaczyk, line 2652: Breaking change :warning:

wallet/src/types.rs
```diff
@@ -19,7 +19,7 @@ pub use witnet_data_structures::{
         Block as ChainBlock, CheckpointBeacon, DataRequestInfo, DataRequestOutput, Epoch, Hash,
         HashParseError, Hashable, Input as TransactionInput, KeyedSignature, OutputPointer,
         PublicKey, PublicKeyHash, PublicKeyHashParseError, RADAggregate, RADRequest, RADRetrieve,
-        RADTally, SuperBlock, ValueTransferOutput as VttOutput,
```
tmpolaczyk, line 22: I approve this change

girazoki: me too

data_structures/src/transaction_factory.rs
```diff
@@ -5,57 +5,63 @@ use crate::{
     },
     error::TransactionError,
     transaction::{DRTransactionBody, VTTransactionBody},
+    utxo_pool::NodeUtxos,
```
girazoki, line 8: fmt

data_structures/src/utxo_pool.rs
```diff
@@ -147,6 +150,51 @@ impl OwnUnspentOutputsPool {
     }
 }
 
+/// Struct that keep the unspent outputs pool and the own unspent outputs pool
```
girazoki, line 153:
```suggestion
/// Struct that keeps the unspent outputs pool and the own unspent outputs pool
```

data_structures/src/transaction_factory.rs
```diff
@@ -5,57 +5,63 @@ use crate::{
...
-        if all_utxos.get(op).unwrap().time_lock > timestamp {
-            continue;
-        }
+        let value = utxos.get_value(op).unwrap();
+        total = total
+            .checked_add(value)
+            .ok_or_else(|| TransactionError::OutputValueOverflow)?;
 
-        if timestamp < *ts {
+        let time_lock = utxos.get_time_lock(op).unwrap();
```
girazoki, line 55:
```suggestion
        If let Some(time_lock) = get_time_lock
```

wallet/src/repository/error.rs
```diff
@@ -43,6 +44,8 @@ pub enum Error {
     HashParseError(#[cause] types::HashParseError),
     #[fail(display = "failed because wallet is still syncing: {}", _0)]
     StillSyncing(String),
+    #[fail(display = "failed creating a transaction: {}", _0)]
+    TransactionCreation(#[cause] TransactionError),
```
girazoki, line 48:
```suggestion
    TransactionCreationError(#[cause] TransactionError),
```

lrubiorod: Other repository errors has not the Error sufix

data_structures/src/utxo_pool.rs
```diff
@@ -147,6 +150,51 @@ impl OwnUnspentOutputsPool {
...
+#[derive(Debug)]
+pub struct NodeUtxos<'a> {
+    pub all_utxos: &'a UnspentOutputsPool,
+    pub own_utxos: &'a mut OwnUnspentOutputsPool,
+}
+
+impl<'a> UtxoFantasy for NodeUtxos<'a> {
+    fn sort_iter(&self, strategy: UtxoSelectionStrategy) -> Vec<OutputPointer> {
```
tmpolaczyk, line 161: The result of this function is always used as an iterator, so we can just return an iterator here. This should allow us to remove the unneeded collect in the `Random` implementation.

```suggestion
    fn sort_iter(&self, strategy: UtxoSelectionStrategy) -> Box<dyn Iterator Item=OutputPointer> {
```

data_structures/src/transaction_factory.rs
```diff
@@ -5,57 +5,65 @@ use crate::{
...
     error::TransactionError,
     transaction::{DRTransactionBody, VTTransactionBody},
-    utxo_pool::{OwnUnspentOutputsPool, UnspentOutputsPool, UtxoDiff, UtxoSelectionStrategy},
+    utxo_pool::{
+        NodeUtxos, OwnUnspentOutputsPool, UnspentOutputsPool, UtxoDiff, UtxoSelectionStrategy,
+    },
 };
 use std::{collections::HashSet, convert::TryFrom};
 
+pub trait UtxoFantasy {
```
tmpolaczyk, line 14: Add a short description, something like:

```suggestion
/// Abstraction used to create transactions using a UtxoSet
pub trait UtxoFantasy {
```

data_structures/src/transaction_factory.rs
```diff
@@ -202,19 +219,21 @@ pub fn build_commit_collateral(
...
     block_number_limit: Option<u32>,
     utxo_strategy: UtxoSelectionStrategy,
-) -> Result<(Vec<Input>, Vec<ValueTransferOutput>), TransactionError> {
+) -> Result<(Vec<Input>, Vec<ValueTransferOutput>), TransactionError>
+where
+    T: UtxoFantasy,
+{
     // On error just assume the value is u64::max_value(), hoping that it is
     // impossible to pay for this transaction
     let output_value: u64 = transaction_outputs_sum(&outputs)
```
tmpolaczyk, line 239: I think this was past me being lazy, can you rewrite this and instead of using `saturating_add`, return `OutputValueOverflow` on overflow?

data_structures/src/transaction_factory.rs
```diff
@@ -226,8 +245,7 @@ fn build_inputs_outputs_inner(
         );
 
     let (output_pointers, input_value) = take_enough_utxos(
-        own_utxos,
-        all_utxos,
+        utxos,
         output_value + fee,
```
tmpolaczyk, line 249: This line was missing an overflow check :eyes:

data_structures/src/utxo_pool.rs
```diff
@@ -147,6 +150,51 @@ impl OwnUnspentOutputsPool {
...
+                self.own_utxos.iter().map(|(o, _ts)| o.clone()).collect()
+            }
+        }
+    }
+
+    fn get_time_lock(&self, outptr: &OutputPointer) -> Option<u64> {
+        let time_lock = self.all_utxos.get(outptr).map(|vto| vto.time_lock);
+        let time_lock_by_used = self.own_utxos.get(outptr).copied();
+
+        match (time_lock, time_lock_by_used) {
```
tmpolaczyk, line 175: I forgot what's the purpose of this match, can you add a comment?

data_structures/src/utxo_pool.rs
```diff
@@ -147,6 +150,51 @@ impl OwnUnspentOutputsPool {
...
+    fn get_value(&self, outptr: &OutputPointer) -> Option<u64> {
+        self.all_utxos.get(outptr).map(|vto| vto.value)
+    }
+
+    fn get_included_block_number(&self, outptr: &OutputPointer) -> Option<u32> {
+        self.all_utxos.included_in_block_number(outptr)
+    }
+
+    fn set_used_output_pointer(&mut self, outptr: &OutputPointer, ts: u64) {
+        if let Some(current_ts) = self.own_utxos.get_mut(outptr) {
```
tmpolaczyk, line 192: I don't like that this can silently fail. Can we use an unwrap here? Or is it better to return an error and let the called handle it?

wallet/src/repository/error.rs
tmpolaczyk: Clippy sometimes complains about using an `Error` suffix, but on the other hand there is a `HashParseError` there... In my opinion you can leave it like this

wallet/src/repository/wallet/mod.rs
```diff
@@ -36,6 +43,75 @@ struct AccountMutation {
...
+
+    fn get_value(&self, outptr: &OutputPointer) -> Option<u64> {
+        self.state
+            .utxo_set
+            .get(&outptr.into())
+            .map(|vto| vto.amount)
+    }
+
+    fn get_included_block_number(&self, _outptr: &OutputPointer) -> Option<u32> {
+        None
```
tmpolaczyk, line 90: What are the implications if this returning `None`?

tmpolaczyk: I see this being used as

```rust
if utxos.get_included_block_number(op).unwrap() > block_number_limit {
```

So we probably want to use `unwrap_or(0)` there?

data_structures/src/utxo_pool.rs
lrubiorod: the function set_used_output_pointer is called when the utxos were previously selected so it should not fail. I wouldnt like too much add an Error, to keep the function as simple as possible

wallet/src/repository/wallet/mod.rs
lrubiorod: it is only used for collaterals in the part of the node, to check the collateral age. In case of wallet it will be never used so thats the reason of the None. I could add the unwrap_or(0) for safuness but I think it would not change the current behaviour

tmpolaczyk: Ok, so if `get_included_block_number` returns `None` it means that we don't know the collateral age. In that case instead of `unwrap_or(0)` let's skip that UTXO, using something like `if x.is_none() { continue; }`

data_structures/src/utxo_pool.rs
tmpolaczyk: Then I prefer using an unwrap, we know that this cannot fail so if it fails it means there is an error in our code.

data_structures/src/utxo_pool.rs
lrubiorod: We could open a new issue for that :P

wallet/src/repository/wallet/mod.rs
```diff
@@ -950,153 +1020,187 @@ where
...
-        if let Some((pkh, time_lock)) = recipient {
-            outputs.push(types::VttOutput {
-                pkh,
-                value,
-                time_lock,
-            });
-        }
+        let (output_pointers, input_value) = take_enough_utxos(
+            &mut wallet_utxos,
+            output_value + fee,
```
tmpolaczyk, line 1157: This also needs overflow checks, and `output_value` should use `checked_add`, same as in the node

wallet/src/types.rs
aesedepece: idk

wallet/src/repository/wallet/state.rs
```diff
@@ -70,10 +70,12 @@ pub struct State {
     pub transaction_next_id: u32,
     /// Current UTXO set (including pending movements)
     pub utxo_set: model::UtxoSet,
+    /// Tagged used UTXO set
```
aesedepece, line 73: I think `used` is in conflict with the `U` in `Utxo` standing for `unspent` :stuck_out_tongue_winking_eye: 

Maybe we can simply name this `used_outputs` and `model::UsedOutputs`.

data_structures/src/transaction_factory.rs
```diff
@@ -5,57 +5,70 @@ use crate::{
...
     transaction::{DRTransactionBody, VTTransactionBody},
-    utxo_pool::{OwnUnspentOutputsPool, UnspentOutputsPool, UtxoDiff, UtxoSelectionStrategy},
+    utxo_pool::{
+        NodeUtxos, OwnUnspentOutputsPool, UnspentOutputsPool, UtxoDiff, UtxoSelectionStrategy,
+    },
 };
 use std::{collections::HashSet, convert::TryFrom};
 
+/// Abstraction used to create transactions using a UtxoSet
+pub trait UtxoFantasy {
```
aesedepece, line 15:
```suggestion
/// Abstraction that facilitates creation of new transactions from a set of unspent outputs.
/// Transaction factories are expected to operate on this trait so that their business logic can be applied on many heterogeneous data structures that may implement it. 
pub trait OutputsCollection {
```

data_structures/src/transaction_factory.rs
```diff
@@ -5,57 +5,70 @@ use crate::{
...
+}
+
 /// Select enough UTXOs to sum up to `amount`.
 ///
 /// On success, return a list of output pointers and their sum.
 /// On error, return the total sum of the output pointers in `own_utxos`.
-pub fn take_enough_utxos(
-    own_utxos: &mut OwnUnspentOutputsPool,
-    all_utxos: &UnspentOutputsPool,
+pub fn take_enough_utxos<T>(
```
aesedepece, line 27: I think this can now be refactored into a [default implementation](https://doc.rust-lang.org/book/ch10-02-traits.html#default-implementations) in the `UtxoFantasy` trait, without any need for generics.

aesedepece: Although maybe making it a separate thing has some advantage for testing or something else? IDK. Please @tmpolaczyk chime in.

wallet/src/model.rs
```diff
@@ -232,6 +233,13 @@ impl OutPtr {
 
         types::Hash::SHA256(array_bytes)
     }
+
+    pub fn output_pointer(&self) -> OutputPointer {
```
aesedepece, line 237:
```suggestion
    /// Create an `OutputPointer` that points to the same output that an `OutPtr` does.
    pub fn output_pointer(&self) -> OutputPointer {
```

aesedepece: For an alternate, more idiomatic implementation of this:
```rust
impl From<&OutPtr> for OutputPointer {
    pub fn from(out_ptr: &OutPtr) -> OutputPointer {
        OutputPointer {
            transaction_id: out_ptr.transaction_id(),
            output_index: out_ptr.output_index,
        }
    }
}
```

wallet/src/model.rs
```diff
@@ -292,6 +300,7 @@ impl fmt::Display for Beacon {
 }
 
 pub type UtxoSet = HashMap<OutPtr, OutputInfo>;
+pub type UsedUtxoSet = HashMap<OutPtr, u64>;
```
aesedepece, line 303: See my comment on `wallet/src/repository/wallet/state.rs`

aesedepece (changes requested):

wallet/src/repository/wallet/mod.rs
```diff
@@ -918,28 +995,20 @@ where
...
-                )?);
-
-                Ok(types::KeyedSignature {
-                    signature,
-                    public_key,
-                })
-            })
-            .collect();
+        let signatures =
+            self.create_signatures_from_inputs_and_update_balance(inputs, sign_data, &mut state);
```
tmpolaczyk, line 1011: This says "update_balance", but I think the old version of `create_vtt` did not update the balance?

tmpolaczyk: Ah, previously the unconfirmed balance was being updated in `create_transaction_components`. Now `create_transaction_components` does not update the balance.

wallet/src/repository/wallet/tests/mod.rs
```diff
@@ -372,16 +377,18 @@ fn test_create_transaction_components_without_a_change_address() {
...
         .unwrap();
 
-    assert_eq!(1, vtt.value);
-    assert_eq!(0, vtt.change);
-    assert_eq!(vec![out_pointer], vtt.used_utxos);
-    assert_eq!(1, vtt.sign_keys.len());
-    assert_eq!(1, vtt.inputs.len());
-    assert_eq!(1, vtt.outputs.len());
+    assert_eq!(1, inputs.len());
+    assert_eq!(1, outputs.len());
```
tmpolaczyk, line 391: We can test the output value, but I think we cannot test the fee

```suggestion
    assert_eq!(1, outputs.len());
    assert_eq!(value, outputs_sum(outputs));
```

wallet/src/model.rs
```diff
@@ -292,6 +302,7 @@ impl fmt::Display for Beacon {
 }
 
 pub type UtxoSet = HashMap<OutPtr, OutputInfo>;
+pub type UsedOutputs = HashMap<OutPtr, u64>;
```
tmpolaczyk, line 305:
```suggestion
/// Map of output pointer to timestamp.
/// Used to mark outputs that have been recently used in a transaction.
/// They will not be used again until this timestamp.
pub type UsedOutputs = HashMap<OutPtr, u64>;
```

wallet/src/types.rs
mariocao: I am against renaming types, completely misleading in our case!

wallet/src/repository/wallet/mod.rs
lrubiorod: It modifies an mutable balance, so I think it updates

tmpolaczyk: > Did you test it?

No. But there should be tests for that already, somewhere?

wallet/src/repository/wallet/tests/mod.rs
lrubiorod: In the next PR of @girazoki, the logic fee changes and there will be so many tests checking fee logic, so maybe it is not really need it now

wallet/src/repository/error.rs
mariocao: I agree with @tmpolaczyk

data_structures/src/utxo_pool.rs
mariocao: `UtxoFantasy` :rofl:

wallet/src/repository/wallet/mod.rs
lrubiorod: removed it, because it really didnt update the balance

data_structures/src/utxo_pool.rs
```diff
@@ -147,6 +150,53 @@ impl OwnUnspentOutputsPool {
     }
 }
 
+/// Struct that keeps the unspent outputs pool and the own unspent outputs pool
+#[derive(Debug)]
+pub struct NodeUtxos<'a> {
+    pub all_utxos: &'a UnspentOutputsPool,
+    pub own_utxos: &'a mut OwnUnspentOutputsPool,
```
mariocao, line 157: Maybe I would comment them: the `OwnUnspentOutputsPool` field has only the output pointers and not value transfer information (that's why you need the `UnspentOutputsPool`).

wallet/src/repository/wallet/mod.rs
```diff
@@ -36,6 +43,75 @@ struct AccountMutation {
     utxo_removals: Vec<model::OutPtr>,
 }
 
+/// Struct that keep the unspent outputs pool and the own unspent outputs pool
+#[derive(Debug)]
+pub struct WalletUtxos<'a> {
+    pub state: &'a mut State,
+}
```
mariocao, line 50: It seems strange to have the complete wallet state inside the `WalletUtxos`.

mariocao (changes requested):

wallet/src/repository/wallet/mod.rs
```diff
@@ -950,153 +1009,166 @@ where
...
+        // Mark UTXOs as used so we don't double spend
+        // Save the timestamp until it could be spend it
+        wallet_utxos.set_used_output_pointer(&inputs, timestamp + tx_pending_timeout);
+
+        // In case of VTTransaction, a new internal address is used
+        // In case of DRTransaction, the first input pkh will be used
+        let change_pkh = if dr_output.is_none() {
+            self._gen_internal_address(state, None)?.pkh
+        } else {
+            let first_output = inputs.first().clone().unwrap().output_pointer();
```
tmpolaczyk, line 1148:
```suggestion
            let first_input = inputs.first().clone().unwrap().output_pointer();
```

data_structures/src/transaction_factory.rs
```diff
@@ -5,83 +5,123 @@ use crate::{
...
-    } else {
-        Err(TransactionError::NoMoney {
-            total_balance: total,
-            available_balance: acc,
-            transaction_value: amount,
-        })
+    /// Generic inputs/outputs builder: can be used to build
+    /// value transfer transactions and data request transactions.
+    #[allow(clippy::too_many_arguments)]
+    fn build_inputs_outputs_inner(
```
tmpolaczyk, line 95: Now that this is function is included in the OutputsCollection trait, maybe it's time to remove the "inner" from the name?

wallet/src/repository/wallet/mod.rs
```diff
@@ -950,153 +1009,166 @@ where
...
+            None,
+            utxo_strategy,
+        )?;
+
+        Ok((inputs, outputs))
     }
 
-    fn create_transaction_components(
+    #[allow(clippy::too_many_arguments)]
+    fn build_inputs_outputs_inner_wallet(
```
mariocao, line 1112: Comment why this function is slightly different from the one in data_structures. :)

wallet/src/model.rs
```diff
@@ -292,6 +302,10 @@ impl fmt::Display for Beacon {
 }
 
 pub type UtxoSet = HashMap<OutPtr, OutputInfo>;
+/// Map of output pointer to timestamp.
```
mariocao, line 305:
```suggestion

/// Map of output pointer to timestamp.
```

wallet/src/repository/wallet/state.rs
```diff
@@ -70,10 +70,12 @@ pub struct State {
...
+    pub used_outputs: model::UsedOutputs,
```
mariocao, line 74: This field should be cleared when resyncing the wallet in `clear_chain_data`.

mariocao: Additionally, this field should be taken into consideration when a revert occurs in the chain.

In other words, in the function `clear_pending_state`, the `usedOutputs` should be reviewed and/or removed (if their timestamp threshold has passed) because they may have some inconsistent state (e.g. an UTXO created in a reverted block).

mariocao: Removing old `usedOutputs` is also encouraged to keep a lower memory footprint`

tmpolaczyk (approved):

mariocao (changes requested):

mariocao (approved):

wallet/src/repository/wallet/state.rs
aesedepece, line 73:
```suggestion
    /// Outputs that have been locally tagged as used. This prevents double spending an output while a first spending transaction is pending.  
```

data_structures/src/transaction_factory.rs
```diff
@@ -5,83 +5,123 @@ use crate::{
...
+    fn build_inputs_outputs(
+        &mut self,
+        outputs: Vec<ValueTransferOutput>,
+        dr_output: Option<&DataRequestOutput>,
+        fee: u64,
+        timestamp: u64,
+        // The block number must be lower than this limit
+        block_number_limit: Option<u32>,
+        utxo_strategy: UtxoSelectionStrategy,
+    ) -> Result<(Vec<Input>, Vec<ValueTransferOutput>, u64, u64), TransactionError> {
```
aesedepece, line 104: May the positive return type be worth a type alias?

data_structures/src/transaction_factory.rs
```diff
@@ -154,19 +196,26 @@ pub fn build_drt(
...
-        own_pkh,
-        all_utxos,
         timestamp,
-        tx_pending_timeout,
         None,
         UtxoSelectionStrategy::Random,
     )?;
 
+    // Mark UTXOs as used so we don't double spend
+    // Save the timestamp until it could be spend it
```
aesedepece, line 213:
```suggestion
    // Save the timestamp after which the transaction will be considered timed out and the output will become available for spending it again
```

data_structures/src/transaction_factory.rs
```diff
@@ -185,56 +234,23 @@ pub fn build_commit_collateral(
...
-    // impossible to pay for this transaction
-    let output_value: u64 = transaction_outputs_sum(&outputs)
-        .unwrap_or(u64::max_value())
-        .saturating_add(
-            dr_output
-                .map(|o| o.checked_total_value().unwrap_or(u64::max_value()))
-                .unwrap_or_default(),
-        );
+    // Mark UTXOs as used so we don't double spend
+    // Save the timestamp until it could be spend it
```
aesedepece, line 251:
```suggestion
    // Save the timestamp after which the transaction will be considered timed out and the output will become available for spending it again
```

aesedepece (changes requested):

aesedepece (approved):

PR #1695, opened 2020-10-29
Arc and memoize blocks & superblocks

aesedepece:


wallet/src/actors/worker/methods.rs
```diff
@@ -1009,16 +1015,12 @@ impl Worker {
 
     pub fn index_block(
         &self,
-        block: types::ChainBlock,
+        block: Arc<types::ChainBlock>,
```
tmpolaczyk, line 1018: If a function does not use `Arc::clone`, prefer using a normal reference.

```suggestion
        block: &types::ChainBlock,
```

wallet/src/actors/worker/methods.rs
```diff
@@ -1033,29 +1035,35 @@ impl Worker {
         let vtt_txns = block
             .txns
             .value_transfer_txns
+            .clone()
             .into_iter()
```
tmpolaczyk, line 1039:
```suggestion
            .into_iter()
            .cloned()
```

wallet/src/actors/worker/methods.rs
aesedepece: What about now? :grimacing:

tmpolaczyk: Confused. I think we can remove the block clone if we change `block_txns` to be an iterator, and make `index_txns` accept an `impl Iterator<Item = &Transaction>`.

aesedepece: Done!

tmpolaczyk: Now `index_block` does not need an `Arc<types::ChainBlock>`, only a reference

wallet/src/actors/worker/methods.rs
```diff
@@ -863,7 +869,7 @@ impl Worker {
 
     pub fn handle_block(
         &self,
-        block: types::ChainBlock,
+        block: Arc<types::ChainBlock>,
```
tmpolaczyk, line 872: If `index_block` takes a reference, so does this method:

```suggestion
        block: &types::ChainBlock,
```

wallet/src/actors/worker/methods.rs
```diff
@@ -731,15 +734,18 @@ impl Worker {
                 let get_block_future = self.get_block(id.clone());
                 let (block, confirmed) = futures03::executor::block_on(get_block_future)?;
 
+                // Wrap block into an atomic reference count for the sake of avoiding expensive clones
+                let block_arc = Arc::new(block);
```
tmpolaczyk, line 738: And if `handle_block` takes a reference, this `Arc` is not needed

wallet/src/actors/worker/methods.rs
aesedepece: But man, all of this code lives in the worker. The way for blocks to arrive to the worker is wrapped in an `Arc` so as to avoid cloning the block for each wallet that may be unlocked at that time (1 wallet = 1 worker). And once you put something into an `Arc`, turning it into a plain reference is probably not the best idea, right?

tmpolaczyk: Turning an arc into a reference is free. If this was a public function, then it would be more sense to use references because in that case we avoid creating arcs when we only have references. But if this is an internal function and we plan on using `Arc<Block>` everywhere, then there is no problem.

aesedepece:
Rebased

tmpolaczyk (approved):

PR #1698, opened 2020-10-30
feat(jsonrpc): Add transaction size in bytes to getTransaction method

drcpu-github:
Added transaction size in bytes to the getTransaction method as that allows me to calculate fee per byte and generate transaction pool fee plots in the explorer.

### DIFF
--- a/node/src/actors/json_rpc/json_rpc_methods.rs
+++ b/node/src/actors/json_rpc/json_rpc_methods.rs
@@ -653,6 +653,8 @@ pub fn get_block(hash: Result<(Hash,), jsonrpc_core::Error>) -> JsonRpcResultAsy
 pub struct GetTransactionOutput {
     /// Transaction
     pub transaction: Transaction,
+    /// Weight of the transaction
+    pub weight: u32,
     /// Hash of the block that contains this transaction in hex format,
     /// or "pending" if the transaction has not been included in any block yet
     pub block_hash: String,
@@ -671,8 +673,10 @@ pub fn get_transaction(hash: Result<(Hash,), jsonrpc_core::Error>) -> JsonRpcRes
             .send(GetItemTransaction { hash })
             .then(move |res| match res {
                 Ok(Ok((transaction, pointer_to_block))) => {
+                    let weight = transaction.weight();
                     let output = GetTransactionOutput {
                         transaction,
+                        weight,
                         block_hash: pointer_to_block.block_hash.to_string(),
                     };
                     let value = match serde_json::to_value(output) {
@@ -691,8 +695,10 @@ pub fn get_transaction(hash: Result<(Hash,), jsonrpc_core::Error>) -> JsonRpcRes
                     Box::new(chain_manager.send(GetMemoryTransaction { hash }).then(
                         |res| match res {
                             Ok(Ok(transaction)) => {
+                                let weight = transaction.weight();
                                 let output = GetTransactionOutput {
                                     transaction,
+                                    weight,
                                     block_hash: "pending".to_string(),
                                 };
                                 let value = match serde_json::to_value(output) {
--- a/data_structures/src/transaction.rs
+++ b/data_structures/src/transaction.rs
@@ -172,6 +172,15 @@ impl Transaction {
     pub fn size(&self) -> u32 {
         u32::try_from(self.to_pb().write_to_bytes().unwrap().len()).unwrap()
     }
+
+    /// Get the weight of the transaction
+    pub fn weight(&self) -> u32 {
+        match self {
+            Transaction::ValueTransfer(vt_txn) => vt_txn.weight(),
+            Transaction::DataRequest(dr_txn) => dr_txn.weight(),
+            _ => 0,
+        }
+    }
 }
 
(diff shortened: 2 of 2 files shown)


tmpolaczyk:
You probably want to use transaction weight instead of transaction size.

This distinction is important because of data requests: a data request with 10 witnesses and one with 100 witnesses may have the same size in bytes, but the one with 100 witnesses will result in 10 times the number of commits and reveals, so we need to take that into account when prioritizing transactions.

`VTTransaction` and `DRTransaction` have a `.weight()` method, but the other kinds of transactions do not, so their weight can be considered to be 0.

node/src/actors/json_rpc/json_rpc_methods.rs
```diff
@@ -672,7 +674,8 @@ pub fn get_transaction(hash: Result<(Hash,), jsonrpc_core::Error>) -> JsonRpcRes
             .then(move |res| match res {
                 Ok(Ok((transaction, pointer_to_block))) => {
                     let output = GetTransactionOutput {
-                        transaction,
+                        transaction: transaction.clone(),
+                        size: transaction.size(),
```
tmpolaczyk, line 678: Usually this kind of clones can be avoided by changing the order of operations. In this case you can do `let size = transaction.size();` And then this simply becomes:
```suggestion
                        transaction,
                        size,
```

drcpu-github:
> You probably want to use transaction weight instead of transaction size.
> 
> This distinction is important because of data requests: a data request with 10 witnesses and one with 100 witnesses may have the same size in bytes, but the one with 100 witnesses will result in 10 times the number of commits and reveals, so we need to take that into account when prioritizing transactions.
> 
> `VTTransaction` and `DRTransaction` have a `.weight()` method, but the other kinds of transactions do not, so their weight can be considered to be 0.

Yes, I'm aware of the distinction, but I purposely took size rather than weight because the size in bytes is what a miner will have to save in the blockchain, unless I'm mistaken? Transaction weight is more of an internal variable to prioritize transactions. However, for the purpose of graphing transaction fees, it probably does make more sense to use weight and not size as that's how the protocol prioritization system works and what other people looking at those graphs have to take into account.

tmpolaczyk:
Using the example of a data request with 10 witnesses and one with 100 witnesses, they may be both 400 bytes in size, but they also pay for all the included commits and reveals, let's say that's 400 bytes per commit + 200 bytes per reveal, then one data request will result in a total of 6400 bytes and the other one in 60400 bytes. So the transaction weight is a way to approximate the total size that will be included in the blockchain, over several blocks, and that's why it is used to prioritize transactions.

> Would it be possible to add a fee field?

We could add a `priority` field which would be `fee/weight` but only for transactions that are currently in the `TransactionsPool`, the ones with `"blockHash":"pending"`. This could be used to plot the weighted fee distribution and recommend fast/slow fees to users. But this `priority` field would disappear when the transaction is included in a block, so I'm not sure if it's a good idea.

If we want to show the fee for all the transactions, then it can be done by storing the fee in the database. That would require a longer discussion and it may be complicated to implement, but I agree that it would be useful.

If the fee is not saved in the database we would need to first get the value of each input and then calculate the fee. This means one extra query to the database for each input of the transaction. This could be implemented as a new JSON-RPC method, something like `getTransactionFee`. This way we would not make the existing `getTransaction` method too slow, and if we see that performance is good we can just include it as a `fee` field in `getTransaction`.

drcpu-github:
> Using the example of a data request with 10 witnesses and one with 100 witnesses, they may be both 400 bytes in size, but they also pay for all the included commits and reveals, let's say that's 400 bytes per commit + 200 bytes per reveal, then one data request will result in a total of 6400 bytes and the other one in 60400 bytes. So the transaction weight is a way to approximate the total size that will be included in the blockchain, over several blocks, and that's why it is used to prioritize transactions.

Makes sense, overlooked that.
> 
> > Would it be possible to add a fee field?
> 
> We could add a `priority` field which would be `fee/weight` but only for transactions that are currently in the `TransactionsPool`, the ones with `"blockHash":"pending"`. This could be used to plot the weighted fee distribution and recommend fast/slow fees to users. But this `priority` field would disappear when the transaction is included in a block, so I'm not sure if it's a good idea.
> 
> If we want to show the fee for all the transactions, then it can be done by storing the fee in the database. That would require a longer discussion and it may be complicated to implement, but I agree that it would be useful.

I guess the main selling point for this is that it would reduce possible errors on my side (and other tool developers), but I am not sure if that is enough of a selling point for the modifications required.

> If the fee is not saved in the database we would need to first get the value of each input and then calculate the fee. This means one extra query to the database for each input of the transaction. This could be implemented as a new JSON-RPC method, something like `getTransactionFee`. This way we would not make the existing `getTransaction` method too slow, and if we see that performance is good we can just include it as a `fee` field in `getTransaction`.

This is basically what I'm doing now in my Python backend: fetching all input and output transactions, calculating their summed value and the difference is the fee. However, this gives me quite some overhead (and is somewhat error-prone). Adan pointed out that it may just be better to maintain my own UTXO set.

Anyway, as you point out, all of that is food for a longer discussion and thus another PR. If this one is good enough for you, feel free to merge it.

tmpolaczyk (approved):

PR #1709, opened 2020-11-04
feat(wallet): stop syncing when logging in with another session

### LINKED ISSUE #1701: [Wallet] Don’t start sync upon unlock if already syncing

girazoki:
Close #1701

### DIFF
--- a/wallet/src/actors/app/methods.rs
+++ b/wallet/src/actors/app/methods.rs
@@ -314,10 +314,15 @@ impl App {
 
     /// Load a wallet's private information and keys in memory.
     pub fn unlock_wallet(
-        &self,
+        &mut self,
         wallet_id: String,
         password: types::Password,
     ) -> ResponseActFuture<types::UnlockedWallet> {
+        // If a synchronization from a previous session is still running, set `stop_syncing` to
+        // `true` so as to signal that it must stop as soon as possible
+        if let Some(wallet) = self.state.get_current_wallet_session(wallet_id.clone()) {
+            wallet.set_stop_syncing().expect("Lock error")
+        }
         let f = self
             .params
             .worker
--- a/wallet/src/actors/worker/methods.rs
+++ b/wallet/src/actors/worker/methods.rs
@@ -752,7 +752,9 @@ impl Worker {
 
             // Keep asking for new batches of blocks until we get less than expected, which signals
             // that there are no more blocks to process.
-            if batch_size < i128::from(limit) {
+            if batch_size < i128::from(limit)
+                || wallet.lock_and_read_state(|state| state.stop_syncing)?
+            {
                 break;
             } else {
                 log::info!(
--- a/wallet/src/repository/wallet/mod.rs
+++ b/wallet/src/repository/wallet/mod.rs
@@ -309,6 +309,7 @@ where
             db_movements_to_update: Default::default(),
             transient_external_addresses: Default::default(),
             transient_internal_addresses: Default::default(),
+            stop_syncing: false,
         });
 
         Ok(Self {
@@ -407,6 +408,14 @@ where
         };
 
         Ok((Arc::new(address), next_index))
+    }
+
+    /// Set stop syncing flag to true
+    pub fn set_stop_syncing(&self) -> Result<()> {
+        let mut state = self.state.write()?;
+        state.stop_syncing = true;
+
+        Ok(())
     }
 
     /// Generate an address in the external keychain (WIP-0001).
--- a/wallet/src/actors/app/state.rs
+++ b/wallet/src/actors/app/state.rs
@@ -161,4 +161,12 @@ impl State {
 
         self.wallets.insert(wallet_id, wallet);
     }
+
+    /// Get latest session for a wallet ID
+    pub fn get_current_wallet_session(
+        &mut self,
+        wallet_id: String,
+    ) -> Option<&types::SessionWallet> {
+        self.wallets.get(&wallet_id)
+    }
 }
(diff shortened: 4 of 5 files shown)


wallet/src/actors/app/methods.rs
```diff
@@ -314,10 +314,14 @@ impl App {
...
     /// Load a wallet's private information and keys in memory.
     pub fn unlock_wallet(
-        &self,
+        &mut self,
         wallet_id: String,
         password: types::Password,
     ) -> ResponseActFuture<types::UnlockedWallet> {
+        // If there is a synchronization from a previous session running, set unsyncing to false so that it can stop
+        if let Some(wallet) = self.state.get_current_wallet_session(wallet_id.clone()) {
+            wallet.unset_syncing().expect("Lock error")
```
tmpolaczyk, line 323: Do we have any guarantee that the other session will read this flag before the new session starts synchronizing? Otherwise, as soon as the new session starts synchronizing it will set the flag back to true, and we are have the same problem again.

tmpolaczyk: Is there one `syncing` flag for each session? In that case there is no problem.

girazoki: AFAIK yes

wallet/src/actors/worker/methods.rs
```diff
@@ -749,9 +752,14 @@ impl Worker {
             )]);
             self.notify_client(&wallet, sink.clone(), events).ok();
 
+            // If the batch size smaller than the limit, we reached the end. Set syncing to false
+            if batch_size < i128::from(limit) {
+                wallet.unset_syncing()?;
```
tmpolaczyk, line 757: If there is any error during synchronization, this will never be called. To avoid any problems, we can set and unset the syncing flag in the `sync` function instead of here. This is why we have a `sync_inner` function.

girazoki: lets do that

wallet/src/repository/wallet/mod.rs
```diff
@@ -347,7 +348,14 @@ where
...
 
-    /// Generate an address in the internal keychain (WIP-0001).
+    /// Set syncing flag to true
+    pub fn set_stop_syncing(&self) -> Result<()> {
+        let mut state = self.state.write()?;
+
+        self._set_stop_syncing(&mut state)
+    }
+
+    /// set syncing to true
```
tmpolaczyk, line 358:
```suggestion
     /// Generate an address in the internal keychain (WIP-0001). 
```

wallet/src/repository/wallet/mod.rs
```diff
@@ -347,7 +348,14 @@ where
         self._gen_external_address(&mut state, label)
     }
 
-    /// Generate an address in the internal keychain (WIP-0001).
+    /// Set syncing flag to true
```
tmpolaczyk, line 351:
```suggestion
    /// Set stop_syncing flag to true
```

tmpolaczyk (approved):

tmpolaczyk:
With the current implementation there is a small chance that the second synchronization will begin before the first synchronization is stopped, resulting in two simultaneous synchronizations. This does not seem to be a problem right now, but we need to keep that in mind.

clbartoli (approved):

wallet/src/actors/app/methods.rs
```diff
@@ -314,10 +314,14 @@ impl App {
 
     /// Load a wallet's private information and keys in memory.
     pub fn unlock_wallet(
-        &self,
+        &mut self,
         wallet_id: String,
         password: types::Password,
     ) -> ResponseActFuture<types::UnlockedWallet> {
+        // If there is a synchronization from a previous session running, set unsyncing to false so that it can stop
```
aesedepece, line 321:
```suggestion
        // If a synchronization from a previous session is still running, set `stop_syncing` to `true` so as to signal that it must stop as soon as possible
```

aesedepece (changes requested):

wallet/src/repository/wallet/mod.rs
```diff
@@ -240,6 +240,7 @@ where
             db_movements_to_update: Default::default(),
             transient_external_addresses: Default::default(),
             transient_internal_addresses: Default::default(),
+            stop_syncing: false,
```
mariocao, line 243: OMG, we are creating a monster... at some point we should refactor that struct. :rofl:

wallet/src/repository/wallet/mod.rs
```diff
@@ -1117,6 +1125,12 @@ where
         Ok(address)
     }
 
+    fn _set_stop_syncing(&self, state: &mut State) -> Result<()> {
+        state.stop_syncing = true;
+
+        Ok(())
+    }
+
```
mariocao, line 1133: If it is not used maybe I would remove this function and move the logic to `set_stop_syncing`. What do you think guys?

mariocao (changes requested):

mariocao (approved):

wallet/src/actors/app/methods.rs
```diff
@@ -314,10 +314,14 @@ impl App {
 
     /// Load a wallet's private information and keys in memory.
     pub fn unlock_wallet(
-        &self,
+        &mut self,
         wallet_id: String,
         password: types::Password,
     ) -> ResponseActFuture<types::UnlockedWallet> {
+        // If a synchronization from a previous session is still running, set `stop_syncing` to `true` so as to signal that it must stop as soon as possible
```
aesedepece, line 321: Clippy will likely complain about the length of this line

aesedepece (approved):

tmpolaczyk (approved):

PR #1710, opened 2020-11-05
[Wallet]: Restart sync wallet when node synced

### LINKED ISSUE #1704: [Wallet] Upon SYNCED notification while not syncing, clear pending and trigger sync

clbartoli:
When the wallet receives a `synced` notification from the node the Wallet clears the pending data and synchronizes again.
This PR also changes the function `unlock_wallet`, which will only start the wallet synchronization if the node is synced.

This PR is using #1677 and it works for the case in which the node has consensus unknown, however it doesn't work for the case of a restarted node since the wallet is not yet able to know if the node disconnects (#1705 ).

Closes #1704

### DIFF
--- a/wallet/src/actors/app/methods.rs
+++ b/wallet/src/actors/app/methods.rs
@@ -3,7 +3,7 @@ use std::sync::Arc;
 use actix::utils::TimerFunc;
 use futures::future;
 
-use witnet_data_structures::chain::InventoryItem;
+use witnet_data_structures::chain::{InventoryItem, StateMachine};
 
 use crate::actors::{
     worker::{HandleBlockRequest, HandleSuperBlockRequest, NodeStatusRequest, NotifyStatus},
@@ -340,13 +340,17 @@ impl App {
                 slf.state
                     .create_session(session_id.clone(), wallet_id.clone(), wallet.clone());
 
-                // Start synchronization for this wallet
-                let sink = slf.state.get_sink(&session_id);
-                slf.params.worker.do_send(worker::SyncRequest {
-                    wallet_id,
-                    wallet,
-                    sink,
-                });
+                // If the node is synced start synchronization for this wallet
+                if slf.state.node_state == Some(StateMachine::Synced)
+                    || slf.state.node_state == None
+                {
+                    let sink = slf.state.get_sink(&session_id);
+                    slf.params.worker.do_send(worker::SyncRequest {
+                        wallet_id,
+                        wallet,
+                        sink,
+                    });
+                };
 
                 fut::ok(types::UnlockedWallet { data, session_id })
             });
@@ -568,6 +572,7 @@ impl App {
             let sink = self.state.get_sink(&wallet.session_id);
             self.handle_node_status_in_worker(&status, wallet, sink.clone());
         }
+        self.state.update_node_state(status);
 
         Ok(())
     }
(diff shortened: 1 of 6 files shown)


wallet/src/actors/worker/methods.rs
```diff
@@ -1013,7 +1015,13 @@ impl Worker {
         wallet: types::SessionWallet,
         sink: types::DynamicSink,
     ) -> Result<()> {
-        log::debug!("The node has changed its status into {:?}", status);
+        log::info!("The node has changed its status into {:?}", status);
+        if status == StateMachine::Synced && !wallet.is_syncing()? {
+            wallet.clear_pending_data().ok();
```
tmpolaczyk, line 1020: This will delete any pending transactions? What happens if we call `self.sync` without clearing the pending data? It should work, right?

wallet/src/actors/worker/methods.rs
```diff
@@ -1013,7 +1015,13 @@ impl Worker {
         wallet: types::SessionWallet,
         sink: types::DynamicSink,
     ) -> Result<()> {
-        log::debug!("The node has changed its status into {:?}", status);
+        log::info!("The node has changed its status into {:?}", status);
+        if status == StateMachine::Synced && !wallet.is_syncing()? {
```
tmpolaczyk, line 1019: Note that if the node sends a notification when the state changes from `Synced` to `Synced`, this will trigger an unneeded synchronization. We can prevent that from happening in the node, or we can detect if the wallet is more than N blocks behind the node, and only sync if we see that it is behind.

clbartoli: I think this does not happen, the StateMachine is updated to Synced only when consensus_beacon is equal to our_beacon and the StateMachine is AlmostSynced, so it can not happen that "changes" from Synced to Synced

wallet/src/actors/worker/methods.rs
clbartoli: Not sure, the issue #1704 specifies to clear the pending data. Anyway it depends on the data, I think in this case we are considering data that is not included in the superblock, but maybe we should make sure if the data is actually pending or if it was included in a superblock while the node was nos synced.

wallet/src/actors/app/methods.rs
```diff
@@ -545,6 +548,27 @@ impl App {
...
+            .wallets
+            .iter()
+            .map(|(_, wallet)| wallet.clone())
+            .collect();
+
+        for wallet in &wallets {
+            let sink = self.state.get_sink(&wallet.session_id);
+            self.handle_node_status_in_worker(&status, wallet, sink.clone());
+            self.state.update_node_state(status);
+        }
```
tmpolaczyk, line 567: This can be moved outside of the for loop

```suggestion
        }
        self.state.update_node_state(status);
```

wallet/src/actors/app/methods.rs
```diff
@@ -335,13 +335,15 @@ impl App {
...
-                });
+                // If the node is synced start synchronization for this wallet
+                if slf.state.node_state == Some(StateMachine::Synced) {
+                    let sink = slf.state.get_sink(&session_id);
+                    slf.params.worker.do_send(worker::SyncRequest {
+                        wallet_id,
+                        wallet,
+                        sink,
+                    });
+                };
```
tmpolaczyk, line 346: What happens if the node is not synced? I guess the synchronization will start later, but when?

clbartoli: It will start when the wallet receives the synced notification from the node

tmpolaczyk (approved):

aesedepece (approved):

PR #1714, opened 2020-11-06
feat(storage): implement iteration support

tmpolaczyk:
This PR adds a `prefix_iterator` to the `Storage` trait, which allows iterating over all the storage keys that start with the same prefix.

This can be used to implement storage migrations by iterating over all the existing keys and updating them to the new format. Also it can be used to check if the storage is empty, or assert that there are no keys with some specific prefix. In this PR it is not used anywhere.

TODO: add tests

### DIFF
--- a/storage/src/backends/crypto.rs
+++ b/storage/src/backends/crypto.rs
@@ -52,7 +52,7 @@ impl<T: Storage> Storage for Backend<T> {
             })
     }
 
-    fn put(&mut self, key: Vec<u8>, value: Vec<u8>) -> Result<()> {
+    fn put(&self, key: Vec<u8>, value: Vec<u8>) -> Result<()> {
         let hash_key = calculate_sha256(key.as_ref());
         let iv = cipher::generate_random(IV_LENGTH)?;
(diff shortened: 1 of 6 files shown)


storage/src/backends/crypto.rs
```diff
@@ -65,10 +66,14 @@ impl<T: Storage> Storage for Backend<T> {
...
     }
 
-    fn delete(&mut self, key: &[u8]) -> Result<()> {
+    fn delete(&self, key: &[u8]) -> Result<()> {
         let hash_key = calculate_sha256(key);
         self.backend.delete(hash_key.as_ref())
     }
+
+    fn prefix_iterator<'a, 'b: 'a>(&'a self, _prefix: &'b [u8]) -> Result<StorageIterator<'a>> {
+        bail!("Iteration is not supported when using encrypted storage")
```
tmpolaczyk, line 75: This is a problem: it looks like there is a hidden feature that allows to use an encrypted storage for the node. I would try to deprecate that in some way to avoid future problems. Otherwise, we must always assume that iteration can fail.

tmpolaczyk:
This was already merged as part of #2159

PR #1715, opened 2020-11-06
fix(net): add cooldown to jsonrpc client reconnect

### LINKED ISSUE #1707: [Wallet] Fix reconnection loop (too many open files)

tmpolaczyk:
Fix #1707 

The issue was that the jsonrpc client library was leaking sockets. I fixed this in https://github.com/witnet/async-jsonrpc-client/tree/fix-tcp-leak but note that this changes the way the jsonrpc client works, so we should make sure this does not break anything.

Also, added a cooldown to the reconnect method, because otherwise the wallet enters in a sort of infinite loop. But the "too many open files" error should be fixed, even without the cooldown.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -161,7 +161,7 @@ checksum = "23b62fc65de8e4e7f52534fb52b0f3ed04746ae267519eef2a83941e8085068b"
 [[package]]
 name = "async-jsonrpc-client"
 version = "0.1.0"
-source = "git+https://github.com/witnet/async-jsonrpc-client#db42be175031677b910efa129f42b23d6a000d41"
+source = "git+https://github.com/witnet/async-jsonrpc-client?branch=fix-tcp-leak#600a2d696af265511868f77c01e448c1d678650e"
 dependencies = [
  "error-chain",
  "futures 0.1.30",
@@ -171,8 +171,8 @@ dependencies = [
  "serde",
  "serde_derive",
  "serde_json",
+ "tokio 0.1.22",
  "tokio-codec",
- "tokio-core",
  "tokio-timer 0.2.13",
 ]
 
--- a/bridges/ethereum/Cargo.toml
+++ b/bridges/ethereum/Cargo.toml
@@ -5,7 +5,7 @@ authors = ["Witnet Foundation <info@witnet.foundation>"]
 edition = "2018"
 
 [dependencies]
-async-jsonrpc-client = { git = "https://github.com/witnet/async-jsonrpc-client", features = ["tcp"] }
+async-jsonrpc-client = { git = "https://github.com/witnet/async-jsonrpc-client", features = ["tcp"], branch = "fix-tcp-leak" }
 env_logger = "0.7.1"
 ethabi = "9.0.0"
 futures = "0.1.28"
--- a/net/Cargo.toml
+++ b/net/Cargo.toml
@@ -7,7 +7,7 @@ workspace = ".."
 
 [dependencies]
 actix = { version = "0.8.3", default-features = false }
-async-jsonrpc-client = { git = "https://github.com/witnet/async-jsonrpc-client", features = ["tcp"] }
+async-jsonrpc-client = { git = "https://github.com/witnet/async-jsonrpc-client", features = ["tcp"], branch = "fix-tcp-leak" }
 failure = "0.1.8"
 futures = "0.1.29"
 jsonrpc-core = "15.1.0"
--- a/wallet/Cargo.toml
+++ b/wallet/Cargo.toml
@@ -6,7 +6,7 @@ version = "0.3.2"
 workspace = ".."
 
 [dependencies]
-async-jsonrpc-client = { git = "https://github.com/witnet/async-jsonrpc-client", features = ["tcp"] }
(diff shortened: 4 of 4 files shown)


net/src/client/tcp/actors/jsonrpc.rs
```diff
@@ -27,8 +27,12 @@ pub struct JsonRpcClient {
     active_subscriptions: Arc<Mutex<HashMap<String, Subscribe>>>,
     pending_subscriptions: HashMap<String, Subscribe>,
     url: String,
+    last_reconnection: Instant,
 }
 
+use std::sync::atomic::AtomicU32;
+static RECONNECTION_ATTEMPTS: AtomicU32 = AtomicU32::new(0);
```
tmpolaczyk, line 34: This is only used for debugging, remove before merging

net/src/client/tcp/actors/jsonrpc.rs
```diff
@@ -58,6 +63,17 @@ impl JsonRpcClient {
 
     /// Replace the TCP connection with a fresh new connection.
     pub fn reconnect(&mut self, ctx: &mut <Self as Actor>::Context) {
+        let now = Instant::now();
+        // Attempt to reconnect at most one every 10 seconds
```
aesedepece, line 67:
```suggestion
        // Attempt to reconnect at most once every 10 seconds
```

net/src/client/tcp/actors/jsonrpc.rs
```diff
@@ -58,6 +63,17 @@ impl JsonRpcClient {
...
+        let reconnection_cooldown = Duration::from_secs(10);
```
aesedepece, line 68: Just out of curiosity, can this be made into a `const`?

net/src/client/tcp/actors/jsonrpc.rs
```diff
@@ -58,6 +63,17 @@ impl JsonRpcClient {
...
+        if now.duration_since(self.last_reconnection) < reconnection_cooldown {
+            log::debug!("Not reconnecting yet");
+            return;
+        }
+        let old_reconnection_attempts = RECONNECTION_ATTEMPTS.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
+        log::info!("More or less {} reconnection attempts", old_reconnection_attempts);
```
aesedepece, line 74:
```suggestion
        log::info!("Approximately {} reconnection attempts so far", old_reconnection_attempts);
```

aesedepece (changes requested):

tmpolaczyk:
It looks like this won't work, after 30 minutes of reconnection attempts, the wallet process has 213 open threads, and the number keeps growing on every reconnection attempt

    cat /proc/$(pgrep witnet)/status | grep -i Threads

net/src/client/tcp/actors/jsonrpc.rs
tmpolaczyk: Probably yes, I also considered using the default timeout of 5 seconds here.

mariocao:
It seems that the PR works! :)

However, while doing some tests I encountered that it needs like 2 attempts to recover the connectivity to the node.

How to reproduce it:
1. Start node
2. Start wallet: `TCP socket is now connected to 127.0.0.1:21338`
3. Stop node: currently the wallet does not know that node is down
4. Unlock a wallet (was previously created): `Ignoring reconnect request: last reconnection attempt was a few seconds ago` and it seems sockets are not being leaked.
5. Restart node (but the wallet does not detect that the node is up again)
6. Trigger re-sync, but it fails: `Node Error: request failed: Transport error: Error sending request`
7. Then,  the wallet by itself subscribes TWICE to the notification:

```
[2020-11-13T09:57:15Z INFO  witnet_net::client::tcp::actors::jsonrpc] Client "superblocks" subscription created
[2020-11-13T09:57:15Z INFO  witnet_net::client::tcp::actors::jsonrpc] Client "superblocks" subscription created
[2020-11-13T09:57:15Z INFO  witnet_net::client::tcp::actors::jsonrpc] Client "blocks" subscription created
[2020-11-13T09:57:15Z INFO  witnet_net::client::tcp::actors::jsonrpc] Client "status" subscription created
[2020-11-13T09:57:15Z INFO  witnet_net::client::tcp::actors::jsonrpc] Client "status" subscription created
[2020-11-13T09:57:15Z INFO  witnet_net::client::tcp::actors::jsonrpc] Client "blocks" subscription created
```

And then, the wallet receives twice all notifications, for example:
```
[2020-11-13T09:58:32Z INFO  witnet_wallet::actors::worker::methods] Superblock #9183 notification received. Consolidating 0 pending blocks...
[2020-11-13T09:58:32Z INFO  witnet_wallet::actors::worker::methods] Superblock #9183 notification received. Consolidating 0 pending blocks...
```

tmpolaczyk:
Commit 19be0c678483f8840faad55cd31e328b70ef776a has already been merged, so this PR is only needed to fix the leak that causes a crash if the node is disconnected for more than ~1 hour. I was unable to reproduce @mariocao comment after the latest changes in master, because the wallet now detects when the node is offline and when it is back online, thanks to PR #1719 and also it is impossible to trigger a resync in sheikah if the node is disconnected.

PR #1716, opened 2020-11-10
Export wallet keys encrypted

### LINKED ISSUE #1699: [Wallet]: export private key (xprv)
The goal of this issue is to create a method in the wallet capable of exporting the accounts private key in XPRV format, encrypted with a user-defined password. Ideally this password should be different to the one provided to the wallet to encrypt the database

### LINKED ISSUE #1700: [Wallet]: Import encrypted XPRV key
The goal of this issue is to import a XPRV key in an encrypted form, by decrypting it first with the user-defined password

girazoki:
Allows to export private key (master key) or private keys (account keys) in an encrypted manner. It also adds support to import the given key to create a wallet. The keys are encrypted with AES-CBC.

Closes #1699 
Closes #1700


wallet/src/actors/app/methods.rs
```diff
@@ -818,4 +829,141 @@ impl App {
...
+
+            Ok(types::SeedSource::XprvDouble((
+                internal.into(),
+                external.into(),
+            )))
+        }
+        _ => Err(validation_error(app::field_error(
+            "seed_data",
+            "Invalid seed data prefix",
+        ))),
```
tmpolaczyk, line 967: All this logic inside the `xprvdouble` branch should be a separate function, and it should be tested. Something like `split_xprvdouble`.

tmpolaczyk: I see that it's already tested in `test_valid_xprv_double_decryption`, so doesn't need many tests. Maybe check that it doesn't panic with some strange inputs like

```
""
"xprv"
"xprvxprv"
"xprv1xprv2"
```

But it should work fine :+1:

girazoki: The issue for such a test is that I need a valid key, a valid encryption and a valid bech32 encoding with something different than "xprv" or "xprvdouble". I will try to make this work but I might have to obtain a test vector manually for that

girazoki: I forgot to mention that this happends because there is a bech32 decode that returns a different error if the decoding went wrong

wallet/src/actors/app/handlers/create_wallet.rs
```diff
@@ -14,6 +13,7 @@ pub struct CreateWalletRequest {
     seed_source: String,
     seed_data: types::Password,
     overwrite: Option<bool>,
+    backup_password: Option<types::Password>,
```
mariocao, line 16: I would comment that this password only is required if the seed source is `xprv` and it refers to the key used to encrypt the wallet master key (or in case of `xprvdouble` both external and internal keys)

wallet/src/repository/wallet/tests/factories.rs
```diff
@@ -3,7 +3,10 @@ use witnet_data_structures::chain::Hash;
 use super::*;
 use crate::db::HashMapDb;
 
-pub fn wallet(data: Option<HashMapDb>) -> (Wallet<db::HashMapDb>, db::HashMapDb) {
+pub fn wallet(
+    data: Option<HashMapDb>,
+    store_master_key: bool,
+) -> (Wallet<db::HashMapDb>, db::HashMapDb) {
```
tmpolaczyk, line 9: This looks like a hack that's only needed because of one test. In some other tests we solve this by allowing to mutate the input, for example in `test_block` in `validations/src/tests.rs`. So a more future-proof approach would be to accept a closure here:

```suggestion
    f: F,
) -> (Wallet<db::HashMapDb>, db::HashMapDb)
where F: FnOnce(&mut Wallet<db::HashMapDb>)
{
```

But this would require all tests to pass an empty `||` as an argument, which is not nice.

So I think a better approach would be to keep a `wallet` function with only one argument, and call a different function like "wallet2", which accepts two arguments, like this:

```rust
pub fn wallet(
// ...
{
    wallet2(data, true)
}
```

This is the hack we use to emulate default arguments in Rust. In this case, we default `store_master_key = true`. This way all the tests remain unchanged, and the one test that needs `store_master_key = false` can simply call `wallet2` (or `wallet_inner` or `wallet_with_args` or whatever).

wallet/src/actors/app/methods.rs
```diff
@@ -818,4 +829,141 @@ impl App {
 
         Box::new(f)
     }
+
+    /// Export wallet private key, encrypted with password
```
mariocao, line 833: I like the term "master key" (or something similar) to make it clearer that the exported key is the master (root) private key used to derive the addresses key pairs.

wallet/src/actors/app/methods.rs
```diff
@@ -818,4 +829,141 @@ impl App {
...
+        description,
+        name,
+        overwrite,
+        password,
+        seed_source,
+    })
+    .map_err(validation_error)
+}
+
+/// Validate XPRV
```
mariocao, line 902: Improve comment :)

wallet/src/actors/app/tests/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+
+    assert_eq!(expected, result);
+}
+
+#[test]
+fn test_valid_xprv_decryption() {
+    let seed_data: types::Password = "xprv1506zvl8u2r23zq8a3ayuzncwjawrx8etu5r2vqgz6r8ncdgt032w6ars0lc6jm47mj9tmcwg6wsg539992vhpqglamzqcpcq23h2ljvexsltv480utty2ma4lzmmuqy6zqfjkprnefr2kcu85lr006u9dmqxs84wx5lecr2c6lpwcg2atwvd60e295eqx245a9h7h72gt5r7gceg6avldxcejpt45ugl9cqe0aqzgsjpssmg23hxrglfu5vmu0f5my0xmqn5kmtq3m3wrgqkatf6uydwnnlp".to_string().into();
+    let backup_password = Some("password".to_string().into());
+    let result = app::methods::validate_xprv(seed_data, backup_password);
+    assert!(result.is_ok());
```
tmpolaczyk, line 142: Check that the result is the expected one

wallet/src/actors/app/tests/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+}
+
+#[test]
+fn test_valid_xprv_double_decryption() {
+    let seed_data: types::Password = "xprvdouble1ae5gfvwm339antauxg9zf7ads86ex6nj00syqghdsw5fmgu4lfx0s4zs6tt2txhznq8g47tzdhwh6pq2xmq2r92qed5cyykh2wesgzhldyzusksclcf6uq54jyzcm86f3p3nu5jmqm0vdhdf7hmac9ylkgjjlhs3fc3vd7tqsn53evszlseslxrztp00lg5vxrsj2l8caskv6xrs5gw8xnnlhzw5pq0j4yd0rvgw422fz6xeteru54n0lwfprmnwu6zl2e7nktarr6dh5n22ztk305veu4eegnxvr7a96dcrgm7cdqde2gmf2jgveppp77hpzkulx4af0kz2mawcmyxe97csqsjm5h2fx9aw8anfgsn3jp40h8gjy5ap5fgddfr808k7ldspf3xvxfkw8elx9rshhlwuyk29cmnsd3sazak27dndnumdwj9hp34kh7g86kgtarzcsr5dzl9".to_string().into();
+    let backup_password = Some("password".to_string().into());
+    let result = app::methods::validate_xprv(seed_data, backup_password);
+    assert!(result.is_ok());
```
tmpolaczyk, line 150: Check that the result is the expected one

wallet/src/actors/app/handlers/create_wallet.rs
tmpolaczyk: Should we return an error if this field is provided but it is not used?

wallet/src/repository/wallet/mod.rs
```diff
@@ -1659,6 +1660,37 @@ where
         Ok(!(state.transient_internal_addresses.is_empty()
             && state.transient_external_addresses.is_empty()))
     }
+
+    pub fn export_master_key(&self, password: &[u8]) -> Result<String> {
```
tmpolaczyk, line 1664: We should always use the `Password` type, is there any reason to use `&[u8]` here?

```suggestion
    pub fn export_master_key(&self, password: types::Password) -> Result<String> {
```

girazoki: very true

wallet/src/actors/app/methods.rs
```diff
@@ -818,4 +829,147 @@ impl App {
...
+    seed_data: types::Password,
+    backup_password: Option<types::Password>,
+) -> Result<types::SeedSource> {
+    let backup_password = backup_password.ok_or_else(|| {
+        validation_error(app::field_error(
+            "backup_password",
+            "Backup password not found for XPRV key",
+        ))
+    })?;
+    let seed_data_string = std::str::from_utf8(seed_data.as_ref()).map_err(|_| {
```
tmpolaczyk, line 915: This can only happen if seed_data is not valid UTF8, which I think is impossible because it is a `types::Password` which is just a type alias for `ProtectedString`, and `ProtectedString` must be valid UTF8.

But `ProtectedString` implements `AsRef<str>`, so you should be able to just do:

```
let seed_data_string: &str = seed_data.as_ref();
```

wallet/src/actors/app/methods.rs
```diff
@@ -818,4 +829,147 @@ impl App {
...
+        validation_error(app::field_error(
+            "seed_data",
+            "Could not convert seed data to XPRV string",
+        ))
+    })?;
+    let (hrp, ciphertext) = bech32::decode(seed_data_string).map_err(|_| {
+        validation_error(app::field_error(
+            "seed_data",
+            "Could not decode bench32 key",
```
tmpolaczyk, line 924:
```suggestion
            "Could not decode bech32 key",
```

wallet/src/repository/error.rs
```diff
@@ -41,6 +41,10 @@ pub enum Error {
     BlockConsolidation(String),
     #[fail(display = "hash parsing failed: {}", _0)]
     HashParseError(#[cause] types::HashParseError),
+    #[fail(display = "Bech32 serialization error: {}", _0)]
+    Bech32(#[cause] bech32::Error),
+    #[fail(display = "Crypto operation failed: {}", _0)]
+    CrytpoError(#[cause] crypto::Error),
```
tmpolaczyk, line 47:
```suggestion
    CryptoError(#[cause] crypto::Error),
```

wallet/src/repository/wallet/tests/factories.rs
girazoki: good

girazoki: I implemented the latter

wallet/src/actors/app/methods.rs
girazoki: great!

wallet/src/actors/app/handlers/create_wallet.rs
```diff
@@ -1,11 +1,12 @@
...
 use serde::{Deserialize, Serialize};
-
-use witnet_crypto::mnemonic::Mnemonic;
+use std::str;
 
 use crate::actors::app;
 use crate::types;
 
+/// Create Wallet request, where name, description and overwrite are optional and backup_password i
+/// s only needed if seed_source is xprv
```
tmpolaczyk, line 9:
```suggestion
/// Create Wallet request, where name, description and overwrite are optional and backup_password
/// is only needed if seed_source is xprv
```

wallet/src/repository/wallet/tests/mod.rs
```diff
@@ -1063,3 +1063,32 @@ fn test_create_vtt_with_locked_balance() {
...
+            .export_master_key(password)
+            .unwrap()
+            .starts_with("xprvdouble"),
+        false
+    );
+}
+
+#[test]
+fn test_export_xprvdouble_key() {
+    let (wallet, _db) = factories::wallet_with_args(None, false);
```
tmpolaczyk, line 1087: Add a small description to this tests:

```suggestion
    // Create a wallet that does not store the master key.
    // This is used to emulate a bug in previous versions of the wallet.
    // In that case, the exported master key format is not "xprv", it is "xprvdouble"
    let (wallet, _db) = factories::wallet_with_args(None, false);
```

wallet/src/repository/wallet/mod.rs
```diff
@@ -1659,6 +1660,37 @@ where
         Ok(!(state.transient_internal_addresses.is_empty()
             && state.transient_external_addresses.is_empty()))
     }
+
+    pub fn export_master_key(&self, password: types::Password) -> Result<String> {
+        let state = self.state.read()?;
+        let (tag, key) = if let Some(master_key) = self.db.get_opt(&keys::master_key())? {
+            let master_key_string = match master_key.to_slip32(&types::KeyPath::default()) {
+                Ok(x) => x,
+                Err(_e) => return Err(Error::TransactionBalanceOverflow),
```
tmpolaczyk, line 1669:
```suggestion
                Err(_e) => return Err(Error::SerializationError),
```

wallet/src/repository/keys.rs
```diff
@@ -207,6 +207,12 @@ pub fn address_info(
     ))
 }
 
+/// Master key
+#[inline]
+pub fn master_key() -> Key<String, ExtendedSK> {
+    Key::new("master-key".to_string())
```
tmpolaczyk, line 213:
```suggestion
pub fn master_key() -> Key<&'static str, ExtendedSK> {
    Key::new("master-key")
```

wallet/src/actors/app/handlers/create_wallet.rs
mariocao: Could make sense in the validation function. :+1:

wallet/src/repository/error.rs
```diff
@@ -44,6 +44,12 @@ pub enum Error {
     HashParseError(#[cause] types::HashParseError),
     #[fail(display = "failed creating a transaction: {}", _0)]
     TransactionCreation(#[cause] TransactionError),
+    #[fail(display = "Bech32 serialization error: {}", _0)]
+    Bech32(#[cause] bech32::Error),
+    #[fail(display = "Crypto operation failed: {}", _0)]
+    CryptoError(#[cause] crypto::Error),
+    #[fail(display = "Key serialization failed failed")]
```
tmpolaczyk, line 51: faileddouble :rofl: . And what's the "key" in this context? I guess the master key?

```suggestion
    #[fail(display = "Key serialization failed")]
```

girazoki: good catch. Depends on the database, if the master key exists it refers to the master key. Otherwise, it refers to the account keys, but I will modify the messsage as "master key serialization failed"

tmpolaczyk (approved):

wallet/src/actors/app/handlers/export_master_key.rs
```diff
@@ -0,0 +1,32 @@
...
+#[derive(Debug, Serialize, Deserialize)]
+pub struct ExportMasterKeyRequest {
+    pub wallet_id: String,
+    pub password: types::Password,
+    pub session_id: types::SessionId,
+}
+
+#[derive(Serialize)]
+pub struct ExportMasterKeyResponse {
+    private_key: String,
```
tmpolaczyk, line 15:
```suggestion
    master_key: String,
```

mariocao (approved):

PR #1717, opened 2020-11-12
feat(wallet): support multiple outputs in create_vtt

mariocao:
This PR allows creating VTT with multiple outputs (address, value, time_lock).

### DIFF
--- a/wallet/src/actors/app/handlers/create_vtt.rs
+++ b/wallet/src/actors/app/handlers/create_vtt.rs
@@ -5,27 +5,30 @@ use crate::actors::app;
 use crate::types;
 use crate::types::{Hashable as _, ProtobufConvert as _};
 
-use witnet_data_structures::chain::{Environment, PublicKeyHash};
+use witnet_data_structures::chain::Environment;
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct VttOutputParams {
+    pub address: String,
+    pub amount: u64,
+    pub time_lock: Option<u64>,
+}
 
 #[derive(Debug, Serialize, Deserialize)]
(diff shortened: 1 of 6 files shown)


mariocao:
I fixed some tests and added a `map_err` to log if something fails. :smile:

wallet/src/actors/app/handlers/create_vtt.rs
```diff
@@ -65,41 +67,51 @@ impl Handler<CreateVttRequest> for app::App {
...
-    Ok(pkh)
-}
+/// - destination address must be in the same network (testnet/mainnet)
+fn validate_output_addresses(
+    testnet: bool,
+    outputs: &[VttOutputParams],
+) -> Result<Vec<types::ValueTransferOutput>, app::ValidationErrors> {
+    outputs.iter().try_fold(vec![], |mut acc, output| {
+        types::PublicKeyHash::from_bech32(
+            if testnet {
```
girazoki, line 96: We can set the environment outside the loop right? that way we dont need to do the testnet mainnet comparison for every output

wallet/src/repository/wallet/tests/mod.rs
```diff
@@ -1101,3 +1111,64 @@ fn test_create_vtt_with_locked_balance() {
         err,
     );
 }
+
+#[test]
+fn test_create_vtt_with_multiple_outputs() {
```
girazoki, line 1116: Is there a test to check for non-well serialized addresses?

mariocao: I added tests in the `create_vtt` validation logic. :)

girazoki (approved):

PR #1719, opened 2020-11-12
[Wallet]: Add periodic node request to know if node disconnected

### LINKED ISSUE #1705: [Wallet] Implement a ping/pong between wallet <> node

clbartoli:
Add the function `periodic_node_request` that sends a `syncStatus` request, if returns an error (meaning the node is disconnected) it sends an event.

Closes #1705

### DIFF
--- a/node/src/actors/json_rpc/json_rpc_methods.rs
+++ b/node/src/actors/json_rpc/json_rpc_methods.rs
@@ -896,6 +896,8 @@ pub struct SyncStatus {
     pub current_epoch: Option<u32>,
     /// Is the node synchronized?
     pub synchronized: bool,
+    /// Node State
+    pub node_state: Option<StateMachine>,
 }
 
 /// Get node status
@@ -912,6 +914,19 @@ pub fn status() -> JsonRpcResultAsync {
             Ok(Err(())) => Err(internal_error(())),
             Err(e) => Err(internal_error(e)),
         });
+    let node_state_fut = chain_manager
+        .send(GetState)
+        .map_err(internal_error_s)
+        .then(|res| match res {
+            Ok(Ok(StateMachine::Synced)) => Ok(Some(StateMachine::Synced)),
+            Ok(Ok(StateMachine::AlmostSynced)) => Ok(Some(StateMachine::AlmostSynced)),
+            Ok(Ok(StateMachine::WaitingConsensus)) => Ok(Some(StateMachine::WaitingConsensus)),
+            Ok(Ok(StateMachine::Synchronizing)) => Ok(Some(StateMachine::Synchronizing)),
+            // Ok(Ok(..)) => Ok(None),
+            Ok(Err(())) => Err(internal_error(())),
+            Err(e) => Err(internal_error(e)),
+        });
+        
     let chain_beacon_fut = chain_manager
         .send(GetHighestCheckpointBeacon)
         .then(|res| match res {
@@ -927,11 +942,12 @@ pub fn status() -> JsonRpcResultAsync {
         Err(e) => Err(internal_error_s(e)),
     });
 
-    let j = Future::join3(synchronized_fut, chain_beacon_fut, current_epoch_fut)
-        .map(|(synchronized, chain_beacon, current_epoch)| SyncStatus {
+    let j = Future::join4(synchronized_fut, chain_beacon_fut, current_epoch_fut, node_state_fut)
+        .map(|(synchronized, chain_beacon, current_epoch, node_state)| SyncStatus {
             chain_beacon,
             current_epoch,
             synchronized,
+            node_state,
         })
         .and_then(|res| match serde_json::to_value(res) {
             Ok(x) => futures::finished(x),
--- a/node/src/actors/json_rpc/json_rpc_methods.rs
+++ b/node/src/actors/json_rpc/json_rpc_methods.rs
@@ -17,7 +17,7 @@ use serde::{Deserialize, Serialize};
 
 use witnet_crypto::key::KeyPath;
 use witnet_data_structures::{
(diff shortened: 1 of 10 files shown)


node/src/actors/json_rpc/json_rpc_methods.rs
```diff
@@ -927,19 +929,27 @@ pub fn status() -> JsonRpcResultAsync {
...
+        synchronized_fut,
+        chain_beacon_fut,
+        current_epoch_fut,
+        node_state_fut,
+    )
+    .map(
+        |(synchronized, chain_beacon, current_epoch, node_state)| SyncStatus {
             chain_beacon,
             current_epoch,
             synchronized,
```
tmpolaczyk, line 942: We should be able to remove the `synchronized_fut` if we do this:

```suggestion
            synchronized: node_state == Some(StateMachine::Synced),
```

wallet/src/actors/app/methods.rs
```diff
@@ -749,6 +756,58 @@ impl App {
...
+                for wallet in &wallets {
+                    let sink = act.state.get_sink(&wallet.session_id);
+                    act.params
+                        .worker
+                        .do_send(NotifyStatus(wallet.clone(), sink, events.clone()))
+                }
+            })
+            .map(|_res, _act, _ctx| {
+                // TODO: send event with node state
+            });
```
tmpolaczyk, line 803:
```suggestion
```

node/src/actors/json_rpc/json_rpc_methods.rs
```diff
@@ -1733,6 +1743,7 @@ mod tests {
                 "masterKeyExport",
                 "nodeStats",
                 "peers",
+                "ping",
```
tmpolaczyk, line 1746:
```suggestion
```

wallet/src/actors/app/methods.rs
```diff
@@ -754,6 +756,58 @@ impl App {
...
+            .flatten()
+            .inspect(|res| {
+                log::debug!("Periodic request result: {:?}", res);
+                let status = serde_json::from_value::<types::SyncStatus>(res.clone());
+                log::debug!("The result of the node status is {:?}", status.unwrap());
+            })
+            .into_actor(self)
+            .map_err(move |err, act, _ctx| {
+                log::warn!("Periodic request failed: {}", &err);
+                log::info!("The node in disconnected");
```
tmpolaczyk, line 792:
```suggestion
                log::info!("The node is disconnected");
```

wallet/src/actors/app/methods.rs
clbartoli: This will be used later to send a notification to Sheikah

wallet/src/actors/app/methods.rs
```diff
@@ -771,20 +783,29 @@ impl App {
             .flatten()
             .inspect(|res| {
                 log::debug!("Periodic request result: {:?}", res);
+                let status = serde_json::from_value::<types::SyncStatus>(res.clone());
+                log::debug!("The result of the node status is {:?}", status.unwrap());
```
girazoki, line 787: Can we handle this error? If this unwrap cannot fail, please say it, otherwise lets try to handle it

wallet/src/actors/app/methods.rs
```diff
@@ -771,20 +783,29 @@ impl App {
...
             })
-            .map_err(|err| {
+            .into_actor(self)
+            .map_err(move |err, act, _ctx| {
                 log::warn!("Periodic request failed: {}", &err);
-                // TODO: detect node disconnected
-            })
-            .map(|_res| {
-                // Ignore result
+                log::info!("The node is disconnected");
```
girazoki, line 792: Can we do this only when the error is a timeout error?

node/src/actors/json_rpc/json_rpc_methods.rs
```diff
@@ -912,6 +914,19 @@ pub fn status() -> JsonRpcResultAsync {
...
         });
+    let node_state_fut = chain_manager
+        .send(GetState)
+        .map_err(internal_error_s)
+        .then(|res| match res {
+            Ok(Ok(StateMachine::Synced)) => Ok(Some(StateMachine::Synced)),
+            Ok(Ok(StateMachine::AlmostSynced)) => Ok(Some(StateMachine::AlmostSynced)),
+            Ok(Ok(StateMachine::WaitingConsensus)) => Ok(Some(StateMachine::WaitingConsensus)),
+            Ok(Ok(StateMachine::Synchronizing)) => Ok(Some(StateMachine::Synchronizing)),
+            // Ok(Ok(..)) => Ok(None),
```
girazoki: uncomment

wallet/src/actors/app/methods.rs
clbartoli: I've checked the possible errors, and I think the node can be considered disconnected as well.  The possible errors when sending a request are:
- `InvalidUrl`: The wallet is pointing an invalid url, so it OK to consider the node is disconnected. The error is handled during the wallet subscriptions.
- RequestTimedOut: The node is not responding, we can consider the node is disconnected.
-RequestFailed: this is due to a connection error or an event loop that is down, since in this request there is no event loop, we can assume it is given by a connection error and so it is OK to send the NodeDisconnected event.
-SerializeFailed: since this request has no params this error is not valid in this case.

girazoki: Lets log the error at least then, as log::error

girazoki: as we spoke, this is already being done. Nothing else then!

tmpolaczyk (approved):

lrubiorod (approved):

girazoki (approved):

PR #1692, opened 2020-10-28
Wallet add weight fee

girazoki:
Add necessary UTXOs to adapt to fee added by transaction weight

Closes #1635

PR #1720, opened 2020-11-12
Weighted fee

### LINKED ISSUE #1635: [Wallet] Implement weighted fee for `create_vtt` and `create_data_request`
Wallet clients are submitting the fee of Value Transfers without taking into consideration the transaction weight it will have (it depends on the number of inputs and outputs ). Therefore, it is difficult for clients to define a fee that will have a proportional impact on how transactions are prioritized by node miners.

lrubiorod:
This PR supersides @girazoki 's PR #1692 

Close #1692 
Close #1635

### DIFF
--- a/data_structures/src/error.rs
+++ b/data_structures/src/error.rs
@@ -3,7 +3,7 @@
 use failure::Fail;
 use std::num::ParseIntError;
 
-use crate::chain::{Epoch, Hash, HashParseError, OutputPointer, PublicKeyHash};
+use crate::chain::{DataRequestOutput, Epoch, Hash, HashParseError, OutputPointer, PublicKeyHash};
 
 /// The error type for operations on a [`ChainInfo`](ChainInfo)
 #[derive(Debug, PartialEq, Fail)]
@@ -262,10 +262,14 @@ pub enum TransactionError {
     ValueTransferWeightLimitExceeded { weight: u32, max_weight: u32 },
     /// Data Request weight limit exceeded
     #[fail(
-        display = "Data Request Transaction weight ({}) exceeds the limit ({})",
-        weight, max_weight
+        display = "Data Request Transaction weight ({}) exceeds the limit ({})\n > {:?}",
+        weight, max_weight, dr_output
     )]
-    DataRequestWeightLimitExceeded { weight: u32, max_weight: u32 },
+    DataRequestWeightLimitExceeded {
+        weight: u32,
+        max_weight: u32,
+        dr_output: DataRequestOutput,
+    },
 }
 
 /// The error type for operations on a [`Block`](Block)
--- a/data_structures/src/transaction.rs
+++ b/data_structures/src/transaction.rs
@@ -13,10 +13,6 @@ use std::convert::TryFrom;
 use std::sync::{Arc, RwLock};
 use witnet_crypto::{hash::calculate_sha256, merkle::FullMerkleTree};
 
-pub trait MemoizedHashable {
-    fn hashable_bytes(&self) -> Vec<u8>;
-    fn memoized_hash(&self) -> &MemoHash;
-}
 // These constants were calculated in:
 // TODO: add link to WIP about transaction weights
 pub const INPUT_SIZE: u32 = 133;
@@ -27,6 +23,11 @@ pub const TALLY_WEIGHT: u32 = 100;
 pub const ALPHA: u32 = 1;
 pub const BETA: u32 = 1;
 pub const GAMMA: u32 = 10;
+
+pub trait MemoizedHashable {
+    fn hashable_bytes(&self) -> Vec<u8>;
+    fn memoized_hash(&self) -> &MemoHash;
+}
 
 /// A shareable wrapper for hash that may or may not be already computed.
 /// This low level structure does not include the implementation for compute-on-read, as that is up
--- a/data_structures/src/transaction.rs
+++ b/data_structures/src/transaction.rs
@@ -19,14 +19,14 @@ pub trait MemoizedHashable {
 }
 // These constants were calculated in:
 // TODO: add link to WIP about transaction weights
-const INPUT_SIZE: u32 = 133;
-const OUTPUT_SIZE: u32 = 36;
-const COMMIT_WEIGHT: u32 = 400;
-const REVEAL_WEIGHT: u32 = 200;
-const TALLY_WEIGHT: u32 = 100;
-const ALPHA: u32 = 1;
-const BETA: u32 = 1;
-const GAMMA: u32 = 10;
+pub const INPUT_SIZE: u32 = 133;
+pub const OUTPUT_SIZE: u32 = 36;
+pub const COMMIT_WEIGHT: u32 = 400;
+pub const REVEAL_WEIGHT: u32 = 200;
+pub const TALLY_WEIGHT: u32 = 100;
+pub const ALPHA: u32 = 1;
+pub const BETA: u32 = 1;
+pub const GAMMA: u32 = 10;
 
 /// A shareable wrapper for hash that may or may not be already computed.
 /// This low level structure does not include the implementation for compute-on-read, as that is up
--- a/data_structures/src/transaction_factory.rs
+++ b/data_structures/src/transaction_factory.rs
@@ -4,11 +4,12 @@ use crate::{
         ValueTransferOutput,
     },
     error::TransactionError,
-    transaction::{DRTransactionBody, VTTransactionBody},
+    transaction::{DRTransactionBody, VTTransactionBody, INPUT_SIZE},
     utxo_pool::{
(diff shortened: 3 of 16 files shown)


data_structures/src/error.rs
```diff
@@ -262,10 +262,14 @@ pub enum TransactionError {
     ValueTransferWeightLimitExceeded { weight: u32, max_weight: u32 },
     /// Data Request weight limit exceeded
     #[fail(
-        display = "Data Request Transaction weight ({}) exceeds the limit ({})",
-        weight, max_weight
+        display = "Data Request Transaction ({:?}) weight ({}) exceeds the limit ({})",
```
mariocao, line 265: Just asking myself if there is any cleaner way of identifying the DR transaction... `dr_output` in Debug might be ugly.

data_structures/src/error.rs
```diff
@@ -262,10 +262,14 @@ pub enum TransactionError {
...
+        dr_output, weight, max_weight
     )]
-    DataRequestWeightLimitExceeded { weight: u32, max_weight: u32 },
+    DataRequestWeightLimitExceeded {
+        weight: u32,
+        max_weight: u32,
+        dr_output: DataRequestOutput,
```
mariocao, line 271: order :rofl:

wallet/src/actors/app/handlers/create_data_req.rs
```diff
@@ -11,6 +11,7 @@ pub struct CreateDataReqRequest {
     wallet_id: String,
     request: DataRequestOutput,
     fee: u64,
+    fee_type: types::FeeType,
```
mariocao, line 14:
```suggestion
    fee_type: Option<types::FeeType>,
```
By default it could be weighted fee, so that we don't break the API :)

wallet/src/actors/app/handlers/create_vtt.rs
```diff
@@ -15,6 +15,7 @@ pub struct CreateVttRequest {
     label: Option<String>,
     amount: u64,
     fee: u64,
+    fee_type: types::FeeType,
```
mariocao, line 18: Same here :)

wallet/src/params.rs
```diff
@@ -43,6 +45,8 @@ impl Default for Params {
             genesis_hash: Hash::default(),
             genesis_prev_hash: Hash::default(),
             sync_address_batch_length: 10,
+            max_vt_weight: 20_000,
+            max_dr_weight: 80_000,
```
mariocao, line 49: Is it possible to set them as constants from data_structures? Same as `INPUT_SIZE`, etc...

data_structures/src/error.rs
tmpolaczyk: The obvious way is the transaction hash

wallet/src/params.rs
lrubiorod: they are consensus constants, so I suppose that I have to read from other part, because INPUT_SIZE is a harcoded constant

data_structures/src/transaction_factory.rs
```diff
@@ -180,13 +264,26 @@ pub fn build_vtt(
...
     utxo_strategy: UtxoSelectionStrategy,
+    max_weight: u32,
 ) -> Result<VTTransactionBody, TransactionError> {
     let mut utxos = NodeUtxos {
         all_utxos,
         own_utxos,
     };
 
-    let tx_info = utxos.build_inputs_outputs(outputs, None, fee, timestamp, None, utxo_strategy)?;
+    // TODO: Apply FeeTypes in the node methods
```
tmpolaczyk, line 274: Let's open an issue

data_structures/src/error.rs
lrubiorod: but that error is also used in the creation of a DataRequestTransaction, that could be achieved before to exist the proper transaction. Maybe I could include the witnesses number, because is the field that more affects to the weight

data_structures/src/error.rs
lrubiorod: order of appearing in the error message or alphabetical order?

wallet/src/params.rs
lrubiorod: it is only used for test

data_structures/src/error.rs
girazoki: For me it is prrety clear, as this error jumps  when the DR TX is being created. If it cannot be created, I understand that we need to log the body, as the tx hash will not point to anything

girazoki: or is this error also used when validating a dr tx?

girazoki (approved):

data_structures/src/transaction.rs
```diff
@@ -19,14 +19,14 @@ pub trait MemoizedHashable {
...
-const BETA: u32 = 1;
-const GAMMA: u32 = 10;
+pub const INPUT_SIZE: u32 = 133;
+pub const OUTPUT_SIZE: u32 = 36;
+pub const COMMIT_WEIGHT: u32 = 400;
+pub const REVEAL_WEIGHT: u32 = 200;
+pub const TALLY_WEIGHT: u32 = 100;
+pub const ALPHA: u32 = 1;
+pub const BETA: u32 = 1;
+pub const GAMMA: u32 = 10;
```
mariocao, line 29: Put before `pub trait MemoizedHashable {`

data_structures/src/transaction_factory.rs
```diff
@@ -121,23 +135,93 @@ pub trait OutputsCollection {
...
-            inputs,
-            outputs,
-            input_value,
-            output_value,
-        })
+        unreachable!("Unexpected exit in build_inputs_outputs method");
     }
 }
 
+pub fn calculate_weight(
```
mariocao, line 188: Comment

wallet/src/repository/error.rs
```diff
@@ -44,6 +44,19 @@ pub enum Error {
...
     TransactionCreation(#[cause] TransactionError),
+    #[fail(display = "failed because wallet is still syncing: {}", _0)]
+    StillSyncing(String),
+    #[fail(
+        display = "Weight limit reached when trying to create a VTT of value {} nanoWits",
+        _0
+    )]
+    MaximumVTTWeightReached(u64),
+    #[fail(display = "Weight limit reached when trying to create DR {:?}", _0)]
+    MaximumDRWeightReached(types::DataRequestOutput),
```
mariocao, line 55: Same as before

mariocao (approved):

data_structures/src/transaction_factory.rs
```diff
@@ -121,23 +136,96 @@ pub trait OutputsCollection {
...
     }
 }
 
+/// Calculate weight from inputs and outputs information
+pub fn calculate_weight(
+    inputs_count: usize,
+    outputs_count: usize,
+    dro: Option<&DataRequestOutput>,
+    max_weight: u32,
+) -> Result<u64, TransactionError> {
```
tmpolaczyk, line 197: We should always use the same units for weight, and I think the correct one is `u32`.

```suggestion
    max_weight: u32,
) -> Result<u32, TransactionError> {
```

data_structures/src/transaction_factory.rs
```diff
@@ -121,23 +136,96 @@ pub trait OutputsCollection {
...
-        let (output_pointers, input_value) =
-            self.take_enough_utxos(amount, timestamp, block_number_limit, utxo_strategy)?;
-        let inputs: Vec<Input> = output_pointers.into_iter().map(Input::new).collect();
+        // For the first estimation: 1 input and 1 output more for the change address
+        let mut current_weight = calculate_weight(1, outputs.len() + 1, dr_output, max_weight)?;
+
+        let max_iterations = 1 + ((u64::from(max_weight) - current_weight) / u64::from(INPUT_SIZE));
+        for _i in 0..max_iterations {
+            let fee = match fee_type {
+                FeeType::Absolute => fee,
```
tmpolaczyk, line 145: I would expect the Absolute fee check to be outside of this for loop, to make this function easier to understand. But that means duplicating some code.

wallet/src/repository/wallet/mod.rs
```diff
@@ -1095,16 +1108,53 @@ where
...
+                .get(keychain as usize)
+                .expect("could not get keychain");
+
+            let extended_sign_key =
+                parent_key.derive(&self.engine, &types::KeyPath::default().index(index))?;
+
+            let public_key: types::PK =
+                types::ExtendedPK::from_secret_key(&self.engine, &extended_sign_key).into();
+
+            witnet_data_structures::chain::PublicKey::from(public_key).pkh()
```
tmpolaczyk, line 1152: This code is a hack from the past, do you know why do we calculate the pkh again instead of using `key_balance.pkh`?

tmpolaczyk (approved):

PR #1721, opened 2020-11-12
feat(jsonrpc): calculate supply (current, locked and total)

drcpu-github:
This PR introduces a JSON RPC method to calculate the current supply of WIT tokens. This method takes the utxo-pool to calculate current supply, locked supply and a hardcoded total supply. The latter is apparently not a constant I can take from somewhere, but I assume it is implicitly defined through the block reward calculation function `block_reward` in `validations.rs`.

The supply that is calculated is a bit of an approximation. I assume the discrepancies in the below examples can be explained by in-flight transactions that mark UTXO's as in use (and thus not a UTXO anymore).

For example between screenshots 1 and 2, there are 32 epochs, which should result in a supply increase of 8000, but it is 8200.
supply_1
supply_2

The difference in supply between screenshots 3 and 4 is even negative (despite a difference of 18 epochs).
supply_3
supply_4

Maybe this could be improved by checking the transaction pool, but I guess I first need to know if my suspicion is actually correct?

Let me know what you think should be improved.

I purposely did not rebase this on master because I guess it's easier to test the PR on the latest release.


tmpolaczyk:
The UTXO set should always be a valid representation of the chain state, so the issues you are describing sound very strange. My guess is that it's because of data requests: creating data requests appears to remove supply, and when data requests resolve they return the missing supply (as reward + returned collateral). I guess we could keep track of that as a field in `DataRequestsPool`. Another reason for having negative supply is when there is a reorganization, because in that case some blocks are removed from the main chain.

The total supply can be calculated as (wits minted in genesis block) + (block_reward for all blocks that should have been mined so far) - (block_reward for missing blocks). This should be equal to the sum of values of the UTXO set (+ the wits missing because of data requests), otherwise we have a bug somewhere. I guess there are two kinds of total supply: the theoretical total supply, and the total supply without the missing blocks.

A method like this will be useful for monitoring if we expose as much information as possible. My main concern would be performance, because this method needs to iterate over all the UTXOs, but we can worry about that later. I would also include "total_block_reward_so_far" and "block_reward_in_missing_blocks" (with better names of course). I think there is no way to burn wits, but if we find one then it would also be interesting to keep track of that.

drcpu-github:
> The UTXO set should always be a valid representation of the chain state, so the issues you are describing sound very strange. My guess is that it's because of data requests: creating data requests appears to remove supply, and when data requests resolve they return the missing supply (as reward + returned collateral). I guess we could keep track of that as a field in `DataRequestsPool`. Another reason for having negative supply is when there is a reorganization, because in that case some blocks are removed from the main chain.
> 

Yes, that is what I meant with the `in-flight transactions`, was that not clear? I'll have a look at introducing such a field so we can double check the figures.

> The total supply can be calculated as (wits minted in genesis block) + (block_reward for all blocks that should have been mined so far) - (block_reward for missing blocks). This should be equal to the sum of values of the UTXO set (+ the wits missing because of data requests), otherwise we have a bug somewhere. I guess there are two kinds of total supply: the theoretical total supply, and the total supply without the missing blocks.
> 
> A method like this will be useful for monitoring if we expose as much information as possible. My main concern would be performance, because this method needs to iterate over all the UTXOs, but we can worry about that later.

This is definitely very resource-intensive, but the patch introduces nothing more than an RPC method. This is not constantly being calculated, so since the user decides he wants to know these values I don't see any issue with it being compute-intensive (unless it could cause a node to miss an epoch or something?).

> I would also include "total_block_reward_so_far" and "block_reward_in_missing_blocks" (with better names of course).

Not sure what you mean by `block_reward_in_missing_blocks`?

> I think there is no way to burn wits, but if we find one then it would also be interesting to keep track of that.

I've also wondered about burning WIT (if that is possible).

drcpu-github:
@tmpolaczyk I implemented a counter to track the total amount of collateral locked up in processing data requests. The main problem is that this is a breaking change with respect to the underlying storage. I guess that is pretty much a deal-breaker because you don't want to force the whole network to resync?

tmpolaczyk:
> Not sure what you mean by `block_reward_in_missing_blocks`?

Ah sorry, let me rephrase: there may be some missing blocks in the blockchain. The reward of this blocks is lost and will never be minted again. So we should keep track of the sum of the reward of all the missing blocks, because this changes the total supply.

> The main problem is that this is a breaking change with respect to the underlying storage.

Oh no, I didn't think about that. If we don't want to break the storage we can calculate that value on the fly, by iterating over all the active data requests. Something like this as a method of `DataRequestPool`:

```rust
let mut sum = 0;
for (_dr_hash, dr_state) in self.data_request_pool {
    sum += dr_state.data_request.checked_total_value().unwrap();
    // TODO: handle commit_fee and reveal_fee
    // if reveal stage and more than 0 commits: sum -= commit_fee
    // if tally stage: sum -= reveal_fee * num_reveals
    // TODO: handle locked collateral
    // sum += num_commits * dr_collateral;
}
```

But that will probably be hard to implement because there are many edge cases. Also I would add this value as a separate field in `SupplyInfo`.

drcpu-github:
> > Not sure what you mean by `block_reward_in_missing_blocks`?
> 
> Ah sorry, let me rephrase: there may be some missing blocks in the blockchain. The reward of this blocks is lost and will never be minted again. So we should keep track of the sum of the reward of all the missing blocks, because this changes the total supply.

Oh right, you mean that epochs without blocks due to rollbacks etc will actually lead to a reduction in total supply because the block reward calculation is based on epochs rather than blocks minted. Never actually thought about that. That definitely means I need to include a counter for that in the supply method.

> 
> > The main problem is that this is a breaking change with respect to the underlying storage.
> 
> Oh no, I didn't think about that. If we don't want to break the storage we can calculate that value on the fly, by iterating over all the active data requests. Something like this as a method of `DataRequestPool`:
> 
> ```rust
> let mut sum = 0;
> for (_dr_hash, dr_state) in self.data_request_pool {
>     sum += dr_state.data_request.checked_total_value().unwrap();
>     // TODO: handle commit_fee and reveal_fee
>     // if reveal stage and more than 0 commits: sum -= commit_fee
>     // if tally stage: sum -= reveal_fee * num_reveals
>     // TODO: handle locked collateral
>     // sum += num_commits * dr_collateral;
> }
> ```
> 

Yeah that sounds quite complicated. I already has issues with data requests not being resolved etc, breaking the collateral_locked counter. That's why I use `checked_sub`, otherwise I see overflows etc. If you don't mind that this is a breaking change to the storage, I don't either. 😉 

> But that will probably be hard to implement because there are many edge cases. Also I would add this value as a separate field in `SupplyInfo`.

I can do that. 👍

drcpu-github:
Implemented all counters and added some more details.
supply_5
supply_6

tmpolaczyk:
Super cool! I will see if I can do something to avoid breaking the storage.

drcpu-github:
@tmpolaczyk I updated the branch to print a prettier table using the num-format crate for locale formatting. Did you by any chance have time to look at the breaking change to the data request pool?

PS: The Rust formatter really asks for stupid format modifications at times ...

tmpolaczyk:
You can try something like this to avoid the breaking change:

https://github.com/tmpolaczyk/witnet-rust/commit/45db5bb60fab0e1ffec4a5732876816bad472c03

I tried to add tests to the `GetSupplyInfo` method because that should make it easier to find and fix any possible edge cases, but it is hard to test the `ChainManager` so I gave up. This means that the supply may not be 100% accurate. So not sure what to do. It would be nice to have some way to mark some cli methods as experimental, or add a warning that the results may not be 100% accurate, or something to avoid people panicking when they see the supply go down...

drcpu-github:
Yeah, the supply going down definitely confused me at first. I guess your method works fine overall speaking, only edge cases where data requests don't get enough commits etc could potentially break it.

One other possibility is to just shelf this method until the next breaking release, just like with the modifications @lrubiorod proposed? It's not like there is a big community ask for it and I would just use if in my explorer. For me the breaking change is not a big issue as I can always sync nodes in non-mining mode and replace the explorer nodes when possible.

src/cli/node/json_rpc_client.rs
```diff
@@ -247,6 +249,79 @@ pub fn get_utxo_info(
...
+        "".to_string(),
+        current_supply.to_formatted_string(&Locale::en)
+    ]);
+    supply_table.add_row(row![
+        "Locked supply".to_string(),
+        "".to_string(),
+        locked_supply.to_formatted_string(&Locale::en)
+    ]);
+    supply_table.add_row(row![
+        "Total supply".to_string(),
```
aesedepece, line 316:
```suggestion
        "Maximum supply".to_string(),
```

src/cli/node/json_rpc_client.rs
```diff
@@ -247,6 +249,79 @@ pub fn get_utxo_info(
...
+    ]);
+    supply_table.add_row(row![
+        "Collateralized data requests".to_string(),
+        supply_info
+            .collateralized_data_requests
+            .to_formatted_string(&Locale::en),
+        collateralized_data_requests_total_wit.to_formatted_string(&Locale::en)
+    ]);
+    supply_table.add_row(row![
+        "Current supply".to_string(),
```
aesedepece, line 306: Does `current` include `locked` or not?
- If included, I'd rename it to `Total`, but I'm also sort of fine with `Current`
- If not, I'd rename it to `Circulating`

aesedepece (changes requested):

src/cli/node/json_rpc_client.rs
drcpu-github: It does not, so circulating it is.

src/cli/node/json_rpc_client.rs
drcpu-github: Done.

drcpu-github:
I have applied the suggested changes to the supply names. Are there plans to merge this (taking into account that the current implementation breaks storage)? If not, I'm not going to bother rebasing, because that'll likely be a significant effort given that the commit is about 8 months old.

tmpolaczyk:
I think that it will be easier to just implement it from scratch, so let's close this. But it's a shame to lose all this discussion, so I will create an issue and copy the relevant parts there: the desired API, edge cases, and impediments.

tmpolaczyk:
Closing, use #1989 for any further discussion

PR #1725, opened 2020-11-17
feat(wallet): mark UTXOs as used in sendTransaction instead of create…

### LINKED ISSUE #1718: [Wallet] If a user cancels the `create_vtt`, the used UTXOs stayed marked as already spent (but only for a period of time)
### Context:

The create and send value transfer were split into 2 different API calls (and steps):
 1. `create_vtt`: receives VTT parameters and returns the transaction object, so that the end-used may check all transaction fields.
 2. `send_transaction`: receives the transaction object and sends it to the node for its broadcasting.

### Problem:

Right now, the step 1 is marking UTXOs as used. If the Sheikah user, cancels the interrupts the send transaction process by clicking on "cancel", then step 2 is not executed but UTXO stay marked as used for a period of time.

Main consequence, is that after cancelling a send transaction, if the user has not enough additional UTXOS, it has to wait (~7 minutes) to be able to retry sending again a transaction.

### Potential solution:

A possible solution is that step 1 does not mutate any state in the wallet, and only step 2 really does the logic of marking UTXOs as used. However, as pointed out by some colleagues, this issues might be a good opportunity to discuss if the send transaction flow makes sense and if it could be simplified.

Making these changes on step 2 would require to add some logic in the `send_transaction` method from the `App` actor and most probably in the function of `add_local_movement` of `repository/wallet/mod.rs`.

lrubiorod:
…Transaction

Close #1718

### DIFF
--- a/wallet/src/repository/wallet/mod.rs
+++ b/wallet/src/repository/wallet/mod.rs
@@ -17,6 +17,7 @@ use witnet_data_structures::{
         OutputPointer, PublicKeyHash, ValueTransferOutput,
     },
     get_environment,
+    transaction::Transaction,
     transaction_factory::{insert_change_output, FeeType, OutputsCollection},
     utxo_pool::UtxoSelectionStrategy,
 };
@@ -1075,7 +1076,6 @@ where
         fee_type: FeeType,
     ) -> Result<(Vec<Input>, Vec<ValueTransferOutput>)> {
         let utxo_strategy = UtxoSelectionStrategy::Random;
-        let tx_pending_timeout = u64::from(state.epoch_constants.checkpoints_period) * 10;
         let timestamp = u64::try_from(get_timestamp()).unwrap();
 
         let (inputs, outputs) = self.build_inputs_outputs_wallet(
@@ -1085,7 +1085,6 @@ where
             fee_type,
             state,
             timestamp,
-            tx_pending_timeout,
             None,
             utxo_strategy,
             self.params.max_vt_weight,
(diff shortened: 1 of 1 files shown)


wallet/src/repository/wallet/mod.rs
```diff
@@ -1317,6 +1311,26 @@ where
     ) -> Result<Option<model::BalanceMovement>> {
         let mut state = self.state.write()?;
 
+        // Mark UTXOs as used so we don't double spend
+        // Save the timestamp until it could be spend it
```
girazoki, line 1315:
```suggestion
        // Save the timestamp to after which the UTXO can be spent again
```

wallet/src/repository/wallet/mod.rs
```diff
@@ -1317,6 +1311,26 @@ where
     ) -> Result<Option<model::BalanceMovement>> {
         let mut state = self.state.write()?;
 
+        // Mark UTXOs as used so we don't double spend
+        // Save the timestamp to after which the UTXO can be spent again
+        let tx_pending_timeout = u64::from(state.epoch_constants.checkpoints_period) * 10;
+        let timestamp = u64::try_from(get_timestamp()).unwrap();
+        let mut wallet_utxos = WalletUtxos {
+            utxo_set: &UtxoSet::default(),
```
tmpolaczyk, line 1319: This is not needed because `wallet_utxos.set_used_output_pointer` only modifies the `used_outputs` field, but it looks like a hack so it's better to simply use a reference to the utxo set here

```suggestion
            utxo_set: &state.utxo_set,
```

wallet/src/repository/wallet/mod.rs
```diff
@@ -31,6 +31,8 @@ use crate::{
 };
 
 use super::*;
+use crate::model::UtxoSet;
+use witnet_data_structures::transaction::Transaction;
```
tmpolaczyk, line 35: fmt

tmpolaczyk (approved):

girazoki (approved):

PR #1726, opened 2020-11-17
feat(cli): include weighted_fee in node send commands

lrubiorod:


src/cli/node/with_node.rs
```diff
@@ -432,6 +445,9 @@ pub enum Command {
         /// Print the request that would be sent to the node and exit without doing anything
         #[structopt(long = "dry-run")]
         dry_run: bool,
+        /// Use weighted fee (instead of absolute)
+        #[structopt(long = "weighted_fee")]
+        weighted_fee: bool,
```
tmpolaczyk, line 450: This means that the syntax will be

    witnet node send --fee=1 --weighted_fee

I think that ideally the syntax should be

    witnet node send --weighted-fee=1

This can be implemented if we make both fields `Option<u64>`. There should be a way to tell `structopt` to accept "either --fee or --weighted-fee", I think it was called "conflicts_with" or something like that.

node/src/actors/messages.rs
```diff
@@ -192,6 +192,8 @@ pub struct BuildVtt {
     pub vto: Vec<ValueTransferOutput>,
     /// Fee
     pub fee: u64,
+    /// weighted fee flag
+    pub weighted_fee: bool,
```
tmpolaczyk, line 196: Similar to the CLI, I would prefer the JSON-RPC API to accept either `"fee":1` or `"weighted_fee":1`, if that's possible.

tmpolaczyk:
We need to discuss the desired API, and whether or not set the default behavior to use a "weighted" fee.

node/src/actors/messages.rs
lrubiorod: it would be a breking change then

src/cli/node/json_rpc_client.rs
```diff
@@ -463,10 +464,18 @@ pub fn send_vtt(
         None => UtxoSelectionStrategy::Random,
     };
 
+    let (fee, weighted_fee) = match (fee, weighted_fee) {
+        (Some(x), None) => (x, false),
+        (None, Some(x)) => (x, true),
+        (None, None) => bail!("Any fee included"),
```
mariocao, line 470: Maybe something clearer :)
```suggestion
        (None, None) => bail!("A fee should be specified"),
```

lrubiorod:
Close until doing a properly discussion

PR #1727, opened 2020-11-19
fix(mempool): only clear pending transactions hashset after a superbl…

lrubiorod:
…ock consolidation

### DIFF
--- a/node/src/actors/chain_manager/handlers.rs
+++ b/node/src/actors/chain_manager/handlers.rs
@@ -95,9 +95,6 @@ impl Handler<EpochNotification<EveryEpochPayload>> for ChainManager {
             chain_beacon.hash_prev_block
         );
 
-        // Clear pending transactions HashSet
-        self.transactions_pool.clear_pending_transactions();
-
         // Handle case consensus not achieved
         if !self.peers_beacons_received {
             log::warn!("No beacon messages received from peers. Moving to WaitingConsensus state");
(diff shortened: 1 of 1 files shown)


girazoki (approved):

node/src/actors/chain_manager/handlers.rs
```diff
@@ -1129,6 +1126,11 @@ impl Handler<PeersBeacons> for ChainManager {
             // During epoch 0 there is no need to create the superblock 0
             if current_epoch != 0 && current_epoch % superblock_period == 0 {
                 self.create_and_broadcast_superblock(ctx, current_epoch);
+
+                // Clear pending transactions HashSet
+                // Only clear after a superblock consolidation to avoid receiving repeated transactions
+                // during a reverting process
+                self.transactions_pool.clear_pending_transactions();
```
tmpolaczyk, line 1133: This will only clear the pending transactions if the state machine is Synced or AlmostSynced. Previously, the pending transactions were cleared on every epoch, regardless of the node state. I don't know if this may be a big problem, but imagine a node that takes 24 hours to synchronize, it will have 24 hours worth of hashes stored in memory.

Also, I'm not sure if the comment is true. You can receive one transaction in epoch 109, then in epoch 110 revert and then in epoch 110 receive the same transaction again.

mariocao (changes requested):
After discussing this PR, we concluded we have to review if this PR still makes sense after the last mempool changes.

PR #1728, opened 2020-11-19
feat(mining): do not include requests in a new block if there are oth…

lrubiorod:
…ers to resolve


data_structures/src/chain.rs
```diff
@@ -2034,10 +2034,14 @@ impl TransactionsPool {
 
     /// Returns a tuple with a vector of commit transactions that achieve the minimum specify
     /// by the data request, and the value of all the fees obtained with those commits
```
tmpolaczyk, line 2036:
```suggestion
    /// by the data request, the value of all the fees obtained with those commits, and
    /// the list of data requests that have reached the commit limit
```

girazoki (approved):

PR #1729, opened 2020-11-19
feat(chain manager): allow configuring a minimum reward to collateral ratio

girazoki:


node/src/actors/chain_manager/mining.rs
```diff
@@ -246,8 +247,15 @@ impl ChainManager {
             .wait(ctx);
     }
 
+    /// Check whether the reward covers the minimum specified collateral
+    pub fn check_reward_to_collateral_filter(&self, dr_state: &DataRequestState) -> bool {
+        let dr_reward_to_collateral_ratio =
+            (dr_state.data_request.witness_reward / dr_state.data_request.collateral) * 1000;
```
tmpolaczyk, line 253: You need to use floats here, otherwise `witness_reward / collateral` will return `0` if `witness_reward < collateral`
```suggestion
            (f32::from(dr_state.data_request.witness_reward) / f32::from(dr_state.data_request.collateral)) * 1000.0;
```

girazoki:
Closed due to further analysis required, specially indicating whether this should be enforced by protocol

PR #1730, opened 2020-11-19
feat(node): stop validating commits on witness limit

tmpolaczyk:
* Commit validation is expensive because of the VRF proof.
* If a data request does not reach the desired amount of commits, it is retried on the next epoch with a backup factor, which means more commits.

We currently validate all those extra commits, even though only some of them may be eventually included in a block. This PR tries to improve that by validating only the required amount of commits. Any extra commits will be ignored and will not be broadcasted.

This changes the commit priority in blocks: previously, each node uses hashmap iteration to decide which commits are included in a block, which is supposed to be random. But with this PR, each node will use the first commits that arrive to them, which means prioritizing faster nodes or nodes geographically close to them.

### DIFF
--- a/node/src/actors/chain_manager/mod.rs
+++ b/node/src/actors/chain_manager/mod.rs
@@ -976,6 +976,27 @@ impl ChainManager {
                 }
             }
 
+            // Stop validating new commit transactions as soon as we achieve the limit of commits because
+            // the VRF validation is expensive. The limit is the number of witnesses of the data request.
+            if let Transaction::Commit(co_tx) = &msg.transaction {
+                let dr_pointer = co_tx.body.dr_pointer;
+                if let Some(dr_state) = self
+                    .chain_state
+                    .data_request_pool
+                    .data_request_state(&dr_pointer)
+                {
+                    let dr_output = &dr_state.data_request;
+                    let commit_limit = usize::from(dr_output.witnesses);
+                    let num_commits_in_pool = self
+                        .transactions_pool
+                        .num_commits_for_data_request(&dr_pointer);
+                    if num_commits_in_pool >= commit_limit {
+                        log::debug!("Ignoring commit {} because of commit limit", co_tx.hash());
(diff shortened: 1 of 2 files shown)


node/src/actors/chain_manager/mod.rs
```diff
@@ -976,6 +976,27 @@ impl ChainManager {
...
+            // the VRF validation is expensive. The limit is the number of witnesses of the data request.
+            if let Transaction::Commit(co_tx) = &msg.transaction {
+                let dr_pointer = co_tx.body.dr_pointer;
+                if let Some(dr_state) = self
+                    .chain_state
+                    .data_request_pool
+                    .data_request_state(&dr_pointer)
+                {
+                    let dr_output = &dr_state.data_request;
+                    let commit_limit = usize::from(dr_output.witnesses);
```
tmpolaczyk, line 989: I thought about adding a small factor here, either `2 * witnesses` or `witnesses + 10`.

node/src/actors/chain_manager/mod.rs
```diff
@@ -976,6 +976,27 @@ impl ChainManager {
...
+                    let num_commits_in_pool = self
+                        .transactions_pool
+                        .num_commits_for_data_request(&dr_pointer);
+                    if num_commits_in_pool >= commit_limit {
```
tmpolaczyk, line 993: We can add some randomness here, for example we could validate 10% of the commits when over the limit. So if we have 1000 commits for a data request with 100 witnesses, this would result in a total of about 190 commits: the first 100 + 10% of the remaining 900.

```suggestion
                    if num_commits_in_pool >= commit_limit && rng.chance(0.1) {
```

tmpolaczyk:
Upon further analysis it looks like the problem is not that there are too many commits for one data request, but that there are too many concurrent data requests. Therefore, this PR does not help much.

lrubiorod:
Closed without merging according to the reasons explained by @tmpolaczyk

PR #1731, opened 2020-11-19
[Wallet]: send note status notification from the periodic request

clbartoli:
This PR removes the wallet status subscription and sends the node status changed from the periodic request. 
Since the periodic request is returning the node status, the wallet can know when the node status has changed and send the NodeStatusRequest.

### DIFF
--- a/node/src/actors/json_rpc/json_rpc_methods.rs
+++ b/node/src/actors/json_rpc/json_rpc_methods.rs
@@ -261,9 +261,6 @@ pub fn jsonrpc_io_handler(
                     "superblocks" => {
                         add_subscription("superblocks", subscriber);
                     }
-                    "status" => {
-                        add_subscription("status", subscriber);
-                    }
                     e => {
                         log::debug!("Unknown subscription method: {}", e);
                         // Ignore errors with `.ok()` because an error here means the connection was closed
--- a/node/src/actors/json_rpc/json_rpc_methods.rs
+++ b/node/src/actors/json_rpc/json_rpc_methods.rs
@@ -261,6 +261,9 @@ pub fn jsonrpc_io_handler(
                     "superblocks" => {
                         add_subscription("superblocks", subscriber);
                     }
+                    "status" => {
+                        add_subscription("status", subscriber);
+                    }
                     e => {
                         log::debug!("Unknown subscription method: {}", e);
                         // Ignore errors with `.ok()` because an error here means the connection was closed
--- a/wallet/src/actors/app/methods.rs
+++ b/wallet/src/actors/app/methods.rs
@@ -512,7 +512,6 @@ impl App {
         match topic.as_str() {
             "blocks" => self.handle_block_notification(value),
             "superblocks" => self.handle_superblock_notification(value),
-            "status" => self.handle_node_status_notification(value),
             _ => {
                 log::debug!("Unhandled `{}` notification", topic);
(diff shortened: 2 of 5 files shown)


wallet/src/actors/app/methods.rs
```diff
@@ -764,6 +730,77 @@ impl App {
...
+                    let sink = act.state.get_sink(&wallet.session_id);
+                    act.params
+                        .worker
+                        .do_send(NotifyStatus(wallet.clone(), sink, events.clone()))
+                }
+            })
+            .map(move |res, act, _ctx| {
+                let status = serde_json::from_value::<types::SyncStatus>(res);
+                // Notify if the node status is changed
+                if status.as_ref().is_ok()
```
girazoki, line 782:
```suggestion
                if let OK(status) = status
```

wallet/src/actors/app/state.rs
```diff
@@ -173,7 +173,11 @@ impl State {
     }
 
     /// Updates the node state
-    pub fn update_node_state(&mut self, node_state: StateMachine) {
-        self.node_state = Some(node_state);
+    pub fn update_node_state(&mut self, node_state: Option<StateMachine>) {
+        if node_state == None {
+            self.node_state = None;
```
girazoki, line 178: This can be rewritten as self.node_state = node_state. But we can check whether we already can change it from app/methods, thus avoiding this function

node/src/actors/chain_manager/mod.rs
```diff
@@ -1460,7 +1459,6 @@ impl ChainManager {
             ),
         }
 
-        self.notify_node_status(next_state);
```
tmpolaczyk, line 1463: Breaking change, the node should still support "status" subscriptions even if the wallet is not using it.

node/src/actors/json_rpc/json_rpc_methods.rs
```diff
@@ -261,9 +261,6 @@ pub fn jsonrpc_io_handler(
                     "superblocks" => {
                         add_subscription("superblocks", subscriber);
                     }
-                    "status" => {
```
tmpolaczyk, line 264: This is also a breaking change in the node

wallet/src/actors/app/methods.rs
```diff
@@ -800,21 +764,40 @@ impl App {
...
+                            });
+                        }
+                    }
+                } else {
+                    log::error!("Periodic request result serialization failed");
+                }
             });
         ctx.spawn(f);
 
         // Try to contact the node once every 10 seconds
```
tmpolaczyk, line 799:
```suggestion
        // Try to contact the node once every 15 seconds
```

wallet/src/actors/worker/methods.rs
```diff
@@ -1074,7 +1074,7 @@ impl Worker {
         wallet: types::SessionWallet,
         sink: types::DynamicSink,
     ) -> Result<()> {
-        log::info!("The node has changed its status into {:?}", status);
+        log::info!("The current node status is {:?}", status);
```
tmpolaczyk, line 1077: We will see this every 15 seconds, should we move the log to debug?

```suggestion
        log::debug!("The current node status is {:?}", status);
```

girazoki (approved):

tmpolaczyk (approved):

PR #1736, opened 2020-11-20
send event after tx broadcast

girazoki:
Send an event to sheikah after a local tx is sent to the network


tmpolaczyk:
I don't understand why is this needed. Sheikah sends a transaction by calling the `sendTransaction` wallet method. When this method finishes, it sends the `SendTransactionResponse` as the result of the JSON-RPC call. Therefore, sheikah can know if the transaction is sent, by looking at the result of the JSON-RPC call. This new `LocalTransactionSent` notification doesn't make any sense.

girazoki:
I agree with you, I assumed that if the issue existed there was a reason why Sheikah was not refreshing the getTransactions page by itself

aesedepece:
That's correct, no need for new messages. We simply need to return a `Movement` object inside `SendTransactionResponse` so that the client can push that into their transactions list.

lrubiorod:
@girazoki, @aesedepece and @tmpolaczyk. Could it be closed?

girazoki:
Superseeded  by #1754

PR #1737, opened 2020-11-20
feat(validations): do not validate signatures from mempool transactions

lrubiorod:


validations/src/validations.rs
```diff
@@ -13,6 +13,7 @@ use witnet_crypto::{
     merkle::{merkle_tree_root as crypto_merkle_tree_root, ProgressiveMerkleTree},
     signature::{verify, PublicKey, Signature},
 };
+use witnet_data_structures::chain::TransactionsPool;
```
girazoki, line 16: FMT

lrubiorod:
There is a problem if the hash is the same but the signatures change, because the hash doesnt include the signature.

lrubiorod:
tagged in #1352

PR #1740, opened 2020-11-23
feat(superblock): spread supervotes along the superepoch

lrubiorod:


node/src/actors/chain_manager/mod.rs
```diff
@@ -741,11 +742,19 @@ impl ChainManager {
...
-                        // Broadcast vote one epoch checkpoint later.
+                        // Broadcast vote between one and ("superblock_period" - 3) epoch checkpoints later.
                         // This is used to prevent the race condition described in issue #1573
+                        let mut rng = rand::thread_rng();
                         let checkpoints_period = act.consensus_constants().checkpoints_period;
+                        let superblock_period = act.consensus_constants().superblock_period;
+                        let end_range = if superblock_period > 4 {
+                            (superblock_period - 3) * checkpoints_period
+                        } else {
+                            checkpoints_period + 1
```
tmpolaczyk, line 753: We can use this to spread the votes across one epoch:

```suggestion
                            checkpoints_period * 2
```

lrubiorod: if we use a superblock period of 4 or less we should send the votes a soon as possible to keep a safe margin

tmpolaczyk: Ok, seems fair

tmpolaczyk (approved):

girazoki (approved):

PR #1748, opened 2020-11-26
refactor(node): remove unnecessary synchronization bool in sycstatus method

### LINKED ISSUE #1734: Refactor SyncStatus method
The syncstatus method is returning:
- node state : Option<StateMachine>
- synchronized: Bool
Since knowing the node state we can know if it is synced, the synchronized param could be removed

clbartoli:
Closes #1734

### DIFF
--- a/src/cli/node/json_rpc_client.rs
+++ b/src/cli/node/json_rpc_client.rs
@@ -22,7 +22,8 @@ use witnet_crypto::{
 use witnet_data_structures::{
     chain::{
         Block, ConsensusConstants, DataRequestInfo, DataRequestOutput, Environment, KeyedSignature,
-        NodeStats, OutputPointer, PublicKey, PublicKeyHash, SyncStatus, ValueTransferOutput,
+        NodeStats, OutputPointer, PublicKey, PublicKeyHash, StateMachine, SyncStatus,
+        ValueTransferOutput,
     },
     proto::ProtobufConvert,
     transaction::Transaction,
@@ -995,7 +996,7 @@ pub fn get_node_stats(addr: SocketAddr) -> Result<(), failure::Error> {
     let sync_status: SyncStatus = parse_response(&response)?;
 
     if let Some(current_epoch) = sync_status.current_epoch {
-        if sync_status.synchronized {
+        if sync_status.node_state == StateMachine::Synced {
             println!(
                 "The node is synchronized and the current epoch is {}",
                 current_epoch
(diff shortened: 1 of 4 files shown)


tmpolaczyk:
To me

         if sync_status.synchronized { 

looks better than

         if sync_status.node_state.unwrap() == StateMachine::Synced { 

so I think we should just keep the synchronized flag

src/cli/node/json_rpc_client.rs
```diff
@@ -995,7 +996,7 @@ pub fn get_node_stats(addr: SocketAddr) -> Result<(), failure::Error> {
     let sync_status: SyncStatus = parse_response(&response)?;
 
     if let Some(current_epoch) = sync_status.current_epoch {
-        if sync_status.synchronized {
+        if sync_status.node_state.unwrap() == StateMachine::Synced {
```
lrubiorod, line 999: when the node_state is None?

girazoki: I guess that can only happen if the state is not initialized yet, so maybe we can have a heck for that with another if let Some(status) = status. That is another level of indentation but I guess it looks better than an unwrap. Or maybe at least we should do an unwrap_or_else, returning an error.

clbartoli: Actually it is never none, since when the node is initialized the state is waitingConsensus. I've changed it so that the node_state is not an option anymore

girazoki: good

lrubiorod (approved):

src/cli/node/json_rpc_client.rs
girazoki: breaking change then

lrubiorod:
Due to be a breaking change it should be merged with all the breaking changes for the December release

girazoki (approved):

mariocao (approved):

aesedepece:
Isn't this ready to be merged?

girazoki:
yep

tmpolaczyk:
I think this may break the compatibility of old versions of the wallet with new versions of the node

aesedepece:
> I think this may break the compatibility of old versions of the wallet with new versions of the node

Interesting. Currently, I think this could be a no-problem, as even if a user refuses to upgrade Sheikah, the latest wallet version is always used, and we're assuming that the WF-operated public nodes are always in that same version.

PR #1749, opened 2020-11-26
Add tests to inventory manager

### LINKED ISSUE #1746: Fix: getTransaction doesnt update block information
When we see a Transaction for the first time in a block, it is saved into the storage with that block information and never updates. So if it happened a revert, and that transaction is included in a new block, the getTransaction method continues returning the information about the first block where it was included.

Possible solutions:
- Update Info
- Do not save into the storage until superblock confirmation

tmpolaczyk:
Close #1746

### DIFF
--- a/node/src/actors/inventory_manager/handlers.rs
+++ b/node/src/actors/inventory_manager/handlers.rs
@@ -312,3 +312,173 @@ impl Handler<GetItemSuperblock> for InventoryManager {
         self.handle_get_item_superblock(msg)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::{
+        actors::chain_manager::mining::build_block, config_mngr, storage_mngr,
+        utils::test_actix_system,
+    };
+    use futures_util::compat::Compat01As03;
+    use std::sync::Arc;
+    use witnet_config::config::{Config, StorageBackend};
+    use witnet_data_structures::{
+        chain::{
+            CheckpointBeacon, EpochConstants, Input, KeyedSignature, OutputPointer, PublicKeyHash,
+            TransactionsPool, ValueTransferOutput,
+        },
+        data_request::DataRequestPool,
+        transaction::{Transaction, VTTransaction, VTTransactionBody},
+        utxo_pool::UnspentOutputsPool,
+        vrf::BlockEligibilityClaim,
+    };
+
+    const INITIAL_BLOCK_REWARD: u64 = 250 * 1_000_000_000;
+    const HALVING_PERIOD: u32 = 3_500_000;
+
+    static MILLION_TX_OUTPUT: &str =
+        "0f0f000000000000000000000000000000000000000000000000000000000000:0";
+    static MY_PKH_1: &str = "wit18cfejmk3305y9kw5xqa59rwnpjzahr57us48vm";
+
+    fn build_block_with_vt_transactions(block_epoch: u32) -> Block {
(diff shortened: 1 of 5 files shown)


node/src/actors/inventory_manager/handlers.rs
```diff
@@ -276,3 +276,316 @@ impl Handler<GetItemSuperblock> for InventoryManager {
...
+    use witnet_config::config::Config;
+    use witnet_config::config::StorageBackend;
+    use witnet_data_structures::chain::{
+        CheckpointBeacon, EpochConstants, Input, KeyedSignature, OutputPointer, PublicKeyHash,
+        TransactionsPool, ValueTransferOutput,
+    };
+    use witnet_data_structures::data_request::DataRequestPool;
+    use witnet_data_structures::transaction::{Transaction, VTTransaction, VTTransactionBody};
+    use witnet_data_structures::utxo_pool::UnspentOutputsPool;
+    use witnet_data_structures::vrf::BlockEligibilityClaim;
```
lrubiorod, line 297: multiple fmt :P

node/src/actors/inventory_manager/handlers.rs
```diff
@@ -276,3 +276,316 @@ impl Handler<GetItemSuperblock> for InventoryManager {
...
+                pointer2: None,
+            };
+            let state = Arc::new(RwLock::new(state));
+            let state2 = state.clone();
+            let state3 = state.clone();
+            let state4 = state.clone();
+            let state5 = state.clone();
+            let state6 = state.clone();
+            let state7 = state.clone();
+            let state8 = state.clone();
```
lrubiorod, line 432: bridge style!!

tmpolaczyk:
This PR would have been easier if there was some easy way to create blocks in tests. I guess we will eventually need some way to modify the `ChainManager` state for testing purposes, and generate fake protocol messages, so creating that can be an interesting refactor.

node/src/actors/inventory_manager/handlers.rs
```diff
@@ -276,3 +276,188 @@ impl Handler<GetItemSuperblock> for InventoryManager {
...
+            let res = Compat01As03::new(f).await.unwrap();
+            res.unwrap();
+
+            // Get first transaction again
+            let f = inventory_manager.send(GetItemTransaction { hash: tx_hash1 });
+            let res = Compat01As03::new(f).await.unwrap();
+
+            // Now, the transaction pointer should point to the second block
+            let (_tx, tx_pointer2) = res.unwrap();
+            assert_eq!(tx_pointer2.block_hash, block_hash2);
```
girazoki, line 460: I am still wondering why this does not work in mainnet, although I guess investigating that becomes another issue

girazoki (approved):

node/src/actors/inventory_manager/handlers.rs
tmpolaczyk: I was thinking about adding more tests until we find the issue. The current test checks that the problem is not in `InventoryManager`, so the next step would be to test `ChainManager`, but we need more information to reproduce the issue.

tmpolaczyk: I marked issue #1746 as closed because we couldn't reproduce it, I will try to merge this PR anyway because it could be useful in the future, if we make some changes to the `InventoryManager` and want to test it.

mariocao (approved):

PR #1751, opened 2020-11-30
ref(wallet): remove unnamed tuples in wallet

### LINKED ISSUE #1744: Refactor: remove unnamed tuples from wallet
Unnamed tuples should only be used when it is obvious what does each field refer to. For example, here the purpose of the `bool` and the `Option<String>` is not obvious:

https://github.com/witnet/witnet-rust/blob/1cd3fa35109fcfe6da19156bbcce49f37c4b23da/wallet/src/actors/worker/handlers/gen_address.rs#L6

Thankfully, if we scroll down a few lines the fields are given a name:

https://github.com/witnet/witnet-rust/blob/1cd3fa35109fcfe6da19156bbcce49f37c4b23da/wallet/src/actors/worker/handlers/gen_address.rs#L15-L19

But this is completely pointless because we can give the fields a name:

```rust
pub struct GenAddress {
    pub wallet: types::SessionWallet,
    pub external: bool,
    pub label: Option<String>,
}
```

And then use it like this (note that `()` has been replaced with `{}`):

```rust
     GenAddress { wallet, external, label }: GenAddress, 
```

In some cases this is not a problem, for example here it is obvious what does each field refer to: the first is a wallet and the second is just extra parameters:

https://github.com/witnet/witnet-rust/blob/1cd3fa35109fcfe6da19156bbcce49f37c4b23da/wallet/src/actors/worker/handlers/create_vtt.rs#L6

But often we add one extra field to one of this structs in a PR, and then it becomes a problem. So let's try to avoid that.

clbartoli:
Change structs in wallet in order to avoid unnamed tuples 
closes #1744

### DIFF
--- a/wallet/src/actors/app/methods.rs
+++ b/wallet/src/actors/app/methods.rs
@@ -115,7 +115,11 @@ impl App {
         .and_then(move |wallet, slf: &mut Self, _| {
             slf.params
                 .worker
-                .send(worker::GenAddress(wallet, external, label))
+                .send(worker::GenAddress {
+                    wallet,
+                    external,
+                    label,
+                })
                 .flatten()
                 .map_err(From::from)
                 .into_actor(slf)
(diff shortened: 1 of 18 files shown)


tmpolaczyk (approved):

mariocao (approved):
Great!

PR #1754, opened 2020-11-30
send event after tx broadcast

girazoki:


wallet/src/types.rs
```diff
@@ -238,6 +238,8 @@ pub enum Event {
     /// A list of hashes of blocks that are now considered orphaned.
     BlocksOrphan(Vec<String>),
     /// A new movement (transaction) affecting balance.
+    LocalTransactionSent(model::BalanceMovement),
```
tmpolaczyk, line 241: Should we create a new event, or should we reuse the existing `Movement`?

girazoki: Depends on whether it makes a difference for sheikah

wallet/src/actors/app/methods.rs
```diff
@@ -631,25 +631,37 @@ impl App {
...
         let f = fut::result(
             self.state
                 .get_wallet_by_session_and_id(&session_id, &wallet_id),
         )
-        .and_then(move |wallet, act: &mut Self, _| {
-            act.send_inventory_transaction(transaction.clone())
-                .and_then(move |jsonrpc_result, _slf, _ctx| {
+        .and_then(move |wallet, slf: &mut Self, _| {
+            slf.send_inventory_transaction(transaction.clone())
+                .and_then(move |jsonrpc_result, slf: &mut Self, _ctx| {
```
tmpolaczyk, line 644: Let's try to follow the same convention as in the node

```suggestion
                .and_then(move |jsonrpc_result, act, _ctx| {
```

wallet/src/actors/app/methods.rs
```diff
@@ -631,25 +631,37 @@ impl App {
...
                     match wallet.add_local_movement(&model::ExtendedTransaction {
-                        transaction,
+                        transaction: transaction.clone(),
```
tmpolaczyk, line 646: The clone shouldn't be needed?

```suggestion
                        transaction,
```

tmpolaczyk (approved):

mariocao (approved):

PR #1756, opened 2020-12-01
Reinsert tx mempool

### LINKED ISSUE #1702: Smart mempool cleaning after reorganization
When reorg happens, collect transactions with inputs that contain unconfirmed output pointers, remove them from mempool and reinsert them once UTXO set has rolled back

### LINKED ISSUE #1763: Remove run_later re-inserting transactions
After PR #1756, mempool transactions and unconfirmed transactions are included again in the mempool using a run_later of 5 epochs. Those 5 epochs are needed to the node have time enough to be synced again before to receive the reinserted transactions, because if not, they would be discarded.

It would be helpful to use a method that allows to process the reinserted transactions as soon as the node would be synced again regardless it spend more or less epochs to achieve that, instead of a 5 fixed epochs

lrubiorod:
Close #1702 
Close #1763

### DIFF
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -1587,6 +1587,37 @@ type PrioritizedHash = (OrderedFloat<f64>, Hash);
 type PrioritizedVTTransaction = (OrderedFloat<f64>, VTTransaction);
 type PrioritizedDRTransaction = (OrderedFloat<f64>, DRTransaction);
 
+#[derive(Debug, Clone, Default)]
+struct UnconfirmedTransactions {
+    current: Vec<Hash>,
+    previous: Vec<Hash>,
+}
+
+impl UnconfirmedTransactions {
+    fn update(&mut self) {
+        std::mem::swap(&mut self.previous, &mut self.current);
+        self.current.clear();
+    }
+
+    fn insert(&mut self, h: Hash) {
+        self.current.push(h);
+    }
+
+    fn clear(&mut self) {
+        self.current.clear();
+        self.previous.clear();
+    }
+
+    fn remove_all(&mut self) -> Vec<Hash> {
+        let mut v = vec![];
+        v.extend(self.current.iter());
+        v.extend(self.previous.iter());
+        self.clear();
+
+        v
+    }
+}
+
 /// A pool of validated transactions that supports constant access by
 /// [`Hash`](Hash) and iteration over the
 /// transactions sorted from by transactions with bigger fees to
@@ -1625,6 +1656,8 @@ pub struct TransactionsPool {
     // Minimum fee required to include a VTT into a block. We check for this fee in the
     // TransactionPool so we can choose not to insert a transaction we will not mine anyway.
     minimum_vtt_fee: u64,
+    // Map for unconfirmed transactions
+    unconfirmed_transactions: UnconfirmedTransactions,
 }
 
 impl Default for TransactionsPool {
@@ -1648,6 +1681,7 @@ impl Default for TransactionsPool {
             vt_to_dr_factor: 1.0,
             // Default is to include all transactions into the pool and blocks
             minimum_vtt_fee: 0,
+            unconfirmed_transactions: Default::default(),
         }
     }
 }
@@ -1728,6 +1762,7 @@ impl TransactionsPool {
             weight_limit: _,
             vt_to_dr_factor: _,
             minimum_vtt_fee: _,
+            unconfirmed_transactions,
         } = self;
 
         vt_transactions.clear();
@@ -1742,6 +1777,7 @@ impl TransactionsPool {
         output_pointer_map.clear();
         *total_vt_weight = 0;
         *total_dr_weight = 0;
+        unconfirmed_transactions.clear();
     }
 
     /// Returns the number of value transfer transactions in the pool.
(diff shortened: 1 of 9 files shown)


node/src/actors/chain_manager/mod.rs
```diff
@@ -1735,6 +1751,103 @@ impl ChainManager {
...
+                transactions.len(),
+                epoch,
+                transactions
+            );
+
+            ctx.run_later(Duration::from_secs(10), move |act, _ctx| {
+                mempool_transactions.extend(transactions);
+                for transaction in mempool_transactions {
+                    let _x =
+                        act.add_transaction(AddTransaction { transaction }, get_timestamp(), false);
```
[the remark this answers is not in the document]
tmpolaczyk: Also, the call to `get_timestamp()` should be outside the for loop

data_structures/src/chain.rs
```diff
@@ -1587,6 +1587,37 @@ type PrioritizedHash = (OrderedFloat<f64>, Hash);
...
 
+#[derive(Debug, Clone, Default)]
+struct UnconfirmedTransactions {
+    current: Vec<Hash>,
+    previous: Vec<Hash>,
+}
+
+impl UnconfirmedTransactions {
+    fn update(&mut self) {
+        self.previous = self.current.clone();
```
tmpolaczyk, line 1598:
```suggestion
        std::mem::swap(&mut self.previous, &mut self.current);
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -1735,6 +1749,111 @@ impl ChainManager {
...
+            );
+
+            // FIXME(#1763): Do not use a run_later
+            let five_epochs = u64::from(
+                act.consensus_constants()
+                    .checkpoints_period
+                    .saturating_mul(5),
+            );
+            ctx.run_later(Duration::from_secs(five_epochs), move |_act, ctx| {
+                mempool_transactions.extend(transactions);
```
tmpolaczyk, line 1842: Let's move the debug log here:

```suggestion
                mempool_transactions.extend(transactions);
                log::debug!(
                    "Re-adding {} transactions into mempool since epoch {}",
                    mempool_transactions.len(),
                    epoch,
                );
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -206,6 +206,10 @@ pub struct ChainManager {
     temp_superblock_votes: Vec<SuperBlockVote>,
     /// Commits and reveals to process later
     temp_commits_and_reveals: Vec<Transaction>,
+    /// Value transfers and data requests to process later
+    temp_vts_and_drs: Vec<Transaction>,
```
tmpolaczyk, line 210: Since we will be writing transactions to the back of this vector and reading them from the front, like a FIFO queue, the logical choice is to use a `VecDeque`.

```suggestion
    temp_vts_and_drs: VecDeque<Transaction>,
```

The main benefit is that `drain` will not need to move the remaining elements. Example with 4 elements, A, B, C, D, and removing the first 2:

```
Using Vec:
ABCD
.drain(..2)
CD
^ performs a memcpy to move CD to the position of AB

Using VecDeque
ABCD
.drain(..2)
  CD
^ does not copy anything, CD is still at the same position in memory
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -1736,6 +1754,97 @@ impl ChainManager {
...
+            }
+            Err(e) => {
+                log::error!("Error in GetBlocksEpochRange: {}", e);
+                futures::future::err(())
+            }
+        })
+        .and_then(move |block_hashes| {
+            // For each block, collect all the transactions that may be valid if this block is
+            // reverted. This includes value transfer transactions and data request transactions.
+            // And some reveals.
```
tmpolaczyk, line 1787:
```suggestion
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -1736,6 +1754,97 @@ impl ChainManager {
...
+            let aux = block_hashes.into_iter().map(move |hash| {
+                inventory_manager
+                    .send(GetItemBlock { hash })
+                    .then(move |res| match res {
+                        Ok(Ok(block)) => {
+                            let mut transactions = vec![];
+
+                            // TODO: be functional
```
tmpolaczyk, line 1795:
```suggestion
let transactions = block.txns.value_transfer_txns.iter().map(|vtt| Transaction::ValueTransfer(vtt.clone()).chain(block.txns.data_request_txns.iter().map(|drt| Transaction::DataRequest(vtt.clone())).collect();
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -1736,6 +1754,97 @@ impl ChainManager {
...
+    /// superepoch and the transactions included in unconfirmed blocks and it allows them to be
+    /// included again in the mempool when the node wold be Synced again
+    #[must_use]
+    fn reinsert_unconfirmed_transactions(
+        &mut self,
+        epoch: Epoch,
+    ) -> ResponseActFuture<Self, (), ()> {
+        let inventory_manager = InventoryManager::from_registry();
+
+        let mut mempool_transactions = self.transactions_pool.remove_unconfirmed_transactions();
```
tmpolaczyk, line 1768: We need to double check that it's impossible to add a transaction to the `TransactionsPool` between this call, which removes the unconfirmed transactions, and the call to `initialize_from_storage()` which restores the previous `ChainState`. Otherwise it could be possible to insert a transaction which is not valid after the revert.

tmpolaczyk (approved):

PR #1757, opened 2020-12-01
refactor: optimize get_rep_ordered_ars_list

tmpolaczyk:


data_structures/src/chain.rs
```diff
@@ -2962,42 +2994,50 @@ impl ReputationEngine {
...
+                pkh_bytes.extend(alpha_bytes);
 
-            a_bytes.extend(alpha_bytes);
-            b_bytes.extend(alpha_bytes);
+                calculate_sha256(&pkh_bytes).into()
+            };
 
-            let new_hash_a: Hash = calculate_sha256(&a_bytes).into();
-            let new_hash_b: Hash = calculate_sha256(&b_bytes).into();
+            let new_hash_a = cached_hash_a.get().unwrap_or_else(|| {
```
girazoki, line 3025: Maybe explain that hashes are only calculated once and cached

aesedepece:
May I understand that for this to be promoted from draft to proper PR, some `unwrap` needs to be removed? Or do we intend to add some more functionality / refactor something else?

tmpolaczyk:
Ah sorry, yes this is almost done. The only problem is that I use a few large JSON files for the benchmarks and I didn't want to push them to the repo. So the plan is to remove that benchmarks and merge. Meaning that this PR can already be reviewed.

aesedepece (approved):

lrubiorod (approved):

PR #1759, opened 2020-12-02
fix(wallet): includes future::lazy in a block_on call

lrubiorod:


wallet/src/actors/worker/methods.rs
```diff
@@ -531,7 +531,8 @@ impl Worker {
             .map(|output| self.query_transaction(output.transaction_id.to_string()));
         let retrieve_responses = async { futures03::future::try_join_all(txn_futures).await };
         let transactions: Vec<types::Transaction> =
-            futures03::executor::block_on(retrieve_responses)?;
+            futures::future::lazy(|| futures03::executor::block_on(retrieve_responses)).wait()?;
```
mariocao, line 534: What about the other `block_on` calls?

lrubiorod: it was for a specific error found. Those other calls, I am not sure if it is needed

tmpolaczyk (approved):

PR #1771, opened 2020-12-09
fix(node): fix updating node stats during synchronization

lrubiorod:
Before to stop it:
image

After clean storage and restarting again:
image

### DIFF
--- a/node/src/actors/chain_manager/mining.rs
+++ b/node/src/actors/chain_manager/mining.rs
@@ -227,6 +227,7 @@ impl ChainManager {
                     // This will run all the validations again
 
                     let block_hash = block.hash();
+                    // FIXME(#1773): Currently last_block_proposed is not used, but removing it is a breaking change
                     act.chain_state.node_stats.last_block_proposed = block_hash;
                     act.chain_state.node_stats.block_proposed_count += 1;
                     log::info!(
--- a/node/src/actors/chain_manager/mod.rs
+++ b/node/src/actors/chain_manager/mod.rs
@@ -549,6 +549,15 @@ impl ChainManager {
                 return;
             }
         };
+
+        let own_pkh = match self.own_pkh {
+            Some(x) => x,
+            None => {
+                log::error!("No OwnPkh loaded in ChainManager");
+                return;
+            }
+        };
+
         match self.chain_state {
             ChainState {
                 chain_info: Some(ref mut chain_info),
@@ -608,10 +617,11 @@ impl ChainManager {
                     &mut self.chain_state.data_request_pool,
                     &mut self.transactions_pool,
                     utxo_diff,
-                    self.own_pkh,
+                    own_pkh,
                     &mut self.chain_state.own_utxos,
                     epoch_constants,
                     &mut self.chain_state.node_stats,
+                    self.sm_state,
                 );
 
                 let miner_pkh = block.block_header.proof.proof.pkh();
@@ -672,11 +682,6 @@ impl ChainManager {
                             .data_request_pool
                             .update_data_request_stages();
 
-                        if block_hash == self.chain_state.node_stats.last_block_proposed {
-                            self.chain_state.node_stats.block_mined_count += 1;
-                            log::info!("Congratulations! Your block was consolidated into the block chain by an apparent majority of peers");
-                        }
-
                         show_info_dr(&self.chain_state.data_request_pool, &block);
 
                         for reveal in reveals {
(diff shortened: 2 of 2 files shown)


node/src/actors/chain_manager/mining.rs
```diff
@@ -227,6 +227,7 @@ impl ChainManager {
                     // This will run all the validations again
 
                     let block_hash = block.hash();
+                    // TODO. Currently last_block_proposed is not used, but removing it is a breaking change
```
mariocao, line 230: Maybe change the TODO as FIXME to track it and change it when there is a new release with breaking changes. :)

node/src/actors/chain_manager/mod.rs
```diff
@@ -703,6 +708,17 @@ impl ChainManager {
...
                     }
                 }
+
+                if miner_pkh == own_pkh {
+                    self.chain_state.node_stats.block_mined_count += 1;
+                    if self.sm_state == StateMachine::Synced {
+                        log::info!("Congratulations! Your block was consolidated into the block chain by an apparent majority of peers");
+                    } else {
+                        // During synchronization, we assume that every consolidated block has,
+                        // at least, one proposed block.
```
mariocao, line 718:
```suggestion
                        // During synchronization, we assume that every consolidated block has, at least, one proposed block.
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -2140,21 +2160,19 @@ fn update_pools(
     transactions_pool.clear_reveals();
 
     // Update own_utxos:
```
mariocao, line 2162:
```suggestion
    // Update own_utxos
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -549,6 +549,15 @@ impl ChainManager {
                 return;
             }
         };
+
+        let own_pkh = match self.own_pkh {
+            Some(x) => x,
+            None => {
+                log::error!("No OwnPkh loaded in ChainManager");
```
mariocao, line 556: This is the kind of error that we should never see, isn't it? :P What happens if it is None?

node/src/actors/chain_manager/mod.rs
```diff
@@ -703,6 +708,17 @@ impl ChainManager {
...
                 }
+
+                if miner_pkh == own_pkh {
+                    self.chain_state.node_stats.block_mined_count += 1;
+                    if self.sm_state == StateMachine::Synced {
+                        log::info!("Congratulations! Your block was consolidated into the block chain by an apparent majority of peers");
+                    } else {
+                        // During synchronization, we assume that every consolidated block has,
+                        // at least, one proposed block.
+                        self.chain_state.node_stats.block_proposed_count += 1;
```
tmpolaczyk, line 719: I thought this is the count of blocks proposed by this node? So it should be assumed to be 0 unless the miner pkh is the same as the node pkh?

node/src/actors/chain_manager/mod.rs
lrubiorod: If it is None, something wrong happened. We have different ways to handle those "non optional" option fields (match, if let, unwrap...) So, we should unify in a refactor PR, but by the moment...

node/src/actors/chain_manager/mod.rs
lrubiorod: block_proposed_count is a counter for all the proposed blocks by the node, in order that you can check how many times you create and broadcast a valid block and how many times it was included it in the blockchain, the block_mined_count

lrubiorod: In this case, if you detect a mined block by you during synchronization, you assume that you proposed a candidate in the past

tmpolaczyk: Ok, so I would assume that the number of proposed blocks is 0, because the probability of proposing a block is very low. Right?

tmpolaczyk: Ah, nevermind, I missed the `if miner_pkh == own_pkh`

node/src/actors/chain_manager/mining.rs
```diff
@@ -227,6 +227,7 @@ impl ChainManager {
                     // This will run all the validations again
 
                     let block_hash = block.hash();
+                    // FIXME(#1773). Currently last_block_proposed is not used, but removing it is a breaking change
```
tmpolaczyk, line 230:
```suggestion
                    // FIXME(#1773): Currently last_block_proposed is not used, but removing it is a breaking change
```

tmpolaczyk (approved):

mariocao (approved):

PR #1772, opened 2020-12-09
feat(error): return timeout event if timeout error

### LINKED ISSUE #1750: Timeouts should potentially have a different error than SyncError
This issue comprises 2 sub-tasks:
- [x] Implement a SyncTimeout instead of a SyncError
- [x] Investigate why those timeouts are happening

girazoki:
This PR solves the issue about getting a timeout while receiving blocks, as this might trigger an inconsistent state in the wallet. The solution provided sets the node state to 'None' when we receive a timeout, so as to trigger a new sync when the periodic ping pong request detects is online again

Closes #1750

### DIFF
--- a/wallet/src/actors/app/methods.rs
+++ b/wallet/src/actors/app/methods.rs
@@ -356,7 +356,7 @@ impl App {
             .flatten()
             .map_err(From::from)
             .into_actor(self)
-            .and_then(move |res, slf: &mut Self, _ctx| {
+            .and_then(move |res, slf: &mut Self, ctx| {
                 let types::UnlockedSessionWallet {
                     wallet,
                     session_id,
@@ -371,11 +371,19 @@ impl App {
                     || slf.state.node_state == None
                 {
                     let sink = slf.state.get_sink(&session_id);
-                    slf.params.worker.do_send(worker::SyncRequest {
-                        wallet_id,
-                        wallet,
-                        sink,
-                    });
+                    slf.params
+                        .worker
+                        .send(worker::SyncRequest {
+                            wallet_id,
+                            wallet,
+                            sink,
+                        })
+                        .flatten()
+                        .into_actor(slf)
+                        .map_err(|e, act: &mut Self, _ctx| {
+                            act.handle_sync_error(&(e.into()));
+                        })
+                        .spawn(ctx)
                 };
 
                 fut::ok(types::UnlockedWallet { data, session_id })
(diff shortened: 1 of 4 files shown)


wallet/src/actors/app/methods.rs
```diff
@@ -371,11 +371,19 @@ impl App {
...
+                            wallet_id,
+                            wallet,
+                            sink,
+                        })
+                        .flatten()
+                        .into_actor(slf)
+                        .map_err(|e, act: &mut Self, _ctx| {
+                            act.handle_sync_error(&(e.into()));
+                        })
+                        .wait(ctx)
```
tmpolaczyk, line 386: I think this `wait` will make the `App` actor to stop processing any messages until the synchronization is done (or there is a timeout). So maybe this breaks the ability to generate addresses while the wallet is syncing. Not sure, I will have to try.

girazoki: interesting...I can also trry this

tmpolaczyk: It looks like this will result in a deadlock if you try to unlock a different wallet while the first wallet is syncing... So let's try to change the `wait` into a `spawn`.

tmpolaczyk: Using `spawn` seems to work as expected.

girazoki: good

wallet/src/actors/worker/methods.rs
```diff
@@ -677,8 +677,14 @@ impl Worker {
                 sync_end,
                 e
             );
-
-            let events = Some(vec![types::Event::SyncError(sync_start, sync_end)]);
+            let events = {
+                if let Error::JsonRpcTimeoutError = e {
+                    log::error!("Timeout error");
```
mariocao, line 682: Maybe more details in the log message (during synchronization)

wallet/src/actors/worker/error.rs
```diff
@@ -26,6 +26,8 @@ pub enum Error {
     Send(#[cause] futures::sync::mpsc::SendError<std::string::String>),
     #[fail(display = "node error: {}", _0)]
     Node(#[cause] failure::Error),
+    #[fail(display = "timeout error")]
```
mariocao, line 29: Maybe something as...
```suggestion
    #[fail(display = "JsonRPC timeout error")]
```

mariocao (approved):

tmpolaczyk (approved):

PR #1784, opened 2020-12-14
Rad replace separators

aesedepece:
Supersedes #1610

### DIFF
--- a/rad/src/operators/array.rs
+++ b/rad/src/operators/array.rs
@@ -80,30 +80,30 @@ pub fn get_bytes(input: &RadonArray, args: &[Value]) -> Result<RadonBytes, RadEr
     let item = get(input, args)?;
     item.try_into()
 }
-pub fn get_integer(input: &RadonArray, args: &[Value]) -> Result<RadonInteger, RadError> {
-    let item = get(input, args)?;
-    let value = if args.len() == 3 {
-        replace_separators(item, args[1].clone(), args[2].clone())
-    } else {
-        item
-    };
 
-    value.try_into()
-}
 pub fn get_float(input: &RadonArray, args: &[Value]) -> Result<RadonFloat, RadError> {
-    let item = get(input, args)?;
-    let value = if args.len() == 3 {
-        replace_separators(item, args[1].clone(), args[2].clone())
-    } else {
-        item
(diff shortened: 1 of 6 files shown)


tmpolaczyk:
I think this only applies to RadonMaps, so we need to repeat the same code in RadonArray for the operators ArrayGetInteger and ArrayGetFloat, and also in RadonString for operators StringAsFloat and StringAsInteger.

aesedepece:
> I think this only applies to RadonMaps, so we need to repeat the same code in RadonArray for the operators ArrayGetInteger and ArrayGetFloat, and also in RadonString for operators StringAsFloat and StringAsInteger.

True

aesedepece:
This grew a bit. Sorry not sorry :stuck_out_tongue_winking_eye:

aesedepece:
Due to the latest changes, this should no longer be a breaking change.

tmpolaczyk:
To make this change even less breaking, would it be possible to active it after a certain block number? Otherwise I think that if someone uses one of the new operators in the tally stage, it can cause old nodes to reject that block and create a hard fork.

aesedepece:
> To make this change even less breaking, would it be possible to active it after a certain block number? Otherwise I think that if someone uses one of the new operators in the tally stage, it can cause old nodes to reject that block and create a hard fork.

Sounds reasonably paranoid :ok_hand:

aesedepece:
I will rebase this on top of #1789 after hard fork date is decided (#1791).

aesedepece:
@tmpolaczyk on second thought, getting knowledge of current epoch at the operator level inside the radon executor would require overly disruptive changes to the code that would add complexity, reduce clarity and serve no other purpose than specifically this.

tmpolaczyk:
Yes, it's hard to implement because this is the first change to RADON in mainnet. I think that eventually we will need to add that logic anyway, as we keep improving RADON. But if it's too much overhead for now, then I prefer to delay this PR until we have some major RADON 2.0 update, to avoid delaying the release because of this.

aesedepece:
Let's do that.

aesedepece:
This should be safe to adopt now that we have TAPI. I'll close this PR and open a new issue.

Issue #1780, opened 2020-12-14
Review timelock field in DataRequest 


girazoki:
This issue should make sure that the timelock field is not being used by the protocol

tmpolaczyk:
Duplicate of #1774

PR #1785, opened 2020-12-14
Remove data requests by epoch mapping

tmpolaczyk:
See issue #1780

### DIFF
--- a/node/src/actors/chain_manager/mining.rs
+++ b/node/src/actors/chain_manager/mining.rs
@@ -263,12 +263,6 @@ impl ChainManager {
         let data_request_timeout = self.data_request_timeout;
         let timestamp = u64::try_from(get_timestamp()).unwrap();
 
-        // Data Request mining
-        let dr_pointers = self
-            .chain_state
-            .data_request_pool
-            .get_dr_output_pointers_by_epoch(current_epoch);
-
         let rep_eng = self.chain_state.reputation_engine.as_ref().unwrap();
         let is_ars_member = rep_eng.is_ars_member(&own_pkh);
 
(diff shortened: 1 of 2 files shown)


node/src/actors/chain_manager/mining.rs
```diff
@@ -277,7 +271,10 @@ impl ChainManager {
         let total_active_reputation = rep_eng.total_active_reputation();
         let num_active_identities =
             u32::try_from(rep_eng.ars().active_identities_number()).unwrap();
-        log::debug!("{} data requests for this epoch", dr_pointers.len());
+        log::debug!(
+            "{} data requests for this epoch",
```
girazoki, line 275:
```suggestion
            "{} active data requests",
```

node/src/actors/chain_manager/mining.rs
```diff
@@ -291,13 +288,14 @@ impl ChainManager {
...
-                .map(|dr_state| (dr_pointer, dr_state.clone()))
-        }) {
+        for (dr_pointer, dr_state) in self
+            .chain_state
+            .data_request_pool
+            .data_request_pool
+            .iter()
+            // Clone the dr_state because it must be valid for a static lifetime
+            .map(|(dr_pointer, dr_state)| (*dr_pointer, dr_state.clone()))
+        {
```
lrubiorod, line 298: The problem here if that requests that are in Reveal and Tally stage appear here as available but we only want requests in Commit stage

tmpolaczyk: Good catch, so we still need to keep track of all the data requests in commit stage. So instead of a `BTreeMap<Epoch, HashSet<Hash>>` we can use a `HashSet<Hash>`, or we can use `filter_map` to iterate over all the requests and check the stage.

tmpolaczyk:
The `epoch` field of `DataRequestState` could also be removed, but I think it would be better to do it after the hard fork to avoid accidentally breaking anything. So I'm closing this PR for now.

PR #1786, opened 2020-12-14
fix(node): panic on deserialization error instead of overwriting chain state

tmpolaczyk:
This error can happen while developing: we can accidentally break compatibility of some data structure. Prior to this PR, the chain state was replaced with a new chain state and there was no way to recover the old chain state. Now, the node panics so we can detect this error.

This was not a panic before because we wanted to make the transition from one testnet to the next as smooth as possible.


aesedepece:
I dig this.

aesedepece (approved):

lrubiorod (approved):

PR #1755, opened 2020-11-30
The Second Hack

lrubiorod:

PR #1788, opened 2020-12-15
The Second Hack

girazoki:
Supersedes #1755


lrubiorod:
This PR was an emergency PR in case of something was wrong during Christmas

PR #1789, opened 2020-12-15
feat(validations): after first hard fork

girazoki:


validations/src/mainnet_validations.rs
```diff
@@ -30,6 +31,10 @@ pub fn in_emergency_period(
     }
 }
 
+pub fn after_first_hard_fork(epoch: Epoch) -> bool {
+    epoch > FIRST_HARD_FORK
```
tmpolaczyk, line 35: What's "after"?

```suggestion
    epoch >= FIRST_HARD_FORK
```

validations/src/mainnet_validations.rs
```diff
@@ -12,6 +12,7 @@ const FIRST_EMERGENCY_COMMITTEE: [&str; 7] = [
     "wit1etherz02v4fvqty6jhdawefd0pl33qtevy7s4z",
 ];
 
+const FIRST_HARD_FORK: u32 = "INSERT EPOCH HERE";
```
tmpolaczyk, line 15:
```suggestion
// FIXME(#issuenumber): decide first hard fork epoch
const FIRST_HARD_FORK: u32 = u32::MAX;
```

validations/src/mainnet_validations.rs
tmpolaczyk: It was a question, I don't know if `>=` is more correct than `>`

validations/src/mainnet_validations.rs
lrubiorod: Now, clippy doesn't like "absurd comparisons" :P

tmpolaczyk: Good, so we know that it's impossible for `after_first_hard_fork` to return true

tmpolaczyk: Joking aside, we can just add `#[allow(clippy::absurd_extreme_comparisons)]`

girazoki: Guys, its OK if clippy does not pass, as this branch is not gonna be merged as it is. Once we have an epoch we just type it there and clippy passes again

validations/src/mainnet_validations.rs
girazoki: I am totally against triggering a semantic discussion on this haha

validations/src/mainnet_validations.rs
tmpolaczyk: I thought we wanted to merge this branch and start implementing the breaking changes, and decide on the exact date when we are confident that the implementation is good

girazoki: If it is more helpful for testing purposes I can add it, never thought such a small PR was gonna generate that much discussion haaha

lrubiorod (approved):

validations/src/mainnet_validations.rs
```diff
@@ -12,6 +12,8 @@ const FIRST_EMERGENCY_COMMITTEE: [&str; 7] = [
     "wit1etherz02v4fvqty6jhdawefd0pl33qtevy7s4z",
 ];
 
+// FIXME(#17): decide first hard fork epoch
```
lrubiorod, line 15:
```suggestion
// FIXME(#1791): decide first hard fork epoch
```

PR #1790, opened 2020-12-15
feat(validation): add dr open weight to block validation

### LINKED ISSUE #1777: Enforce data request limit validation (can't start a request lifecycle while too many requests are in force)

lrubiorod:
Close #1777

### DIFF
--- a/validations/src/tests.rs
+++ b/validations/src/tests.rs
@@ -17,6 +17,7 @@ use witnet_data_structures::{
         calculate_tally_change, calculate_witness_reward, create_tally, DataRequestPool,
     },
     error::{BlockError, DataRequestError, Secp256k1ConversionError, TransactionError},
+    mainnet_validations::FIRST_HARD_FORK,
     radon_error::RadonError,
     radon_report::{RadonReport, ReportContext, TypeLike},
     transaction::*,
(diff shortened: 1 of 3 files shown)


tmpolaczyk:
Add tests

validations/src/tests.rs
```diff
@@ -5434,13 +5435,14 @@ fn test_block_with_drpool<F: FnMut(&mut Block) -> bool>(
     mut_block: F,
     dr_pool: DataRequestPool,
 ) -> Result<(), failure::Error> {
-    test_block_with_drpool_and_utxo_set(mut_block, dr_pool, UnspentOutputsPool::default())
+    test_block_with_drpool_and_utxo_set(mut_block, dr_pool, UnspentOutputsPool::default(), 1000)
```
tmpolaczyk, line 5438: The default should probably be after the FIRST_HARD_FORK, this way we ensure that all the existing tests still pass?

validations/src/tests.rs
```diff
@@ -6010,6 +6011,7 @@ fn block_duplicated_commits() {
         },
         dr_pool,
         utxo_set,
+        1000,
```
tmpolaczyk, line 6014: Same here, this is epoch 1000 which is before FIRST_HARD_FORK

validations/src/tests.rs
tmpolaczyk: But then we will have to change it again to SECOND_HARD_FORK in the future...

tmpolaczyk (approved):

PR #1794, opened 2020-12-16
chore: improve compile time of tests

tmpolaczyk:
And prevent some tests from being run twice.

See here: https://doc.rust-lang.org/cargo/guide/project-layout.html

> If a binary, example, bench, or integration test consists of multiple source files, place a main.rs file along with the extra modules within a subdirectory of the src/bin, examples, benches, or tests directory. The name of the executable will be the directory name.

So if we just create a `lib.rs` file like the one that was in `p2p/tests/lib.rs`, all the tests in `p2p/tests/*.rs` will be compiled separately, and since the `lib.rs` test includes all the others, they will be run twice.

The solution is to move the tests to `p2p/tests/integration/` and move the old `lib.rs` to `p2p/tests/integration/main.rs`. Unfortunately cargo does not show any warnings if you add a test file but forget to add it to `main.rs`.


tmpolaczyk:
According to github actions this PR has made it slower, so further investigation is needed.

aesedepece:
> According to github actions this PR has made it slower, so further investigation is needed.

Then I'm tagging this as a draft, with your permission.

tmpolaczyk:
https://github.com/dtolnay/automod

This crate can solve the problem of manually adding tests to the `mod.rs` file

tmpolaczyk:
Closing this because the improvement is not worth it, opened #1918 to remove the duplicate tests

PR #1795, opened 2020-12-16
fix(validations): include epoch in test_block

lrubiorod:


tmpolaczyk (approved):
The `test_block` function needs a refactor someday

PR #1797, opened 2020-12-16
fix(wallet): small bug in a test

clbartoli:
This PR adds a small fix in a test that was being ignored in the wallet file model.rs


girazoki (approved):

tmpolaczyk:
As far as I know there is no simple way to prevent this kind of mistakes...

tmpolaczyk (approved):

lrubiorod (approved):

PR #1801, opened 2020-12-18
feat(node): remove outbounds in case of a wrong batch

lrubiorod:


tmpolaczyk:
Instead of dropping all outbounds, I think it would be better to just drop the one peer that sent a wrong batch. We could also "ice" it.

lrubiorod:
> Instead of dropping all outbounds, I think it would be better to just drop the one peer that sent a wrong batch. We could also "ice" it.

The problem here is that if that outbound peer has started to send you blocks, it has to be in consensus with the majority of the others, so probably, the other peers would send you the same batch. And if we remove one by one, you take the risk of the good peers would be removed because there are not in consensus with the rest of your outbounds

node/src/actors/chain_manager/handlers.rs
```diff
@@ -458,6 +473,13 @@ impl Handler<AddBlocks> for ChainManager {
                                 // Process remaining blocks
                                 let (batch_succeeded, num_processed_blocks) = act.process_blocks_batch(ctx, &sync_target, &remainig_blocks);
                                 if !batch_succeeded {
+                                    // We receive an invalid batch of blocks. So we need to throw away our
+                                    // outbound peers in order to find new ones that can give us the blocks
+                                    // consolidated by the network
+                                    act.drop_all_outbounds();
+                            // Ice the invalid blocks' batch sender
+                            act.ice_peer(sender);
```
tmpolaczyk, line 481: Fmt

node/src/actors/chain_manager/mod.rs
```diff
@@ -1841,6 +1842,30 @@ impl ChainManager {
...
+
+    /// This function send a message to SessionsManager to drop all outbounds peers
+    pub fn drop_all_outbounds(&self) {
+        let peers_to_unregister = self
+            .last_received_beacons
+            .iter()
+            .map(|(addr, _)| *addr)
+            .collect();
+        let sessions_manager_addr = SessionsManager::from_registry();
+        sessions_manager_addr.do_send(DropOutboundPeers {
```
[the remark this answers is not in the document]
tmpolaczyk: Or do we prefer to drop all the peers that were in consensus with the peer that sent us the bad block?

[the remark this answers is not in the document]
lrubiorod: Before to start synchronization, nodes should be in consensus, and the nodes are not would be removed. So they would be equals. When the node is synced, it is allowed to have nodes that are not in consensus but they have consensus at superblock, but in the moment of synchronization it is not

tmpolaczyk (approved):

node/src/actors/chain_manager/handlers.rs
```diff
@@ -403,6 +404,13 @@ impl Handler<AddBlocks> for ChainManager {
                         let (batch_succeeded, num_processed_blocks) =
                             self.process_first_batch(ctx, &sync_target, &blocks);
                         if !batch_succeeded {
+                            // We receive an invalid batch of blocks. So we need to throw away our
+                            // outbound peers in order to find new ones that can give us the blocks
+                            // consolidated by the network
+                            self.drop_all_outbounds();
+                            // Ice the invalid blocks' batch sender
+                            self.ice_peer(sender);
+
```
aesedepece, line 413: Do you think it could be worth refactoring this code fragment into a `drop_all_outbounds_and_ice_batch_sender` function or method that operates on `self` / `act` and `sender` so as to avoid code repetition?

tmpolaczyk: The whole `AddBlocks` handler needs a refactor...

lrubiorod: the idea was to have some functions that could be used in another parts of the code, so that's the reason to do not a so specific function

node/src/actors/chain_manager/mod.rs
```diff
@@ -1841,6 +1842,40 @@ impl ChainManager {
...
+            let peers_manager_addr = PeersManager::from_registry();
+            peers_manager_addr.do_send(RemoveAddressesFromTried {
+                addresses: vec![addr],
+                ice: true,
+            });
+        }
+    }
+
+    /// This function wraps the drop_all_outbound and ice_peer.
+    /// It is called when we receive an invalid batch of blocks. So we need to throw away our
```
aesedepece, line 1871:
```suggestion
    /// This is called when we receive an invalid batch of blocks. It will throw away our
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -1841,6 +1842,40 @@ impl ChainManager {
...
+    /// outbound peers in order to find new ones that can give us the blocks consolidated by the
+    /// network, and ice the node that sent the invalid batch
```
aesedepece, line 1873:
```suggestion
    /// network, and ice the node that sent the invalid batch.
```

aesedepece (changes requested):

node/src/actors/chain_manager/mod.rs
```diff
@@ -1841,6 +1842,40 @@ impl ChainManager {
 
         Box::new(fut)
     }
+
+    /// This function send a message to SessionsManager to drop all outbounds peers
```
aesedepece, line 1846:
```suggestion
    /// Send a message to `SessionsManager` to drop all outbound peers.
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -1841,6 +1842,40 @@ impl ChainManager {
...
+            .iter()
+            .map(|(addr, _)| *addr)
+            .collect();
+        let sessions_manager_addr = SessionsManager::from_registry();
+        sessions_manager_addr.do_send(DropOutboundPeers {
+            peers_to_drop: peers_to_unregister,
+        });
+    }
+
+    /// This function send a message to PeersManager to ice a specific peer
```
aesedepece, line 1859:
```suggestion
    /// Send a message to `PeersManager` to ice a specific peer.
```

aesedepece (changes requested):

node/src/actors/chain_manager/mod.rs
```diff
@@ -1841,6 +1842,40 @@ impl ChainManager {
...
+        if let Some(addr) = addr {
+            let peers_manager_addr = PeersManager::from_registry();
+            peers_manager_addr.do_send(RemoveAddressesFromTried {
+                addresses: vec![addr],
+                ice: true,
+            });
+        }
+    }
+
+    /// This function wraps the drop_all_outbound and ice_peer.
```
aesedepece, line 1870:
```suggestion
    /// Execute `drop_all_outbounds` and `ice_peer` at once.
    /// 
```

aesedepece (changes requested):

tmpolaczyk (approved):

node/src/actors/chain_manager/mod.rs
```diff
@@ -1841,6 +1842,40 @@ impl ChainManager {
...
+            let peers_manager_addr = PeersManager::from_registry();
+            peers_manager_addr.do_send(RemoveAddressesFromTried {
+                addresses: vec![addr],
+                ice: true,
+            });
+        }
+    }
+
+    /// Execute `drop_all_outbounds` and `ice_peer` at once.
+    /// This is called when we receive an invalid batch of blocks. It will throw away our outbound
```
aesedepece, line 1871:
```suggestion
    /// This is called when we receive an invalid batch of blocks. It will throw away our outbound
    /// 
```

aesedepece (approved):

PR #1806, opened 2020-12-23
Removing pub uses from types in Wallet

lrubiorod:


wallet/src/types.rs
```diff
@@ -1,49 +1,37 @@
...
 };
 
-pub use witnet_net::client::tcp::jsonrpc::Request as RpcRequest;
 use witnet_protected::{Protected, ProtectedString};
-pub use witnet_rad::{error::RadError, types::RadonTypes, RADRequestExecutionReport};
-
-use crate::{model, types::signature::Signature};
 
 use super::{db, repository};
+use witnet_data_structures::chain::Signature;
```
tmpolaczyk, line 34: fmt

tmpolaczyk (approved):

wallet/src/types.rs
lrubiorod: O_O

PR #1807, opened 2020-12-23
[WIP] Revalidate block chain

tmpolaczyk:
Add an easy way to test that the old blocks are still valid when the validation logic changes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4801,6 +4801,7 @@ dependencies = [
  "failure",
  "itertools",
  "log 0.4.11",
+ "witnet_config",
  "witnet_crypto",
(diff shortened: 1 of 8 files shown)


aesedepece:
This will be specially useful when assessing whether a network upgrade is a soft-fork or a hard-fork. Thanks @tmpolaczyk for this Christmas present! :gift:

tmpolaczyk:
The rewind command added in PR #1865 could be used to implement a similar feature. The validations are skipped when rewinding, but they could be enabled with an optional flag or something.

So I'm closing this PR, but it would be nice to refactor the consolidation logic out of ChainManager sometime.

PR #1808, opened 2020-12-24
Allow to resync node using blocks from storage

tmpolaczyk:
This PR has two parts:

* Do not request blocks that are already in storage when synchronizing

Check if the node already has the blocks in the database and only request the missing ones. This alone is not very useful because the database only contains consolidated blocks, but it can be useful if we implement another way to add blocks to the database.

* Create deleteChainState JSON-RPC method

This new JSON-RPC method allows to reset the `ChainState` of the node, triggering a full resynchronization. This can be used to recover nodes that have a corrupted `ChainState`, or are missing some blocks or transactions. Note that all the blocks will be validated again so this process will take many hours, being only slightly faster than a normal re-sync.

This feature can also be useful to recover nodes that have consolidated a forked chain and cannot sync automatically.

Performance-wise this is not a good trade-off because it introduces a redundant read from the database that will never succeed for nodes that have never used deleteChainState.

### DIFF
--- a/node/src/actors/chain_manager/mod.rs
+++ b/node/src/actors/chain_manager/mod.rs
@@ -314,6 +314,7 @@ impl ChainManager {
         .into_actor(self)
         .and_then(|(), act, ctx| {
             log::info!("Successfully persisted empty chain state into storage");
+            act.update_state_machine(StateMachine::WaitingConsensus);
             act.initialize_from_storage(ctx);
             fut::ok(())
         })
--- a/node/src/actors/chain_manager/mod.rs
+++ b/node/src/actors/chain_manager/mod.rs
@@ -305,6 +305,25 @@ impl ChainManager {
             .wait(ctx);
     }
 
+    fn delete_chain_state_and_reinitialize(&mut self, ctx: &mut Context<Self>) {
+        let empty_state = ChainState::default();
+        storage_mngr::put(
+            &storage_keys::chain_state_key(self.get_magic()),
+            &empty_state,
+        )
+        .into_actor(self)
+        .and_then(|(), act, ctx| {
+            log::info!("Successfully persisted empty chain state into storage");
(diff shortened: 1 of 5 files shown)


aesedepece:
I always longed for this feature, under the assumption that checking the block for existence in the storage and reading it from there was faster than querying a peer for it; just because the peer needs in turn to perform the same operation of reading it from disk.

I didn't bother though to verify this assumption or benchmark how significant of an improvement could that be. As you mentioned, in the context of a "full resync", I do agree that the time difference would likely be quite mild.

node/src/actors/chain_manager/mod.rs
```diff
@@ -305,6 +305,25 @@ impl ChainManager {
...
+    fn delete_chain_state_and_reinitialize(&mut self, ctx: &mut Context<Self>) {
+        let empty_state = ChainState::default();
+        storage_mngr::put(
+            &storage_keys::chain_state_key(self.get_magic()),
+            &empty_state,
+        )
+        .into_actor(self)
+        .and_then(|(), act, ctx| {
+            log::info!("Successfully persisted empty chain state into storage");
+            act.initialize_from_storage(ctx);
```
tmpolaczyk, line 317: This should also set the node to waiting consensus state:

```suggestion
            act.initialize_from_storage(ctx);
            act.sm_state = StateMachine::WaitingConsensus;
```

aesedepece:
This is already on master, right? @tmpolaczyk

tmpolaczyk:
Not exactly. We have the functionality to rewind the chain back to a previous epoch without asking the blocks again, but that's separate from the p2p protocol. For example, in the case where a node crashes and restarts, it may ask for blocks that are already in the storage.

But the `deleteChainState` command is no longer needed, so we can close this draft and open a new PR if we want to start optimizing the p2p protocol.

PR #1810, opened 2021-01-05
Update wallet to actix 0.11

tmpolaczyk:

PR #1811, opened 2021-01-07
Update node to actix 0.11

tmpolaczyk:
Followup to #1810

### DIFF
--- a/node/src/actors/connections_manager/resolver.rs
+++ b/node/src/actors/connections_manager/resolver.rs
@@ -1,0 +1,412 @@
+#![allow(missing_docs)]
+//! This module used to belong to the actix repository:
+//! https://github.com/actix/actix/blob/fdaa5d50e25ffc892f5c1c6fcc51097796debecf/src/actors/resolver.rs
+//! But it has been deprecated because "actix as an actor framework should not be bind to a dns dependency.":
+//! https://github.com/actix/actix/issues/451
+//!
+//! DNS resolver and connector utility actor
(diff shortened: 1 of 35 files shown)


node/src/actors/connections_manager/resolver.rs
```diff
@@ -0,0 +1,412 @@
...
+//!
+//! DNS resolver and connector utility actor
+use std::collections::VecDeque;
+use std::fmt;
+use std::future::Future;
+use std::io;
+use std::net::SocketAddr;
+use std::pin::Pin;
+use std::task::{self, Poll};
+use std::time::Duration;
```
lrubiorod, line 15: fmt :P

node/src/storage_mngr.rs
```diff
@@ -70,23 +94,24 @@ where
...
+            Ok(kv_bytes) if kv_bytes.is_empty() => Ok(()),
+            Ok(kv_bytes) => addr.send(PutBatch(kv_bytes)).await?,
+            Err(e) => Err(e),
+        }
     }
 }
 
 /// Delete value associated to key
-pub fn delete<K>(key: &K) -> impl Future<Item = (), Error = failure::Error>
+pub async fn delete<K>(key: &K) -> Result<(), failure::Error>
```
tmpolaczyk, line 107: This function is not used, but it needs to be rewritten to be similar to `get`. Because using `async fn` will result in lifetime errors if the output future lives longer than the input key.

lrubiorod (approved):

PR #1812, opened 2021-01-08
feat(rad): upgrade surf to 2.1

tmpolaczyk:


lrubiorod:
image

It looks like that there is a problem with bitstamp json

tmpolaczyk:
> image
> 
> It looks like that there is a problem with bitstamp json

Can reproduce using the default `examples/bitcoin_price.json` and a local testnet with 1 node.

tmpolaczyk:
Closing for now, will re-open when surf releases a version that works

PR #1817, opened 2021-01-11
fix(wallet): fix AddressInfo received_amount and received_payments 

### LINKED ISSUE #1815: Incorrect payment counter for address
@drcpu-github commented on [Fri Dec 25 2020](https://github.com/witnet/sheikah/issues/1587)

**Describe the bug**
When sending multiple VTT's to one address in the Sheikah wallet (you can put me on Santa's naughty list for doing that), the payments counter is incorrect. I sent 10 VTT's to an address, but the counter only shows that address received 8 payments. I can see all transactions in the transaction view, so they are correctly indexed / discovered, but not counted in the address pane's payment counter. I think this happens when more than one VTT is included in the same block. In my example the VTT's were spread over 8 different blocks and 2 blocks each contained 2 VTT's. 

**To Reproduce**
Send multiple VTT's to one address. They should be included in the same block. The payment counter in the address pane of the Sheikah wallet only counts one per block.

**Expected behavior**
Preferably the correct amount of payments should be shown. 😄 

**Desktop (please complete the following information):**
 - OS: Windows
 - Version Sheikah v1.7.2

tmpolaczyk:
Fix #1815

### DIFF
--- a/wallet/src/repository/wallet/mod.rs
+++ b/wallet/src/repository/wallet/mod.rs
@@ -376,7 +376,6 @@ where
         }
         .to_string();
         let info = model::AddressInfo {
-            db_key: keys::address_info(account, keychain, index),
             label,
             received_payments: vec![],
             received_amount: 0,
@@ -392,7 +391,7 @@ where
             batch.put(&keys::address(account, keychain, index), &address)?;
             batch.put(&keys::address_path(account, keychain, index), &path)?;
             batch.put(&keys::address_pkh(account, keychain, index), &pkh)?;
-            batch.put(&info.db_key, &info)?;
+            batch.put(&keys::address_info(account, keychain, index), &info)?;
             batch.put(
                 &keys::pkh(&pkh),
                 &model::Path {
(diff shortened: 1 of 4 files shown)


wallet/src/repository/wallet/mod.rs
```diff
@@ -391,7 +391,8 @@ where
             batch.put(&keys::address(account, keychain, index), &address)?;
             batch.put(&keys::address_path(account, keychain, index), &path)?;
             batch.put(&keys::address_pkh(account, keychain, index), &pkh)?;
-            batch.put(&info.db_key, &info)?;
+            let info_db_key = keys::address_info(account, keychain, index);
+            batch.put(&info_db_key, &info)?;
```
lrubiorod, line 395: should be not use the "let" to use the same nomenclature than the others?

lrubiorod (approved):

clbartoli (approved):

aesedepece (approved):
`utACK`

PR #1822, opened 2021-01-14
feat(node): Show unconfirmed balance

### LINKED ISSUE #1814: Balance UX in the node (should tell confirmed vs. unconfirmed)

clbartoli:
The cli command `balance` now shows the balance that has been confirmed (tx included in the superblock) and the balance that has not been confirmed yet.

It includes tests for get_total_balance

Closes #1814

### DIFF
--- a/data_structures/src/transaction_factory.rs
+++ b/data_structures/src/transaction_factory.rs
@@ -21,6 +21,15 @@ pub struct TransactionInfo {
     pub fee: u64,
 }
 
+// Structure that the includes the confirmed and pending balance of a node
+#[derive(Copy, Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub struct NodeBalance {
+    /// Total amount of a node's funds after last confirmed superblock
+    pub confirmed: u64,
+    /// Total amount of node's funds after last block
+    pub total: u64,
+}
+
 /// Fee type distinguished between absolute or Weighted (fee/weight unit)
 #[derive(Copy, Clone, Debug, Serialize, Deserialize)]
 pub enum FeeType {
@@ -236,9 +245,14 @@ pub fn calculate_weight(
 }
 
 /// Get total balance
-pub fn get_total_balance(all_utxos: &UnspentOutputsPool, pkh: PublicKeyHash) -> u64 {
+pub fn get_total_balance(
+    all_utxos: &UnspentOutputsPool,
+    old_all_utxos: Option<&UnspentOutputsPool>,
+    pkh: PublicKeyHash,
+) -> NodeBalance {
     // FIXME: this does not scale, we need to be able to get UTXOs by PKH
-    all_utxos
+    // Get the balance of the current utxo set
+    let new_balance = all_utxos
         .iter()
         .filter_map(|(_output_pointer, (vto, _))| {
             if vto.pkh == pkh {
@@ -247,7 +261,27 @@ pub fn get_total_balance(all_utxos: &UnspentOutputsPool, pkh: PublicKeyHash) -> 
                 None
             }
         })
-        .sum()
+        .sum();
+
+    // Get the balance of the confirmed utxo set from storage
+    let old_balance = match old_all_utxos {
+        Some(old_all_utxos) => old_all_utxos
+            .iter()
+            .filter_map(|(_output_pointer, (vto, _))| {
+                if vto.pkh == pkh {
+                    Some(vto.value)
+                } else {
+                    None
+                }
+            })
+            .sum(),
+        None => 0,
+    };
+
+    NodeBalance {
+        confirmed: old_balance,
+        total: new_balance,
+    }
 }
 
 /// If the change_amount is greater than 0, insert a change output using the supplied `pkh`.
@@ -485,6 +519,13 @@ mod tests {
         PublicKeyHash::from_public_key(&PublicKey {
             compressed: 2,
             bytes: [0x01; 32],
+        })
+    }
+
+    fn bob_pkh() -> PublicKeyHash {
+        PublicKeyHash::from_public_key(&PublicKey {
+            compressed: 2,
+            bytes: [0x04; 32],
         })
     }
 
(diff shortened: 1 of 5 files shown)


aesedepece:
This is totally correct for the case of incoming transactions, i.e. some UTXOs to your PKH will eventually get confirmed by a superblock. Example:
- I have 10 wits
- I receive 1 wit
- Confirmed shows 10
- Unconfirmed shows 1
- After a while, it either shows 11/0 or goes back to 10/0 in case of rollback

However, **this seems incorrect to me for the case of outgoing transactions**. Example:

- I have 10 wits
- I send 1 to a different user
- Confirmed shows 9
- Unconfirmed shows 0

I was expecting confirmed to show 10, and unconfirmed to show -1.

In a separate concern, I find it hard to reason about what "total balance" means in this context. Again, for receiving, OK, you have 10, 1 is pending confirmation, so "total" stands for your future balance; but when paying out, "total" is puzzling to me.

data_structures/src/transaction_factory.rs
```diff
@@ -236,9 +236,14 @@ pub fn calculate_weight(
 }
 
 /// Get total balance
-pub fn get_total_balance(all_utxos: &UnspentOutputsPool, pkh: PublicKeyHash) -> u64 {
+pub fn get_total_balance(
+    all_utxos: &UnspentOutputsPool,
+    old_all_utxos: Option<UnspentOutputsPool>,
```
tmpolaczyk, line 241:
```suggestion
    old_all_utxos: Option<&UnspentOutputsPool>,
```

node/src/actors/chain_manager/handlers.rs
```diff
@@ -1292,20 +1293,43 @@ impl Handler<GetDataRequestInfo> for ChainManager {
...
-            &self.chain_state.unspent_outputs_pool,
-            pkh,
-        ))
+        let magic = self.get_magic();
+
+        let res = storage_mngr::get::<_, ChainState>(&storage_keys::chain_state_key(magic))
+            .into_actor(self)
+            .then(move |chain_state_from_storage, act, _| {
+                let result = match chain_state_from_storage {
+                    Ok(Some(x)) => transaction_factory::get_total_balance(
```
tmpolaczyk, line 1314: The `Ok(Some(x))` and `Ok(None)` branches are the same, so maybe they can be merged as:

```suggestion
                    Ok(x) => transaction_factory::get_total_balance(
```

src/cli/node/json_rpc_client.rs
```diff
@@ -99,9 +99,26 @@ pub fn get_balance(addr: SocketAddr, pkh: Option<PublicKeyHash>) -> Result<(), f
     );
     let response = send_request(&mut stream, &request)?;
     log::info!("{}", response);
-    let amount = parse_response::<u64>(&response)?;
-
-    println!("{} wits", Wit::from_nanowits(amount));
+    let amount = parse_response::<(u64, u64)>(&response)?;
+    if amount.0 >= amount.1 {
```
tmpolaczyk, line 103: This if is needed because `Wit` must be positive, and here we also want to print negative balance. Should we create a new `SignedWit` type that accepts negative amounts?

data_structures/src/transaction_factory.rs
```diff
@@ -236,9 +236,14 @@ pub fn calculate_weight(
...
+    pkh: PublicKeyHash,
+) -> (u64, u64) {
```
tmpolaczyk, line 243: Instead of `(u64, u64)` let's use a struct with named fields, similar to `WalletBalance` from the wallet

```suggestion
) -> NodeBalance {
```

data_structures/src/transaction_factory.rs
```diff
@@ -247,7 +261,27 @@ pub fn get_total_balance(all_utxos: &UnspentOutputsPool, pkh: PublicKeyHash) ->
                 None
             }
         })
-        .sum()
+        .sum();
+
+    // Get the balance of the confirmed utxo set from storage
+    let old_balance = match old_all_utxos {
+        Some(old_all_utxos) => old_all_utxos
```
girazoki, line 268: I tend to use If let Some(old_all_utxos) = old_all_utxos for Options but its just a matter of feeling (I would rather use match for more than 2 patterns).  But it is a matter of style

src/cli/node/json_rpc_client.rs
```diff
@@ -99,9 +100,26 @@ pub fn get_balance(addr: SocketAddr, pkh: Option<PublicKeyHash>) -> Result<(), f
...
+    } else {
+        println!(
+            "Total balance:       {} wits\n\
+        Confirmed balance:   {} wits\n\
+        Pending balance:     -{} wits",
+            Wit::from_nanowits(amount.total),
+            Wit::from_nanowits(amount.confirmed),
+            Wit::from_nanowits(amount.confirmed - amount.total)
+        );
+    }
```
aesedepece, line 122: There's some code repetition here, I think we could unify both branches, as `total - confirmed` will be negative for all cases in which `total < confirmed`. Although maybe you are doing this to avoid casting them to `i64`, right? If so, I guess ths is fine then.

On other matters, I'm still unsure about the "Total balance" thing. Beyond my initial concern that the wording "total" doesn't represent what it actually is (what your balance will be when pending transactions are eventually confirmed), I have the feeling that this still could lead someone into thinking they have received a payment while it is still not confirmed. Maybe I'm nitpicking here, but I think we're assuming the user to be educated enough about finality of transactions, when the truth is that the demographics for Witnet node operators are quite diverse and many of them won't totally understand what it means.

My hunch is that showing confirmed and pending balance may be enough. If someone has a use case in which they  do need to know their potential balance after eventual confirmation, I think they can do the math on their own.

However, if we finally decide to keep the total balance, in all cases, I would totally print it last, as in that way it'll be easier for the user to figure out that it's in fact the addition of the confirmed and pending balances.

aesedepece (changes requested):

tmpolaczyk:
It would be nice if `node getBalance --help` provided an explaination of what is confirmed balance and what is total balance.

src/cli/node/json_rpc_client.rs
clbartoli: I think is alright not to include the Total balance, but in that case I would specify the sign of the pending balance (as it is now),to avoid confusion

node/src/actors/chain_manager/handlers.rs
```diff
@@ -1292,20 +1294,39 @@ impl Handler<GetDataRequestInfo> for ChainManager {
...
+            .then(move |chain_state_from_storage, act, _| {
+                let result = match chain_state_from_storage {
+                    Ok(x) => transaction_factory::get_total_balance(
+                        &act.chain_state.unspent_outputs_pool,
+                        // If there is no chain state in storage, send None and set the balance to 0
+                        x.as_ref().map(|x| &x.unspent_outputs_pool),
+                        pkh,
+                    ),
+                    Err(e) => {
+                        panic!("Error while getting chain state from storage {}", e);
```
tmpolaczyk, line 1322: No need to panic because this function can return an error

src/cli/node/json_rpc_client.rs
aesedepece: Good. I was assuming the sign would appear automatically when printing a negative `i64` number.

clbartoli: I've added a function (`negative_wits`) to print a negative i64 in the case the pending balance is < 0. Another option could be change the Wit structure from u64 to i64. However, since at the moment it looks like it only affects this specific case, i think adding the function `negative_wits` should be enough

node/src/actors/chain_manager/handlers.rs
```diff
@@ -12,6 +12,7 @@ use witnet_data_structures::{
     error::{ChainInfoError, TransactionError::DataRequestNotFound},
     transaction::{DRTransaction, Transaction, VTTransaction},
     transaction_factory,
+    transaction_factory::NodeBalance,
```
lrubiorod, line 15: fmt

node/src/actors/chain_manager/handlers.rs
```diff
@@ -1292,20 +1294,35 @@ impl Handler<GetDataRequestInfo> for ChainManager {
...
+            .into_actor(self)
+            .map(move |chain_state_from_storage, act, _| {
+                match chain_state_from_storage {
+                    Ok(x) => Ok(transaction_factory::get_total_balance(
+                        &act.chain_state.unspent_outputs_pool,
+                        // If there is no chain state in storage, send None and set the balance to 0
+                        x.as_ref().map(|x| &x.unspent_outputs_pool),
+                        pkh,
+                    )),
+                    Err(e) => Err(e),
```
lrubiorod, line 1321:
```suggestion
                chain_state_from_storage.map(|x| {
                    transaction_factory::get_total_balance(
                        &act.chain_state.unspent_outputs_pool,
                        // If there is no chain state in storage, send None and set the balance to 0
                        x.as_ref().map(|x| &x.unspent_outputs_pool),
                        pkh,
                    })
```

lrubiorod: could be expressed as a map?

tmpolaczyk (approved):

aesedepece (approved):
`ACK`

PR #1825, opened 2021-01-15
fix(radon): remove unwrap in CBOR encode hash function to avoid panics

lrubiorod:

### DIFF
--- a/rad/src/types/mod.rs
+++ b/rad/src/types/mod.rs
@@ -212,7 +212,13 @@ impl PartialEq for RadonTypes {
 impl std::hash::Hash for RadonTypes {
     // FIXME(953): Unify all CBOR libraries
     fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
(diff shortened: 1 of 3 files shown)


rad/src/types/mod.rs
```diff
@@ -212,7 +212,13 @@ impl PartialEq for RadonTypes {
 impl std::hash::Hash for RadonTypes {
     // FIXME(953): Unify all CBOR libraries
     fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
-        self.encode().map(|vec| vec.hash(state)).unwrap();
+        match self.encode() {
+            Ok(vec) => vec.hash(state),
+            Err(e) => {
+                let error_vec = e.to_string().as_bytes().to_vec();
```
tmpolaczyk, line 218: String implements Hash, so you can just do:

```suggestion
                let error_vec = e.to_string();
```

tmpolaczyk (approved):

Issue #1827, opened 2021-01-18
Too many newcomer blocks without transactions


girazoki:
Recently there have been many blocks in the network that do not include any TX apart from the tally and the mint transactions (mandatory). After investigating these block producers, we have observed that all of them come from newcomers. The proportion of blocks coming from newcomers is bigger than we expected, so after reviewing our equations we have found that the probability for newcomers is wrongly calculated in accordance to our DoS protections. Currently, the eligibility probability is being calculated utilizing the active nodes as a denominator, but then only those blocks produced by reputed nodes are given priority over the rest. With current numbers this yields:

Active Nodes: ~6000

Reputed Nodes: ~2000

RF=3

P(no Reputed node produces block) = P(one reputed node does not produce block)^(number of reputed nodes) = (1 - (3/6000))^2000 = 0.37

In other words, currently there is a 37% chance that a newcomer can insert its node in the ARS, and thus also censor TXs by creating empty blocks.  This is far from the numbers we analyzed, and in accordance the following preemptive measures have been discussed:

- Utilize as a denominator for the eligibility the Reputed number of nodes, instead of the ARS length. This would increases the probability for reputed nodes to become eligible, and thus would make it more difficult for an outsider to censor TXs. The only caveat is that it increases the number of candidates.

- Utilize as a tie-breaker the fact that a node IS IN THE ARS, instead of a node having reputation. This also decreases the number of odds for an outsider to censor TXs, as usually the ARS is larger than the number of reputed nodes.

- Increase the replication factor while removing the back up factor. We are not even near of being epochs without block candidates, so the back up factor removal is something we should start considering. Further, an incerase on the replication factor gives more probability for the reputed/ARS nodes to find a valid candidate and thus prevent an outisder from continously proposing blocks.

- Take into account the block weight in terms of data requests, giving of course a lower priority to empty candidates.

This should follow a proper discussion, and most likely yield a hard fork. I see option 1 and option 2 as exclusive, while 3  and 4can be added on top any of the first two

Resources: https://github.com/witnet/research/blob/master/reputation/docs/randpoe.md

drcpu-github:
[...]

PR #1829, opened 2021-01-19
Second Hard Fork proposal

lrubiorod:
This proposal of Second Hard Fork:
- Modifies mining probabilities using total identities with non-zero reputation instead of total active identities

https://github.com/witnet/WIPs/blob/master/wip-0009.md

### DIFF
--- a/data_structures/src/mainnet_validations.rs
+++ b/data_structures/src/mainnet_validations.rs
@@ -13,6 +13,8 @@ const FIRST_EMERGENCY_COMMITTEE: [&str; 7] = [
 
 /// 22 January 2021 @ 09:00:00 UTC
 pub const FIRST_HARD_FORK: Epoch = 192000;
+/// 16 March 2021 @ 09:00:00 UTC
+pub const SECOND_HARD_FORK: Epoch = 293680;
 
 /// Return a hard-coded signing committee if the provided epoch belongs to an emergency period.
 /// 750 and 1344: Between those indices, a special committee of 7 nodes was set.
@@ -35,6 +37,11 @@ pub fn in_emergency_period(
 /// Returns a boolean indicating whether the epoch provided is after the first hard fork date
 pub fn after_first_hard_fork(epoch: Epoch, environment: Environment) -> bool {
     epoch >= FIRST_HARD_FORK && Environment::Mainnet == environment
+}
+
+/// Returns a boolean indicating whether the epoch provided is after the second hard fork date
+pub fn after_second_hard_fork(epoch: Epoch, environment: Environment) -> bool {
+    epoch >= SECOND_HARD_FORK && Environment::Mainnet == environment
 }
 
 #[cfg(test)]
--- a/node/src/actors/chain_manager/mining.rs
+++ b/node/src/actors/chain_manager/mining.rs
@@ -28,11 +28,10 @@ use crate::{
     },
     signature_mngr,
 };
-use witnet_data_structures::chain::Hash;
 use witnet_data_structures::{
     chain::{
         Block, BlockHeader, BlockMerkleRoots, BlockTransactions, Bn256PublicKey, CheckpointBeacon,
-        CheckpointVRF, DataRequestOutput, EpochConstants, Hashable, Input, PublicKeyHash,
+        CheckpointVRF, DataRequestOutput, EpochConstants, Hash, Hashable, Input, PublicKeyHash,
         ReputationEngine, TransactionsPool, ValueTransferOutput,
     },
     data_request::{calculate_witness_reward, create_tally, DataRequestPool},
--- a/validations/src/validations.rs
+++ b/validations/src/validations.rs
@@ -2112,13 +2112,16 @@ pub fn total_block_reward(initial_block_reward: u64, halving_period: u32) -> u64
 ///
 /// Note that this only compares the block candidates, it does not validate them. A block must be
 /// the best candidate and additionally it must be valid in order to be the consolidated block.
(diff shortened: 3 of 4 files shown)


validations/src/validations.rs
```diff
@@ -1753,10 +1753,22 @@ pub fn validate_block(
         // with the genesis_block_hash
         validate_genesis_block(block, consensus_constants.genesis_hash).map_err(Into::into)
     } else {
-        let total_identities = u32::try_from(rep_eng.ars().active_identities_number())?;
+        let (total_identities, replication_factor) =
+            if after_second_hard_fork(current_epoch, get_environment()) {
```
tmpolaczyk, line 1757: Not sure if `current_epoch == block_epoch` here

```suggestion
            if after_second_hard_fork(block_epoch, get_environment()) {
```

aesedepece (changes requested):
`NACK`

In the context of a broad, decentralized network, one would normally try to avoid any further hard fork until the one that enables nodes to signal actual support for specific hard forks; so as to prevent severe network disruption in case of a failed upgrade (let's see how the Jan 22nd upgrade goes).

In my mind, a second Witnet network upgrade / hard fork would focus on implementing a mechanism that delays activation of further consensus rules until some threshold of miners have upgraded to the newer version that includes the new validation rules.

That being said, if the changes in this PR are critical and the lack thereof really endangers the network, I'm open to be convinced of the urgency, and I'll personally push for this hard fork over the one I mentioned.

Needless to say, that both hard forks can come together as long as the one from this PR has a posterior activation date. That's something I'd also personally `ACK`, and I would be totally fine if this PR was merged into a branch different than master until a PR implementing what I mentioned before lands on master.

node/src/actors/chain_manager/mining.rs
```diff
@@ -108,6 +108,19 @@ impl ChainManager {
             .consensus_constants
             .epochs_with_initial_difficulty;
 
+        let (total_identities, replication_factor) =
+            if after_second_hard_fork(current_epoch, get_environment()) {
+                (
+                    u32::try_from(rep_engine.trs().num_identities()).unwrap(),
```
tmpolaczyk, line 114: Are we sure that `rep_eng.trs().num_identities()` will never be 0? Maybe we should add a minimum difficulty to prevent saturating the network with candidates when this value is low.

lrubiorod:
> `NACK`
> 
> In the context of a broad, decentralized network, one would normally try to avoid any further hard fork until the one that enables nodes to signal actual support for specific hard forks; so as to prevent severe network disruption in case of a failed upgrade (let's see how the Jan 22nd upgrade goes).
> 
> In my mind, a second Witnet network upgrade / hard fork would focus on implementing a mechanism that delays activation of further consensus rules until some threshold of miners have upgraded to the newer version that includes the new validation rules.
> 
> That being said, if the changes in this PR are critical and the lack thereof really endangers the network, I'm open to be convinced of the urgency, and I'll personally push for this hard fork over the one I mentioned.
> 
> Needless to say, that both hard forks can come together as long as the one from this PR has a posterior activation date. That's something I'd also personally `ACK`, and I would be totally fine if this PR was merged into a branch different than master until a PR implementing what I mentioned before lands on master.

I agree, we should prioritize the creation of a mechanism of "soft" hard forks. In this case, the urgency of this change is that in the current state, the percentage of newcomers is highly than spectated, so it makes the network more vulnerable to an attacker with a massive quantity of nodes.

aesedepece:
I'm connecting this to #1827 as it seems quite related to me.

aesedepece:
I created an issue in the WIPs repo for drafting a WIP or a couple of WIPs covering the changes this proposal:
https://github.com/witnet/WIPs/issues/39

I took the liberty of calling this "RandPoE 2.0" :rofl:

tmpolaczyk:
Add a comment to `ConsensusConstants` saying that `mining_backup_factor` is not used and it is equal to `mining_replication_factor` in mainnet.

tmpolaczyk:
Also if possible add tests to validations that show how a block that was valid before the second hard fork is invalid after it, please!

validations/src/tests.rs
```diff
@@ -5870,6 +5871,218 @@ fn block_difficult_proof() {
     );
 }
 
+#[test]
+fn block_difficult_proof_after_second_hard_fork() {
+    let dr_pool = DataRequestPool::default();
+    let vrf = &mut VrfCtx::secp256k1().unwrap();
+
+    // Create a reputation engine with 512 identities
```
tmpolaczyk, line 5879: ?

```suggestion
    // Create a reputation engine with 16 identities
```

data_structures/src/mainnet_validations.rs
```diff
@@ -11,8 +11,10 @@ const FIRST_EMERGENCY_COMMITTEE: [&str; 7] = [
     "wit1etherz02v4fvqty6jhdawefd0pl33qtevy7s4z",
 ];
 
-/// 22 January 2021 @ 09:00:00 UTC
+/// 22th January 2021 @ 09:00:00 UTC
```
tmpolaczyk, line 14:
```suggestion
/// 22nd January 2021 @ 09:00:00 UTC
```

tmpolaczyk: But I prefer just `22`

tmpolaczyk (approved):
Approved but let's merge to a new branch, because the fork date is not yet final and we may want to add more fixes or features.

lrubiorod:
Merged in second_hard_fork branch

PR #1830, opened 2021-01-20
fix(radon): handle empty arrays in deviation filter

lrubiorod:


aesedepece:
I'm OK with throwing `UnsupportedFilter` when trying to remove outliers from an empty array; although this corner case could well use a more specific error, because otherwise the reason for lack of support could be non obvious.

data_structures/src/radon_report.rs
```diff
@@ -407,5 +407,9 @@ mod tests {
         ];
 
         assert_eq!(metadata.liars, expected);
+
+        // empty vec does not modify anything
```
aesedepece, line 411: Please elaborate on this. Maybe something like this? (I don't know if I understood correctly)
```suggestion
        // An empty vector of new liars should cause no changes to the existing list of of liars 
```

aesedepece (changes requested):

rad/src/filters/deviation.rs
```diff
@@ -40,6 +40,14 @@ pub fn standard_filter(
     match value.first() {
         None => Ok(RadonTypes::from(input.clone())),
         Some(RadonTypes::Array(arr2)) => {
+            // Empty array
+            if arr2.value().is_empty() {
+                return Err(RadError::UnsupportedFilter {
+                    array: input.clone(),
+                    filter: RadonFilters::DeviationStandard.to_string(),
+                });
```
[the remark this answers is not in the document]
lrubiorod: It is not an empty array, is an array of an empty array... that could be a strange case

rad/src/filters/deviation.rs
```diff
@@ -619,4 +627,40 @@ mod tests {
...
+
+    #[test]
+    fn test_filter_deviation_standard_array_of_empty_arrays() {
+        let num_witnesses = 1;
+        // Reveal value:
+        // 1D array: []
+        let input = vec![];
+        // Tally input:
+        // 2D array: [[]; num_witnesses]
+        let input_vec = vec![RadonTypes::Array(RadonArray::from(input)); num_witnesses];
```
tmpolaczyk, line 639: This is probably a different issue, but this input also panics with message "Only 2D arrays are supported"

```suggestion
        let input_vec = vec![RadonTypes::Array(RadonArray::from(vec![
            RadonTypes::Integer(RadonInteger::from(9)),
            RadonTypes::Array(RadonArray::from(vec![])),
        ])); num_witnesses];
```

lrubiorod: I am checking that with this PR it also works

tmpolaczyk: My bad, it should be like this:

```rust
let inner_arr = RadonTypes::Array(RadonArray::from(vec![
            RadonTypes::Integer(RadonInteger::from(9)),
            RadonTypes::Array(RadonArray::from(vec![
                RadonTypes::Integer(RadonInteger::from(9)),
            ])),
        ]));
```

aesedepece (approved):
`utACK`

tmpolaczyk (approved):

PR #1831, opened 2021-01-21
feat(jsonrpc): add confirmed flag to getTransaction method

### LINKED ISSUE #1828: JSONRPC/CLI method for transaction details should show confirmation / finality status
There's no simple way for a node operator to check for confirmation / finality status of a transaction.

In the same way as the `transaction` JSONRPC/CLI method shows the block in which a transaction was last found, it should also show whether that block has been confirmed by a superblock or not.

tmpolaczyk:
Close #1828 

Just like the `getBlock` method, now the result of `getTransaction` includes a "confirmed" field. This field is set to `true` if the block that includes this transaction has been confirmed by a superblock, and `false` otherwise.

### DIFF
--- a/node/src/actors/inventory_manager/handlers.rs
+++ b/node/src/actors/inventory_manager/handlers.rs
@@ -7,10 +7,10 @@ use crate::actors::messages::{
     StoreInventoryItem, SuperBlockNotify,
 };
 use crate::storage_mngr;
-use witnet_data_structures::chain::{
-    Block, Hash, Hashable, InventoryEntry, InventoryItem, PointerToBlock,
+use witnet_data_structures::{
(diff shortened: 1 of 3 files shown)


node/src/actors/inventory_manager/handlers.rs
```diff
@@ -7,10 +7,10 @@ use crate::actors::messages::{
     StoreInventoryItem, SuperBlockNotify,
 };
 use crate::storage_mngr;
-use witnet_data_structures::chain::{
-    Block, Hash, Hashable, InventoryEntry, InventoryItem, PointerToBlock,
+use witnet_data_structures::{
+    chain::{Block, Epoch, Hash, Hashable, InventoryEntry, InventoryItem, PointerToBlock},
```
lrubiorod, line 11: like!

lrubiorod (approved):

PR #1838, opened 2021-01-27
feat(wallet): allow to avoid using unconfirmed UTXOs in transactions

### LINKED ISSUE #1832: [Wallet] Only allow transacting when superblock-confirmed balance is greater than 0
At the moment the Wallet checks the total balance when sending a vtt or a DR. However the wallet should check the available balance, meaning the balance that has been confirmed in a superblock, and ignore the pending balance.

tmpolaczyk:
Close #1832 

Keep track of the transactions that modify the utxo set but are not confirmed yet, so that we can ignore unconfirmed utxo when selecting utxos for a new transaction.

This feature is disabled by default, can be enabled by setting

    wallet.use_unconfirmed_utxos = false

in witnet.toml

### DIFF
--- a/wallet/src/actors/app/error.rs
+++ b/wallet/src/actors/app/error.rs
@@ -115,12 +115,18 @@ impl From<actors::worker::Error> for Error {
             actors::worker::Error::KeyGen(e @ crypto::Error::InvalidKeyPath(_)) => {
                 validation_error(field_error("seedData", e.to_string()))
             }
-            actors::worker::Error::Repository(_e @ repository::Error::InsufficientBalance) => {
-                validation_error(field_error(
-                    "balance",
-                    "Wallet account has not enough balance",
-                ))
-            }
+            actors::worker::Error::Repository(repository::Error::InsufficientBalance {
+                total_balance,
+                available_balance,
+                transaction_value,
+            }) => validation_error(field_error(
+                json! {{
+                    "total_balance": total_balance,
+                    "available_balance": available_balance,
+                    "transaction_value": transaction_value,
+                }},
+                "Wallet account has not enough balance",
+            )),
             actors::worker::Error::JsonRpcTimeoutError => Error::JsonRpcTimeoutError,
             _ => internal_error(err),
         }
--- a/wallet/src/repository/wallet/mod.rs
+++ b/wallet/src/repository/wallet/mod.rs
@@ -1230,10 +1230,16 @@ where
         utxo_strategy: &UtxoSelectionStrategy,
         max_weight: u32,
     ) -> Result<(Vec<Input>, Vec<ValueTransferOutput>)> {
+        let empty_hashset = HashSet::default();
+        let unconfirmed_transactions = if self.params.use_unconfirmed_utxos {
+            &empty_hashset
+        } else {
+            &state.pending_transactions
+        };
         let mut wallet_utxos = WalletUtxos {
             utxo_set: &state.utxo_set,
             used_outputs: &mut state.used_outputs,
-            unconfirmed_transactions: &state.pending_transactions,
+            unconfirmed_transactions,
         };
 
         let tx_info = wallet_utxos.build_inputs_outputs(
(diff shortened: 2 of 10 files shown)


wallet/src/repository/wallet/mod.rs
```diff
@@ -42,6 +42,7 @@ use crate::{
 
 use super::*;
 use std::collections::hash_map::Entry;
+use std::collections::HashSet;
```
lrubiorod, line 45: fmt :P

wallet/src/repository/wallet/mod.rs
```diff
@@ -1173,8 +1209,11 @@ where
         let mut wallet_utxos = WalletUtxos {
             utxo_set: &state.utxo_set,
             used_outputs: &mut state.used_outputs,
+            unconfirmed_transactions: &state.pending_transactions,
         };
 
+        println!("{:?}", wallet_utxos);
+
```
lrubiorod, line 1216: println?

tmpolaczyk: For debugging :smiley:

tmpolaczyk:
This PR is blocked because the "available" balance does not match the balance that is calculated using the confirmed UTXOs. For example, the available balance can show 5000 nanowits and creating a transaction with less than that value can fail.

wallet/src/actors/app/error.rs
```diff
@@ -115,12 +115,18 @@ impl From<actors::worker::Error> for Error {
...
+                available_balance,
+                transaction_value,
+            }) => validation_error(field_error(
+                json! {{
+                    "total_balance": total_balance,
+                    "available_balance": available_balance,
+                    "transaction_value": transaction_value,
+                }},
+                "Wallet account has not enough balance",
+            )),
```
tmpolaczyk, line 129: This was the easiest way to send an object to the frontend, but this json object is serialized as a string, so the result is really ugly, JSON inside JSON:

    {"code":400,"message":"Validation Error","data":[["{\"available_balance\":17221973255,\"total_balance\":17221973255,\"transaction_value\":18000000853}","Wallet account has not enough balance"]]}

I don't know how to add additional fields to this error message, so any better alternatives welcome.

.github/workflows/push.yml
```diff
@@ -19,6 +19,12 @@ jobs:
         -D clippy::cast-sign-loss
         -D clippy::checked-conversions
         -A clippy::field-reassign-with-default
+        -A clippy::upper-case-acronyms
```
tmpolaczyk, line 22: Reminder to remove this commit before merging, this PR should go after #1888

mariocao:
What's the status of this PR?

wallet/src/repository/wallet/mod.rs
```diff
@@ -59,14 +58,36 @@ struct AccountMutation {
...
+            let pointer: OutputPointer = out_ptr.into();
+            if self
+                .unconfirmed_transactions
+                .contains(&pointer.transaction_id)
+            {
+                None
+            } else {
+                Some(pointer)
+            }
+        };
```
mariocao, line 76: We are still not sure if this if something a user would like or not... :smile: 

- This feature will make that transactions that require some unconfirmed UTXO, will have to wait at least 20 blocks.
- Some users might NOT want to able to create transactions from unconfirmed UTXOs (e.g. trades)
- Other users might prefer to have the freedom to create several transaction without having to wait for confirmation.

wallet/src/repository/wallet/mod.rs
```diff
@@ -1203,9 +1230,16 @@ where
         utxo_strategy: &UtxoSelectionStrategy,
         max_weight: u32,
     ) -> Result<(Vec<Input>, Vec<ValueTransferOutput>)> {
+        let empty_hashset = HashSet::default();
+        let unconfirmed_transactions = if self.params.use_unconfirmed_utxos {
+            &empty_hashset
+        } else {
+            &state.pending_transactions
+        };
```
lrubiorod, line 1238: Is it not the opposite logic? It means, if we want to use unconfirmed_utxos we include the pending transactions

lrubiorod: I didnt say nothing, it is to create the hashset to check if we can use or not, so it is ok

tmpolaczyk: It seems backwards but this is the logic: `unconfirmed_transactions` is the list of transactions that will not be used. If `unconfirmed_transactions` is empty, we will use all the unconfirmed UTXOs (this is the old behavior before this PR). If `unconfirmed_transactions` is equal to pending transactions, we will not use unconfirmed UTXOs in new transactions.

lrubiorod (approved):

Issue #1836, opened 2021-01-26
[Wallet] Show confirmed balance when the wallet is resynced


gabaldon:
The balance shown when resyncing the wallet does not show the confirmed balance until a superblock is confirmed or the user logs in the wallet again.
When resyncing, It should show the balance confirmed in previous superblocks.

aesedepece:
Good one. All blocks received while either syncing or resyncing are already confirmed by a superblock, so any balance derived from them is inherently confirmed.

PR #1846, opened 2021-02-01
fix(wallet): fix show transactions after resync

### LINKED ISSUE #1818: [Wallet] Wallet only shows genesis transactions after resync
Steps to reproduce:

* Unlock a synced wallet with some transactions from the genesis block and some transactions from a later block
* Press resync button
* After it shows the genesis transactions, close the wallet
* Unlock the wallet again and wait
* The wallet will sync but only the genesis transactions will be shown.

mariocao:
Closes #1818

- Removes `resynchronization` flag
- Reuses a `needs_indexing` flag in `handle_block` function (single entry point for sync, resync and new block from node)
- Updates tests

### DIFF
--- a/wallet/src/actors/worker/handlers/handle_block.rs
+++ b/wallet/src/actors/worker/handlers/handle_block.rs
@@ -18,6 +18,6 @@ impl Handler<HandleBlockRequest> for worker::Worker {
     type Result = <HandleBlockRequest as Message>::Result;
 
     fn handle(&mut self, msg: HandleBlockRequest, _ctx: &mut Self::Context) -> Self::Result {
-        self.handle_block(msg.block, false, false, msg.wallet, msg.sink)
+        self.handle_block(msg.block, false, msg.wallet, msg.sink)
     }
 }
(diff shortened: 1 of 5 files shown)


clbartoli:
When stopping the wallet while resyncing some of the balances are shown as  unconfirmed even if they were already confirmed. This could happen because by the time the resync was stopped those balances were not yet confirmed

image

mariocao:
> When stopping the wallet while resyncing some of the balances are shown as unconfirmed even if they were already confirmed. This could happen because by the time the resync was stopped those balances were not yet confirmed

Definitely there is something weird with the balances! :smile: I will check how related it is to the issue https://github.com/witnet/witnet-rust/issues/1836

mariocao:
After discussing and debugging with @lrubiorod and @clbartoli we found out the following issues:
 - The wallet is not tracking the available balance while synchronizing (issue #1836)
 - The wallet is not handling well "double" synchronization after unlocking the wallet for a second time (issue #1849)
 - Sometimes Sheikah goes back to the wallet list and ask to unlock wallet (despite being already unlocked) (issue https://github.com/witnet/sheikah/issues/1644)

lrubiorod (approved):

aesedepece:
I'm trying this

aesedepece (approved):
It works!

PR #1848, opened 2021-02-02
feat(radon): modify RadonMap::values() to return an RadonArray sorted…

### LINKED ISSUE #1847: Modify RadonMap::values() to return values sorted by key

lrubiorod:
… by keys

Close #1847

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4835,6 +4835,7 @@ dependencies = [
  "failure",
  "futures 0.3.8",
  "hex 0.4.2",
+ "itertools",
  "json",
  "log 0.4.11",
  "num_enum",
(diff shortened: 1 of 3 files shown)


tmpolaczyk:
What about RadonMap::keys()? Should that be sorted as well? Although in that case users can sort the array using RadonArray::sort().

Could you provide a few examples with use cases of this operators to see why they need to be sorted by key?

lrubiorod:
> What about RadonMap::keys()? Should that be sorted as well? Although in that case users can sort the array using RadonArray::sort().

I agree, it should be sorted to have a deterministic response, but it also could sorted with the RadonArray::sort()


> Could you provide a few examples with use cases of this operators to see why they need to be sorted by key?

convert {positive: 1 negative: 0} into [0, 1]

lrubiorod:
Solved in #1855

PR #1850, opened 2021-02-03
Handle radon panics

lrubiorod:
This PR has to be merged after #1829


data_structures/src/radon_report.rs
```diff
@@ -120,7 +120,7 @@ where
     /// The index of the last argument in a call that has been processed.
     pub call_argument_index: Option<u8>,
     /// The index of the last call that has been processed.
-    pub call_index: Option<u8>,
+    pub call_index: Option<usize>,
     /// The operator in the last call that has been processed.
     pub call_operator: Option<u8>,
```
tmpolaczyk, line 125: Let's change all the other `u8` to `usize` as well

node/src/actors/chain_manager/mining.rs
```diff
@@ -647,6 +662,13 @@ impl ChainManager {
                             // Mailbox error
                             .map_err(|e| log::error!("Couldn't run tally: {}", e))?;
 
+                        // RadError::Unknown should not be included in a Tally until second hard fork
```
tmpolaczyk, line 665: This is only enforced in mining, not in validations. So if a node disables this check and broadcasts a block with a RadError::Unknown tally and it will be accepted as valid before the hard fork date.

lrubiorod: A node that tries to validate a tally with a RadError::Unknown produced by a radon panic and it is not updated, it will panic so...

lrubiorod:
Due to the new actix update, this PR is not need it. The essential changes are included in #1860

PR #1851, opened 2021-02-04
Fix possible overflow when calculating absolute value

### LINKED ISSUE #1839: i32::abs can overflow when input is i32::MIN
The following code panics in debug mode and returns -2147483648 in release mode:

```rust
let x = i32::MIN.abs();
```

Same applies to the other integer types (`i8` to `i128`), but not to the floating point types. So `f64::abs` is safe to use.

Most of the time, we want to take the absolute value to compare it with an unsigned value. In that case, there is an unstable method in the standard library that returns an unsigned integer:

https://doc.rust-lang.org/std/primitive.i32.html#method.unsigned_abs

We cannot use it because it's unstable, but this is the implementation:

```rust
pub const fn unsigned_abs(self) -> $UnsignedT {
    self.wrapping_abs() as $UnsignedT
}
```

So we should review all the uses of `abs` to ensure that they are correct. This are all the uses of `abs` on integers that I found:

https://github.com/witnet/witnet-rust/blob/b97a14c69e103d4e49cbf1482b578983563d77e5/node/src/actors/json_rpc/json_rpc_methods.rs#L514

Here it is used to convert to `u32` so we should use unsigned_abs.

https://github.com/witnet/witnet-rust/blob/b805a701644d6e9900974f12c20afb8b3c0e6e47/rad/src/operators/string.rs#L149

This is `|exponent| < 38`, so we should use unsigned_abs.

https://github.com/witnet/witnet-rust/blob/a5e23ff3b4abda7d443c116b418e55e23a4eea2a/node/src/actors/session/handlers.rs#L835

This is `|a - b| > diff` so we also need to be careful about overflow when doing the subtraction.

tmpolaczyk:
Fix #1839

### DIFF
--- a/node/src/actors/session/handlers.rs
+++ b/node/src/actors/session/handlers.rs
@@ -822,6 +822,28 @@ fn check_beacon_compatibility(
     }
 }
 
+/// Check that the received timestamp is close enough to the current timestamp
+fn check_timestamp_drift(
+    current_ts: i64,
+    received_ts: i64,
+    max_ts_diff: i64,
+) -> Result<(), HandshakeError> {
+    if max_ts_diff == 0 {
+        return Ok(());
+    }
+
+    let valid_ts_range =
+        current_ts.saturating_sub(max_ts_diff)..=current_ts.saturating_add(max_ts_diff);
+    if !valid_ts_range.contains(&received_ts) {
+        return Err(HandshakeError::DifferentTimestamp {
+            current_ts,
+            timestamp_diff: received_ts.saturating_sub(current_ts),
+        });
+    }
+
+    Ok(())
+}
+
 /// Function called when Version message is received
 fn handshake_version(
     session: &mut Session,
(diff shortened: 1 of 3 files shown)


node/src/actors/session/handlers.rs
```diff
@@ -832,7 +832,10 @@ fn handshake_version(
     // Check timestamp drift
     let received_ts = command_version.timestamp;
     let max_ts_diff = session.config.connections.handshake_max_ts_diff;
-    if max_ts_diff != 0 && (current_ts - received_ts).abs() > max_ts_diff {
+    if max_ts_diff != 0
+        && (current_ts.saturating_sub(max_ts_diff)..=current_ts.saturating_add(max_ts_diff))
```
lrubiorod, line 836: are you sure that it follows the same logic?

tmpolaczyk: No, but I am sure that it does not panic. I can extract this line to a function and add tests, because after this change it is not very clear.

tmpolaczyk: Oh you are right, it should be the inverse condition.

lrubiorod (approved):

PR #1852, opened 2021-02-04
fix(rad): fix deserialization of nested JSON booleans

tmpolaczyk:
Before this PR, a data request that tries to parse `"[true]"` using `StringParseJSONArray` will return an empty RadonArray. That's because booleans are mapped to null because of a missing conversion. This PR fixes this by adding that missing conversion.

Since there is no RadonNull type, all null fields are silently ignored when parsing arrays and maps using `StringParseJSON`. Added a few tests to document this behavior.

### DIFF
--- a/rad/src/operators/string.rs
+++ b/rad/src/operators/string.rs
@@ -156,7 +156,8 @@ fn json_to_cbor(value: &json::JsonValue) -> Value {
             }
         }
         json::JsonValue::String(value) => Value::Text(String::from(value.as_str())),
-        _ => Value::Null,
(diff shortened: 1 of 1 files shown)


lrubiorod:
is it yet a draft?

tmpolaczyk:
> is it yet a draft?

No, but we should make sure that this change does not accidentally break anything on mainnet

lrubiorod (approved):

lrubiorod:
I synchronize a node from zero with these PR and I didnt see any problem

mariocao (approved):

PR #1859, opened 2021-02-11
fix(clippy): apply clippy suggestions included in rust 1.50

lrubiorod:


tmpolaczyk (approved):
:+1: but rename the commit to "chore"

PR #1860, opened 2021-02-12
fix: tally creation cannot fail

tmpolaczyk:
If there is a problem when serializing the tally result, it will be replaced with a RadError::Unknown

TODO: decide if this needs to be a hard fork


validations/src/tests.rs
```diff
@@ -5330,6 +5364,50 @@ fn validate_calculate_witness_reward() {
...
+
+    assert_eq!(bytes, expected);
+}
+
+#[test]
+fn tally_unserializable_value() {
+    // When the result of the tally cannot be serialized, the result is a RadError::Unknown, as
+    // returned by `tally_bytes_on_encode_error`
+
+    // Reveal value: negative(18446744073709551361)
```
lrubiorod, line 5380: {"jsonrpc":"2.0","method":"sendRequest","id":"1","params":{"dro":{"data_request":{"time_lock":0,"retrieve":[{"kind":"HTTP-GET","url":"http://127.0.0.1:8000/negative_big_int.json","script":[129, 24, 115]}],"aggregate":{"filters":[],"reducer":3},"tally":{"filters":[],"reducer":3}},"witness_reward":1000,"witnesses":2,"commit_and_reveal_fee":1,"min_consensus_percentage":51,"collateral":1000000000},"fee":10}}

lrubiorod:
It looks fine to me!! Should we add a conditional branch of the second hard fork?

tmpolaczyk:
> It looks fine to me!! Should we add a conditional branch of the second hard fork?

Where?

lrubiorod:
I am not sure if we want this behavior or not, but in previous discussions we talk about do not create a Tally that comes from a panic in order to have a similar behavior between updated and non-updated nodes, until the second hark fork date was achieved

tmpolaczyk:
Added that change in a new commit, so we can still decide if it is needed or not.

lrubiorod (approved):

PR #1863, opened 2021-02-18
Refactor RadManager handlers to improve tests

tmpolaczyk:
This ensures that the tests that use `run_dr_locally_with_data` use the same logic as a witnet node when resolving data requests


tmpolaczyk:
Closing because the code has changed and this need to be rebased.

PR #1865, opened 2021-02-23
Rewind

tmpolaczyk:
This allows to recover nodes that have consolidated a wrong chain.

Usage: to go back to block 248839, use this command:

```
witnet node rewind --epoch 248839
```

TODO:

* What should happen if the node stops during a rewind? Save progress to storage?
* Fix panic if rewind command is called twice before the first one finishes.
* Allow to rewind with validations enabled (CLI flag --validate)? Currently all the blocks are assumed to be valid.
* Allow to rewind with an epoch that is greater than the current epoch? Currently it works (it processes all the blocks), but we could optimize it and not rewind in that case.
* [x] ~~Persist chain state after rewind success! Otherwise, if there is no superblock consensus, the node will revert to epoch 0.~~
* When setting the peer limit to 0, any open connections will result in iced peers.

### DIFF
--- a/node/src/actors/chain_manager/handlers.rs
+++ b/node/src/actors/chain_manager/handlers.rs
@@ -29,8 +29,8 @@ use crate::{
             GetBlocksEpochRange, GetDataRequestInfo, GetHighestCheckpointBeacon,
             GetMemoryTransaction, GetMempool, GetMempoolResult, GetNodeStats, GetReputation,
             GetReputationResult, GetState, GetSuperBlockVotes, GetUtxoInfo, IsConfirmedBlock,
-            PeersBeacons, ReputationStats, SendLastBeacon, SessionUnitResult, SetLastBeacon,
-            TryMineBlock,
+            PeersBeacons, ReputationStats, Rollback, SendLastBeacon, SessionUnitResult,
+            SetLastBeacon, TryMineBlock,
         },
         sessions_manager::SessionsManager,
         storage_keys,
@@ -162,7 +162,7 @@ impl Handler<EpochNotification<EveryEpochPayload>> for ChainManager {
                         }) = best_candidate
                         {
                             // Persist block and update ChainState
-                            self.consolidate_block(ctx, block, utxo_diff);
+                            self.consolidate_block(ctx, block, utxo_diff, false);
                         } else if msg.checkpoint > 0 {
                             let previous_epoch = msg.checkpoint - 1;
                             log::warn!(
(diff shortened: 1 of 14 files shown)


node/src/actors/chain_manager/handlers.rs
```diff
@@ -23,6 +23,7 @@ use witnet_util::timestamp::get_timestamp;
 use witnet_validations::validations::validate_rad_request;
 
 use super::{ChainManager, ChainManagerError, StateMachine, SyncTarget};
+use crate::actors::messages::SetPeersLimits;
```
lrubiorod, line 26: fmt

node/src/actors/sessions_manager/handlers.rs
```diff
@@ -12,6 +12,7 @@ use ansi_term::Color::Cyan;
 use tokio_util::codec::FramedRead;
 
 use super::{NotSendingPeersBeaconsBecause, SessionsManager};
+use crate::actors::messages::SetPeersLimits;
```
lrubiorod, line 15: fmt

node/src/actors/sessions_manager/handlers.rs
```diff
@@ -480,3 +481,12 @@ impl Handler<DropOutboundPeers> for SessionsManager {
         self.drop_outbound_peers(msg.peers_to_drop.as_ref());
     }
 }
+impl Handler<SetPeersLimits> for SessionsManager {
+    type Result = <DropOutboundPeers as Message>::Result;
+
+    fn handle(&mut self, msg: SetPeersLimits, _ctx: &mut Context<Self>) -> Self::Result {
+        self.sessions.set_limits(msg.inbound, msg.outbound);
+        // TODO: drop all inbound and outbound peers above new limit
```
lrubiorod, line 489: ??

lrubiorod (approved):

aesedepece (changes requested):
As per [WIP-0010](https://github.com/witnet/WIPs/pull/44), the method is expected to be named `rewind`.

aesedepece:
Do you think is it worth mentioning that the `--epoch` argument is exclusive? In other words, that we are forgetting all blocks whose epoch is strictly greater than the one provided as an argument, which excludes the block for that specific epoch?

tmpolaczyk:
> Do you think is it worth mentioning that the `--epoch` argument is exclusive? In other words, that we are forgetting all blocks whose epoch is strictly greater than the one provided as an argument, which excludes the block for that specific epoch?

It makes sense if you think about it as "rolling back to block N", so we could say that the epoch argument will be the epoch of the top block of the chain after the rewind is complete.

aesedepece (approved):

PR #1867, opened 2021-02-23
feat(node): drop peers after using `initializePeers`

mariocao:


node/src/actors/json_rpc/json_rpc_methods.rs
```diff
@@ -1512,6 +1517,11 @@ mod mock_actix {
             // We cannot test methods which use `send`, so return an error
             Err(MailboxError::Closed)
         }
+
+        pub fn do_send<T: Message>(&self, _msg: T) {
+            // We cannot test methods which use `do_send`, so return an error
+            panic!("We cannot test actix `do_send` methods")
```
lrubiorod, line 1523: Why this panic?

tmpolaczyk: This is the `mock_actix`, which is used to test json rpc methods. There are no tests that use `do_send`, but it is needed to compile the tests.

lrubiorod (approved):

PR #1868, opened 2021-02-23
WIP-0010 Recovery Script

aesedepece:
To be used like this on a plain witnet-rust node:
```console
./docker/witnet-rust/wip0010-recovery.sh <path_to_witnet_binary> <path_to_witnet.toml_config_file>
```

To build a docker image that automatically runs the recovery script:
```console
docker buildx build --load -t witnet-recovery docker/witnet-rust/
docker stop witnet_node
docker rm witnet_node
docker run --name witnet node <bindings and ports go here> -dit witnet-recovery
```


docker/witnet-rust/wip0010-recovery.sh
```diff
@@ -0,0 +1,104 @@
...
+if $WITNET_BINARY node blockchain --epoch 248839 --limit 2 2>&1 | grep -q '#248921 had digest 7556670d'; then
+  log "The local witnet node at $HOST seems to be on the leading chain. No recovery action is needed"
+  exit 0
+fi
+
+echo "$RECOVERY_BANNER"
+
+# TODO: replace known_peers in witnet.toml
+# TODO: clear peers and buckets
+# TODO: rewind
```
aesedepece, line 104: @witnet/devs I will appreciate input from people who already got their nodes fixed on what specific sequence of CLI commands will work best for this.

lrubiorod: In my case, update the witnet.toml with the new IPs, restart the node, execute the rewind cli command and execute the initializePeers cli command to delete all the peers and only inject the updated ones

tmpolaczyk: Instead of restarting, can we use the `addPeers` command?

aesedepece: That, I wonder.

drcpu-github: Commenting because I did use the addPeers command (though I did use extra steps too).

1) I enabled UFW to disable all outgoing traffic.
2) Used `clearPeers` to make sure all peers were gone. The UFW is necessary here to make sure you have not connected and started syncing with forked peers yet.
3) Use `addPeers` with a set of known peers on the correct branch (I did not modify witnet.toml).
4) Wait until the node is properly connected to good peers and synchronizing, after which I disabled UFW again.
5) Profit.

tmpolaczyk: Then the script can do:

1. `rewind` - this will set the outbound limit to 0, so we can be sure that all the peers are gone
2. `clearPeers` - to remove the old forked peers
3. `addPeers` - to add the new peers without having to edit the witnet.toml

mariocao: sounds good just because the `rewind` sets the outbound limit to 0 :)

aesedepece: That's what it does now :)

docker/witnet-rust/Dockerfile
```diff
@@ -3,7 +3,8 @@ FROM --platform=$TARGETPLATFORM ubuntu:focal
 # Install basic environment dependencies
 RUN apt-get update && apt-get install -y --no-install-recommends \
     ca-certificates \
-    curl
+    curl \
+    netcat
```
tmpolaczyk, line 7: I don't see `netcat` being used in any script

tmpolaczyk: Oh, I found it, it's used as `nc` to check if the node JSON-RPC is working.

docker/witnet-rust/wip0010-recovery.sh
```diff
@@ -0,0 +1,153 @@
...
+  else
+    log "ERROR: Failed to connect to local witnet node at $HOST"
+    log "Retrying in $TIME_TO_NEXT_RETRY seconds"
+    sleep "$TIME_TO_NEXT_RETRY"
+    TIME_TO_NEXT_RETRY=$(( 2 * TIME_TO_NEXT_RETRY ))
+  fi
+do true; done
+
+# Check whether the local witnet node is below WIP-0010 "common checkpoint" (#248839)
+if [[ "$($WITNET_BINARY node blockchain --epoch 248839 --limit 1 2>&1 | wc -l)" == "5" ]]; then
```
tmpolaczyk, line 107: This seems a bit hacky, wouldn't it be better to grep for `block for epoch`?

Output if there is a block above that checkpoint:

```
Loading config from: witnet.toml
Setting log level to: ERROR, source: Env
block for epoch #248839 had digest 47eb3069ad98b81c89bd0691f89885387d51505ed3f45ba2b0a240582bb75b0c
```

Output if there are no blocks above that checkpoint:
```
Loading config from: witnet.toml
Setting log level to: ERROR, source: Env
```

docker/witnet-rust/wip0010-recovery.sh
```diff
@@ -0,0 +1,153 @@
...
+  exit 1
+fi
+if [ ! -f "$WITNET_CONFIG_FILE" ]; then
+  log "ERROR: The provided witnet configuration file (second argument to this script) is not a valid configuration file: $WITNET_CONFIG_FILE"
+  exit 2
+fi
+
+# Read configuration (e.g. node server address) from config file
+log "Using configuration file at $WITNET_CONFIG_FILE"
+HOST=$(grep "server_addr" "$WITNET_CONFIG_FILE" | sed -En "s/server_addr = \"(.*)\"/\1/p" | sed -En "s/0.0.0.0/127.0.0.1/p" )
```
tmpolaczyk, line 87: This evaluates to empty string when the `witnet.toml` has a line like

```
server_addr = "127.0.0.1:21337"
```

The first `sed` works, but the second one doesn't.

```
$ echo 'server_addr = "127.0.0.1:21337"' | sed -En "s/server_addr = \"(.*)\"/\1/p"
127.0.0.1
$ echo 'server_addr = "127.0.0.1:21337"' | sed -En "s/server_addr = \"(.*)\"/\1/p" | sed -En "s/0.0.0.0/127.0.0.1/p"
```

docker/witnet-rust/wip0010-recovery.sh
```diff
@@ -0,0 +1,153 @@
...
+  START=$1
+  NOW=$2
+  PROGRESS=$3
+  if [ "$PROGRESS" == "00" ]; then
+    echo "will be shown as synchronization moves forward..."
+  else
+    ELAPSED=$(( NOW - START ))
+    SPEED=$((PROGRESS * 1000 / ELAPSED))
+    REMAINING_PROGRESS=$(( 10000 - PROGRESS ))
+    REMAINING_TIME=$((REMAINING_PROGRESS / (SPEED / 1000) ))
```
tmpolaczyk, line 64: This line seems to cause some minor problems at the beginning:

```
[WIP0010_RECOVERY] Successfully added healthy peers
[WIP0010_RECOVERY] Still rewinding and synchronizing. Progress: 0.00%. ETA: will be shown as synchronization moves forward...
./wip0010-recovery.sh: línea 64: REMAINING_PROGRESS / (SPEED / 1000) : división por 0 (el elemento de error es "(SPEED / 1000) ")
[WIP0010_RECOVERY] Still rewinding and synchronizing. Progress: . ETA: 
./wip0010-recovery.sh: línea 64: REMAINING_PROGRESS / (SPEED / 1000) : división por 0 (el elemento de error es "(SPEED / 1000) ")
[WIP0010_RECOVERY] Still rewinding and synchronizing. Progress: . ETA: 
```

tmpolaczyk: Actually, I think this only happens if the rewind fails and the chain manager dies. Then `nodeStats` returns

    Error: ServerError { code: -32603, message: "MailboxError(Mailbox has closed)" }

docker/witnet-rust/Dockerfile
aesedepece: It's called `nc`

docker/witnet-rust/wip0010-recovery.sh
aesedepece: Hahahaha big F, it should be like this:
```console
echo 'server_addr = "4.0.0.0:21337"' | sed -En "s/server_addr = \"(.*)\"/\1/p" | sed -E "s/0\.0\.0.\0/127.0.0.1/g"
```

docker/witnet-rust/wip0010-recovery.sh
tmpolaczyk: So this was a consequence of #1870, we can assume that the rewind will not fail.

tmpolaczyk:
Tested the rewind feature and the script works fine but there was a problem later. After reaching the rewind target (248839/248839), the node tries to find outbound peers to connect to. I have set `outbound_limit = 6` to make this easier. The first set of 6 peers results in lack of consensus and all the peers are disconnected. After that, 3 other peers are tried but they all fail because of a handshake timeout. Then the node is stuck with 0 inbounds, 0 outbound, and 0 known peers. Not sure what happened to the other peers.

By reviewing the logs it looks like only the first 3 addresses were added in the original `addPeers` command. I tried to run the same command manually again and it worked. The problem seems to be the pattern `command | grep`, which kills the command as soon as `grep` finds one match.

docker/witnet-rust/wip0010-recovery.sh
```diff
@@ -0,0 +1,156 @@
...
+# Rewind local chain back to the WIP-0010 "common checkpoint" (#248839)
+log "Triggering rewind of local block chain back to epoch #248839"
+$WITNET_BINARY node rewind --epoch 248839 &>/dev/null
+REWIND_START=$(date +%s)
+
+# Flush existing peers and inject new peers in runtime
+$WITNET_BINARY node clearPeers 2>&1 | grep -q "Successful" &&
+log "Successfully cleared existing peers from buckets" ||
+log "ERROR: Failed to clear existing peers from buckets"
+printf "%b" "$KNOWN_PEERS" | sed -En "s/\s*\"(.*)\"\,/\1/p" | "$WITNET_BINARY" node addPeers 2>&1 | grep -q "Successful" &&
```
tmpolaczyk, line 139: This doesn't work because `addPeers` processes the input line by line, so when the first line is "Successful", grep finds a match and kills the process. The result is that only the first 2 or 3 addresses are added.

Solution: if `$KNOWN_PEERS` was a single line, all the addresses from that line would be added at once.
Solution 2: grep should have a flag to keep searching after the first match

aesedepece: I made it totally best-effort, but yeah, it's totally worth it putting it in a single line. Is some separator needed? Whitespace? Comma?

tmpolaczyk: You can use any separator except `.` and `:`, use your favorite format.

docker/witnet-rust/wip0010-recovery.sh
```diff
@@ -0,0 +1,156 @@
...
+    echo "will be shown as synchronization moves forward..."
+  else
+    ELAPSED=$(( NOW - START ))
+    SPEED=$((PROGRESS * 1000 / ELAPSED))
+    if [ "$SPEED" == "0" ]; then
+        SPEED=1
+    fi
+    REMAINING_PROGRESS=$(( 10000 - PROGRESS ))
+    REMAINING_TIME=$((REMAINING_PROGRESS / SPEED / 1000))
+    echo $(( REMAINING_TIME / 60 )) minutes $((REMAINING_TIME % 60)) seconds
```
tmpolaczyk, line 68: Now it says `ETA: 0 minutes 0 seconds`, let's just revert it to how it was before :see_no_evil:

docker/witnet-rust/wip0010-recovery.sh
aesedepece: Thanks! Done!

docker/witnet-rust/wip0010-recovery.sh
aesedepece: I think I was dividing where I should be multiplying.

aesedepece: `[WIP0010_RECOVERY] Still rewinding and synchronizing. Progress: 6.25%. ETA: 115 minutes 18 seconds`

tmpolaczyk: Fixed in f5dcf21

tmpolaczyk (approved):

mariocao (approved):

PR #1871, opened 2021-02-26
feat(node): change minimum superblock committee size

### LINKED ISSUE #1872: Modify superblock minimum committee size
As mentioned in https://github.com/witnet/WIPs/issues/45, modify the superblock minimum committee size from `1`  to `50`.

tmpolaczyk:
Allow to change the minimum superblock committee size, because the current minimum of 1 can lead to forks.

The new minimum and the activation date of this feature need to be discussed.

Close #1872

### DIFF
--- a/node/src/actors/chain_manager/mod.rs
+++ b/node/src/actors/chain_manager/mod.rs
@@ -1356,6 +1356,14 @@ impl ChainManager {
                     // the list itself is ordered by decreasing reputation
                     let ars_identities = ARSIdentities::new(ars_members);
 
+                    // After the second hard fork, the superblock committee size must be at least 50
+                    let min_committee_size = if after_second_hard_fork(block_epoch, get_environment()) {
+                        50
+                    } else {
+                        // Before that hard fork, the minimum was 1 identity
+                        1
+                    };
+
                     // Committee size should decrease if sufficient epochs have elapsed since last confirmed superblock
                     let committee_size = current_committee_size_requirement(
                         consensus_constants.superblock_signing_committee_size,
@@ -1364,7 +1372,8 @@ impl ChainManager {
                         consensus_constants.superblock_committee_decreasing_step,
                         chain_info.highest_superblock_checkpoint.checkpoint,
                         superblock_index,
-                        last_superblock_signed_by_bootstrap
+                        last_superblock_signed_by_bootstrap,
+                        min_committee_size,
                     );
                     log::debug!("The current signing committee size is {}", committee_size);
 
@@ -2491,6 +2500,7 @@ fn last_superblock_signed_by_bootstrap(consensus_constants: &ConsensusConstants)
 }
 // Returns the committee size to be applied given the default committee size, decreasing period
 // and  step, last consolidated epoch and the current checkpoint
+#[allow(clippy::too_many_arguments)]
 fn current_committee_size_requirement(
     default_committee_size: u32,
     last_committee_size: u32,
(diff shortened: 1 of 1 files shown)


node/src/actors/chain_manager/mod.rs
```diff
@@ -1342,6 +1342,22 @@ impl ChainManager {
                     // the list itself is ordered by decreasing reputation
                     let ars_identities = ARSIdentities::new(ars_members);
 
+                    // TODO: decide hard fork date and minimum value
+                    // After the Nth hard fork, the minimum superblock committee size must be 1/3 of 100
+                    let superblock_hard_fork_index = 99999999;
```
aesedepece, line 1347: Why is this an index and not an epoch?

aesedepece, line 1347: We'd better put all activation epochs as constants in the same file.

node/src/actors/chain_manager/mod.rs
```diff
@@ -1342,6 +1342,22 @@ impl ChainManager {
...
+                    let min_committee_size = if superblock_index > superblock_hard_fork_index {
+                        // Note that this ignores the case when the ARS size is less that 100.
+                        // For example, if the ARS size is 10, then the minimum will be 100/3 = 33,
+                        // so all 10 identities are expected to vote, and 2/3 of that 10 identities
+                        // must be in consensus.
+                        // If we want to ask for 1/3 of 10 in that case, so 3 identities , we need to use
+                        // min(ars_identities.len(), consensus_constants.superblock_signing_committee_size) / 3
+                        consensus_constants.superblock_signing_committee_size / 3
```
aesedepece, line 1355: This leaves me wondering: Is there any evidence of the ARS size ever going that low? Is it even possible in a successfully bootstrapped network?

node/src/actors/chain_manager/mod.rs
```diff
@@ -2485,22 +2503,25 @@ fn current_committee_size_requirement(
...
     last_checkpoint_signed_by_bootstrap: u32,
+    min_committee_size: u32,
 ) -> u32 {
     assert!(last_consolidated_checkpoint <= current_checkpoint, "Something went wrong as the last consolidated checkpoint is bigger than our current checkpoint {} > {}", last_consolidated_checkpoint, current_checkpoint);
     // If the last consolidated superblock or the current checkpoint is below last_checkpoint_signed_by_bootstrap, return the default committee size
     if last_consolidated_checkpoint <= last_checkpoint_signed_by_bootstrap {
         default_committee_size
     } else if current_checkpoint - last_consolidated_checkpoint >= decreasing_period {
+        // Decrease committee size
+        // The minimum committee size is configurable, but it must be at least 1
```
aesedepece, line 2514: Wouldn't say it's "configurable". It is indeed from a code perspective, but this statement gives the false idea that each user is free to put whatever minimum size they want.

aesedepece (changes requested):

node/src/actors/chain_manager/mod.rs
drcpu-github: In a normal scenario, it should never go below 2000 as that is the number of epochs a node is considered active when mining a block? The only cases I can think of would be long-ish rollbacks (e.g. longer than 2000 epochs) due to consensus failures causing a large majority of the nodes to fail, but even that sounds somewhat unlikely.

tmpolaczyk:
The discussion in #1879 seems to indicate that this change is no longer needed as the reduction process will be deprecated.

aesedepece:
> The discussion in #1879 seems to indicate that this change is no longer needed as the reduction process will be deprecated.

That's only under the assumption that #1879 will be eventually adopted :stuck_out_tongue_winking_eye:

lrubiorod (approved):

lrubiorod:
image

tmpolaczyk:
@lrubiorod it's a sign that we shouldn't merge this

aesedepece (approved):
Looks perfect to me.

PR #1880, opened 2021-03-04
Rescue committee

lrubiorod:


data_structures/src/superblock.rs
```diff
@@ -319,6 +377,25 @@ impl SuperBlockState {
         }
     }
 
+    /// Since we do not check signatures here, a superblock vote is valid if the signing identity
```
clbartoli, line 380: I think the comment needs to be updated, at least i dont really understand what  is valid in this function

node/src/actors/chain_manager/mod.rs
```diff
@@ -218,6 +218,10 @@ pub struct ChainManager {
     max_reinserted_transactions: usize,
     /// Last received Beacons
     last_received_beacons: Vec<(SocketAddr, Option<LastBeacon>)>,
+    /// a
+    pending_superblock_inventory_msgs: Vec<AddItem>,
+    /// b
+    pending_superblock_notify_msgs: Vec<SuperBlockNotify>,
```
tmpolaczyk, line 224: This could be a single `Vec<(AddItem, SuperBlockNotify)>`, because we always insert there two messages at the same time, and we always process and clear both vectors at the same time.

data_structures/src/superblock.rs
```diff
@@ -345,34 +422,132 @@ impl SuperBlockState {
...
-                signing_committee_length,
-            ) {
-                // There is no consensus, but if the missing votes vote the same as the
-                // majority, there can be consensus
-                SuperBlockConsensus::Unknown
+            // N3
+            if let Some(rescue_superblock) = rescue_consensus {
+                // R1
+                if most_voted_superblock == self.current_superblock_beacon.hash_prev_block {
+                    SuperBlockConsensus::SameAsLocal(false)
```
tmpolaczyk, line 470: Add a comment explaining why is this false

data_structures/src/superblock.rs
```diff
@@ -37,7 +37,7 @@ pub enum AddSuperBlockVote {
 #[derive(Debug)]
 pub enum SuperBlockConsensus {
     /// The local superblock has the majority of votes, everything ok
-    SameAsLocal,
+    SameAsLocal(bool),
```
tmpolaczyk, line 40: What is the purpose of this bool?

data_structures/src/superblock.rs
```diff
@@ -345,34 +422,132 @@ impl SuperBlockState {
...
+                )
+            };
+
+        matches!(
+            (
+                two_thirds_consensus_normal,
+                two_thirds_consensus_with_missing_votes,
+            ),
+            (false, true)
+        )
```
tmpolaczyk, line 548: Clippy should detect this and suggest:

```suggestion
        !two_thirds_consensus_normal && two_thirds_consensus_with_missing_votes
```

node/src/actors/chain_manager/actor.rs
```diff
@@ -252,6 +253,12 @@ impl ChainManager {
                     }
                 }
 
+                // TODO: use conditional hard fork
+                let rescue_committee = RESCUE_COMMITTEE
```
tmpolaczyk, line 257: This should probably be

```suggestion
                let rescue_committee = get_rescue_committee(get_environment());
```

Otherwise it's not possible to start the node in testnet mode because of "Malformed signing committee"

data_structures/src/superblock.rs
```diff
@@ -345,34 +423,131 @@ impl SuperBlockState {
...
-                most_voted_num_votes + num_missing_votes,
-                signing_committee_length,
-            ) {
-                // There is no consensus, but if the missing votes vote the same as the
-                // majority, there can be consensus
-                SuperBlockConsensus::Unknown
+            // N3
+            if let Some(rescue_superblock) = rescue_consensus {
+                // R1
+                if most_voted_superblock == self.current_superblock_beacon.hash_prev_block {
```
tmpolaczyk, line 470:
```suggestion
                if rescue_superblock == self.current_superblock_beacon.hash_prev_block {
```

data_structures/src/superblock.rs
```diff
@@ -445,6 +620,16 @@ impl SuperBlockState {
             }
         }
 
+        let rescue_old_votes = self.rescue_votes_mempool.clear_and_remove_votes();
```
tmpolaczyk, line 623: This returns the votes from `self.rescue_votes_mempool.received_votes` but we never insert any votes there.

Currently all the votes are inserted into `self.votes_mempool.received_votes`, so we could move this add this check to the other loop, `for sbv in old_votes` something like:

    if is_rescue_valid() { insert_rescue_vote() } else if is_valid() { insert_vote() }

And remove the `for sbv in rescue_old_votes` loop. But note that we still need to call `self.rescue_votes_mempool.clear_and_remove_votes()` to clear old votes.

lrubiorod: > This returns the votes from `self.rescue_votes_mempool.received_votes` but we never insert any votes there.

The rescue votes are inserted using the function "insert_rescue_vote" that is called in "add_vote"

tmpolaczyk: `insert_rescue_vote` does not modify `self.rescue_votes_mempool.received_votes`

lrubiorod: ahhh true!!

node/src/actors/chain_manager/handlers.rs
```diff
@@ -89,6 +91,11 @@ impl Handler<EpochNotification<EveryEpochPayload>> for ChainManager {
         let current_epoch = msg.checkpoint;
         self.current_epoch = Some(current_epoch);
 
+        // Set the right Rescue Committee according to the Epoch
+        self.chain_state
+            .superblock_state
+            .set_rescue_committe(get_rescue_committee(current_epoch, get_environment()));
```
tmpolaczyk, line 97: There may be a race condition here. If we receive some votes before setting the rescue committee they will be ignored.

lrubiorod: The consolidation of the superblock is in epochs finished in 0, and the first epoch to sign a superblockvote is in epochs finished in 1, so if the change of one committee to another is in an epoch finished in 0, it should not be a problem

tmpolaczyk: Maybe this could be fixed by setting the committee when the chain manager starts (just like it was being set before), and also setting it here.

tmpolaczyk: I was thinking about the case when the node starts and receives some votes before the first epoch notification.

lrubiorod: The problem here is that we obtain the epoch information in the EpochNotification handler

lrubiorod:
This PR is currently closed by some issues with the bridge

PR #1881, opened 2021-03-10
feat(superblock): improve randomness in committee selection

### LINKED ISSUE #1467: Non-uniform distribution when selecting superblock committee
The algorithm to randomly select a subset of the ARS to sign superblocks does not follow a uniform distribution:

https://github.com/witnet/witnet-rust/blob/fe5411af6b80f3e360dc5216518e7493a3e7c876/data_structures/src/superblock.rs#L434-L440

Here `first` should be a uniform value between `0` and `ars_identities.len() - 1` that depends on `current_superblock_index` and `superblock_hash`.

The current implementation is wrong because `u32::from(*superblock_hash.as_ref().get(0).unwrap())` will be a value between 0 and 255, and `u32::from(*index_hash.as_ref().get(0).unwrap())` will also be a value between 0 and 255. Adding this two values together will result in a value between 0 and 510, but following a triangular distribution so values closer to 255 will be much more likely than values close to 0 or 510.

Finally, this value between 0 and 510 is mapped to the range between 0 and ars_identities.len() - 1 using the modulo operator. This introduces modulo bias when `ars_identitites.len()` is smaller than 510 and it doesn't work when `ars_identities.len()` is greater than 510.

For reference, this is an example of a function that generates a uniform value:

https://github.com/witnet/bls-signatures-rs/blob/bec995125502465de546f91c238637290a2e92e8/src/bn256/mod.rs#L97

lrubiorod:
There are 2 main changes:

1. Selection of the first member:
- Old: We take the first byte of the superblock index hash, and the first byte of the superblock hash and we add them (It is not a uniform distribution)
- New: We use those 2 bytes in the next function, byte_a * K + byte_b, where K = 256

2. Selection of the rest members:
- Old: the size of ARS is divided by the size of the committee to obtain the steps between the elements, to achieve a distributed committee. (The problem here, is that there are only a few possible committee combinations, in case of 2000 identities and a committee of 100, only there will be 20 possibles committees)
- New: we use the same logic but adding a random offset to every element, this randomness comes from the bytes of the superblock index hash

Close #1467

### DIFF
--- a/data_structures/src/superblock.rs
+++ b/data_structures/src/superblock.rs
@@ -1,6 +1,10 @@
-use crate::chain::{
-    AltKeys, BlockHeader, Bn256PublicKey, CheckpointBeacon, Hash, Hashable, PublicKeyHash,
-    SuperBlock, SuperBlockVote,
+use crate::{
+    chain::{
+        AltKeys, BlockHeader, Bn256PublicKey, CheckpointBeacon, Epoch, Hash, Hashable,
+        PublicKeyHash, SuperBlock, SuperBlockVote,
+    },
+    get_environment,
+    mainnet_validations::after_second_hard_fork,
 };
 use std::{
     collections::{HashMap, HashSet},
@@ -393,6 +397,7 @@ impl SuperBlockState {
         last_block_in_previous_superblock: Hash,
         alt_keys: &AltKeys,
         sync_superblock: Option<SuperBlock>,
+        block_epoch: Epoch,
     ) -> SuperBlock {
         let key_leaves = hash_key_leaves(&ars_identities.get_rep_ordered_bn256_list(alt_keys));
 
@@ -408,6 +413,7 @@ impl SuperBlockState {
                 sb.signing_committee_length,
                 superblock_index,
                 self.current_superblock_beacon.hash_prev_block,
+                block_epoch,
             );
 
             sb
@@ -418,6 +424,7 @@ impl SuperBlockState {
                 signing_committee_size,
                 superblock_index,
                 self.current_superblock_beacon.hash_prev_block,
+                block_epoch,
             );
 
             mining_build_superblock(
@@ -499,6 +506,7 @@ pub fn calculate_superblock_signing_committee(
     signing_committee_size: u32,
     current_superblock_index: u32,
     superblock_hash: Hash,
+    block_epoch: Epoch,
 ) -> HashSet<PublicKeyHash> {
     // If the number of identities is lower than committee_size all the members of the ARS sign the superblock
     if ars_identities.len() < usize::try_from(signing_committee_size).unwrap() {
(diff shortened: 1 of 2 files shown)


aesedepece:
I'd appreciate a description on what the new algorithm looks like vs. the former. Every change we make is an opportunity to document an underdocumented part of the protocol.

data_structures/src/superblock.rs
```diff
@@ -544,6 +570,39 @@ fn magic_partition<T: Clone>(v: &[T], first: usize, size: usize) -> Vec<T> {
...
+    if first >= v.len() {
+        return hs_subset;
+    }
+    let each = v.len() / size;
+
+    let mut step_index = 0_usize;
+    let mut step = rand_distribution[step_index] as usize;
+    let mut a = first;
+    let mut b = (a + step) % v.len();
+    while hs_subset.len() < size {
```
tmpolaczyk, line 590: This could be an infinite loop if `v.len() < size`, but currently that's not possible because the size if checked before calling this function. But I guess we can add an `assert!(v.len() >= size)` to be safe.

data_structures/src/superblock.rs
```diff
@@ -544,6 +570,39 @@ fn magic_partition<T: Clone>(v: &[T], first: usize, size: usize) -> Vec<T> {
...
+    T: Clone + Eq + std::hash::Hash,
+{
+    let mut hs_subset = HashSet::default();
+    if first >= v.len() {
+        return hs_subset;
+    }
+    let each = v.len() / size;
+
+    let mut step_index = 0_usize;
+    let mut step = rand_distribution[step_index] as usize;
```
tmpolaczyk, line 587: Since we always access the `rand_distribution` variable sequentially, we can simplify the code by using iterators:

```suggestion
    let mut rand_gen = rand_distribution.iter().map(|x| usize::from(*x)).cycle();
    let mut step = rand_gen.next().unwrap();
```

This way we do not need the `step_index` variable. The unwrap is safe because `.cycle()` creates an infinite iterator that always returns something.

data_structures/src/superblock.rs
```diff
@@ -499,27 +506,46 @@ pub fn calculate_superblock_signing_committee(
...
+        if after_second_hard_fork(block_epoch, get_environment()) {
+            // Start counting the members of the subset from 'first_byte_index_hash'x256 + 'first_byte_sb_hash'
+            let mut first =
+                u64::from(first_byte_index_hash) * 256u64 + u64::from(first_byte_sb_hash);
+            // We need to choose a first member from all the potential ARS members
+            first %= ars_identities.len() as u64;
+
+            // Get the subset
+            magic_partition_2(
+                &ars_identities.ordered_identities.to_vec(),
```
tmpolaczyk, line 529: This `to_vec()` was not needed

```suggestion
                &ars_identities.ordered_identities,
```

data_structures/src/superblock.rs
tmpolaczyk: Although this change will make it slightly more difficult to rewrite this function in solidity, so not sure if it's worth it.

lrubiorod: If the logic is the same, we could keep it simple

clbartoli (approved):

data_structures/src/superblock.rs
```diff
@@ -499,27 +506,49 @@ pub fn calculate_superblock_signing_committee(
...
+        let second_byte_index_hash = *index_hash.as_ref().get(1).unwrap();
+
+        if after_second_hard_fork(block_epoch, get_environment()) {
+            // Start counting the members of the subset from:
+            // 'first_byte_sb_hash'x263 + 'first_byte_index_hash'x257 + 'second_byte_index_hash'
+            let mut first = u64::from(first_byte_sb_hash) * 263u64
+                + u64::from(first_byte_index_hash) * 257u64
+                + u64::from(second_byte_index_hash);
+            // We need to choose a first member from all the potential ARS members
+            first %= ars_identities.len() as u64;
```
tmpolaczyk, line 528: With the new algorithm the value of `first` is not very important. And taking into account the complexity of this logic, it would be nice to remove it if possible.

But if we remove it then the committees will not depend on the superblock hash and I assume that that's required. So I suggest to use `index_hash ^ superblock_hash` as the "rand_distribution", instead of `index_hash`. And hardcode `first` to `0`. What do you think?

lrubiorod: If the rand_distribution use bytes as a random offsets, the biggest offset would be +255, so if the distance between candidates in the initial distribution without randomness is bigger than that there would be some positions that never will be chosen to participate in the committee

tmpolaczyk (approved):

PR #1883, opened 2021-03-16
Merge second hard fork

### LINKED ISSUE #1467: Non-uniform distribution when selecting superblock committee

lrubiorod:
Those are the changes for the second hard fork:
 - ARS members have priority respect to non ARS in block candidate selection
 - Committee size will be only reduced until size 50 
 - Improved randomness selecting the signing committee
 - Tally can not fail (panic or error), it will return a RadError::Unknown
 - Minimum difficulty

Close #1467


data_structures/src/mainnet_validations.rs
```diff
@@ -13,6 +13,8 @@ const FIRST_EMERGENCY_COMMITTEE: [&str; 7] = [
 
 /// 22 January 2021 @ 09:00:00 UTC
 pub const FIRST_HARD_FORK: Epoch = 192000;
+/// 16 March 2021 @ 09:00:00 UTC
+pub const SECOND_HARD_FORK: Epoch = 293680;
```
tmpolaczyk, line 17: This date needs to be updated

lrubiorod:
One problem found:
- After a big rollback, bigger than an activity period, the ARS would be totally clean, and the only member would be the address that mine the last block. With the current changes (ARS priority), if the node has reputation (and there are a high probability that it happened), taking into account that ARS members has priority respect to the others, and the difficulty to mine 1/(ars_members = 1), this address would became in the only miner until there would be a request that update the ars members.

A proposal to solve that is to add a minimum difficulty to mine a block, to avoid that a only miner take all the rewards. And this minimum could be the initial_difficulty value that it was used in the bootstrapping period

data_structures/src/superblock.rs
```diff
@@ -572,6 +602,51 @@ fn magic_partition<T: Clone>(v: &[T], first: usize, size: usize) -> Vec<T> {
...
+// magic_partition(v, 3, 3, r), v=[0, 1, 2, 3, 4, 5], r=[1].
+// Will return elements at index 4, 0, 2.
+fn magic_partition_2<T>(v: &[T], first: usize, size: usize, rand_distribution: &[u8]) -> HashSet<T>
+where
+    T: Clone + Eq + std::hash::Hash,
+{
+    let mut hs_subset = HashSet::default();
+    if first >= v.len() {
+        return hs_subset;
+    }
```
aesedepece, line 615: Why returning an empty set in this case?

Is it just to comply with the function signature without introducing a `Result` because in theory it is impossible that `f >= v.len() == true`? Or is there any good reason for it? I'm trying to understand if this is a protocol thing or is only due to specifics of the Rust implementation.

data_structures/src/superblock.rs
```diff
@@ -572,6 +602,51 @@ fn magic_partition<T: Clone>(v: &[T], first: usize, size: usize) -> Vec<T> {
     v_subset
 }
 
+// Take size element out of v.len() starting with element at index first plus an offset:
```
aesedepece, line 605:
```suggestion
// Sample `size` elements uniformly out of a slice `v` starting with element at index `first`, then picking the rest with a pseudorandom offset.
```

data_structures/src/superblock.rs
tmpolaczyk: I believe it was used to handle the case of `v.len() == 0`, but now that should have been checked before calling this function. So this is probably dead code that can be removed.

lrubiorod: Yes, we can modify by an unreachable!()

tmpolaczyk (approved):

tmpolaczyk (approved):

tmpolaczyk (approved):

data_structures/src/data_request.rs
```diff
@@ -467,16 +506,34 @@ where
...
+        } else {
+            calculate_witness_reward_before_second_hard_fork(
+                commits_count,
+                reveals_count,
+                liars_count + non_reveals_count,
+                errors_count,
+                dr_output.witness_reward,
+                collateral,
+            )
+        };
```
mariocao, line 527: At some point in the future, this code can be refactored by calling a `calculate_witness_reward` function with the block_epoch as argument.

PR #1884, opened 2021-03-16
feat(validations): reuse initial difficulty when there are a few iden…

lrubiorod:
…tities

### DIFF
--- a/validations/src/tests.rs
+++ b/validations/src/tests.rs
@@ -6647,7 +6647,7 @@ fn test_block_with_drpool_and_utxo_set<F: FnMut(&mut Block) -> bool>(
         b.block_sig = sign_tx(PRIV_KEY_1, &b.block_header);
     }
     let mut signatures_to_verify = vec![];
-    validate_block(
+    validate_block_with_minimum_difficulty(
         &b,
         current_epoch,
         vrf_input,
@@ -6655,6 +6655,7 @@ fn test_block_with_drpool_and_utxo_set<F: FnMut(&mut Block) -> bool>(
         &mut signatures_to_verify,
         &rep_eng,
         &consensus_constants,
+        2,
     )?;
     verify_signatures_test(signatures_to_verify)?;
     let mut signatures_to_verify = vec![];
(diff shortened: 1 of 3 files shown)


validations/src/validations.rs
```diff
@@ -1906,7 +1906,10 @@ pub fn calculate_randpoe_threshold(
 ) -> (Hash, f64) {
     let max = u64::max_value();
     let initial_difficulty = std::cmp::max(1, initial_difficulty);
-    let target = if block_epoch <= epochs_with_initial_difficulty {
+    let target = if block_epoch <= epochs_with_initial_difficulty
+        || (after_second_hard_fork(block_epoch, get_environment())
+            && total_identities < initial_difficulty)
+    {
         max / u64::from(initial_difficulty)
     } else if total_identities == 0 || replication_factor >= total_identities {
```
mariocao, line 1914: Maybe the second is not needed! :)

mariocao: Double-check what happens for `testnet` environments

lrubiorod:
In this PR we use the 'initial_difficulty' consensus constant as threshold to reactivate the initial difficulty again. But we should discuss if we want to use another parameters to the **threshold** and to the **difficulty**

What do you think @witnet/devs?

validations/src/validations.rs
lrubiorod: > Maybe the second branch is not needed! :)

After second hard fork is not needed but it is needed for the previous blocks

> Double-check what happens for `testnet` environments

In case of testnet, second_hard_fork returns false, so this branch does not exist

lrubiorod:
As a proposal, we could use a low threshold (around 100), but applying the initial difficulty (2000)

tmpolaczyk:
So the problem is that when there are 3 identities or less in the ARS, each of them has a 100% chance of mining and the network may enter a state where only those identities mine blocks.

The proposed solution is to set a limit to the mining probability, 1/2000 which is 0.05%. This assumes that the network will never have less than 2000 identities, because otherwise (for example if the ARS does gradually go down to 1000) there will be many missing blocks.

A simpler solution is to never allow 100% chance of mining, for example limit it to 50%. This solves the problem of getting stuck with an ARS size of 3, but it may result in excessive flooding of block candidates when the ARS size is much smaller than the total number of mining nodes.

It's hard to imagine a scenario where the ARS size goes down to 3 so this problem is purely theoretical, therefore I prefer a solution that has the minimum impact on the normal operation of the network. I would consider an ARS of size 1000 or even 500 identities to be expected, and I think that setting the threshold to 2000 it too aggressive. So I like the idea to set the threshold to 100 identities, and if the ARS size is smaller than that then the network goes into a fixed-difficulty mode. But I would prefer a smaller difficulty, for example 25, because I don't expect candidate flooding to be a problem in that scenario.

drcpu-github:
I wonder if one of the effects of the second hard fork where we heavily prioritize blocks produces by nodes that are already in the ARS would be that it overall becomes smaller (compared to now)? And if this happens, whether there is an infliction point where a small ARS actually perpetuates it shrinking further (which would mean this proposal is definitely needed).

I would be very careful with setting the difficulty too low because we did see in the testnets that this was very problematic to arrive at a consensus. Given that we know that there are significantly more nodes outside the ARS than inside I wonder if setting the difficulty lower than the threshold is a good idea.

aesedepece:
If the number of identities in the system is much bigger than the ARS size, the ARS doesn't necessarily shrinks because for every epoch in which there are no block candidate from the ARS, there's at least one from the outsiders.

Then every time an identity goes out of the ARS, a spot is left for some new outsider to take.

WIll this result in the ARS stabilizing around a smaller size? Most probably. Can we expect that number to be much smaller? I don't think so.

In either case, we get a stable ARS size, regardless of the specific number. This has of course some effect on short term rewards distribution, but as the identities can't simply force their stay in the ARS, the assumption is that, long term, their rewards share remains the same.

lrubiorod:
So... in the edge cases (several rollbacks), where the ARS is reduced to 1 or a few number of identities. What difficulty do you consider is enough to keep a safu ARS?

drcpu-github:
I don't think this has an easy answer to be honest:
1) If the rollbacks are an effect of forking because of late blocks or too many blocks, we probably still want a high enough difficulty to not increase the number of proposed blocks and worsen the situation.
2) If the rollbacks are instigated by a large number of nodes going down, we don't want too high of a difficulty to make sure we can still produce enough blocks (otherwise the network is effectively down).

tmpolaczyk:
So let's try 125?

validations/src/tests.rs
```diff
@@ -6675,6 +6676,64 @@ fn test_block_with_drpool_and_utxo_set<F: FnMut(&mut Block) -> bool>(
     Ok(())
 }
 
+#[allow(clippy::too_many_arguments)]
```
tmpolaczyk, line 6679: Add a TODO here because the idea is to eventually turn `minimum_difficulty` into a consensus constant, right? Because then we can remove this function.

tmpolaczyk (approved):

PR #1885, opened 2021-03-18
feat: respect min_consensus in data requests

tmpolaczyk:
Data requests have a `min_consensus_percentage` field that is used to ensure that the result of the data request has a minimum quality.

For example, a data request will 100 witnesses and `min_consensus_percentage = 75` needs 75 valid reveals. Valid means that they are not RadonError, and they all have the same type. The problem comes in the next step: after running the tally script some reveals can be marked as "liars" if they deviate too much from the consensus, according to the filter scripts. So for example using a mode filter with 75 values out of which 73 are all different and 2 are equal, the result would be a value that has only been reported by 2 witnesses out of 100, and the remaining 98 witnesses will be penalized.

This PR changes the validation logic to ensure that the `min_consensus_percentage` is applied to the number of honest witnesses. This means that with a `min_consensus_percentage = 51`, which is the minimum, it will be impossible to penalize more than 49% of the participants of a data request. If there is a high number of liars the data request resolves to "InsufficientConsensus". Note that this only applies to tally filters. The mode reducer will keep working as usual so it will be possible to have a result that is only reported by 2 witnesses, but the remaining 73 will not be penalized.

This is a breaking change, so it will be enabled after the second hard fork

### DIFF
--- a/data_structures/src/data_request.rs
+++ b/data_structures/src/data_request.rs
@@ -10,7 +10,7 @@ use crate::{
     },
     error::{DataRequestError, TransactionError},
     get_environment,
-    mainnet_validations::after_first_hard_fork,
+    mainnet_validations::{after_first_hard_fork, after_second_hard_fork},
     radon_report::{RadonReport, Stage, TypeLike},
     transaction::{CommitTransaction, DRTransaction, RevealTransaction, TallyTransaction},
 };
@@ -389,7 +389,7 @@ pub fn calculate_tally_change(
         + dr_output.commit_and_reveal_fee * (witnesses - commits_count)
 }
 
-pub fn calculate_witness_reward(
+pub fn calculate_witness_reward_before_second_hard_fork(
     commits_count: usize,
     reveals_count: usize,
     // Number of values that are out of consensus plus non-revealers
@@ -406,6 +406,34 @@ pub fn calculate_witness_reward(
         (collateral, 0)
     } else {
         let honests_count = (commits_count - liars_count - errors_count) as u64;
+        let liars_count = liars_count as u64;
+        let slashed_collateral_reward = collateral * liars_count / honests_count;
+        let slashed_collateral_remainder = (collateral * liars_count) % honests_count;
+
+        (
+            reward + collateral + slashed_collateral_reward,
+            slashed_collateral_remainder,
+        )
+    }
+}
+
+pub fn calculate_witness_reward(
+    commits_count: usize,
+    // Number of values that are out of consensus plus non-revealers
+    liars_count: usize,
+    // To calculate the reward, we consider errors_count as the number of errors that are
+    // out of consensus, it means, that they do not deserve any reward
+    errors_count: usize,
+    reward: u64,
+    collateral: u64,
+) -> (u64, u64) {
+    let honests_count = (commits_count - liars_count - errors_count) as u64;
+
+    if commits_count == 0 {
+        (0, 0)
+    } else if honests_count == 0 {
+        (collateral, 0)
+    } else {
         let liars_count = liars_count as u64;
         let slashed_collateral_reward = collateral * liars_count / honests_count;
         let slashed_collateral_remainder = (collateral * liars_count) % honests_count;
@@ -451,6 +479,7 @@ pub fn create_tally<RT, S: ::std::hash::BuildHasher>(
     committers: HashSet<PublicKeyHash, S>,
     collateral_minimum: u64,
     tally_bytes_on_encode_error: Vec<u8>,
+    block_epoch: Epoch,
 ) -> TallyTransaction
 where
     RT: TypeLike,
@@ -477,16 +506,34 @@ where
 
         // Collateral division rest goes for the miner
         let non_reveals_count = commits_count - reveals_count;
-        let (reward, _rest) = calculate_witness_reward(
-            commits_count,
-            reveals_count,
-            liars_count + non_reveals_count,
-            errors_count,
-            dr_output.witness_reward,
-            collateral,
-        );
+        let is_after_second_hard_fork = after_second_hard_fork(block_epoch, get_environment());
+        let (reward, _rest) = if is_after_second_hard_fork {
+            calculate_witness_reward(
+                commits_count,
+                liars_count + non_reveals_count,
+                errors_count,
+                dr_output.witness_reward,
+                collateral,
+            )
+        } else {
+            calculate_witness_reward_before_second_hard_fork(
+                commits_count,
+                reveals_count,
+                liars_count + non_reveals_count,
+                errors_count,
+                dr_output.witness_reward,
+                collateral,
+            )
+        };
+        // Check if we need to reward revealers
+        let any_honest_revealers = if is_after_second_hard_fork {
+            let honests_count = commits_count - liars_count - errors_count - non_reveals_count;
 
-        let mut outputs: Vec<ValueTransferOutput> = if reveals_count > 0 {
+            honests_count > 0
+        } else {
+            reveals_count > 0
+        };
+        let mut outputs: Vec<ValueTransferOutput> = if any_honest_revealers {
             revealers
                 .iter()
                 .zip(liars.iter().zip(errors.iter()))
@@ -528,7 +575,16 @@ where
                 })
                 .collect()
         } else {
-            // In case of no revealers, collateral returns to their owners
+            // In case of no honests, collateral returns to their owners
+
+            if is_after_second_hard_fork {
+                // After second hard fork, mark all the revealers as errors to avoid penalizing them:
+                // If honests_count == 0, all revealers are out of consensus errors
+                for revealer in revealers {
+                    error_committers.push(revealer);
+                }
+            }
+
             out_of_consensus
                 .iter()
                 .map(|&committer| ValueTransferOutput {
--- a/node/src/actors/rad_manager/handlers.rs
+++ b/node/src/actors/rad_manager/handlers.rs
@@ -1,11 +1,15 @@
 //! Message handlers for `RadManager`
 
 use actix::{Handler, ResponseFuture};
-use witnet_data_structures::radon_report::{RadonReport, ReportContext};
+use witnet_data_structures::{
+    get_environment,
+    mainnet_validations::after_second_hard_fork,
+    radon_report::{RadonReport, ReportContext},
+};
 use witnet_rad::{error::RadError, script::RadonScriptExecutionSettings, types::RadonTypes};
 use witnet_validations::validations::{
-    construct_report_from_clause_result, evaluate_tally_precondition_clause,
-    TallyPreconditionClauseResult,
+    construct_report_from_clause_result, evaluate_tally_postcondition_clause,
+    evaluate_tally_precondition_clause, TallyPreconditionClauseResult,
 };
 
 use crate::actors::messages::{ResolveRA, RunTally};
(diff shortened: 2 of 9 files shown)


drcpu-github:
> For example, a data request will 100 witnesses and `min_consensus_percentage = 75` needs 75 valid reveals. Valid means that they are not RadonError, and they all have the same type. The problem comes in the next step: after running the tally script some reveals can be marked as "liars" if they deviate too much from the consensus, according to the filter scripts. So for example using a mode filter with 75 values out of which 73 are all different and 2 are equal, the result would be a value that has only been reported by 2 witnesses out of 100, and the remaining 98 witnesses will be penalized.

Maybe I am misunderstanding this paragraph, but is this supposed to be a description of the current situtation? Because right now, with the mode operator, the witnesses that reveal a value that is not in consensus with the majority option (even though that could be only 5% of the witnesses) are not penalized. See [1] for an example where only 4/28 witnesses (there were 32 in total) reveal the same value.

[1] https://witnet.network/search/3e1e2a266e3a2226ff33240626eae43d4224ed7ce4669e6e509e0d931dde2e32

tmpolaczyk:
@drcpu-github that data request does not have any "tally filters" so it is not affected by this change. You can see the "bitcoin_last_hash.json" as an example that uses tally filters, it uses a mode filter and a mode reducer. Tally filters are used to penalize values that are out of consensus, so this PR proposes to limit the maximum number of penalized witnesses to mitigate possible attacks.

drcpu-github:
Okay, so that means I did misunderstand this, thanks. 😄

lrubiorod:
In this PR, we are tagging all revealers to an out of consensus error (it means, no reward, no slash) in case of a RadError::InsufficientConsensus, and I think we should do the same with the RadErrorUnknown produced by a panic in Tally creation

data_structures/src/data_request.rs
```diff
@@ -391,7 +391,6 @@ pub fn calculate_tally_change(
 
 pub fn calculate_witness_reward(
     commits_count: usize,
-    reveals_count: usize,
```
lrubiorod, line 394: remember not remove it, to be able to use it with the after_second_hard_fork branch

node/src/actors/rad_manager/handlers.rs
```diff
@@ -40,7 +40,7 @@ impl Handler<ResolveRA> for RadManager {
                     .collect();
 
             let clause_result = evaluate_tally_precondition_clause(retrieve_responses, 0.2, 1);
-
+            // TODO: add postcondition here?
```
lrubiorod, line 43: liars and errors are only used in tally, so I think it is no needed

validations/src/tests.rs
```diff
@@ -5312,56 +5504,714 @@ fn validate_calculate_witness_reward() {
...
+
+    // Include commits and reveals in DataRequestPool
+    include_commits(&mut dr_pool, commits_count, commits);
+    include_reveals(&mut dr_pool, reveals_count, reveals);
+
+    let dr_pkh = dr_public_key.pkh();
+    let change = 1000 * 4;
+
+    // Tally value: float(NAN)
+    //let tally_value = vec![0xF9, 0x7E, 0x00];
```
lrubiorod, line 5648: it is need?

validations/src/validations.rs
```diff
@@ -482,6 +482,63 @@ pub fn evaluate_tally_precondition_clause(
...
+            }
+
+            // TODO: mutate current report instead of creating new
+            RadonReport::from_result(
+                Err(RadError::InsufficientConsensus {
+                    achieved: achieved_consensus,
+                    required: minimum_consensus,
+                }),
+                &ReportContext::from_stage(Stage::Tally(metadata)),
+            )
```
lrubiorod, line 535: Could you use the report_from_error function here?

validations/src/validations.rs
```diff
@@ -1112,7 +1183,10 @@ pub fn validate_tally_transaction<'a>(
                 .into());
             }
         } else {
-            if reveals_count > 0 {
+            if honests_count > 0 {
+                // TODO: non-revealers can now be rewarded, but only if the tally result is
+                // InsufficientConsensus
```
lrubiorod, line 1188: or another RadError produced in tally construction

data_structures/src/data_request.rs
```diff
@@ -477,16 +505,37 @@ where
...
-            reveals_count,
-            liars_count + non_reveals_count,
-            errors_count,
-            dr_output.witness_reward,
-            collateral,
-        );
-
-        let mut outputs: Vec<ValueTransferOutput> = if reveals_count > 0 {
+        let _after_second_hard_fork = true;
+        let (reward, _rest) = if _after_second_hard_fork {
```
lrubiorod, line 509: why do not you use the real after_second_hard_fork function?

tmpolaczyk: Because I need the block epoch

validations/src/validations.rs
```diff
@@ -533,7 +584,9 @@ pub fn construct_report_from_clause_result(
         // Failed to evaluate the precondition clause. `RadonReport::from_result()?` is the last
         // chance for errors to be intercepted and used for consensus.
         Err(e) => {
-            RadonReport::from_result(Err(e), &ReportContext::from_stage(Stage::Tally(metadata)))
+            // If there is an error during the precondition, all revealers are set to error and liar.
+            // This is an error that is not penalized and not rewarded.
+            radon_report_from_error(e, reports_len)
```
lrubiorod, line 589: should we add after_second_hard_fork condition?

validations/src/validations.rs
```diff
@@ -934,16 +987,29 @@ fn create_expected_report(
...
             evaluate_tally_precondition_clause(results, non_error_min, commits_count);
-        construct_report_from_clause_result(clause_result, &tally, results_len)
+        let report = construct_report_from_clause_result(clause_result, &tally, results_len);
+        evaluate_tally_postcondition_clause(report, non_error_min, commits_count)
     }) {
         Ok(x) => x,
         Err(_e) => {
             // If there is a panic during tally creation: set tally result to RadError::Unknown
-            RadonReport::from_result(Err(RadError::Unknown), &ReportContext::default())
+            radon_report_from_error(RadError::Unknown, reveals.len())
```
lrubiorod, line 996: should we add after_second_hard_fork condition?

validations/src/validations.rs
```diff
@@ -1112,37 +1189,89 @@ pub fn validate_tally_transaction<'a>(
...
+            if _after_second_hard_fork {
+                if honests_count > 0 {
+                    // TODO: non-revealers can now be rewarded, but only if the tally result is
+                    // InsufficientConsensus
+
+                    // Make sure every rewarded address is a revealer
+                    if dr_state.info.reveals.get(&output.pkh).is_none() {
+                        return Err(TransactionError::RevealNotFound.into());
+                    }
+                    // Make sure every rewarded address passed the tally function, a.k.a. "is honest" / "is not a liar"
```
lrubiorod, line 1201: should we include this check conditions in a function? Something like check_revealer(dr_state, output) -> Result((), TransactionError).

On this way, the code could be more readable, because we only have to put:
check_revealer(..)?;

tmpolaczyk: I tried to extract all the logic from inside the for loop into a "validate_tally_output" function, but it needs 10 arguments so I'm not sure if it's worth it.

validations/src/validations.rs
```diff
@@ -438,6 +440,7 @@ pub fn evaluate_tally_precondition_clause(
                         unreachable!("Mode filter should always return a `RadonArray`");
                     }
                     Err(RadError::ModeTie { values, max_count }) => {
+                        // TODO: this calculates consensus as x / num_reveals, should it be x / num_commits?
```
lrubiorod, line 443: It should

validations/src/validations.rs
```diff
@@ -512,12 +563,12 @@ pub fn construct_report_from_clause_result(
                 RadonScriptExecutionSettings::all_but_partial_results(),
             ) {
                 Ok(x) => x,
-                Err(e) => RadonReport::from_result(
-                    Err(RadError::TallyExecution {
+                Err(e) => radon_report_from_error(
```
lrubiorod, line 566: It should use the after_second_hard_fork condition

validations/src/validations.rs
lrubiorod: my proposal was one function by each conditional branch, but I am also not sur if it's worth it

validations/src/tests.rs
```diff
@@ -4844,64 +5106,168 @@ fn tally_dishonest_reward() {
...
+        E,
+    );
+    let script = RADTally {
+        filters: vec![RADFilter {
+            op: RadonFilters::Mode as u32,
+            args: vec![],
+        }],
+        reducer: RadonReducers::Mode as u32,
+    };
+    let report = construct_report_from_clause_result(clause_result, &script, 3, E);
```
lrubiorod, line 5268: If you are using an E > AFTER_SECOND_HARD_FORK, you should include the postcondition clause. (Same in other parts below)

tmpolaczyk: Done

validations/src/validations.rs
```diff
@@ -1155,11 +1315,7 @@ pub fn validate_tally_transaction<'a>(
     }
 
     let expected_collateral_value = if commits_count > 0 {
-        (if dr_state.data_request.collateral == 0 {
-            collateral_minimum
-        } else {
-            dr_state.data_request.collateral
-        }) * u64::from(dr_state.data_request.witnesses)
+        collateral * u64::from(dr_state.data_request.witnesses)
```
lrubiorod, line 1317: why do you remove the minimum collateral?

tmpolaczyk: `collateral` is defined as

```rust
    let collateral = if dr_state.data_request.collateral == 0 {
        collateral_minimum
    } else {
        dr_state.data_request.collateral
    };
```

So this change should be equivalent.

validations/src/validations.rs
```diff
@@ -2281,9 +2437,11 @@ mod tests {
     use witnet_rad::types::{float::RadonFloat, integer::RadonInteger};
 
     use super::*;
+    use witnet_data_structures::mainnet_validations::SECOND_HARD_FORK;
```
lrubiorod, line 2440: fmt

validations/src/tests.rs
lrubiorod: ping

validations/src/tests.rs
```diff
@@ -5312,58 +5708,687 @@ fn validate_calculate_witness_reward() {
...
+        vec![24, 60],
+        vec![24, 60],
+        vec![216, 39, 129, 0],
+        vec![216, 39, 129, 0],
+    ];
+    let (pkhs, dr_pkh, dr_pointer, dr_pool) = generic_tally_test_stddev_dr(4, reveals, stddev_cbor);
+    let change = 1000 * 4;
+
+    // Tally value: float(NAN)
+    //let tally_value = vec![0xF9, 0x7E, 0x00];
```
lrubiorod, line 5935: same here

validations/src/tests.rs
```diff
@@ -5312,58 +5708,687 @@ fn validate_calculate_witness_reward() {
...
+    let stddev_cbor = vec![249, 0x39, 0]; // 0.625
+    let (pkhs, dr_pkh, dr_pointer, dr_pool) = generic_tally_test_stddev_dr(
+        4,
+        vec![vec![24, 60], vec![24, 60], vec![24, 61]],
+        stddev_cbor,
+    );
+    let change = 1000 * 4;
+
+    // Tally value: float(NAN)
+    //let tally_value = vec![0xF9, 0x7E, 0x00];
```
lrubiorod, line 5988: same here

lrubiorod (approved):

PR #1887, opened 2021-03-25
Non zero timelocks

### LINKED ISSUE #1886: Validations: forbid non-zero timelock in tally outputs
While working on #1885, we have noticed that the timelock field in tally outputs is not being validated. This means that a malicious miner could potentially "burn" the data request reward by setting a timelock far in the future.

So we need to check if this attack is possible, and fix it as soon as possible. The fix should be applied on top of #1885 which includes a test for this case: `tally_valid_1_reveal_5_commits_with_absurd_timelock`.

lrubiorod:
Close #1886

### DIFF
--- a/validations/src/tests.rs
+++ b/validations/src/tests.rs
@@ -4509,8 +4509,6 @@ fn tally_valid_1_reveal_5_commits_invalid_value() {
 }
 
 #[test]
-// FIXME(#1886): use this test
-#[ignore]
 fn tally_valid_1_reveal_5_commits_with_absurd_timelock() {
     let collateral = ONE_WIT;
     let (pkhs, dr_pkh, dr_pointer, dr_pool) = generic_tally_test(5, vec![vec![0]]);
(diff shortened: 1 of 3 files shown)


validations/src/validations.rs
```diff
@@ -1310,6 +1310,14 @@ pub fn validate_tally_transaction<'a>(
             }
             pkh_rewarded.insert(output.pkh);
         }
+
+        if output.time_lock != 0 {
```
tmpolaczyk, line 1314: Does this need an additional `if after_second_hard_fork`?

validations/src/tests.rs
```diff
@@ -4570,9 +4568,9 @@ fn tally_valid_1_reveal_5_commits_with_absurd_timelock() {
     assert_eq!(
         x.unwrap_err().downcast::<TransactionError>().unwrap(),
         //TransactionError::InvalidTimeLock {
```
tmpolaczyk, line 4570:
```suggestion
```

tmpolaczyk (approved):

PR #1888, opened 2021-03-25
feat(clippy): apply new clippy suggestions for rust 1.51

lrubiorod:


tmpolaczyk:
I would prefer to merge this after the second hard fork branch is merged into master, to avoid unnecessary merge conflicts.

data_structures/src/chain.rs
```diff
@@ -992,6 +992,7 @@ impl Into<ExtendedSK> for ExtendedSecretKey {
 #[protobuf_convert(pb = "witnet::Hash")]
 pub enum Hash {
     /// SHA-256 Hash
+    #[allow(clippy::upper_case_acronyms)]
```
tmpolaczyk, line 995: Maybe it's better to leave `-A clippy::upper-case-acronyms` instead of adding this allow clippy everywhere. Later we can decide if we want to fix it or not.

data_structures/src/chain.rs
```diff
@@ -1113,6 +1113,7 @@ impl fmt::Debug for Hash {
 }
 
 /// SHA-256 Hash
+#[allow(clippy::upper_case_acronyms)]
```
tmpolaczyk, line 1116:
```suggestion
```

rad/src/hash_functions/mod.rs
```diff
@@ -9,6 +9,7 @@ mod sha2;
 
 #[derive(Debug, PartialEq, TryFromPrimitive)]
 #[repr(u8)]
+#[allow(clippy::upper_case_acronyms)]
```
tmpolaczyk, line 12:
```suggestion
```

rad/src/operators/mod.rs
```diff
@@ -20,6 +20,7 @@ pub mod string;
 /// re-assigned without causing a non-backwards-compatible protocol upgrade.
 #[derive(Copy, Clone, Debug, PartialEq, Serialize, TryFromPrimitive)]
 #[repr(u8)]
+#[allow(clippy::upper_case_acronyms)]
```
tmpolaczyk, line 23:
```suggestion
```

tmpolaczyk (approved):
A few leftover `clippy allow` but otherwise great

PR #1890, opened 2021-04-05
Implement latest changes to superblock consensus algorithm

### LINKED ISSUE #1891: Simplify `calculate_first` function
Replace the previously proposed polynomial algorythm with concatenation plus hash.

tmpolaczyk:
* Change algorithm to calculate first (close #1891)
* Remove collision check from `magic_partition_2`, use modulo instead

### DIFF
--- a/data_structures/src/superblock.rs
+++ b/data_structures/src/superblock.rs
@@ -584,27 +584,21 @@ fn magic_partition_2<T>(v: &[T], first: usize, size: usize, rand_distribution: &
 where
     T: Clone,
 {
-    if first >= v.len() {
+    let v_len = v.len();
+    if first >= v_len {
         return vec![];
     }
 
     // Check that the required size is bigger than v
-    assert!(size <= v.len());
+    assert!(size <= v_len);
 
-    let each = v.len() / size;
-
-    let mut step_index = 0_usize;
-    let mut step = rand_distribution[step_index] as usize % each;
-    let mut a = first;
-    let mut b = (a + step) % v.len();
+    let each = v_len / size;
     let mut v_subset = vec![];
-    for _ in 0..size {
-        v_subset.push(v[b].clone());
-
-        step_index = (step_index + 1) % rand_distribution.len();
-        step = rand_distribution[step_index] as usize % each;
-        a = (a + each) % v.len();
-        b = (a + step) % v.len();
+    let rand_len = rand_distribution.len();
+    for i in 0..size {
+        let rand_offset = usize::from(rand_distribution[i % rand_len]) % each;
+        let position = (first + i * each + rand_offset) % v_len;
+        v_subset.push(v[position].clone());
     }
 
     v_subset
--- a/data_structures/src/superblock.rs
+++ b/data_structures/src/superblock.rs
@@ -511,45 +511,46 @@ pub fn calculate_superblock_signing_committee(
     // If the number of identities is lower than committee_size all the members of the ARS sign the superblock
     if ars_identities.len() < usize::try_from(signing_committee_size).unwrap() {
         ars_identities.identities
+    } else if after_second_hard_fork(block_epoch, get_environment()) {
+        // Start counting the members of the subset from:
+        // sha256(superblock_hash || superblock_index) % ars_identities.len()
+        let superblock_hash_and_index_bytes = [
+            superblock_hash.as_ref(),
+            current_superblock_index.to_be_bytes().as_ref(),
+        ]
+        .concat();
+        let superblock_hash_and_index_bytes_hashed =
+            Hash::from(calculate_sha256(&superblock_hash_and_index_bytes));
+        let first = superblock_hash_and_index_bytes_hashed
+            .div_mod(ars_identities.len() as u64)
+            .1 as usize;
+
+        // Get the subset
+        magic_partition_2(
+            &ars_identities.ordered_identities,
+            first,
+            signing_committee_size.try_into().unwrap(),
+            superblock_hash_and_index_bytes_hashed.as_ref(),
+        )
     } else {
         // Hash of the current_index, to avoid potential committee collisions
         let index_hash = Hash::from(calculate_sha256(&current_superblock_index.to_be_bytes()));
         let first_byte_sb_hash = *superblock_hash.as_ref().get(0).unwrap();
         let first_byte_index_hash = *index_hash.as_ref().get(0).unwrap();
-        let second_byte_index_hash = *index_hash.as_ref().get(1).unwrap();
 
-        if after_second_hard_fork(block_epoch, get_environment()) {
-            // Start counting the members of the subset from:
-            // 'first_byte_sb_hash'x263 + 'first_byte_index_hash'x257 + 'second_byte_index_hash'
-            let first = calculate_first_in_committee(
-                first_byte_sb_hash,
-                first_byte_index_hash,
-                second_byte_index_hash,
-                ars_identities.len(),
-            );
+        // Start counting the members of the subset from the superblock_hash plus superblock index hash
+        let mut first = u32::from(first_byte_sb_hash) + u32::from(first_byte_index_hash);
+        // We need to choose a first member from all the potential ARS members
+        first %= ars_identities.len() as u32;
 
-            // Get the subset
-            magic_partition_2(
-                &ars_identities.ordered_identities,
-                first,
-                signing_committee_size.try_into().unwrap(),
-                index_hash.as_ref(),
-            )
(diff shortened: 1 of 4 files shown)


data_structures/src/superblock.rs
```diff
@@ -516,17 +516,18 @@ pub fn calculate_superblock_signing_committee(
         let index_hash = Hash::from(calculate_sha256(&current_superblock_index.to_be_bytes()));
         let first_byte_sb_hash = *superblock_hash.as_ref().get(0).unwrap();
         let first_byte_index_hash = *index_hash.as_ref().get(0).unwrap();
```
lrubiorod, line 518: you can move those lines into else branch to avoid doing a hash when it is not needed

data_structures/src/superblock.rs
```diff
@@ -516,17 +516,18 @@ pub fn calculate_superblock_signing_committee(
...
+                superblock_hash.as_ref(),
+                current_superblock_index.to_be_bytes().as_ref(),
+            ]
+            .concat();
+            let first = Hash::from(calculate_sha256(&superblock_hash_and_index_bytes))
+                .div_mod(ars_identities.len() as u64)
+                .1 as usize;
 
             // Get the subset
             magic_partition_2(
```
lrubiorod, line 533: you should use the new hash instead of the index hash

lrubiorod:
Related to the last commit of no overlapping, did you check that the randomness is has not been reduced?

tmpolaczyk:
> Related to the last commit of no overlapping, did you check that the new randomness is enough?

I tried to write some meaningful tests because it obviously should be less random, but I wasn't able to detect any differences. So I think we need to do something like simulating a 20% network failure and seeing if it recovers in the expected time.

data_structures/src/superblock.rs
tmpolaczyk: Done!

data_structures/src/superblock.rs
```diff
@@ -592,33 +593,21 @@ where
...
 
     let mut step_index = 0_usize;
-    let mut step = rand_distribution[step_index] as usize;
+    let mut step = rand_distribution[step_index] as usize % each;
     let mut a = first;
     let mut b = (a + step) % v.len();
     while hs_subset.len() < size {
-        if !hs_subset.contains(&v[b]) {
-            hs_subset.insert(v[b].clone());
+        hs_subset.insert(v[b].clone());
```
aesedepece, line 600: It seems to me that, under the assumption that `v` contains no repetitions, this algorithm can no longer produce collisions because of `% each`, so probably the data structure doesn't need to be a `HashSet` anymore, and we can probably go with something lighter, right?

Furthermore, if that's true, I think we can simply iterate `[0 → size)` instead of trying to procedurally grow the `hs_subset` accumulator up to a target length, because we have guarantee that every iteration will insert exactly one entry into it.

aesedepece (changes requested):

aesedepece:
These changes have been described here:
https://hackmd.io/L3Q3P0uIQrqZRtUn069QoQ

lrubiorod (approved):

lrubiorod:
> These changes have been described here:
> https://hackmd.io/L3Q3P0uIQrqZRtUn069QoQ

I think it is good enough to create the WIP proposal

data_structures/src/superblock.rs
```diff
@@ -577,13 +580,12 @@ fn magic_partition<T: Clone>(v: &[T], first: usize, size: usize) -> Vec<T> {
 // Take size element out of v.len() starting with element at index first plus an offset:
 // magic_partition(v, 3, 3, r), v=[0, 1, 2, 3, 4, 5], r=[1].
 // Will return elements at index 4, 0, 2.
-fn magic_partition_2<T>(v: &[T], first: usize, size: usize, rand_distribution: &[u8]) -> HashSet<T>
+fn magic_partition_2<T>(v: &[T], first: usize, size: usize, rand_distribution: &[u8]) -> Vec<T>
 where
     T: Clone + Eq + std::hash::Hash,
```
tmpolaczyk, line 585:
```suggestion
    T: Clone,
```

tmpolaczyk:
I tested the case of subset size 50 out of ars size 10_000, and 35% of the nodes down. In theory, the expected number of superblocks to reach consensus once is 2.5716. These are the results:

* 2.5716 in theory 
* 2.6024 for the algorithm using hashset
* 2.6048 for the algorithm using modulo instead of hashset
* 2.5938 for a different algorithm using rand::choose_multiple (completely random)

You can reproduce this by going to this branch https://github.com/tmpolaczyk/witnet-rust/tree/consensus-histogram and running

    cargo run --release -p witnet_data_structures --example magic_partition_randomness

So I think the results are close enough and we can consider the new algorithm random enough.

aesedepece:
> I tested the case of subset size 50 out of ars size 10_000, and 35% of the nodes down. In theory, the expected number of superblocks to reach consensus once is 2.5716. These are the results:
> 
>     * 2.5716 in theory
> 
>     * 2.6024 for the algorithm using hashset
> 
>     * 2.6048 for the algorithm using modulo instead of hashset
> 
>     * 2.5938 for a different algorithm using rand::choose_multiple (completely random)
> 
> 
> You can reproduce this by going to this branch https://github.com/tmpolaczyk/witnet-rust/tree/consensus-histogram and running
> 
> ```
> cargo run --release -p witnet_data_structures --example magic_partition_randomness
> ```
> 
> So I think the results are close enough and we can consider the new algorithm random enough.

Thanks for the insights, that's been helpful!

data_structures/src/superblock.rs
```diff
@@ -592,31 +594,20 @@ where
...
+    for _ in 0..size {
+        v_subset.push(v[b].clone());
 
-    hs_subset
-}
+        step_index = (step_index + 1) % rand_distribution.len();
+        step = rand_distribution[step_index] as usize % each;
+        a = (a + each) % v.len();
+        b = (a + step) % v.len();
+    }
```
aesedepece, line 608: I think `step_index` is redundant if you leverage the `i` in `for i in 0..size` and use it as `rand_distribution[step_index % rand_distribution.len()]`.

aesedepece (changes requested):

data_structures/src/superblock.rs
aesedepece: `step` would simply become `rand_distribution[0_usize] as usize % each`.

aesedepece: The first assignment of `step` would simply become `rand_distribution[0_usize] as usize % each`

data_structures/src/superblock.rs
```diff
@@ -592,31 +594,20 @@ where
...
+    let mut v_subset = vec![];
+    for _ in 0..size {
+        v_subset.push(v[b].clone());
 
-    hs_subset
-}
+        step_index = (step_index + 1) % rand_distribution.len();
+        step = rand_distribution[step_index] as usize % each;
+        a = (a + each) % v.len();
+        b = (a + step) % v.len();
```
aesedepece, line 607: `a` and `b` also look redundant to me, as `b` can be probably derived as `((first + i * each) % v.len() + step) % v.len()`  before pushing into `v_subset`.

This is only of a coding style suggestion, both approaches do look right to me. It's only that I personally find the procedural version harder to read than the parametric—which comes with the nice feature of not requiring mutable variables.

aesedepece: Thinking about this made me realize that there's a mismatch between the code and my description in the WIP:
- Code: `((first + i * each) % v.len() + step) % v.len()`
- WIP: `(first + i * each + step) % v.len()`

aesedepece (changes requested):

data_structures/src/superblock.rs
aesedepece: Any advantage of the first over the second? My intuition is that both will work, but the second is simpler.

tmpolaczyk: > Thinking about this made me realize that there's a mismatch between the code and my description in the WIP:

I think it doesn't matter because `(a % N + b) % N == (a + b) % N`

tmpolaczyk: @aesedepece I don't want to break the current implementation while trying to simplify it, I think we should rewrite the whole function from scratch to prevent mistakes. So could you provide some pseudocode for the parametric version?

tmpolaczyk: Implemented the version without iterators because I guess it will be easier to port to solidity.

aesedepece (approved):

PR #1892, opened 2021-04-07
fix(node): clear commits and reveals during rollbacks

lrubiorod:


tmpolaczyk:
Maybe add a test that checks that all the commits and reveals are cleared after calling remove_unconfirmed_transactions?

tmpolaczyk (approved):

PR #1893, opened 2021-04-07
refactor(node): reuse the initial difficulty as minimum difficulty an…

lrubiorod:
…d rename it


tmpolaczyk:
Since this is just a refactor, can it wait until after the second hard fork is merged?

lrubiorod:
> Since this is just a refactor, can it wait until after the second hard fork is merged?
It's ok for me

config/src/config.rs
```diff
@@ -383,8 +383,8 @@ fn to_partial_consensus_constants(c: &ConsensusConstants) -> PartialConsensusCon
         collateral_age: Some(c.collateral_age),
         superblock_period: Some(c.superblock_period),
         extra_rounds: Some(c.extra_rounds),
-        initial_difficulty: Some(c.initial_difficulty),
-        epochs_with_initial_difficulty: Some(c.epochs_with_initial_difficulty),
+        minimum_difficulty: Some(c.minimum_difficulty),
+        epochs_with_minimum_difficulty: Some(c.epochs_with_minimum_difficulty),
```
tmpolaczyk, line 387: Changing the names of the fields here can break some `witnet.toml`s. Not the ones used by most users only `witnet.toml`s used for testing which have a "consensus_constants" section. So it would be nice to show a warning if we parse a toml with the old name of the field instead of the new one, or something.

lrubiorod: Due to this warning only be used for testing proposals, it could be done in a separated issue if it would be required

tmpolaczyk: Ok, remind me to update my tomls after this is merged.

tmpolaczyk (approved):

PR #1898, opened 2021-04-16
fix(node): clean outbounds in case of possible fork

lrubiorod:


tmpolaczyk (approved):
Untested but looks ok

PR #1901, opened 2021-04-20
refactor: remove unused field from process_data_request

### LINKED ISSUE #1796: Remove unused logic from data_request_pool after first hard fork
See #1774 

> The `time_lock` field is being used and commits are not being accepted before that timestamp, but after 4 commit rounds the data request finishes with "no commits" error, so it is impossible to use the `time_lock` field to schedule data requests for the future.
> 
> So this field can be removed in the next breaking release.
> 
> Also, we have a mapping of epoch to list of data requests that was designed to support this feature and now it is no longer needed:
> 
> https://github.com/witnet/witnet-rust/blob/415ca73a6b49716f12f9a59d008fe26470ddca56/data_structures/src/data_request.rs#L22-L23

I think that after the first hard fork it is safe to remove these fields, and also remove the check introduced in PR #1787 (the check in data_structures should be safe to remove, not sure about the check in validations).

tmpolaczyk:
Fix #1796 

Remove `epoch_constants` parameter from `process_data_request` function.

As mentioned in that issue, in theory the `data_requests_by_epoch` field can be also removed, but that would require database migrations. So let's ignore that for now.

### DIFF
--- a/data_structures/src/data_request.rs
+++ b/data_structures/src/data_request.rs
@@ -5,12 +5,12 @@ use serde::{Deserialize, Serialize};
 
 use crate::{
     chain::{
-        DataRequestInfo, DataRequestOutput, DataRequestStage, DataRequestState, Epoch,
-        EpochConstants, Hash, Hashable, PublicKeyHash, ValueTransferOutput,
+        DataRequestInfo, DataRequestOutput, DataRequestStage, DataRequestState, Epoch, Hash,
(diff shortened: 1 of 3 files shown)


tmpolaczyk:
Tested this using the rewind command, so we know that this does not break the validation of any old blocks.

lrubiorod (approved):

PR #1907, opened 2021-04-26
feat(bridge): add value limit to avoid relaying expensive DRs

### LINKED ISSUE #1904: [Bridge] Define and implement relaying thresholds (avoid funds drainage)
The centralized bridge will include thresholds in order to mitigate attack vectors and limit the funds that can be drained while forwarding Data Requests to the Witnet network (e.g. data request with high reward or collateral values).

These paremeters should be configurable in the `witnet_ethereum_bridge.toml`.

tmpolaczyk:
Close #1904

Add a limit to the total value of a data request: this is the total cost of creating the data request, including all the fees. The bridge node will treat any data requests with a higher value as errors, and report a generic error to solidity.

### DIFF
--- a/bridges/centralized-ethereum/src/actors/dr_sender.rs
+++ b/bridges/centralized-ethereum/src/actors/dr_sender.rs
@@ -129,13 +129,19 @@ impl DrSender {
                         // error and report error as result to ethereum.
                         log::error!("[{}] error: {}", dr_id, err);
                         let result = err.encode_cbor();
+                        // In this case there is no data request transaction, so the dr_tx_hash
+                        // field can be set to anything.
+                        // Except all zeros, because that hash is invalid.
+                        let dr_tx_hash =
+                            "0000000000000000000000000000000000000000000000000000000000000001"
+                                .parse()
+                                .unwrap();
 
-                        // TODO: review if dr_tx_hash = [0;32] makes sense
                         dr_reporter_addr
                             .send(DrReporterMsg {
                                 dr_id,
                                 dr_bytes,
-                                dr_tx_hash: Default::default(),
+                                dr_tx_hash,
                                 result,
                             })
                             .await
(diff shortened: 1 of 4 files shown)


bridges/centralized-ethereum/src/actors/dr_sender.rs
```diff
@@ -153,13 +157,26 @@ impl DrSender {
     }
 }
 
-fn deserialize_and_validate_dr_bytes(dr_bytes: &[u8]) -> Result<DataRequestOutput, String> {
+fn deserialize_and_validate_dr_bytes(
```
lrubiorod, line 160: should we have a test for this function?

bridges/centralized-ethereum/src/config.rs
```diff
@@ -29,6 +29,8 @@ pub struct Config {
     pub witnet_testnet: bool,
     /// Gas limits for some methods. If missing, let the client estimate
     pub gas_limits: Gas,
+    /// Max value that will be accepted in a data request
```
mariocao, line 32:
```suggestion
    /// Max value that will be accepted by the bridge in a data request
```

bridges/centralized-ethereum/src/actors/dr_sender.rs
tmpolaczyk: I think we should wait until the error codes are well defined, otherwise we will be rewriting all the tests. Then we can simply test if the error converted to cbor matches the expected error.

lrubiorod: PR #1909 is already included

tmpolaczyk: Added a few tests

bridges/centralized-ethereum/src/actors/dr_sender.rs
```diff
@@ -116,14 +124,11 @@ impl DrSender {
...
-                        log::error!("[{}] error deserializing data request: {}", dr_id, e);
-                        // TODO: decide on error result, currently using vec with one element [0]
-                        // This cannot be an empty vector because an empty vector means that the
-                        // request has not finished yet.
-                        let result = vec![0];
+                        log::error!("[{}] error: {}", dr_id, err);
+                        let result = err.encode_cbor();
 
                         // TODO: review if dr_tx_hash = [0;32] makes sense
                         dr_reporter_addr
```
clbartoli, line 134: When sending the DrReporterMsg the dr_tx_hash should not be 0

lrubiorod (approved):

PR #1908, opened 2021-04-27
feat(bridge): use dr_gas_price for reporting result

clbartoli:
This PR queries the gas price of a data request so that the same value is used when reporting the result

### DIFF
--- a/bridges/centralized-ethereum/src/actors/dr_reporter.rs
+++ b/bridges/centralized-ethereum/src/actors/dr_reporter.rs
@@ -18,6 +18,8 @@ pub struct DrReporter {
     pub wrb_contract: Option<Contract<web3::transports::Http>>,
     /// eth_account
     pub eth_account: H160,
+    /// report_result_limit
+    pub report_result_limit: Option<u64>,
 }
 
 /// Make actor from EthPoller
@@ -45,6 +47,7 @@ impl DrReporter {
         Ok(Self {
             wrb_contract: Some(wrb_contract),
             eth_account: config.eth_account,
+            report_result_limit: config.gas_limits.report_result,
         })
     }
 }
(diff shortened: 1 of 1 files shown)


tmpolaczyk:
This PR reads the gas price when there is a new request and stores it. But I think the gas price can be updated by the user, so it may change after we read it. A better alternative would be to read the gas price right before reporting the result, this way we always have the most up to date gas price and we don't need to store it.

bridges/centralized-ethereum/src/actors/dr_reporter.rs
```diff
@@ -71,46 +74,98 @@ impl Handler<DrReporterMsg> for DrReporter {
...
-                                dr_state: DrState::Finished,
-                                dr_tx_hash: Some(msg.dr_tx_hash),
-                            },
-                        ))
-                        .await
-                        .ok();
+
+            match dr_gas_price {
+                Ok(_) => {
+                    let gas_price = dr_gas_price;
```
tmpolaczyk, line 96:
```suggestion
                Ok(gas_price) => {
```

This way you can avoid the gas_price.unwrap() below

bridges/centralized-ethereum/src/actors/dr_reporter.rs
```diff
@@ -71,46 +74,98 @@ impl Handler<DrReporterMsg> for DrReporter {
...
+                    let receipt = wrb_contract
+                        .call_with_confirmations(
+                            "reportResult",
+                            (msg.dr_id, dr_hash, msg.result),
+                            eth_account,
+                            contract::Options::with(|opt| {
+                                opt.gas = gas_limit.map(Into::into);
```
tmpolaczyk, line 103: I expected the gas_limit to be the gas_price_limit, but that doesn't make sense because the gas is paid by the requester and can be as high as they want. But just to be sure, do we want to set a gas price limit?

bridges/centralized-ethereum/src/actors/dr_reporter.rs
```diff
@@ -71,46 +74,98 @@ impl Handler<DrReporterMsg> for DrReporter {
...
+            //                     dr_tx_hash: Some(msg.dr_tx_hash),
+            //                 },
+            //             ))
+            //             .await
+            //             .ok();
+            //     }
+            //     Err(e) => {
+            //         log::error!("reportResult{:?}: {:?}", params_str, e);
+            //     }
+            // }
```
tmpolaczyk, line 168: Remember to remove this commented out code

bridges/centralized-ethereum/src/actors/dr_reporter.rs
clbartoli: this gas limit is actually the report result limit set in the configuration. The idea is that the gas cost of reporting a result should not be higher than the reward set when posting a DR

tmpolaczyk (approved):

bridges/centralized-ethereum/src/actors/dr_reporter.rs
lrubiorod: ping

lrubiorod (approved):

PR #1909, opened 2021-04-27
feat(radon): include new radon errors

lrubiorod:


data_structures/src/radon_error.rs
```diff
@@ -52,11 +52,17 @@ pub enum RadonErrors {
     TallyExecution = 0x53,
     /// Invalid reveal serialization (malformed reveals are converted to this value)
     MalformedReveal = 0x60,
+    /// Invalid data request serialization (malformed data requests are converted to this value)
```
aesedepece, line 55: It is not true that malformed data requests are converted to this value.

```suggestion
    /// Requests that cannot be parsed must always get this error as their result.
    /// However, this is not a valid result in a Tally transaction, because invalid requests 
    /// are never included into blocks and therefore never get a Tally in response.
```

data_structures/src/radon_error.rs
```diff
@@ -52,11 +52,17 @@ pub enum RadonErrors {
...
+    MalformedRequest = 0x61,
```
aesedepece, line 56: I wonder whether malformation is the only reason for nonacceptance of a request. For example, an excessive number of witnesses could also be a reason for nonacceptance.

Generally speaking, it feels to me like we should be able to communicate the Ethereum client any reason why a request could be rejected by the Witnet network, either by specifying every case, or lumping all of them together into something generic.

data_structures/src/radon_error.rs
```diff
@@ -52,11 +52,17 @@ pub enum RadonErrors {
...
     // Access errors
     /// Tried to access a value from an index using an index that is out of bounds
     ArrayIndexOutOfBounds = 0x70,
     /// Tried to access a value from a map using a key that does not exist
     MapKeyNotFound = 0x71,
+    // This error it is not really an error. It is need to notify to an external contract that some
+    // of the parameters specified in the request, exceed the capabilities that a bridge node is willing to offer
+    /// Wit requirements for the data request exceed the capabilities that a bridge node is willing to offer
+    ExpensiveRequest = 0x80,
```
aesedepece, line 65: Aha — it looks like there are multiple reasons why a relayer may refuse to submit an otherwise perfectly valid request.

Maybe we can reflect on the thing that is common to all those reasons: the request is lacking enough incentives to compel the relayer to submit the transaction and put their wits at stake.

```suggestion
    /// The request is rejected on the grounds that it may cause the submitter, the
    /// committers or the revealers to spend or stake an amount of value that is
    /// unjustifiably high when compared with the reward they will be getting. 
    PoorIncentives = 0x80,
```

aesedepece: I'm now wondering if this is a RadonError that we may want to allow in commits, reveals and tallies.

Currently our take would be to refrain from committing, and let the request time out with a tally stating that it didn't get enough commits.

Of course, going down the route of allowing this error in commits, reveals and tallies would mean that all nodes would be required to enact the same criteria for considering whether a request "is worth it". That's just the opposite to our current take, in which we preferred to let each node decide how much they want to put their wits into risk.

data_structures/src/radon_error.rs
aesedepece: I think that errors that only belong in inter-client communication (ethereum<>relayer<>witnet-rust) but cannot exist in tallies should all belong in a new category, for which we can use a dedicated prefix, e.g. `0xE...` so they are not confused with proper witnessing errors.

aesedepece (changes requested):

data_structures/src/radon_error.rs
lrubiorod: I think this error is an inter-client communication error that you commented before, because it only affect when there are bridges nodes that refuse to include it in Witnet. In the Witnet protocol, the person who wants to include a request and the person who finally include it is the same, and there are the witnesses who decides to solve or not, but it exist a RadonError for that (InsufficientCommits)

tmpolaczyk: `PoorIncentives` is too broad, I prefer to split this error into the different cases:

* Too expensive for the relayer
* Collateral too high
* Reward too low
* Commit/reveal fees too low

lrubiorod:
I was reviewing the possible errors that a requester can have when includes a request and I added an error to handle:
- Errors from `validate_rad_request` function -> `BridgeMalformedRequest`
- Errors from `calculate_weight` -> Applying a limit of witnesses (100 to be safu) and `BridgeTooManyWitnesses`
- Errors from fee, output... overflow -> Applying a limit of wits to spend by requester and `BridgePoorIncentives`
- Errors from not synced, not money... -> Do not return an error and tries later

tmpolaczyk:
So if the new errors will be serialized as RadError::Unknown, then why include them? We could simply report a RadError::Unknown to ethereum and the result will be the same.

data_structures/src/radon_error.rs
lrubiorod: I think that collateral too high, reward too low and commit/reveal fees too low should be considered by witnesses, not by the bridge node operator, because to that problems, there already are a RadonError::InsufficientCommits in case that witnesses or miners decide not to solve that request

lrubiorod:
> So if the new errors will be serialized as RadError::Unknown, then why include them? We could simply report a RadError::Unknown to ethereum and the result will be the same.

It is only to create a RadError from a RadonError. But in our case, these errors are used to commit directly by the bridge node operator to the contract

rad/src/error.rs
```diff
@@ -448,6 +447,7 @@ impl RadError {
                     message: Some(message),
                 }
             }
+            _ => RadError::Unknown,
```
tmpolaczyk, line 450: I would prefer to be explicit to avoid problems if we add more errors in the future and forget to change it here...

```suggestion
            RadonErrors::Unknown | RadonErrors::BridgeMalformedRequest | etc => RadError::Unknown,
```

data_structures/src/radon_error.rs
```diff
@@ -57,6 +57,16 @@ pub enum RadonErrors {
     ArrayIndexOutOfBounds = 0x70,
     /// Tried to access a value from a map using a key that does not exist
     MapKeyNotFound = 0x71,
+    // Bridge errors: errors that only belong in inter-client communication
```
mariocao, line 60:
```suggestion
    /// Bridge errors: errors that only belong in inter-client communication
```

mariocao, line 60: There is an error that has not been considered here. It was mentioned in the issue https://github.com/witnet/witnet-ethereum-bridge/issues/156:

> data request result length exceeds a bridge-contract-defined limit (in bytes) --> limits funds drainage

Regarding the error described by @lrubiorod about DRs with too many witnesses:

> Errors from calculate_weight -> Applying a limit of witnesses (100 to be safu) and BridgeTooManyWitnesses

I would include this type of error as a `BridgeMalformedRequest` if the calculated weight exceeds the block limit weight. Otherwise, I do not see why we should set a maximum at 100 and not 125 (theoretical maximum with 400 bytes DRs) or even lower.

tmpolaczyk: Instead of looking at the number of witnesses we can calculate the weight, this way the code will be more future proof if we change the block limit or the weight algorithm.

rad/src/error.rs
```diff
@@ -707,7 +710,17 @@ mod tests {
...
+                    if x == RadonErrors::BridgeMalformedRequest
+                        || x == RadonErrors::BridgeTooManyWitnesses
+                        || x == RadonErrors::BridgePoorIncentives
+                    {
+                        // We skip these RadonErrors because there are not related with the protocol
+                        None
+                    } else {
+                        Some(x)
+                    }
+                }
```
aesedepece, line 723: For the fans of `match` xD
```suggestion
                Ok(RadonErrors::BridgeMalformedRequest )
                | Ok(RadonErrors::BridgeTooManyWitnesses)
                | Ok(RadonErrors::BridgePoorIncentives) {
                    // We skip these RadonErrors because there are not related with the protocol
                    None
                },
                Ok(x) => Some(x),
```

rad/src/error.rs
```diff
@@ -707,7 +710,17 @@ mod tests {
         // So just try all the possible `u8` values and return the successful ones
         (0u8..=255).filter_map(|error_code| {
             match RadonErrors::try_from_primitive(error_code) {
-                Ok(x) => Some(x),
+                Ok(x) => {
+                    if x == RadonErrors::BridgeMalformedRequest
+                        || x == RadonErrors::BridgeTooManyWitnesses
+                        || x == RadonErrors::BridgePoorIncentives
+                    {
+                        // We skip these RadonErrors because there are not related with the protocol
```
aesedepece, line 718:
```suggestion
                        // We skip these RadonErrors because they don't belong to the core witnessing protocol
```

aesedepece (changes requested):

data_structures/src/radon_error.rs
```diff
@@ -57,6 +57,16 @@ pub enum RadonErrors {
...
+    /// Requests that cannot be parsed must always get this error as their result.
+    /// However, this is not a valid result in a Tally transaction, because invalid requests
+    /// are never included into blocks and therefore never get a Tally in response.
+    BridgeMalformedRequest = 0xE0,
+    /// Witnesses exceeds 100
+    BridgeTooManyWitnesses = 0xE1,
+    /// The request is rejected on the grounds that it may cause the submitter to spend or stake an
+    /// amount of value that is unjustifiably high when compared with the reward they will be getting
+    BridgePoorIncentives = 0xE2,
```
aesedepece, line 69: We could use input from @tomsanch on whether _poor_ or _weak_ sounds better:
```suggestion
    BridgeWeakIncentives = 0xE2,
```

data_structures/src/radon_error.rs
```diff
@@ -57,6 +57,16 @@ pub enum RadonErrors {
...
+    // Bridge errors: errors that only belong in inter-client communication
+    /// Requests that cannot be parsed must always get this error as their result.
+    /// However, this is not a valid result in a Tally transaction, because invalid requests
+    /// are never included into blocks and therefore never get a Tally in response.
+    BridgeMalformedRequest = 0xE0,
+    /// The request is rejected on the grounds that it may cause the submitter to spend or stake an
+    /// amount of value that is unjustifiably high when compared with the reward they will be getting
+    BridgeWeakIncentives = 0xE1,
+    /// The request result length exceeds a bridge contract defined limit
+    BridgeOversizeResult = 0xE2,
```
mariocao, line 69:
```suggestion
    BridgeOversizedResult = 0xE2,
```

data_structures/src/radon_error.rs
tomsanch: I would go for PoorIncentives

aesedepece: Thanks @tomsanch!

data_structures/src/radon_error.rs
```diff
@@ -57,6 +57,16 @@ pub enum RadonErrors {
...
     /// Tried to access a value from a map using a key that does not exist
     MapKeyNotFound = 0x71,
+    // Bridge errors: errors that only belong in inter-client communication
+    /// Requests that cannot be parsed must always get this error as their result.
+    /// However, this is not a valid result in a Tally transaction, because invalid requests
+    /// are never included into blocks and therefore never get a Tally in response.
+    BridgeMalformedRequest = 0xE0,
+    /// The request is rejected on the grounds that it may cause the submitter to spend or stake an
+    /// amount of value that is unjustifiably high when compared with the reward they will be getting
+    BridgeWeakIncentives = 0xE1,
```
aesedepece, line 67: Tomsanch eventually went for `Poor` over `Weak` :shrug:

aesedepece (changes requested):

tmpolaczyk:
This may introduce a breaking change when parsing reveals, see this test:

```rust
// rad/src/types/mod.rs
#[test]
fn serial_iter_decode_bridge_radon_errors() {
    #[allow(clippy::trivially_copy_pass_by_ref, clippy::unnecessary_wraps)]
    fn malformed_reveal_fn(_: RadError, _: &[u8], _: &()) -> Option<RadonReport<RadonTypes>> {
        Some(RadonReport::from_result(
            Err(RadError::MalformedReveal),
            &ReportContext::default(),
        ))
    }

    let malformed_reveal =
        RadonTypes::RadonError(RadonError::try_from(RadError::MalformedReveal).unwrap());

    // One reveal with value RadonErrors(0xE0)
    // Before PR #1909 this is deserialized as RadError::MalformedReveal
    // After PR #1909 this is deserialized as RadError::Unknown
    let cbor_bytes: Vec<(&[u8], &())> = vec![(&[0xD8, 0x27, 0x81, 0x18, 0xE0], &())];
    let rad_decode_error_as_result: Vec<_> =
        serial_iter_decode(&mut cbor_bytes.into_iter(), malformed_reveal_fn)
            .into_iter()
            .map(|report| report.into_inner())
            .collect();
    assert_eq!(rad_decode_error_as_result, vec![malformed_reveal]);
}
```

Not sure if this can enable attacks or forks

lrubiorod:
> This may introduce a breaking change when parsing reveals, see this test:
> 
> ```rust
> // rad/src/types/mod.rs
> #[test]
> fn serial_iter_decode_bridge_radon_errors() {
>     #[allow(clippy::trivially_copy_pass_by_ref, clippy::unnecessary_wraps)]
>     fn malformed_reveal_fn(_: RadError, _: &[u8], _: &()) -> Option<RadonReport<RadonTypes>> {
>         Some(RadonReport::from_result(
>             Err(RadError::MalformedReveal),
>             &ReportContext::default(),
>         ))
>     }
> 
>     let malformed_reveal =
>         RadonTypes::RadonError(RadonError::try_from(RadError::MalformedReveal).unwrap());
> 
>     // One reveal with value RadonErrors(0xE0)
>     // Before PR #1909 this is deserialized as RadError::MalformedReveal
>     // After PR #1909 this is deserialized as RadError::Unknown
>     let cbor_bytes: Vec<(&[u8], &())> = vec![(&[0xD8, 0x27, 0x81, 0x18, 0xE0], &())];
>     let rad_decode_error_as_result: Vec<_> =
>         serial_iter_decode(&mut cbor_bytes.into_iter(), malformed_reveal_fn)
>             .into_iter()
>             .map(|report| report.into_inner())
>             .collect();
>     assert_eq!(rad_decode_error_as_result, vec![malformed_reveal]);
> }
> ```
> 
> Not sure if this can enable attacks or forks

Could be solved using RadError::MalformedReveal as a result for bridge_error serialization?

aesedepece (approved):

tmpolaczyk (approved):
This seems to work fine! Trying a local network with 1 node in master and 1 node in this pr, when there is a reveal with error code 0xE0 it is parsed as RadError::MalformedReveal. The only difference is that the old node shows a warning: "Could not decode reveal from RevealTransaction", which is expected.

PR #1910, opened 2021-04-29
fix(node): include second_rescue_committee

lrubiorod:
Only in case it would be necessary

Remaining tasks:
- [ ]  Decide identities for the rescue committee
- [ ]  Decide superblock index for activation


lrubiorod:
We achieve consensus!!

Issue #1911, opened 2021-04-29
Add minimum difficulty to data request eligibility


tmpolaczyk:
PR #1884 implements a minimum mining difficulty so that in extreme cases when the ars size is small, the probability of mining a block is not 100%.

However, data request eligibility does not have a minimum difficulty, so in that case the probability of being eligible for resolving a data request can be 100%. This results in extreme flooding on the network because every node with enough collateral will send a commit transaciton, and potential forks because commit transactions are expensive to validate and some nodes may not be able to keep up under that load.

tmpolaczyk:
Unresolved question: should we allow 100% eligibility?

Currently, the identities with most reputation may have a 100% chance of being eligible for some data requests. The formula is something like `target = reputation / total_reputation * threshold_factor`. `target` will be a number between 0.0 and 1.0 that represents the probability of creating a valid VRF proof. The threshold_factor depends on the number of witnesses of the data request, and it ensures that the average number of eligible identities is high enough. So adding a minimum difficulty here (for example 10%) will reduce the average number of commits per data request unless we also adjust the threshold_factor calculation, but that will require an analysis of the algorithm to ensure we don't break anything. But we may want to prevent some identities from having 100% eligibility to increase the decentralization of the network, for example. I think it's not worth it to make complex changes to this algorithm, but we have plenty of time until a "third hard fork" to fix this issue, so any suggestions welcome.

Proposal: only apply minimum difficulty to identities that are not in the ARS

This would solve the spamming problem we have seen when the ARS size is suddenly reduced, while all the other cases will work the same way as before. The only danger would be setting the minimum too low, because that may result in data requests having too few commits, and that will mean that the ARS will not get updated. I guess the value should make sure that on a data request with 125 witnesses there will be at least 250 commits out of 2000 inactive identities, so 12.5%?

Note that with a minimum of 100% the mainnet has eventually recovered, so not setting a minimum is also a possibility.

drcpu-github:
I want to voice two concerns:

1. 100% eligibility should probably not be possible as, in my opinion, that paves the way for manipulating data requests. Imagine if you have multiple high-reputation nodes and a set of them can always commit to answering a certain type of data request. That seems to make it way too easy to manipulate them.

[...]

PR #1912, opened 2021-04-29
feat: limit data request eligibility to 10%

tmpolaczyk:
A temporary solution to issue #1911 which should improve the stability of the mainnet


tmpolaczyk:
Since the mainnet has recovered, I'm closing this pull request so we can discuss this change in more depth and execute it in the third hard fork if needed

PR #1915, opened 2021-05-05
feat(bridge): add maximum data request's result size

lrubiorod:

### DIFF
--- a/witnet_centralized_ethereum_bridge.toml
+++ b/witnet_centralized_ethereum_bridge.toml
@@ -14,6 +14,9 @@ eth_new_dr_polling_rate_ms = 1_000
 wit_tally_polling_rate_ms = 45_000
 # Period to post new requests to Witnet
 wit_dr_sender_polling_rate_ms = 45_000
+# Maximum data request result size (in bytes)
+# TODO: Choose a proper value
+max_result_size = 100
 
 # Max value that will be accepted by the bridge node in a data request
 # This is the maximum amount that the relayer is willing to lose per one data request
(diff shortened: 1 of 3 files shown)


witnet_centralized_ethereum_bridge.toml
```diff
@@ -14,6 +14,9 @@ eth_new_dr_polling_rate_ms = 1_000
 wit_tally_polling_rate_ms = 45_000
 # Period to post new requests to Witnet
 wit_dr_sender_polling_rate_ms = 45_000
+# Maximum data request result size (in bytes)
+# TODO: Choose a proper value
+max_result_size = 100
```
tmpolaczyk, line 19: Thoughts? @clbartoli @mariocao

mariocao: The sample DRs about the Bitcoin price have a length of 96 bytes.
Maybe we could define the maximum size to be reported as 200 bytes (just to be safe).

tmpolaczyk: But this is the result size, not the request size. The price of bitcoin is just a floating point number so it should be something like 9 bytes?

mariocao: Yep, you are right, we were misled by the following `reportResult` [transaction](https://rinkeby.etherscan.io/tx/0xfc2b1f475aa2cbd55fe18e72267bb68b3776bf16d733adb5e041cbd509cba328) where the result bytes take 96 bytes because of padding.

100 bytes should be more than enough and, in case we have to change it in the future, it is a config param and we can easily upgrade it! :) Thanks

mariocao (approved):

PR #1916, opened 2021-05-06
Update rad errors

lrubiorod:


rad/src/error.rs
```diff
@@ -337,6 +337,19 @@ pub enum RadError {
     /// Invalid reveal serialization (malformed reveals are converted to this value)
     #[fail(display = "The reveal was not serialized correctly")]
     MalformedReveal,
+    // These errors would never appear in the witnet protocol
+    /// Requests that cannot be parsed must always get this error as their result.
+    /// However, this is not a valid result in a Tally transaction, because invalid requests
+    /// are never included into blocks and therefore never get a Tally in response.
+    #[fail(display = "BridgeMalformedRequest")]
+    BridgeMalformedRequest,
```
aesedepece, line 345: I have my doubts about this being a bridge specific error. There's probably somewhere else where we coult return this and we're returning a more generic error instead.

I'm OK however to leave this as is if we don't see the clear case for using it elsewhere.

rad/src/types/mod.rs
```diff
@@ -575,27 +575,4 @@ mod tests {
...
-
-        // One reveal with value RadonErrors(0xE0)
-        let cbor_bytes: Vec<(&[u8], &())> = vec![(&[0xD8, 0x27, 0x81, 0x18, 0xE0], &())];
-        let rad_decode_error_as_result: Vec<_> =
-            serial_iter_decode(&mut cbor_bytes.into_iter(), malformed_reveal_fn)
-                .into_iter()
-                .map(|report| report.into_inner())
-                .collect();
-        assert_eq!(rad_decode_error_as_result, vec![malformed_reveal]);
-    }
```
aesedepece, line 600: Why is this test disappearing?

aesedepece (changes requested):

rad/src/types/mod.rs
tmpolaczyk: My test :cry:

rad/src/error.rs
lrubiorod: The problem here is that BridgeMalformedRequest is an error detected by the node that tries to send a request but it obtains an error in validation, not in runtime. So it only happens when it is not the same "person" the requestor and who includes the request

rad/src/types/mod.rs
lrubiorod: Why this test check that if a witness reveal a result with a "Bridge" radon error. In the past it would be translated as a MalformedReveal, but if a real RadError exist, it can not be a MalformedReveal

tmpolaczyk: So technically this is a breaking change? Maybe we can add a match somewhere that will convert this errors to MalformedReveal when used in a reveal transction?

tmpolaczyk:
Also this PR should remove the hacks from the bridge where we serialize the errors manually

bridges/centralized-ethereum/src/actors/dr_reporter.rs
```diff
@@ -84,8 +85,10 @@ impl Handler<DrReporterMsg> for DrReporter {
         };
 
         if msg.result.len() > self.max_result_size {
-            let radon_error = RadonErrors::BridgeOversizedResult as u8;
-            msg.result = vec![0xD8, 0x27, 0x81, 0x18, radon_error]
+            let radon_error =
+                RadError::try_from_kind_and_cbor_args(RadonErrors::BridgeOversizedResult, None)
+                    .unwrap();
```
mariocao, line 90:
```suggestion
                RadError::try_from_kind_and_cbor_args(RadonErrors::BridgeOversizedResult, None)?;
```

bridges/centralized-ethereum/src/actors/dr_reporter.rs
```diff
@@ -84,8 +85,10 @@ impl Handler<DrReporterMsg> for DrReporter {
...
+            msg.result = radon_error.encode_tagged_bytes().unwrap();
```
mariocao, line 91:
```suggestion
            msg.result = radon_error.encode_tagged_bytes()?;
```

mariocao, line 91:
```suggestion
            msg.result = radon_error.encode_tagged_bytes()?;
```

mariocao, line 91: I think we could use `?` syntax to replace some of the `unwrap`s if we modify the return type of the handler. Wdyt?

rad/src/error.rs
mariocao: any conclusion on this? :)

tmpolaczyk: I thought this PR is closed, not sure if we want to implement it because the benefits are small.

lrubiorod:
Close because we are not sure about the benefits of this PR

PR #1928, opened 2021-05-11
feat: create struct to store active WIPs

tmpolaczyk:
This pull request is a refactor of the validations code to use a new `ActiveWips` struct when checking if some validation logic is enabled or not. Currently this only replaces the `after_first_hard_fork` and `after_second_hard_fork` functions, but the idea is to extend it when the Threshold Activation of Protocol Improvements (TAPI) is implemented.

### DIFF
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -1,13 +1,3 @@
-use bech32::{FromBase32, ToBase32};
-use bls_signatures_rs::{bn256, bn256::Bn256, MultiSignature};
-use failure::Fail;
-use ordered_float::OrderedFloat;
-use partial_struct::PartialStruct;
-use secp256k1::{
-    PublicKey as Secp256k1_PublicKey, SecretKey as Secp256k1_SecretKey,
-    Signature as Secp256k1_Signature,
-};
-use serde::{Deserialize, Serialize};
 use std::{
     cell::{Cell, RefCell},
     cmp::Ordering,
@@ -17,6 +7,18 @@ use std::{
     ops::{AddAssign, SubAssign},
     str::FromStr,
 };
+
+use bech32::{FromBase32, ToBase32};
+use bls_signatures_rs::{bn256, bn256::Bn256, MultiSignature};
+use failure::Fail;
+use ordered_float::OrderedFloat;
+use secp256k1::{
+    PublicKey as Secp256k1_PublicKey, SecretKey as Secp256k1_SecretKey,
+    Signature as Secp256k1_Signature,
+};
+use serde::{Deserialize, Serialize};
+
+use partial_struct::PartialStruct;
 use witnet_crypto::{
     hash::{calculate_sha256, Sha256},
     key::ExtendedSK,
@@ -3843,11 +3845,6 @@ pub fn block_example() -> Block {
 
 #[cfg(test)]
 mod tests {
-    use super::*;
-    use crate::{
-        superblock::{mining_build_superblock, ARSIdentities},
-        transaction::{CommitTransactionBody, RevealTransactionBody, VTTransactionBody},
-    };
     use witnet_crypto::{
         merkle::{merkle_tree_root, InclusionProof},
         secp256k1::{
@@ -3855,6 +3852,13 @@ mod tests {
         },
         signature::sign,
     };
+
+    use crate::{
+        superblock::{mining_build_superblock, ARSIdentities},
+        transaction::{CommitTransactionBody, RevealTransactionBody, VTTransactionBody},
+    };
+
+    use super::*;
 
     fn dr_root_superblock_loop_test(
         sb: SuperBlock,
--- a/data_structures/src/mainnet_validations.rs
+++ b/data_structures/src/mainnet_validations.rs
@@ -245,6 +245,32 @@ pub fn after_second_hard_fork(epoch: Epoch, environment: Environment) -> bool {
     epoch >= SECOND_HARD_FORK && Environment::Mainnet == environment
 }
 
+/// Allows to check the active Witnet Improvement Proposals
+#[derive(Clone, Debug)]
+pub struct ActiveWips {
+    pub active_wips: HashMap<&'static str, Epoch>,
+    pub block_epoch: Epoch,
+    pub environment: Environment,
+}
+
+impl ActiveWips {
+    pub fn first_hard_fork(&self) -> bool {
+        after_first_hard_fork(self.block_epoch, self.environment)
+    }
+
+    pub fn second_hard_fork(&self) -> bool {
+        after_second_hard_fork(self.block_epoch, self.environment)
+    }
+
+    pub fn wip0014(&self) -> bool {
+        // TODO: ignore environment?
+        self.active_wips
+            .get("WIP0014")
+            .map(|activation_epoch| self.block_epoch >= *activation_epoch)
+            .unwrap_or(false)
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/data_structures/src/mainnet_validations.rs
+++ b/data_structures/src/mainnet_validations.rs
@@ -248,7 +248,7 @@ pub fn after_second_hard_fork(epoch: Epoch, environment: Environment) -> bool {
 /// Allows to check the active Witnet Improvement Proposals
 #[derive(Clone, Debug)]
 pub struct ActiveWips {
-    pub active_wips: HashMap<&'static str, Epoch>,
+    pub active_wips: HashMap<String, Epoch>,
     pub block_epoch: Epoch,
     pub environment: Environment,
 }
--- a/data_structures/src/mainnet_validations.rs
+++ b/data_structures/src/mainnet_validations.rs
@@ -265,7 +265,6 @@ pub fn after_second_hard_fork(epoch: Epoch, environment: Environment) -> bool {
 pub struct ActiveWips {
     pub active_wips: HashMap<String, Epoch>,
     pub block_epoch: Epoch,
-    pub environment: Environment,
 }
 
 impl ActiveWips {
--- a/data_structures/src/mainnet_validations.rs
+++ b/data_structures/src/mainnet_validations.rs
@@ -70,25 +70,40 @@ impl TapiEngine {
         self.bit_tapi_counter.last_epoch = epoch_to_update;
     }
 
-    pub fn initialize_wip_information(&mut self) -> (Epoch, HashSet<String>) {
-        // Hardcoded information about WIPs
-        self.wip_activation
-            .insert("WIP0008".to_string(), FIRST_HARD_FORK);
-        self.wip_activation
-            .insert("WIP0009-0011-0012".to_string(), SECOND_HARD_FORK);
-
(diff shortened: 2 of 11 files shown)


data_structures/src/chain.rs
```diff
@@ -17,6 +7,18 @@ use std::{
...
+use bls_signatures_rs::{bn256, bn256::Bn256, MultiSignature};
+use failure::Fail;
+use ordered_float::OrderedFloat;
+use secp256k1::{
+    PublicKey as Secp256k1_PublicKey, SecretKey as Secp256k1_SecretKey,
+    Signature as Secp256k1_Signature,
+};
+use serde::{Deserialize, Serialize};
+
+use partial_struct::PartialStruct;
```
aesedepece, line 21: Everything is better if alphabetically ordered :rofl: 
```suggestion
use bech32::{FromBase32, ToBase32};
use bls_signatures_rs::{bn256, bn256::Bn256, MultiSignature};
use failure::Fail;
use ordered_float::OrderedFloat;
use partial_struct::PartialStruct;
use secp256k1::{
    PublicKey as Secp256k1_PublicKey, SecretKey as Secp256k1_SecretKey,
    Signature as Secp256k1_Signature,
};
use serde::{Deserialize, Serialize};
```

data_structures/src/chain.rs
```diff
@@ -3856,6 +3853,13 @@ mod tests {
         signature::sign,
     };
 
+    use crate::{
+        superblock::{mining_build_superblock, ARSIdentities},
+        transaction::{CommitTransactionBody, RevealTransactionBody, VTTransactionBody},
+    };
+
+    use super::*;
```
aesedepece, line 3861: Is this possible?
```suggestion
    use crate::{
        Self::*,
        superblock::{mining_build_superblock, ARSIdentities},
        transaction::{CommitTransactionBody, RevealTransactionBody, VTTransactionBody},
    };
```

data_structures/src/mainnet_validations.rs
```diff
@@ -209,6 +209,32 @@ pub fn after_second_hard_fork(epoch: Epoch, environment: Environment) -> bool {
     epoch >= SECOND_HARD_FORK && Environment::Mainnet == environment
 }
 
+/// Allows to check the active Witnet Improvement Proposals
+#[derive(Clone, Debug)]
+pub struct ActiveWips {
+    pub active_wips: HashMap<String, Epoch>,
+    pub block_epoch: Epoch,
+    pub environment: Environment,
+}
```
aesedepece, line 218: As a general note, there isn't necessarily a 1:1 relation between a "protocol improvement" and an "improvement proposal".

That is, the improvements from multiple WIP documents can be signaled and activated as a single "protocol improvement".

Therefore, using the WIP term in the code seems misleading to me, because e.g. you can't make your node oppose a WIP, because WIPs are opposed by written word and discussion—what you can do here is signal that your node is not ready to activate the protocol changes proposed by the WIP (which cannot exist at the TAPI level before they have been "socially" approved by he community)

I don't have a better name though, so I can't really demand this to be changed and I'd be actually OK to leave it as is.

data_structures/src/mainnet_validations.rs
```diff
@@ -209,6 +209,32 @@ pub fn after_second_hard_fork(epoch: Epoch, environment: Environment) -> bool {
...
+    pub fn first_hard_fork(&self) -> bool {
+        after_first_hard_fork(self.block_epoch, self.environment)
+    }
+
+    pub fn second_hard_fork(&self) -> bool {
+        after_second_hard_fork(self.block_epoch, self.environment)
+    }
+
+    pub fn wip0014(&self) -> bool {
+        // TODO: ignore environment?
```
aesedepece, line 230: This is an interesting point. Different Witnet networks (e.g. mainnet vs. testnet) may activate an improvement or fail to do it at different times.

This should be a no-problem under the assumption that the activation state of an improvement is always derived from the local chain, and that nodes can't jump from one network to another without a full resync.

data_structures/src/mainnet_validations.rs
```diff
@@ -209,6 +209,32 @@ pub fn after_second_hard_fork(epoch: Epoch, environment: Environment) -> bool {
...
+    pub block_epoch: Epoch,
+    pub environment: Environment,
+}
+
+impl ActiveWips {
+    pub fn first_hard_fork(&self) -> bool {
+        after_first_hard_fork(self.block_epoch, self.environment)
+    }
+
+    pub fn second_hard_fork(&self) -> bool {
```
aesedepece, line 225:
```suggestion
   // WIPs 0009, 0011 and 0012 were activated through community coordination on April 28, 2021
    pub fn wips_0009_0011_0012(&self) -> bool {
```

data_structures/src/mainnet_validations.rs
```diff
@@ -209,6 +209,32 @@ pub fn after_second_hard_fork(epoch: Epoch, environment: Environment) -> bool {
...
+/// Allows to check the active Witnet Improvement Proposals
+#[derive(Clone, Debug)]
+pub struct ActiveWips {
+    pub active_wips: HashMap<String, Epoch>,
+    pub block_epoch: Epoch,
+    pub environment: Environment,
+}
+
+impl ActiveWips {
+    pub fn first_hard_fork(&self) -> bool {
```
aesedepece, line 221:
```suggestion
    // WIP 0008 was activated through community coordination on January 22, 2021
    pub fn wip_0008(&self) -> bool {
```

aesedepece (changes requested):

data_structures/src/chain.rs
tmpolaczyk: I don't think so, `crate::` means `/` and `super::` means `../`, combined they form `/../` which doesn't make sense.

aesedepece: If `super::*` means `every sibling of this module` and this is a 2-nd level module, how's that not equivalent to `crate::*`?

data_structures/src/mainnet_validations.rs
tmpolaczyk: I would expect that after an improvement proposal is enabled in mainnet any local network will start with that improvement proposal enabled. So we could hardcode a list of enabled WIPs somewhere, or since this is only used for developing we can even hardcode them in the code.

data_structures/src/mainnet_validations.rs
tmpolaczyk: It would be nice to create 3 separate functions, one for each WIP

data_structures/src/chain.rs
tmpolaczyk: You mean it's redundant? I can remove the `use super::*` and see if it compiles

tmpolaczyk: Ah no, this is inside `mod tests`, so `super::*` refers to `crate::chain::*`. That's a common pattern used in tests so I'm leaving it as `super::*`.

data_structures/src/mainnet_validations.rs
aesedepece: Why so if they were activated atomically?

aesedepece: Just so that when we use it in the code, we explicitly say which WIP introduced for that piece of code?

data_structures/src/chain.rs
aesedepece: Oh, now I understand. Resolved!

data_structures/src/mainnet_validations.rs
tmpolaczyk: Yes, exactly, for documentation purposes. But we can do that later because this PR already changes many lines.

data_structures/src/mainnet_validations.rs
aesedepece: That makes sense, given that WIP-0014§(17) states that:
> Once a protocol improvement is activated through this procedure, the set of protocol rules that it
introduces become part of the core Witnet protocol.

aesedepece:
I have no further objections to merging this. @tmpolaczyk are you changing anything or shall I approve it?

tmpolaczyk:
Let's merge to the tapi branch

data_structures/src/mainnet_validations.rs
lrubiorod: ActiveWpi? (Witnet Protocol Improven)

WPI vs WIP 🤯

lrubiorod: We could also add a "description" field that explain what wips are included in that wpi

data_structures/src/mainnet_validations.rs
```diff
@@ -209,6 +209,34 @@ pub fn after_second_hard_fork(epoch: Epoch, environment: Environment) -> bool {
     epoch >= SECOND_HARD_FORK && Environment::Mainnet == environment
 }
 
+/// Allows to check the active Witnet Improvement Proposals
+#[derive(Clone, Debug)]
+pub struct ActiveWips {
```
lrubiorod, line 214: Is this struct similar to TapiEngine?

tmpolaczyk: The `active_wips` field is exactly the same as `tapi_engine.wip_activation`, but the other fields are not the same.

data_structures/src/mainnet_validations.rs
tmpolaczyk: > WPI vs WIP

No.

data_structures/src/mainnet_validations.rs
lrubiorod: should we include the other fields and the impl in the TapiEngine?

lrubiorod (approved):

data_structures/src/mainnet_validations.rs
```diff
@@ -70,25 +70,40 @@ impl TapiEngine {
...
+                };
+                voting_wips[bit] = Some(wip_0014);
+            }
+            Environment::Testnet | Environment::Development => {
+                // In non-mainnet chains, all the WIPs that are active in mainnet are considered
+                // active since epoch 0. And there is no voting.
+                self.wip_activation.insert("WIP0008".to_string(), 0);
+                self.wip_activation
+                    .insert("WIP0009-0011-0012".to_string(), 0);
+            }
```
tmpolaczyk, line 105: Added a new commit implementing different defaults for testnet

tmpolaczyk:
I left the ones that can be removed, in mining and superblock logic. Because I think there is an issue to remove them so we can do it in a different PR.

lrubiorod (approved):

node/src/actors/messages.rs
```diff
@@ -717,8 +719,9 @@ pub struct RunTally {
     pub min_consensus_ratio: f64,
     /// Number of commits
     pub commits_count: usize,
-    /// Epoch of the block that will include this tally, used to select the correct version of the validation logic
-    pub block_epoch: Epoch,
+    /// Active Witnet Improvement Proposals as of the block that will include this tally.
```
aesedepece, line 722:
```suggestion
    /// Active Witnet protocol improvements as of the block that will include this tally.
```

aesedepece (changes requested):

aesedepece (approved):

PR #1929, opened 2021-05-11
feat(tapi): add counter for block version signaling

lrubiorod:


data_structures/src/chain.rs
```diff
@@ -3004,6 +3005,8 @@ pub struct ChainState {
     pub alt_keys: AltKeys,
     /// Current superblock state
     pub superblock_state: SuperBlockState,
+    /// WIPs votes counter by bits
+    pub bit_wip_counter: HashMap<i8, WipVotesCounter>,
```
tmpolaczyk, line 3009: Use an array instead?

```suggestion
    pub bit_wip_counter: [WipVotesCounter; 32],
```

data_structures/src/mainnet_validations.rs
```diff
@@ -16,6 +18,64 @@ pub const FIRST_HARD_FORK: Epoch = 192000;
...
 pub const SECOND_HARD_FORK: Epoch = 376320;
 
+/// Struct that count the positives votes of a WIP
+#[derive(Debug, Default, Clone, Serialize, Deserialize, PartialEq)]
+pub struct BitVotesCounter {
+    pub votes: u32,
+    pub period: u32,
+    pub wip: String,
+    pub init: u32,
+    pub end: u32,
```
tmpolaczyk, line 28: `period`, `init`, and `end` should have type `Epoch`.

data_structures/src/mainnet_validations.rs
```diff
@@ -16,6 +18,64 @@ pub const FIRST_HARD_FORK: Epoch = 192000;
...
+}
+
+#[derive(Debug, Default, Clone, Serialize, Deserialize, PartialEq)]
+pub struct BitTapiCounter {
+    pub info: HashMap<i8, Vec<BitVotesCounter>>,
```
tmpolaczyk, line 33: Same comment as before, since the key of this map is a bit index between 0 and 31, it can be replaced by an array of 32 elements:

```suggestion
    pub info: [Vec<BitVotesCounter>; 32],
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -829,6 +834,56 @@ impl ChainManager {
...
+            if let Some(mut bit_counter) = self.chain_state.bit_tapi_counter.get_mut(&n, &epoch) {
+                if !self
+                    .chain_state
+                    .wip_activation
+                    .contains_key(&bit_counter.wip)
+                    && is_bit_n_activated(v, n)
+                {
+                    bit_counter.votes += 1;
+
+                    if (epoch + bit_counter.init) % bit_counter.period == 0 {
```
tmpolaczyk, line 848: Not sure if this calculation is correct. Add a test case:

init = 0, period = 100, should check activation at epoch 100
init = 1, period = 100, should check activation at epoch 101 (or also 100?)

Also, why is `end` ignored here?

node/src/actors/chain_manager/mod.rs
```diff
@@ -829,6 +834,56 @@ impl ChainManager {
...
+                        }
+                        bit_counter.votes = 0;
+                    }
+                }
+            }
+        }
+    }
+
+    #[allow(clippy::cast_possible_truncation, clippy::cast_possible_wrap)]
+    fn update_wip_information(&mut self) {
```
tmpolaczyk, line 864: Is this method used to initialize the `BitTapiCounter` once or is this something that should be called periodically to update the structure?

```suggestion
    fn initialize_wip_information(&mut self) {
```

node/src/actors/chain_manager/mod.rs
lrubiorod: Its a mistake, it should be epoch - init.
And end is used when you are doing the get_mut, because if you are out of the period, it returns None

tmpolaczyk: Does that subtraction check for overflow? When init > epoch this check should always return false.

lrubiorod: It is not need, because if you are out of the period you can not reach that line

data_structures/src/mainnet_validations.rs
```diff
@@ -16,6 +18,140 @@ pub const FIRST_HARD_FORK: Epoch = 192000;
...
+                        }
+                        bit_counter.votes = 0;
+                    }
+                }
+            }
+        }
+        self.bit_tapi_counter.last_epoch = epoch;
+    }
+
+    #[allow(clippy::cast_possible_truncation, clippy::cast_possible_wrap)]
```
tmpolaczyk, line 59: I think its no longer needed

```suggestion
```

node/src/actors/chain_manager/actor.rs
```diff
@@ -254,6 +256,17 @@ impl ChainManager {
...
 
+                // Update possible new WIP information
+                let (new_wip_epoch, old_wips) = act.chain_state.tapi_engine.initialize_wip_information();
+                let last_consolidated_epoch = act.get_chain_beacon().checkpoint;
+                if new_wip_epoch < last_consolidated_epoch {
+                    ChainManager::from_registry().do_send(UpdateWipVotes{
+                        init: new_wip_epoch,
+                        end: last_consolidated_epoch,
+                        old_wips
+                    });
```
tmpolaczyk, line 267: I think it's important that this message is completed before the chain manager tries to do anything else, so use `.send().wait(ctx)` instead of `do_send()`.

data_structures/src/mainnet_validations.rs
```diff
@@ -16,6 +18,143 @@ pub const FIRST_HARD_FORK: Epoch = 192000;
...
+    /// bit votes counter by bits
+    pub bit_tapi_counter: BitTapiCounter,
+    /// wip activation
+    pub wip_activation: HashMap<String, Epoch>,
+}
+
+impl TapiEngine {
+    pub fn update_bit_counter(&mut self, v: u32, epoch: Epoch, avoid_wip_list: &HashSet<String>) {
+        // In case of empty epochs, they would be considered as blocks with tapi version to 0
+        if self.bit_tapi_counter.last_epoch != 0 && epoch > self.bit_tapi_counter.last_epoch + 1 {
```
tmpolaczyk, line 33: What happens if `bit_tapi_counter.last_epoch == 0`?

data_structures/src/mainnet_validations.rs
```diff
@@ -16,6 +18,143 @@ pub const FIRST_HARD_FORK: Epoch = 192000;
...
+                self.info[k] = vec![v];
+            }
+        }
+    }
+
+    pub fn contains(&self, bit: usize, wip: &str) -> bool {
+        match self.info.get(bit) {
+            Some(bit_info) => {
+                for i in bit_info {
+                    if i.wip.eq(wip) {
```
tmpolaczyk, line 142:
```suggestion
                    if i.wip == wip {
```

tmpolaczyk: No, wait, this for loop is `.any()`

https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.any

data_structures/src/mainnet_validations.rs
lrubiorod: that it would be updated as a normal update, and last_epoch would be overwrite with the last consolidated block epoch

node/src/actors/chain_manager/actor.rs
```diff
@@ -254,6 +254,13 @@ impl ChainManager {
 
                 act.chain_state = chain_state;
 
+                // Update possible new WIP information
+                let (new_wip_epoch, old_wips) = act.chain_state.tapi_engine.initialize_wip_information();
+                let last_consolidated_epoch = act.get_chain_beacon().checkpoint;
+                if new_wip_epoch < last_consolidated_epoch {
```
tmpolaczyk, line 260:
```suggestion
                if new_wip_epoch < last_consolidated_epoch {
                    // Some blocks have been consolidated before this node updated to the latest version,
                    // so we need to count the missing wip_votes from that blocks
```

data_structures/src/mainnet_validations.rs
```diff
@@ -16,6 +18,145 @@ pub const FIRST_HARD_FORK: Epoch = 192000;
...
+pub struct TapiEngine {
+    /// bit votes counter by bits
+    pub bit_tapi_counter: BitTapiCounter,
+    /// wip activation
+    pub wip_activation: HashMap<String, Epoch>,
+}
+
+impl TapiEngine {
+    pub fn update_bit_counter(&mut self, v: u32, epoch: Epoch, avoid_wip_list: &HashSet<String>) {
+        // In case of empty epochs, they would be considered as blocks with tapi version to 0
```
tmpolaczyk, line 32: Can you add an assert here so we can catch errors when testing?

`assert!(epoch > self.bit_tapi_counter.last_epoch);`

data_structures/src/mainnet_validations.rs
```diff
@@ -16,6 +18,145 @@ pub const FIRST_HARD_FORK: Epoch = 192000;
...
+        // In order to not update bit counter from old blocks where the block version was not used,
+        // the first time (bit_tapi_counter.last_epoch == 0) would be skipped in this conditional branch
+        if self.bit_tapi_counter.last_epoch != 0 && epoch > self.bit_tapi_counter.last_epoch + 1 {
+            let init = self.bit_tapi_counter.last_epoch + 1;
+            let end = epoch;
+            for i in init..end {
+                self.update_bit_counter(0, i, avoid_wip_list);
+            }
+        }
+        for n in 0..32 {
```
tmpolaczyk, line 42: Remove hardcoded number:

```suggestion
        for n in 0..self.bit_tapi_counter.info.len() {
```

I was thinking that it would be better to use an iterator, like

    for (n, bit_counter) in self.bit_tapi_counter.iter_mut(epoch) {

But it is tricky to implement that, and the benefit is small.

data_structures/src/mainnet_validations.rs
```diff
@@ -16,6 +18,158 @@ pub const FIRST_HARD_FORK: Epoch = 192000;
...
+        }
+    }
+
+    pub fn insert(&mut self, k: usize, v: BitVotesCounter) {
+        match self.info.get_mut(k) {
+            Some(bit_info) => {
+                bit_info.push(v);
+            }
+            None => {
+                self.info[k] = vec![v];
```
tmpolaczyk, line 136: This line will always panic, if `self.info.get_mut(k)` is None, then `self.info[k]` is out of bounds.

lrubiorod: legacy code from vecs

data_structures/src/mainnet_validations.rs
tmpolaczyk: Ah, but maybe this conflicts with the function `update_new_wip_votes` because there we want to update old epochs?

tmpolaczyk (approved):
Let's merge this, any unresolved questions can be discussed later

data_structures/src/mainnet_validations.rs
```diff
@@ -16,6 +18,169 @@ pub const FIRST_HARD_FORK: Epoch = 192000;
...
+        v: u32,
+        epoch_to_update: Epoch,
+        block_epoch: Epoch,
+        avoid_wip_list: &HashSet<String>,
+    ) {
+        // In case of empty epochs, they would be considered as blocks with tapi version to 0
+        // In order to not update bit counter from old blocks where the block version was not used,
+        // the first time (bit_tapi_counter.last_epoch == 0) would be skipped in this conditional branch
+        if self.bit_tapi_counter.last_epoch != 0
+            && epoch_to_update > self.bit_tapi_counter.last_epoch + 1
```
aesedepece, line 42: I'm not understanding or can't see the need for this guard.

aesedepece (changes requested):

data_structures/src/mainnet_validations.rs
```diff
@@ -16,6 +18,169 @@ pub const FIRST_HARD_FORK: Epoch = 192000;
...
+
+/// Struct that count the positives votes of a WIP
+#[derive(Debug, Default, Clone, Serialize, Deserialize, PartialEq)]
+pub struct BitVotesCounter {
+    pub votes: u32,
+    pub period: Epoch,
+    pub wip: String,
+    pub init: Epoch,
+    pub end: Epoch,
+}
```
aesedepece, line 113: I see fields here that imply unspecified TAPI features that are not covered by [WIP-0014](https://github.com/witnet/WIPs/blob/81346388b74e301e1a679365dae72d8042fbff7d/wip-0014.md).

data_structures/src/mainnet_validations.rs
```diff
@@ -16,6 +18,169 @@ pub const FIRST_HARD_FORK: Epoch = 192000;
...
+                None
+            }
+            None => None,
+        }
+    }
+
+    pub fn insert(&mut self, k: usize, v: BitVotesCounter) {
+        if k >= self.info.len() {
+            log::error!(
+                "Tapi Engine: This bit position ({}) is invalid. {} has not been included",
```
aesedepece, line 141: What does "included" mean here?

data_structures/src/mainnet_validations.rs
lrubiorod: First time that you update witnet to have a TapiEngine, the last_epoch field starts to 0. It combined with the empty epoch checker would check the version of all of blocks of the chain, when it is not necessary. The idea is that when you start your node you set the current block epoch, or the first wip activation epoch in the "last_epoch" field without to recheck the complete chain

data_structures/src/mainnet_validations.rs
lrubiorod: if you specify in the code a wip that will be in a bit bigger than 32, that wip will not be included in the TapiEngine

data_structures/src/mainnet_validations.rs
lrubiorod: - `votes` is the counter
- `period` should be put to 26_880 (according to WIP-0014)
- `wip` is an identifier
- `init` and `end` are used to handle the duration of the vote process. Due that we are going to use the same bits for different vote processes, I think that if we do not specify, a node could activate automatically one WIP with votes from other WIP assigned in the same bit but in a later time

data_structures/src/mainnet_validations.rs
aesedepece: Thanks for the clarification!

data_structures/src/mainnet_validations.rs
aesedepece: I see. This kind of decisions, please raise them publicly. Otherwise we would end up implementing code that differs from the WIP or vice-versa.

data_structures/src/mainnet_validations.rs
aesedepece: Good.

PR #1932, opened 2021-05-14
Allow to add new fields to ChainState without breaking the storage

tmpolaczyk:
Warning: this PR changes the format of the ChainState, so make a backup of your storage before testing.

This mechanism allows to add extra fields to the `ChainState` struct. It does not allow to remove or modify existing fields, and it does not work with other types, only with `ChainState`. The idea is to exploit a property of `bincode`, that the fields of a struct are serialized as the concatenation of the serialization of every individual field. Therefore, we can read the old serialized chain state, append some extra bytes (the serialization of the new fields), and deserialize the result as the new version of `ChainState`.

There were many options to detect the version of `ChainState` that is stored in the database:
* Try to read as the new version, in case of error try to read as older version.
* Use a different key to store the version: `"chain-state-version"`.
* Add a `db_version` field to `ChainState` and serialize that along the other fields.
* Store version number inline, before storing the serialized bytes.

I decided to use the last option (store version number inline) because it's explicit, performant, and it allows us to easily change the serialization library in the future because bincode is not used to store the version number. The other options may be useful in other cases.

### DIFF
--- a/node/src/storage_mngr/node_migrations.rs
+++ b/node/src/storage_mngr/node_migrations.rs
@@ -1,0 +1,296 @@
+use super::*;
+use witnet_data_structures::chain::ChainState;
+use witnet_data_structures::mainnet_validations::TapiEngine;
+
+macro_rules! as_failure {
+    ($e:expr) => {
+        failure::Error::from_boxed_compat(Box::new($e))
+    };
+}
+
+/// Return the version of the `ChainState` serialization. Returns error on end of file.
+fn check_chain_state_version(chain_state_bytes: &[u8]) -> Result<u32, ()> {
+    if chain_state_bytes.is_empty() {
+        return Err(());
+    }
+
+    // Before versioning support, the first byte of the serialization of ChainState was the tag of
+    // an Option, which is one byte that must be either 0 or 1.
+    if chain_state_bytes[0] == 0 || chain_state_bytes[0] == 1 {
+        Ok(0)
+    } else {
+        // After versioning support, there is a db_version before the serialization.
+        // This field is a u32 (little endian) so it takes the first 4 bytes.
+        // db_version % 256 must never be 0 or 1, because that can be confused with version 0.
+        if chain_state_bytes.len() < 4 {
+            return Err(());
+        }
+        let mut four_bytes = [0; 4];
+        four_bytes.copy_from_slice(&chain_state_bytes[0..4]);
+        let db_version = u32::from_le_bytes(four_bytes);
+
+        Ok(db_version)
+    }
+}
+
+// TODO: change signature to &mut Vec<u8> and edit the bytes in place?
+// The input is assumed to be the serialization of a v0 ChainState
+fn migrate_chain_state_v0_to_v2(old_chain_state_bytes: &[u8]) -> Vec<u8> {
+    let db_version: u32 = 2;
+    let db_version_bytes = db_version.to_be_bytes();
+
+    // Extra fields in ChainState v2:
+    let tapi = TapiEngine::default();
+    let tapi_bytes = bincode::serialize(&tapi).unwrap();
+
+    [&db_version_bytes, old_chain_state_bytes, &tapi_bytes].concat()
+}
+
+fn migrate_chain_state(bytes: &[u8]) -> Result<ChainState, failure::Error> {
+    match check_chain_state_version(&bytes) {
+        Ok(0) => {
+            // Migrate from v0 to v2
+            let bytes = migrate_chain_state_v0_to_v2(&bytes);
+            log::debug!("Successfully migrated ChainState v0 to v2");
+
+            // Latest version
+            // Skip the first 4 bytes because they are used to encode db_version
+            match deserialize(&bytes[4..]) {
+                Ok(v) => Ok(v),
+                Err(e) => Err(as_failure!(e)),
+            }
+        }
+        Ok(2) => {
+            // Latest version
+            // Skip the first 4 bytes because they are used to encode db_version
+            match deserialize(&bytes[4..]) {
+                Ok(v) => Ok(v),
+                Err(e) => Err(as_failure!(e)),
+            }
+        }
+        Ok(unknown_version) => Err(failure::format_err!(
+            "Error when reading ChainState from database: version {} not supported",
+            unknown_version
+        )),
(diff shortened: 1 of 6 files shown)


node/src/storage_mngr/node_migrations.rs
```diff
@@ -0,0 +1,379 @@
...
+{
+    let addr = StorageManagerAdapter::from_registry();
+
+    let key_bytes = match serialize(key) {
+        Ok(x) => x,
+        Err(e) => return futures::future::Either::Left(future::ready(Err(e.into()))),
+    };
+
+    let fut = async move {
+        let opt = addr.send(Get(key_bytes)).await??;
```
lrubiorod, line 132: hesitant code :P

node/src/storage_mngr/node_migrations.rs
```diff
@@ -0,0 +1,379 @@
...
+) -> impl Future<Output = Result<(), failure::Error>> + 'static
+where
+    K: serde::Serialize,
+{
+    let addr = StorageManagerAdapter::from_registry();
+
+    let key_bytes = match serialize(key) {
+        Ok(x) => x,
+        Err(e) => {
+            return futures::future::Either::Left(futures::future::Either::Right(future::ready(
```
lrubiorod, line 167: Etiher::Left of Either::Right?

node/src/storage_mngr/node_migrations.rs
```diff
@@ -0,0 +1,379 @@
...
+        // This is detected as version 0
+        assert_eq!(check_chain_state_version(&t0_bytes), Ok(0));
+
+        let default_chain_info = ChainInfo {
+            environment: Default::default(),
+            consensus_constants: Default::default(),
+            highest_block_checkpoint: Default::default(),
+            highest_superblock_checkpoint: Default::default(),
+            highest_vrf_output: Default::default(),
+        };
```
lrubiorod, line 307:
```suggestion
        let default_chain_info = ChainInfo::default();
```

node/src/storage_mngr/node_migrations.rs
tmpolaczyk: :sob:

aesedepece:
I'm also totally for the last mechanism "Store version number inline, before storing the serialized bytes."

node/src/storage_mngr/node_migrations.rs
```diff
@@ -0,0 +1,367 @@
...
+    }
+
+    // Before versioning support, the first byte of the serialization of ChainState was the tag of
+    // an Option, which is one byte that must be either 0 or 1.
+    if chain_state_bytes[0] == 0 || chain_state_bytes[0] == 1 {
+        Ok(0)
+    } else {
+        // After versioning support, there is a db_version before the serialization.
+        // This field is a u32 (little endian) so it takes the first 4 bytes.
+        // db_version % 256 must never be 0 or 1, because that can be confused with version 0.
```
aesedepece, line 25: This is brilliant :)

aesedepece: Simply solves the one concerned I always had about bincode migrations: telling a versioned structure from the original one that may had virtually anything as its first byte. Luckily, the fact that in the case of the chain state this is an enum, this will totally do the trick!

node/src/storage_mngr/node_migrations.rs
```diff
@@ -0,0 +1,367 @@
...
+    #[derive(Debug, Default, Clone, Serialize, Deserialize, PartialEq)]
+    pub struct BitVotesCounter {
+        pub votes: u32,
+        pub period: Epoch,
+        pub wip: String,
+        pub init: Epoch,
+        pub end: Epoch,
+    }
+
+    // TODO: do not define this structs here, import them after rebasing
```
aesedepece, line 70:
```suggestion
    // TODO: do not define these structs here, import them after rebasing
```

aesedepece (commented):
@tmpolaczyk what's on your mind for assignment of `db_version`? Shall we keep public track of which are the deserialization logic for each version? How do we prevent using versions with `0` or `1` in the LSB?

node/src/storage_mngr/node_migrations.rs
```diff
@@ -0,0 +1,296 @@
...
+    key: &'a K,
+    chain_state: &'b ChainState,
+) -> impl Future<Output = Result<(), failure::Error>> + 'static
+where
+    K: serde::Serialize + 'static,
+{
+    let db_version: u32 = 2;
+    // The first byte of the ChainState db_version must never be 0 or 1,
+    // because that can be confused with version 0.
+    assert!(db_version.to_le_bytes()[0] >= 2);
```
tmpolaczyk, line 167: @aesedepece as long as this assert is not lost during some refactor, any attempts to set the version to an invalid number will fail.

aesedepece: Nice!

tmpolaczyk:
I tested this on a mainnet storage, the migration seems to work fine. And after going back to the previous version the storage is invalid, as expected, with error:

    thread 'main' panicked at 'Error while getting chain state from storage: tag for enum is not valid, found 2', node/src/actors/chain_manager/actor.rs:122:33

Which is not great, but not bad either.

aesedepece:
> I tested this on a mainnet storage, the migration seems to work fine. And after going back to the previous version the storage is invalid, as expected, with error:
> 
> ```
> thread 'main' panicked at 'Error while getting chain state from storage: tag for enum is not valid, found 2', node/src/actors/chain_manager/actor.rs:122:33
> ```
> 
> Which is not great, but not bad either.

Seems OK to me. It certainly rings a bell on the fact that it's a serialization / migration issue.

node/src/storage_mngr/node_migrations.rs
```diff
@@ -0,0 +1,296 @@
+use super::*;
+use witnet_data_structures::chain::ChainState;
+use witnet_data_structures::mainnet_validations::TapiEngine;
```
lrubiorod, line 3: fmt

lrubiorod (approved):

tmpolaczyk:
I'm waiting to merge this after the `TapiEngine` structure is stabilized, to prevent us from accidentally migrating the storage to a work-in-progress version that will become unusable after a few commits.

PR #1933, opened 2021-05-14
Birthdate sync

### LINKED ISSUE #1919: Check if `import_wallet` / `create_wallet` uses `birth_date`

Tommytrg:
Synchronize wallet from birth_date param instead of genesis

Closes #1919

### DIFF
--- a/wallet/src/actors/app/handlers/unlock_wallet.rs
+++ b/wallet/src/actors/app/handlers/unlock_wallet.rs
@@ -19,6 +19,7 @@ pub struct UnlockWalletResponse {
     current_account: u32,
     account_balance: model::WalletBalance,
     session_expiration_secs: u64,
+    birth_date: types::Epoch,
 }
 
 impl Message for UnlockWalletRequest {
@@ -43,6 +44,7 @@ impl Handler<UnlockWalletRequest> for app::App {
                     available_accounts: data.available_accounts,
                     account_balance: data.balance,
                     session_expiration_secs: slf.params.session_expires_in.as_secs(),
+                    birth_date: data.birth_date.checkpoint,
                 }
             },
         );
--- a/wallet/src/actors/app/methods.rs
+++ b/wallet/src/actors/app/methods.rs
@@ -27,6 +27,7 @@ pub struct Validated {
     pub overwrite: bool,
     pub password: types::Password,
     pub seed_source: types::SeedSource,
+    pub birth_date: Option<types::BirthDate>,
 }
 
 impl App {
@@ -271,6 +272,7 @@ impl App {
         name: Option<String>,
         description: Option<String>,
         overwrite: bool,
+        birth_date: Option<types::BirthDate>,
     ) -> ResponseFuture<String> {
         let f = self
             .params
@@ -281,6 +283,7 @@ impl App {
                 password,
                 seed_source,
                 overwrite,
+                birth_date,
             })
             .flatten_err();
 
@@ -989,6 +992,7 @@ pub fn validate(
     description: Option<String>,
     overwrite: Option<bool>,
     backup_password: Option<types::Password>,
+    birth_date: Option<types::BirthDate>,
 ) -> Result<Validated> {
     let source = match seed_source.as_ref() {
         "xprv" => validate_xprv(seed_data, backup_password)
@@ -1016,6 +1020,7 @@ pub fn validate(
         overwrite,
         password,
         seed_source,
+        birth_date,
     })
     .map_err(validation_error)
 }
--- a/wallet/src/actors/worker/error.rs
+++ b/wallet/src/actors/worker/error.rs
@@ -44,6 +44,11 @@ pub enum Error {
         _0, _1
     )]
     NodeBehindLocalTip(u32, u32),
+    #[fail(
+        display = "the provided `birth_date` epoch is greater than the current epoch({} > {})",
+        _0, _1
+    )]
+    InvalidBirthDate(u32, u32),
 }
 
 #[derive(Debug, Fail)]
--- a/wallet/src/actors/worker/handlers/create_wallet.rs
+++ b/wallet/src/actors/worker/handlers/create_wallet.rs
@@ -14,6 +14,8 @@ pub struct CreateWallet {
     pub seed_source: types::SeedSource,
     /// Overwrite flag
     pub overwrite: bool,
+    /// Protocol epoch in which a wallet was created (won't synchronize blocks prior to this epoch)
+    pub birth_date: Option<types::BirthDate>,
 }
 
 impl Message for CreateWallet {
@@ -31,6 +33,7 @@ impl Handler<CreateWallet> for worker::Worker {
             password,
             seed_source,
             overwrite,
+            birth_date,
         }: CreateWallet,
         _ctx: &mut Self::Context,
     ) -> Self::Result {
@@ -40,6 +43,7 @@ impl Handler<CreateWallet> for worker::Worker {
             password.as_ref(),
             &seed_source,
             overwrite,
+            birth_date,
         )
     }
 }
--- a/wallet/src/actors/worker/methods.rs
+++ b/wallet/src/actors/worker/methods.rs
@@ -1,4 +1,4 @@
-use std::convert::TryFrom;
+use std::convert::{TryFrom, TryInto};
 
 use jsonrpc_core as rpc;
 use serde_json::{json, Value};
@@ -23,6 +23,7 @@ use witnet_data_structures::{
 use witnet_futures_utils::TryFutureExt2;
 use witnet_net::client::tcp::jsonrpc;
 use witnet_rad::{script::RadonScriptExecutionSettings, RADRequestExecutionReport};
+use witnet_util::timestamp::get_timestamp;
 
 use super::*;
 use futures_util::compat::Compat01As03;
@@ -82,6 +83,7 @@ impl Worker {
         password: &[u8],
         source: &types::SeedSource,
         overwrite: bool,
+        birth_date: Option<types::BirthDate>,
     ) -> Result<String> {
         let (id, default_account, master_key) = match source {
             types::SeedSource::XprvDouble((internal, external)) => {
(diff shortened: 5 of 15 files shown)


wallet/src/actors/worker/methods.rs
```diff
@@ -1153,3 +1194,10 @@ impl Worker {
         wallet.export_master_key(password).map_err(Error::from)
     }
 }
+
+fn is_old_epoch(birth_date: u32, checkpoint_zero_timestamp: i64, checkpoints_period: u16) -> bool {
+    let seconds_from_genesis = chrono::offset::Utc::now().timestamp() - checkpoint_zero_timestamp;
```
tmpolaczyk, line 1199: We use our custom `get_timestamp` function

```suggestion
    let seconds_from_genesis = get_timestamp() - checkpoint_zero_timestamp;
```

wallet/src/repository/keys.rs
```diff
@@ -132,6 +132,12 @@ pub fn wallet_last_sync() -> Key<&'static str, CheckpointBeacon> {
     Key::new("last-sync")
 }
 
+/// The epoch of the latest consolidated block that was processed by the wallet
```
tmpolaczyk, line 135: Isn't this always equal to `wallet_last_sync`? I thought the wallet only stores consolidated state to the database?

wallet/src/types.rs
```diff
@@ -40,6 +40,8 @@ pub type SessionWallet = Arc<repository::Wallet<db::EncryptedDb>>;
 
 pub type Wallet = repository::Wallet<db::EncryptedDb>;
 
+pub type BirthDateParam = Option<u32>;
```
tmpolaczyk, line 43: This type is useless, just use `Option<Epoch>` everywhere instead.

wallet/src/actors/worker/error.rs
```diff
@@ -44,6 +44,8 @@ pub enum Error {
         _0, _1
     )]
     NodeBehindLocalTip(u32, u32),
+    #[fail(display = "birth_date has not passed yet: {}", _0)]
+    InvalidBirthDate(u32),
```
aesedepece, line 48: Type and description differ in meaning.

wallet/src/actors/worker/handlers/create_wallet.rs
```diff
@@ -14,6 +14,8 @@ pub struct CreateWallet {
     pub seed_source: types::SeedSource,
     /// Overwrite flag
     pub overwrite: bool,
+    /// Wallet creation checkpoint
```
aesedepece, line 17: Let's try to deprecate the "checkpoint" term and rather use "protocol epoch". In that way, we make things simpler by having only one way to name stuff, and maybe we'll be able to reuse the "checkpoint" term in the future for proper chain checkpointing (expedited synchronization).
```suggestion
    /// Protocol epoch in which a wallet was created (won't synchronize blocks prior to this epoch)
```

wallet/src/actors/worker/methods.rs
```diff
@@ -135,6 +136,47 @@ impl Worker {
                 (id, default_account, Some(master_key))
             }
         };
+
+        let superblock_interval: i64 = 21;
```
aesedepece, line 140: Magic number. Ideally this should be a constant somewhere out of this function, and have some documentation on what it means and where did its value come from.

wallet/src/actors/worker/methods.rs
```diff
@@ -135,6 +136,47 @@ impl Worker {
...
+                    self.params.epoch_constants.checkpoints_period,
+                ) {
+                    return Err(Error::InvalidBirthDate(epoch));
+                }
+
+                let gen_fut = self.get_block_chain((epoch as i64) - superblock_interval, 2);
+                let gen_res: Vec<ChainEntry> = futures::executor::block_on(gen_fut)?;
+                let gen_entry = gen_res
+                    .get(0)
+                    .expect("It should always found a consolidated block for a validated epoch");
```
aesedepece, line 155:
```suggestion
                    .expect("It should always find a last consolidated block for a any epoch number");
```

(The block that you get may be the one for the provided epoch number, or the closest to that number that got consolidated, if there were rollbacks or no block candidates)

wallet/src/repository/keys.rs
aesedepece: I think so. Apparently, this is redundant.

wallet/src/repository/keys.rs
```diff
@@ -256,3 +262,9 @@ pub fn transaction_movement(
         account_index, transaction_id
     ))
 }
+
+/// wallet creation
```
aesedepece, line 266:
```suggestion
/// Protocol epoch in which a wallet was created (won't synchronize blocks prior to this epoch) 
```

wallet/src/repository/wallet/state.rs
```diff
@@ -80,6 +80,8 @@ pub struct State {
     pub transient_external_addresses: HashMap<PublicKeyHash, model::Address>,
     /// Flag that determines whether to stop syncing an ongoing synchronization for a wallet session
     pub stop_syncing: bool,
+    /// Wallet creation checkpoint
```
aesedepece, line 83:
```suggestion
    /// Protocol epoch in which a wallet was created (won't synchronize blocks prior to this epoch)
```

wallet/src/types.rs
aesedepece: +1

aesedepece (changes requested):

wallet/src/actors/worker/methods.rs
tmpolaczyk: `2 * consensus_constants.superblock_period + 1`

wallet/src/actors/worker/methods.rs
```diff
@@ -1153,3 +1195,10 @@ impl Worker {
         wallet.export_master_key(password).map_err(Error::from)
     }
 }
+
+fn is_old_epoch(birth_date: u32, checkpoint_zero_timestamp: i64, checkpoints_period: u16) -> bool {
```
tmpolaczyk, line 1199: I don't understand the name of this function, what epoch should be old? It checks if the birth date epoch is greater than the current epoch. Maybe rename it to `validate_birth_date` and return a `Result<()>` using this line in case of error:

    if total_epochs < birth_date {
        return Err(Error::InvalidBirthDate(epoch));
    }

wallet/src/actors/worker/methods.rs
```diff
@@ -135,6 +137,47 @@ impl Worker {
...
+                    return Err(Error::InvalidBirthDate(epoch));
+                }
+
+                let gen_fut = self.get_block_chain((epoch as i64) - superblock_interval, 2);
+                let gen_res: Vec<ChainEntry> = futures::executor::block_on(gen_fut)?;
+                let gen_entry = gen_res
+                    .get(0)
+                    .expect("It should always find a last consolidated block for a any epoch number");
+                let get_gen_future = self.get_block(gen_entry.1.clone());
+                let (block, _confirmed_1) = futures::executor::block_on(get_gen_future)?;
```
tmpolaczyk, line 158: Ideally you should make this function async instead of using `block_on`, but that may be too difficult.

wallet/src/actors/worker/error.rs
```diff
@@ -44,6 +44,8 @@ pub enum Error {
         _0, _1
     )]
     NodeBehindLocalTip(u32, u32),
+    #[fail(display = "provided birth_date epoch is greater than the current: {}", _0)]
```
tmpolaczyk, line 47: Sorry, it may be ambiguous to whether that is the provided or the current one 

```suggestion
    #[fail(display = "provided birth_date epoch ({}) is greater than the current", _0)]
```

aesedepece, line 47: I'd even go further and be explicit about both values: 
```suggestion
    #[fail(display = "The provided `birth_date` epoch is greater than the current epoch ({} > {})", _0, _1)]
```

aesedepece (changes requested):

wallet/src/actors/worker/methods.rs
```diff
@@ -1153,3 +1197,18 @@ impl Worker {
...
+fn validate_birth_date(
+    birth_date: u32,
+    checkpoint_zero_timestamp: i64,
+    checkpoints_period: u16,
+) -> Result<()> {
+    let seconds_from_genesis = get_timestamp() - checkpoint_zero_timestamp;
+    let current_epoch = seconds_from_genesis / i64::from(checkpoints_period);
+
+    if current_epoch < i64::from(birth_date) {
+        Err(Error::InvalidBirthDate(birth_date, current_epoch as u32))
```
tmpolaczyk, line 1210: You can use this:

```suggestion
        Err(Error::InvalidBirthDate(birth_date, u32::try_from(current_epoch).unwrap_or(0)))
```

If for some reason the current epoch is negative, then the error message will show a 0. And if the current epoch is greater than u32::MAX, then it will not enter this branch so we don't care.

wallet/src/actors/app/handlers/unlock_wallet.rs
```diff
@@ -19,6 +19,7 @@ pub struct UnlockWalletResponse {
     current_account: u32,
     account_balance: model::WalletBalance,
     session_expiration_secs: u64,
+    birth_date: u32,
```
tmpolaczyk, line 22:
```suggestion
    birth_date: Epoch,
```

wallet/src/actors/app/methods.rs
```diff
@@ -27,6 +27,7 @@ pub struct Validated {
     pub overwrite: bool,
     pub password: types::Password,
     pub seed_source: types::SeedSource,
+    pub birth_date: Option<u32>,
```
tmpolaczyk, line 30:
```suggestion
    pub birth_date: Option<Epoch>,
```

lrubiorod:
I remember that previously, create a wallet or import from a seed it was the same process. But now, if you import a seed, how can you obtain the birth date?

wallet/src/types.rs
```diff
@@ -198,6 +202,12 @@ pub struct ExtendedKeyedSignature {
 #[derive(Clone, Debug, Deserialize)]
 pub struct ChainEntry(pub u32, pub String);
 
+#[derive(Debug, Serialize, Deserialize)]
+pub enum BirthDate {
+    Current,
```
tmpolaczyk, line 207: You can use serde rename to have a more idiomatic API, if you want

```suggestion
    #[serde(rename = "current")]
    Current,
```

tmpolaczyk (approved):
Untested but looks good. Squash all the commits before merging please.

wallet/src/actors/worker/methods.rs
```diff
@@ -135,6 +137,58 @@ impl Worker {
...
+            }
+            // Use provided epoch as birth date
+            Some(types::BirthDate::Imported(epoch)) => {
+                validate_birth_date(
+                    epoch,
+                    self.params.epoch_constants.checkpoint_zero_timestamp,
+                    self.params.epoch_constants.checkpoints_period,
+                )?;
+
+                let gen_fut = self.get_block_chain((epoch as i64) - confirm_superblock_period, 2);
```
tmpolaczyk, line 170: This doesn't check for overflow, `epoch - confirm_superblock_period` must never be negative

 ```suggestion
                let gen_fut = self.get_block_chain(epoch.saturating_sub(confirm_superblock_period), 2);
```

**Review by tmpolaczyk — CHANGES_REQUESTED** [review]

**Comment by tmpolaczyk** [comment]
Also, when you import an xprv file and then export it again, the birthdate epoch is smaller. But this is a very minor problem and it does not affect the usability of this feature.

**Comment by Tommytrg** [comment]
> I remember that previously, create a wallet or import from a seed it was the same process. But now, if you import a seed, how can you obtain the birth date?

Sheikah will include a step to let the user introduce a birth_date for their wallet. If they don't provide such value, it will be the genesis by default

**Comment by Tommytrg** [comment]
> Also, when you import an xprv file and then export it again, the birthdate epoch is smaller. But this is a very minor problem and it does not affect the usability of this feature.

We are doing that to reduce the error risk from the user input. IMHO, I prefer to synchronize 20 blocks more on each import rather than users not seeing their transactions due to they introduce an imprecise date. What do you think?

**Comment by tmpolaczyk** [comment]
> What do you think?

You can synchronize 20 blocks more but store the birth date as provided by the user

File: wallet/src/actors/worker/methods.rs [hunk 18]
```diff
@@ -135,6 +137,61 @@ impl Worker {
...
+            // Use provided epoch as birth date
+            Some(types::BirthDate::Imported(epoch)) => {
+                validate_birth_date(
+                    epoch,
+                    self.params.epoch_constants.checkpoint_zero_timestamp,
+                    self.params.epoch_constants.checkpoints_period,
+                )?;
+
+                let gen_fut = self.get_block_chain(
+                    i64::from(epoch).saturating_sub(confirm_superblock_period),
```
tmpolaczyk, line 171: Sorry, this is still wrong. It should be i64::from(u32.saturating_sub(u32))

tmpolaczyk (approved):

Issue #1920, opened 2021-05-11
Check viability of `BlockHeader.version` field repurpose for TAPI support signals


aesedepece:

tmpolaczyk:
`BlockHeader.version` can be used for TAPI support signals.

The version field affects the block hash so it is not possible to change it after the block is created (it is tamper proof), and nodes running on old software accept any value in the version field.

Furthermore, the version field can be renamed without breaking the storage or the protobuf messages, but it must be a `u32`.

The only strange detail is that the genesis block has version 1 while all the other mined blocks have version 0.

aesedepece:
> The only strange detail is that the genesis block has version 1 while all the other mined blocks have version 0.

That's a funny fact. Shouldn't be a problem at all :)

PR #1935, opened 2021-05-18
Tapi block validation tests

tmpolaczyk:
Implements the tests described in #1920


aesedepece:
I would love to see an activation test that for a signaling period of 100 epochs if the threshold activation is met, and blocks 95 to 105 are missing, the first block with the new rules is 126. Is this even possible to test?

lrubiorod (approved):

PR #1936, opened 2021-05-18
refactor(tapi): modify Vec<wip> by Option<wip> in TapiEngine

lrubiorod:


aesedepece:
Why this?

lrubiorod:
> Why this?

That is what we spoke about to fix the active wips, and avoid to have all of them. It means, before to this PR, we have all the wip voting proposals in a vector inside of each bit, so every node knows if a wip is active or not counting all the votes associated to that wip during that period (so it would count votes for wips that by example they dindt achieve 80% in the past). With that PR, each bit will have an Option with the current voting wip in each bit, and previous active wips will be included directly without the need to count those votes.

data_structures/src/mainnet_validations.rs
```diff
@@ -110,61 +121,85 @@ pub struct BitVotesCounter {
...
+        for bit_info in self.info.iter().flatten() {
+            length = bit_info.bit + 1;
+        }
+
+        self.current_length = length;
+    }
+
+    pub fn contains(&self, bit: usize, wip: &str) -> bool {
+        match self.info.get(bit) {
+            Some(Some(bit_info)) => bit_info.wip.eq(wip),
```
tmpolaczyk, line 201:
```suggestion
            Some(Some(bit_info)) => bit_info.wip == wip,
```

data_structures/src/mainnet_validations.rs
```diff
@@ -72,29 +72,40 @@ impl TapiEngine {
 
     pub fn initialize_wip_information(&mut self) -> (Epoch, HashSet<String>) {
         // Hardcoded information about WIPs
-        let mut bits = vec![vec![]; 32];
+        // TODO: Include first and second hard fork as activated wips
+        //self.wip_activation.insert("first_hard_fork".to_string(), FIRST_HARD_FORK);
+        //self.wip_activation.insert("second_hard_fork".to_string(), SECOND_HARD_FORK);
```
tmpolaczyk, line 77: Maybe we can use the WIP names?

```suggestion
        self.wip_activation.insert("WIP0008".to_string(), FIRST_HARD_FORK);
        self.wip_activation.insert("WIP0009-0011-0012".to_string(), SECOND_HARD_FORK);
```

lrubiorod: As @aesedepece said in another PR, I am not sure if we should use the wip numbers as identifiers for the protocol improvements or not, so 🤷🏽

lrubiorod: But we could put like that and modify in another PR if we consider useful

tmpolaczyk: Yes, and maybe we could use a `u32` index instead of a `String`

aesedepece:
> > Why this?
> 
> That is what we spoke about to fix the active wips, and avoid to have all of them. It means, before to this PR, we have all the wip voting proposals in a vector inside of each bit, so every node knows if a wip is active or not counting all the votes associated to that wip during that period (so it would count votes for wips that by example they dindt achieve 80% in the past). With that PR, each bit will have an Option with the current voting wip in each bit, and previous active wips will be included directly without the need to count those votes.

I'll ask for spoken explanation tomorrow 😸

tmpolaczyk (approved):

PR #1938, opened 2021-05-21
feat(node): stop actor system if any essential actor panics

### LINKED ISSUE #1934: Investigate why some panics stop the node and other ones don't
For example, this panic:

https://github.com/witnet/witnet-rust/blob/1ac9e4974cbb22ce9930f1449d5f3269675e9e7a/node/src/actors/chain_manager/actor.rs#L121-L123

Is meant to stop the node, as the user is expected to delete the corrupted storage or restore a backup.

However the observed behavior is that the chain manager dies but the rest of the node actors keep running. This results in a node that is not usable.

This is a problem because the docker script that runs the node detects when the node panics and automatically restarts it, but now that won't work as nodes that panic will keep running.

So, investigate why does that happen and how to prevent it.

A simple solution would be to handle this in the actor "stopped" method. If it's possible to detect that the actor is stopping because of a panic, then we can stop the entire system.

tmpolaczyk:
Fix #1934

How to test: add a panic somewhere in the node and check that the node stops after showing the panic message.

### DIFF
--- a/node/src/actors/chain_manager/mod.rs
+++ b/node/src/actors/chain_manager/mod.rs
@@ -87,6 +87,7 @@ use crate::{
         storage_keys,
     },
     signature_mngr, storage_mngr,
+    utils::stop_system_if_panicking,
 };
 
 mod actor;
@@ -223,6 +224,13 @@ pub struct ChainManager {
     last_superblock_consensus: Option<CheckpointBeacon>,
     /// Settings for Threshold Activation of Protocol Improvements
     tapi: Tapi,
+}
+
+impl Drop for ChainManager {
+    fn drop(&mut self) {
+        log::trace!("Dropping ChainManager");
+        stop_system_if_panicking("ChainManager");
+    }
 }
 
 /// Wrapper around a block candidate that contains additional metadata regarding
--- a/node/src/actors/rad_manager/mod.rs
+++ b/node/src/actors/rad_manager/mod.rs
@@ -12,3 +12,11 @@ mod handlers;
 /// RadManager actor
 #[derive(Default)]
 pub struct RadManager;
+
+impl Drop for RadManager {
+    fn drop(&mut self) {
+        log::trace!("Dropping RadManager");
+        // TODO: RadManager is expected to restart on panic, ensure
+        //stop_system_if_panicking("RadManager");
+    }
+}
--- a/node/src/actors/rad_manager/mod.rs
+++ b/node/src/actors/rad_manager/mod.rs
@@ -6,6 +6,8 @@
 //! [Data Requests]: https://docs.witnet.io/protocol/data-requests/overview/
 //! [RAD Engine]: https://docs.witnet.io/protocol/data-requests/overview/#the-rad-engine
 
+use crate::utils::stop_system_if_panicking;
+
 mod actor;
 mod handlers;
 
@@ -16,7 +18,10 @@ pub struct RadManager;
 impl Drop for RadManager {
     fn drop(&mut self) {
         log::trace!("Dropping RadManager");
-        // TODO: RadManager is expected to restart on panic, ensure
-        //stop_system_if_panicking("RadManager");
+        // RadManager handles radon panics so it should never stop because of a panic.
+        // That's handled by ensuring that the panics always happen inside a future.
+        // If for some reason RadManager panics outside of a future, then we want to stop the actor
+        // system.
+        stop_system_if_panicking("RadManager");
     }
 }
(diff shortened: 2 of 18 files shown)


node/src/actors/rad_manager/actor.rs
```diff
@@ -10,6 +10,11 @@ impl Actor for RadManager {
     fn started(&mut self, _ctx: &mut Self::Context) {
         log::debug!("RadManager actor has been started!");
     }
+
+    fn stopped(&mut self, _ctx: &mut Self::Context) {
+        // RadManager is expected to restart on panic
+        //System::current().stop_with_code(1);
```
lrubiorod, line 16: In this case, when the RadManager dies, it looks that it would never recover

node/src/signature_mngr.rs
```diff
@@ -257,6 +257,10 @@ impl Actor for SignatureManager {
 
         self.secp = Some(CryptoEngine::new());
     }
+
+    fn stopped(&mut self, _ctx: &mut Self::Context) {
+        System::current().stop_with_code(1);
```
lrubiorod, line 262: I include a panic in Handler<GetPkh> in line 379 and it doesnt close the main node

node/src/actors/rad_manager/actor.rs
tmpolaczyk: I assumed that it is restarted automatically, will try to add tests to see which actors restart and which ones stop the system when they panic.

node/src/signature_mngr.rs
lrubiorod: I am seeing that it depends in which part of the code panic, in the handlers there would not close the main node, but it will do if the panic is in the functions that send the messages to the handlers

tmpolaczyk: Maybe it's because we also need to add `System::current().stop_with_code` to `StorageManagerAdapter`?

tmpolaczyk: > Maybe it's because we also need to add System::current().stop_with_code to StorageManagerAdapter?

No, that didn't work. But I fixed it by adding an `impl Drop for StorageManager` that detects if the thread is panicking.

tmpolaczyk:
This is a rocksdb error:

```
pure virtual method called
terminate called without an active exception
```

we fixed it in the wallet by waiting for the database to close before killing the process, we can do something similar in the node.

tmpolaczyk:
> This is a rocksdb error:
> 
> ```
> pure virtual method called
> terminate called without an active exception
> ```
> 
> we fixed it in the wallet by waiting for the database to close before killing the process, we can do something similar in the node.

I added a sleep and it seems to work good enough, we can create an issue to fix it properly

node/src/actors/chain_manager/mod.rs
```diff
@@ -71,6 +71,7 @@ use witnet_validations::validations::{
     validate_new_transaction, verify_signatures, VrfSlots,
 };
 
+use crate::utils::stop_system_if_panicking;
```
lrubiorod, line 74: fmt

node/src/actors/rad_manager/mod.rs
```diff
@@ -12,3 +12,11 @@ mod handlers;
 /// RadManager actor
 #[derive(Default)]
 pub struct RadManager;
+
+impl Drop for RadManager {
+    fn drop(&mut self) {
+        log::trace!("Dropping RadManager");
+        // TODO: RadManager is expected to restart on panic, ensure
```
tmpolaczyk, line 19: It currently does not restart on panic.

node/src/actors/rad_manager/actor.rs
tmpolaczyk: Currently all actors stop the system on panic, (except `Session` and `JsonRpc`, which handle client connections and disconnect on panic). `RadManager` executes data requests inside a future, and that makes it immune to radon panics. It's not that the `RadManager` restarts automatically, it's that it never stops.

node/src/signature_mngr.rs
tmpolaczyk: This happens because the `SignatureManager` is started using `SyncArbiter::start`. When a `SignatureManager` handler panics, the panic propagates to the `SyncArbiter`, and that kills the actor without calling its `stopped` method. The `SignatureManagerAdapter` stores a reference (`Addr<SignatureManager>`) to this `SyncArbiter` and when the `SyncArbiter` dies, this reference is no longer valid. We could fix it by doing something like:

```rust
if !self.addr.connected() {
    // The StorageManager died, restart
    self.addr = SyncArbiter::start(...);
}
```

But I think that is too complex, and the alternative solution using `impl Drop` is more robust.

lrubiorod (approved):

tmpolaczyk:
Finally I did not add tests because not sure to how ensure that every actor has the expected behavior.

PR #1939, opened 2021-05-21
refactor(data-structures): rename "version" field to "signals" in BlockHeader

lrubiorod:


data_structures/src/chain.rs
```diff
@@ -636,7 +636,8 @@ impl Hashable for PublicKey {
 #[derive(Debug, Eq, PartialEq, Clone, Serialize, Deserialize, ProtobufConvert, Default)]
 #[protobuf_convert(pb = "witnet::Block_BlockHeader")]
 pub struct BlockHeader {
-    /// The block version number indicating the block validation rules
+    /// 32 bits for binary signaling new witnet protocol improvements
+    // TODO: Rename `version` to `signals` when we support migrations
```
aesedepece, line 640: Why? AFAIK fields can be safely renamed.

data_structures/src/chain.rs
```diff
@@ -636,7 +636,8 @@ impl Hashable for PublicKey {
...
     pub version: u32,
```
tmpolaczyk, line 641: Renaming is supported by both bincode and protobuf, but it would break the JSON-RPC API.

tmpolaczyk:
So this definitely breaks the JSON-RPC API, but in an expected way. We assume that nobody was using the old `version` field, and any client that wants to add a signaling monitor will need to update in order to support it.

However, this will break any client that uses the `getBlock` method, for example the wallet. This is because the version field is needed to deserialize the JSON object as a witnet `Block`. This wouldn't be a problem if the `getBlock` method used protobuf as the serialization format. Possible solutions:

1. Do nothing, close this PR
2. Use `serde(rename = "version")` to keep compatibility in the JSON-RPC API
3. Add an extra version field after the serialization, similar to how we add the "blockHashes" field.
4. Break the API. This means updating all the nodes and all the clients that may be affected. We may want to use this opportunity to change other aspects of the JSON-RPC API.

I prefer 1) because it's the one with less unexpected side effects.

mariocao:
Option 4 is mentioned in the [WIP-0014](https://github.com/witnet/WIPs/blob/master/wip-0014.md) as the improvement labeled as **(1)**.

mariocao (commented):
Option 4 is mentioned in the [WIP-0014](https://github.com/witnet/WIPs/blob/master/wip-0014.md) as the improvement labeled as (1).

aesedepece (approved):

data_structures/src/chain.rs
```diff
@@ -638,8 +638,8 @@ impl Hashable for PublicKey {
 #[derive(Debug, Eq, PartialEq, Clone, Serialize, Deserialize, ProtobufConvert, Default)]
 #[protobuf_convert(pb = "witnet::Block_BlockHeader")]
 pub struct BlockHeader {
-    /// The block version number indicating the block validation rules
-    pub version: u32,
+    /// 32 bits for binary signaling new witnet protocol improvements
```
aesedepece, line 641: Let's link to WIP-0014 here

aesedepece (changes requested):

data_structures/src/chain.rs
mariocao: That would be great

data_structures/src/chain.rs
```diff
@@ -638,8 +638,9 @@ impl Hashable for PublicKey {
...
+    /// https://github.com/witnet/WIPs/blob/master/wip-0014.md
```
aesedepece, line 642:
```suggestion
    /// 32 bits for binary signaling new witnet protocol improvements.
    /// See [WIP-0014](https://github.com/witnet/WIPs/blob/master/wip-0014.md) for more info.
```

aesedepece (changes requested):

validations/src/tests.rs
```diff
@@ -7521,7 +7521,7 @@ fn block_change_version() {
     // invalidate the signature
     let x = test_block(|b| {
         // Flip one bit of the version field
```
tmpolaczyk, line 7523:
```suggestion
        // Flip one bit of the signals field
```

validations/src/tests.rs
```diff
@@ -7521,7 +7521,7 @@ fn block_change_version() {
...
-        b.block_header.version ^= 1;
+        b.block_header.signals ^= 1;
 
         // Do not sign the block again
         // If the miner changes the version field and then signs the block, the block will be valid
```
tmpolaczyk, line 7527:
```suggestion
        // If the miner changes the signals field and then signs the block, the block will be valid
```

validations/src/tests.rs
tmpolaczyk: And also the test name

tmpolaczyk (approved):

aesedepece (approved):

PR #1941, opened 2021-05-25
feat: allow to select from address when creating VTT transactions

### LINKED ISSUE #1940: Allow to select `from` field while creating new VTT transactions
The wallet `create_vtt` endpoint should support the user to optionally define which `from` address should be used as `input` of the VTT transaction.

Regarding the implementation, UTXO selection strategies should be updated to allow an optional `from` value.

tmpolaczyk:
Allow to use an UTXO selection strategy that only takes UTXOs from the specified address.

Close #1940

In the wallet JSON-RPC API: add optional utxo_strategy field to `create_vtt` method. This allows to choose the address of the inputs of the transaction, but it also allows to use different UTXO selection strategies such as "bigger first" or "smaller first", aside from the default "random".

In the node JSON-RPC API: since the node only has one address this field can only be used as a fail-safe: if the node address is the one in the "from" field then the transaction will be created, and if the node address is not the one in the "from" field then the transaction will return an error "InsufficientBalance".

Example call (node):

```json
{
  "jsonrpc": "2.0",
  "method": "sendValue",
  "params": {
    "vto": [
      {
        "pkh": "wit1vfuklxq4p5d7t744evcrj04r68q7j2f8yj0vqx",
        "value": 6000,
        "time_lock": 0
      }
    ],
    "fee": 0,
    "utxo_strategy": {
      "strategy": "Random",
      "from": "wit1qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqwrt3a4"
    }
  },
  "id": "1"
}
```

### DIFF
--- a/data_structures/src/serialization_helpers.rs
+++ b/data_structures/src/serialization_helpers.rs
@@ -9,6 +9,7 @@
 use crate::{
     chain::{GenesisBlockInfo, Hash, OutputPointer, PublicKeyHash, ValueTransferOutput, SHA256},
     get_environment,
+    utxo_pool::UtxoSelectionStrategy,
 };
 use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
 use std::{fmt::Display, str::FromStr};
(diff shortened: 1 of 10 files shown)


wallet/src/actors/app/handlers/create_vtt.rs
```diff
@@ -40,6 +41,8 @@ pub struct CreateVttRequest {
     session_id: types::SessionId,
     wallet_id: String,
     fee_type: Option<FeeType>,
+    #[serde(default)]
+    utxo_strategy: UtxoSelectionStrategy,
```
mariocao, line 45:
```suggestion
    utxo_strategy: Option<UtxoSelectionStrategy>,
```

I might make sense to have the UTXO selection strategy as an optional field to be backwards compatible with older Sheikah versions. AFAIK, Sheikah tries to download latest wallet version even it is an older version.

mariocao: Now I see the `serde(default)`. i will test it! :) Thanks

tmpolaczyk: `serde(default)` will use the default value if it's missing (Random), so it should be backwards compatible.

data_structures/src/utxo_pool.rs
```diff
@@ -206,16 +216,47 @@ impl<'a> OutputsCollection for NodeUtxos<'a> {
...
 /// Strategy to sort our own unspent outputs pool
-#[derive(Clone, Copy, Debug, Serialize, Deserialize, Eq, PartialEq, Hash)]
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
 pub enum UtxoSelectionStrategy {
-    Random,
-    BigFirst,
-    SmallFirst,
+    Random { from: Option<PublicKeyHash> },
+    BigFirst { from: Option<PublicKeyHash> },
+    SmallFirst { from: Option<PublicKeyHash> },
```
mariocao, line 223:
```suggestion
    #[serde(rename = "random")]
    Random { from: Option<PublicKeyHash> },
    #[serde(rename = "big_first")]
    BigFirst { from: Option<PublicKeyHash> },
    #[serde(rename = "small_first")]
    SmallFirst { from: Option<PublicKeyHash> },
```

data_structures/src/serialization_helpers.rs
```diff
@@ -201,3 +202,150 @@ impl From<GenesisBlock> for GenesisBlockInfo {
...
+            UtxoSelectionStrategy::Random { .. } => UtxoSelectionStrategyName::Random,
+            UtxoSelectionStrategy::BigFirst { .. } => UtxoSelectionStrategyName::BigFirst,
+            UtxoSelectionStrategy::SmallFirst { .. } => UtxoSelectionStrategyName::SmallFirst,
+        }
+    }
+}
+
+#[derive(Deserialize, Serialize)]
+#[serde(untagged)]
+enum UtxoSelectionStrategyHelper {
```
mariocao, line 237: When I tried to use the Wallet API with a random strategy didn't work.

Input
```
    "utxo_strategy": {
      "strategy": "Random",
      "from": "wit1qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqwrt3a4"
    }
```

Output:
```
    "message": "Invalid params: data did not match any variant of untagged enum UtxoSelectionStrategyHelper.",
```

tmpolaczyk: This is strange because the tests pass, can you provide the full JSON-RPC call?

mariocao:
```json
{
  "jsonrpc": "2.0",
  "method": "create_vtt",
  "params": {
    "session_id": "663cffd566a614c6ac478d550a05e5c66c366b3ffab40b288d4812ce1a56e6b3",
    "wallet_id": "8f5b85981addad621a86f01a1ddb646ccd90620c95247948ce8d99feefd0496c",
    "fee": 1,
    "label": "",
    "outputs": [
      {
        "address": "twit1nfkythqds4r2hz3le2zaauxtl7yum76jd0ut9c",
        "amount": 1,
        "time_lock": 0
      }
    ],
    "utxo_strategy": {
      "strategy": "Random",
      "from": "wit1qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqwrt3a4"
    }
  },
  "id": 18
}
```

mariocao: Another weird thing is that if I try to create a VTT with the strategy with valid UTXOs there is also an error.

Input:
```json
{
  "jsonrpc": "2.0",
  "method": "create_vtt",
  "params": {
    "session_id": "241ec1bbbadbd7024bccde1746f64124a0db6d5126800ea408aed21def0f7391",
    "wallet_id": "8f5b85981addad621a86f01a1ddb646ccd90620c95247948ce8d99feefd0496c",
    "fee": 1,
    "label": "",
    "outputs": [
      {
        "address": "twit1eghyyar76nuvdfu0h70f4gmxruj2rw4gfnrhnk",
        "amount": 1,
        "time_lock": 0
      }
    ],
    "utxo_strategy": {
      "strategy": "Random",
      "from": "twit1eghyyar76nuvdfu0h70f4gmxruj2rw4gfnrhnk"
    }
  },
  "id": 18
}
```

Output:
```json
{
  "jsonrpc": "2.0",
  "error": {
    "code": 400,
    "message": "Validation Error",
    "data": [
      [
        "balance",
        "Wallet account has not enough balance"
      ]
    ]
  },
  "id": 18
}
```

If I take away the `utxo_strategy` it works and creates the transaction.

tmpolaczyk: The Invalid params error is probably because the address is invalid, and it shows a generic error instead of "invalid address". That can be fixed.

tmpolaczyk: The second issue is strange because the node API works as expected, so it must be a problem in the wallet component

tmpolaczyk: We couldn't reproduce the "Wallet account has not enough balance" error, so assuming that it was because of external factors.

data_structures/src/utxo_pool.rs
tmpolaczyk: Did this for `UtxoSelectionStrategyName`, which is where serialization is implemented. Also, using the old names is still allowed because of backward-compatibility in the node JSON-RPC API.

mariocao (commented):
I did some tests with an UTXO set with different values, and the strategies of `small_first` and `big_first` are working fine.

The `random` strategy is working as before, but I realized that it is not random for each call of `create_vtt`. It can be considered random from the perspective that we are using `HashSet` and we iterate directly from that data structure.

At some point some of us talked about possible theoretical maximum for transaction sizes (given the amount of inputs used) and that it would not be a big problem because a user could always try to send again the transaction, which is not true. :sweat_smile:

mariocao (approved):

PR #1945, opened 2021-05-27
feat(tapi): avoid active wips in pending_upgrades in GetSignalingInfo

lrubiorod:


data_structures/src/mainnet_validations.rs
```diff
@@ -251,9 +251,20 @@ impl BitTapiCounter {
         self.current_length == 0
     }
 
-    pub fn info(&self) -> Vec<BitVotesCounter> {
+    pub fn info(&self, active_wips: &HashMap<String, Epoch>) -> Vec<BitVotesCounter> {
         self.info[..self.current_length]
             .iter()
+            .map(|x| {
```
tmpolaczyk, line 257: This `.map().flatten()` could be rewritten as `.filter_map()`, but if clippy does not complain nether do I.

tmpolaczyk (approved):

PR #1946, opened 2021-05-28
chore: commit Cargo.lock

tmpolaczyk:
Forgot to do it after commit 352fb88d26a573604e427640f9b7839afe2c0070

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -149,17 +149,6 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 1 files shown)


tmpolaczyk:
There should be some way to make CI fail if the Cargo.lock has not been committed, but I can't find an easy way to do it.

tmpolaczyk:
No longer needed because the Cargo.lock was updated by some other PR

PR #1948, opened 2021-05-28
[wip] Include new data request weight

lrubiorod:
Include this PR only if it would be need after the DataRequest fix weight discussion

(wip): it is not complete u_u

### DIFF
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -2429,10 +2429,10 @@ impl TransactionsPool {
                 }
             }
             Transaction::DataRequest(dr_tx) => {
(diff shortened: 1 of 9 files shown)


tmpolaczyk:
This PR has an unexpected problem: the `TransactionsPool` struct has a field `total_dr_weight` which stores the total weight of the data requests in the transactions pool. If the formula to calculate the weight changes, some logic may be wrong (the weight might become negative). So I see two options: use the old weight formula inside the transactions pool, or write some kind of migration to convert weights to the new formula.

lrubiorod:
It is implemented in #1966

PR #1950, opened 2021-05-31
TAPI testing: include another upgrade

tmpolaczyk:


tmpolaczyk:
Closing because the tapi branch has been merged into master in PR #1985, so the next TAPI upgrade will probably be on mainnet.

PR #1953, opened 2021-06-02
feat: temporarily allow microforked peering

aesedepece:
This makes a node tolerate peers that have the same top superblock, even if they are on different blocks, assuming that they will eventually get on the same page at some point after failed superblock consensus.


aesedepece:
This is wild and untested.

mariocao:
Just for the record, the modified function seems to be used only by function `handshake_version` which is also used by the `handle` of the `StreamHandler` for `Session` only for the messages of the kind `Version`. In other words, it should be SAFU. :smile:

tmpolaczyk (approved):

PR #1954, opened 2021-06-02
feat: temporarily use shorter icing periods

aesedepece:
Given the exceptional nature of the current situation, these temporary values are enforced through compiled logic, not configuration.

### DIFF
--- a/config/src/defaults.rs
+++ b/config/src/defaults.rs
@@ -62,7 +62,7 @@ pub trait Defaults {
 
     /// Default period for melt peers
     fn connections_check_melted_peers_period(&self) -> Duration {
-        Duration::from_secs(3600)
+        Duration::from_secs(300)
     }
 
     /// Default period for trying connecting to recently discovered peer addresses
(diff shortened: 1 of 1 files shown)


p2p/src/peers/mod.rs
```diff
@@ -92,8 +92,10 @@ impl Peers {
     /// the process of bootstrapping the network.
     pub fn current_ice_period(&self) -> i64 {
         if self.bootstrapped {
-            i64::try_from(self.ice_period.as_secs())
-                .expect("Ice period should fit in the range of u64")
+            // FIXME: recover the actual logic commented below
+            /*i64::try_from(self.ice_period.as_secs())
+            .expect("Ice period should fit in the range of u64")*/
+            3600
```
mariocao, line 98: As discussed we could put it in the default params :)

aesedepece: Done!

tmpolaczyk (approved):

PR #1968, opened 2021-06-10
Added a flag to fetch simple balance vs confirmed / pending balance

drcpu-github:
Fetching confirmed and pending balance is terribly slow for executing in an automated fashion (e.g. as part of a block explorer or a simple wallet app). I added a flag to the balance RPC call to fall back to the simple balance counter instead of the confirmed / pending balance.

### DIFF
--- a/data_structures/src/transaction_factory.rs
+++ b/data_structures/src/transaction_factory.rs
@@ -25,7 +25,7 @@ pub struct TransactionInfo {
 #[derive(Copy, Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
 pub struct NodeBalance {
     /// Total amount of a node's funds after last confirmed superblock
-    pub confirmed: u64,
+    pub confirmed: Option<u64>,
     /// Total amount of node's funds after last block
     pub total: u64,
 }
@@ -249,6 +249,7 @@ pub fn get_total_balance(
     all_utxos: &UnspentOutputsPool,
     old_all_utxos: Option<&UnspentOutputsPool>,
     pkh: PublicKeyHash,
+    simple: bool,
 ) -> NodeBalance {
     // FIXME: this does not scale, we need to be able to get UTXOs by PKH
     // Get the balance of the current utxo set
@@ -262,6 +263,13 @@ pub fn get_total_balance(
             }
         })
         .sum();
+
+    if simple {
+        return NodeBalance {
+            confirmed: None,
+            total: new_balance,
+        };
+    }
 
     // Get the balance of the confirmed utxo set from storage
     let old_balance = match old_all_utxos {
@@ -279,7 +287,7 @@ pub fn get_total_balance(
     };
 
     NodeBalance {
-        confirmed: old_balance,
+        confirmed: Some(old_balance),
         total: new_balance,
     }
 }
@@ -1195,19 +1203,35 @@ mod tests {
             pay_me(334),
         ];
         let (_own_utxos, all_utxos) = build_utxo_set(outputs, None, vec![]);
-        // If the utxo set from the storage is None it should set the confirmed bbalance to 0
+        // If the utxo set from the storage is None it should set the confirmed balance to 0
         assert_eq!(
-            get_total_balance(&all_utxos, None, own_pkh),
+            get_total_balance(&all_utxos, None, own_pkh, false),
             NodeBalance {
-                confirmed: 0,
+                confirmed: Some(0),
+                total: 1000,
+            }
+        );
+        // When using simple balance, both balances should be 1000
+        assert_eq!(
+            get_total_balance(&all_utxos, None, own_pkh, true),
+            NodeBalance {
+                confirmed: None,
                 total: 1000,
             }
         );
         // Assert the balance is 1000 when the superblock is confirmed
         assert_eq!(
-            get_total_balance(&all_utxos, Some(&all_utxos), own_pkh),
+            get_total_balance(&all_utxos, Some(&all_utxos), own_pkh, false),
             NodeBalance {
-                confirmed: 1000,
+                confirmed: Some(1000),
+                total: 1000,
+            }
+        );
+        // Assert the balance is 1000 when the superblock is confirmed when using simple balance
+        assert_eq!(
+            get_total_balance(&all_utxos, Some(&all_utxos), own_pkh, true),
+            NodeBalance {
+                confirmed: None,
                 total: 1000,
             }
         );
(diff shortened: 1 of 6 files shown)


tmpolaczyk:
So there are two reasons why getBalance is so slow:

* The balance is not indexed, which means that to calculate it we need to iterate over all the UTXOs and sum the ones that belong to the provided PublicKeyHash.
* The distinction of confirmed and unconfirmed balance means that we need to check the UTXOs that existed before the last superblock, and currently this is implemented by reading the old ChainState from storage and iterating again. This will probably be fixed by implementing #1583 or something similar.

This PR fixes the second point, and it's a good idea because most of the time the confirmed balance will be the same as the unconfirmed.

node/src/actors/json_rpc/json_rpc_methods.rs
```diff
@@ -1098,17 +1098,30 @@ pub async fn data_request_report(params: Result<(Hash,), jsonrpc_core::Error>) -
...
+    #[serde(default)] // default to false
+    pub simple: bool,
+}
+
 /// Get balance
-pub async fn get_balance(params: Result<(PublicKeyHash,), jsonrpc_core::Error>) -> JsonRpcResult {
-    let pkh = match params {
-        Ok(x) => x.0,
+pub async fn get_balance(
+    params: Result<Option<GetBalanceParams>, jsonrpc_core::Error>,
```
tmpolaczyk, line 1114: Changing the input type is technically a breaking change, but it may not be a problem because getBalance is only used in the CLI, and we may break the JSON-RPC API in PR #1939 anyway.

But if we want to make this backwards-compatible, this method should take an unparsed `Params`, similar to the `get_block` method, and it should parse the input manually: if params is an array, parse using old logic and set simple to false; if params is an object parse using new logic.

```suggestion
    params: Params,
```

data_structures/src/transaction_factory.rs
```diff
@@ -263,6 +264,13 @@ pub fn get_total_balance(
         })
         .sum();
 
+    if simple {
+        return NodeBalance {
+            confirmed: new_balance,
```
tmpolaczyk, line 269: Instead of duplicating the total balance, the `confirmed` field should be an `Option<u64>`, set to `None` if the confirmed balance is not known.

```suggestion
            confirmed: None,
```

drcpu-github:
> * The balance is not indexed, which means that to calculate it we need to iterate over all the UTXOs and sum the ones that belong to the provided PublicKeyHash.
> * The distinction of confirmed and unconfirmed balance means that we need to check the UTXOs that existed before the last superblock, and currently this is implemented by reading the old ChainState from storage and iterating again. This will probably be fixed by implementing #1583 or something similar.

I actually 'analyzed' where the slowdown is coming from and an overwhelming part comes from fetching the ChainState from storage. Iterating over the UTXOs may not be optimal, but it is nowhere near slow. After the confirmed balance change, fetching balance takes >2s (which is obviously infeasibly slow for tool-based fetching), before the change it took less than 0.1s (on average) if I remember correctly. I'm not quite sure if a change like storing UTXOs in storage will actually help with the speed of a `get_balance` call since it's still fetching from storage, but maybe it does.

I added all of your suggested changes, so when the tests pass, I'll squash them and this PR can be merged.

node/src/actors/json_rpc/json_rpc_methods.rs
```diff
@@ -1098,17 +1098,55 @@ pub async fn data_request_report(params: Result<(Hash,), jsonrpc_core::Error>) -
...
+
+        match params.get(1) {
+            None => simple = false,
+            Some(Value::Bool(ith)) => simple = *ith,
+            Some(_) => {
+                return Err(jsonrpc_core::Error::invalid_params(
+                    "Second argument of `get_balance` must have type `Bool`",
+                ))
+            }
+        };
```
tmpolaczyk, line 1139: I didn't mean to copy get_block exactly, this can be removed

```suggestion
        simple = false;
```

node/src/actors/json_rpc/json_rpc_methods.rs
```diff
@@ -1098,17 +1098,55 @@ pub async fn data_request_report(params: Result<(Hash,), jsonrpc_core::Error>) -
...
+    } else {
+        return Err(jsonrpc_core::Error::invalid_params(
+            "Params of `get_balance` method must have type `Array`",
+        ));
```
tmpolaczyk, line 1143: And here we should parse params as `GetBalanceParams`, otherwise I think in this state the CLI method doesn't work:

```suggestion
        let params: GetBalanceParams = params.parse()?;
        pkh = params.pkh;
        simple = params.simple;
```

tmpolaczyk:
> I'm not quite sure if a change like storing UTXOs in storage will actually help with the speed of a get_balance call since it's still fetching from storage, but maybe it does.

That's a good point, because iterating over something that is in storage should be slower than iterating over something that is in RAM, but I remember testing it and RocksDB is surprisingly fast. The current slowdown is mostly from deserializing the entire ChainState struct when only the `utxo_set` is needed, and that will be skipped if the UTXOs are stored in separate keys.

node/src/actors/json_rpc/json_rpc_methods.rs
```diff
@@ -1098,17 +1098,55 @@ pub async fn data_request_report(params: Result<(Hash,), jsonrpc_core::Error>) -
         .await
 }
 
+/// Params of getBlockChain method
+#[derive(Debug, Default, Deserialize, Serialize)]
+pub struct GetBalanceParams {
+    /// Public key hash
+    #[serde(default)] // default to default pkh
```
tmpolaczyk, line 1105: This field should not use `#[serde(default)]` because it is a required argument.

```suggestion
```

node/src/actors/json_rpc/json_rpc_methods.rs
drcpu-github: I completely misunderstood how that block of code worked, my Rust knowledge is sometimes a bit lacking.

drcpu-github:
```
./witnet -c .witnet/config/escrow.toml node raw

{"jsonrpc": "2.0", "method": "getBalance", "params": ["..."], "id": 1}
{"jsonrpc":"2.0","result":{"confirmed":1966529170,"total":1966529170},"id":1}

{"jsonrpc": "2.0", "method": "getBalance", "params": {"pkh": "...", "simple": true}, "id": 1}
{"jsonrpc":"2.0","result":{"confirmed":null,"total":1966529170},"id":1}

{"jsonrpc": "2.0", "method": "getBalance", "params": {"pkh": "...", "simple": false}, "id": 1}
{"jsonrpc":"2.0","result":{"confirmed":1966529170,"total":1966529170},"id":1}

./witnet -c .witnet/config/witnet.toml node balance

Confirmed balance:   1.966529170 wits
Pending balance:     0.000000000 wits

./witnet -c .witnet/config/witnet.toml node balance --simple

Balance:   1.966529170 wits
```
All of these raw JSON RPC calls + CLI produce the results I expected, so I think this implementation is good. Thanks for the review!

drcpu-github:
> That's a good point, because iterating over something that is in storage should be slower than iterating over something that is in RAM, but I remember testing it and RocksDB is surprisingly fast. The current slowdown is mostly from deserializing the entire ChainState struct when only the `utxo_set` is needed, and that will be skipped if the UTXOs are stored in separate keys.

Yeah, that's the part I'm not too sure about. If the `utxo_set` is a small part of the ChainState, it may result in a noticeably faster completion compared to now, but I would still expect it'll be significantly slower as that data structure is probably quite big. I think this PR is a nice addition overall since it allows for specialization and optimization of balance calls when confirmed vs pending is not required.

tmpolaczyk (approved):

tmpolaczyk:
Thanks for the PR! Don't forget to squash the commits before merging.

drcpu-github:
I squashed the commits, but I cannot merge it myself as far as I know.

PR #1973, opened 2021-06-11
Consensus unknown: set target superblock

### LINKED ISSUE #1963: Improve node behavior in SuperBlockConsensus::Different case
The current behavior when the superblock consensus is different than the node consensus is the following:

* Drop peers with the same highest_block_checkpoint than us
* Restore chain state from storage
* Move to WaitingConsensus

https://github.com/witnet/witnet-rust/blob/641da5bb9f8ede85023bf50c5e07ff51267b35ed/node/src/actors/chain_manager/mod.rs#L1490

Note that we have no guarantees that the new peers will be on the desired chain, so not sure how will the node recover in that case.

I think it's impossible to make the node sync to the target superblock with the current protocol, so I propose a simple change: make sure that the new peers are on the target superblock. The handshake process already includes the superblock beacon, so we could make the SessionsManager only accept peers with the correct superblock beacon. If no valid peers are found after N epochs we can fall back to the old process, but that will have the same problems as before.

So this would be the new behavior:

* Drop peers that are not on the target_superblock_hash
* Only allow to connect to peers that are on target_superblock_hash
* Restore chain state from storage
* Move to WaitingConsensus

tmpolaczyk:
Fix #1963

### DIFF
--- a/node/src/actors/chain_manager/mod.rs
+++ b/node/src/actors/chain_manager/mod.rs
@@ -76,9 +76,10 @@ use crate::{
         json_rpc::JsonRpcServer,
         messages::{
             AddItem, AddItems, AddTransaction, Anycast, BlockNotify, Broadcast, DropOutboundPeers,
-            GetBlocksEpochRange, GetItemBlock, NodeStatusNotify, RemoveAddressesFromTried,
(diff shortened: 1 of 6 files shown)


tmpolaczyk:
Closing because the target superblock feature is being implemented in PR #1975, but we still need to implement "Drop peers that are not on the target_superblock_hash" before closing the issue.

PR #1975, opened 2021-06-11
Improve peering process using SuperBlock consensus information

### LINKED ISSUE #1969: Change block comparison by superblock comparison in handshake

### LINKED ISSUE #1970: Remove icing in Unregister Peers and include icing when needed

### LINKED ISSUE #1971: Include SuperBlock consensus achieved in handshaking

### LINKED ISSUE #1972: Include SuperBlock consensus achieved in PeersBeacons filtering

lrubiorod:
Close #1969 
Close #1971 
Close #1970
Close #1972

### DIFF
--- a/node/src/actors/chain_manager/handlers.rs
+++ b/node/src/actors/chain_manager/handlers.rs
@@ -836,32 +836,21 @@ impl PeersBeacons {
 
     /// Ignore beacons that are old or has a different hash if is the same super epoch
     pub fn ignore_no_consensus_beacons(&mut self, superbeacon: CheckpointBeacon) {
-        let mut valid_peers = vec![];
-
-        for (addr, last_beacon) in &self.pb {
-            if let Some(last_beacon) = last_beacon {
-                // Ignore beacons that would have an old superblock checkpoint
+        for (addr, last_beacon_opt) in &mut self.pb {
+            if let Some(last_beacon) = last_beacon_opt {
+                // Ignore beacons that point to an old superblock checkpoint
                 if last_beacon.highest_superblock_checkpoint.checkpoint < superbeacon.checkpoint
-                    // Ignore beacons that are different from the consensus superblock
+                    // Ignore beacons that do not point to the right consensus superblock
                     || (last_beacon.highest_superblock_checkpoint.checkpoint
                         == superbeacon.checkpoint
                         && last_beacon.highest_superblock_checkpoint != superbeacon)
                 {
-                    // Peers out of consensus there will be handle as no beacon peers
+                    // Peers out of consensus will be treated the same as peers from which we receive no beacon
                     log::debug!("LastBeacon from: {} was set to NO_BEACON", *addr);
-                    valid_peers.push((*addr, None));
-
-                // Do not ignore beacons with newer superblock beacon or with the same as consensus
-                } else {
-                    valid_peers.push((*addr, Some(last_beacon.clone())))
+                    *last_beacon_opt = None;
                 }
-            } else {
-                // No beacons
-                valid_peers.push((*addr, None));
             }
         }
-
-        self.pb = valid_peers;
     }
 
     /// Collects the peers that have not sent us a beacon
@@ -911,7 +900,7 @@ impl Handler<PeersBeacons> for ChainManager {
             .unwrap_or(1);
 
         let peers_with_no_beacon = peers_beacons.peers_with_no_beacon();
-        // Ice peers with no beacon (or out of last superblock consensus)
+        // Ice peers with a beacon that does not point to last consensus superblock, or provided no beacon at all
         for peer in &peers_with_no_beacon {
             self.ice_peer(Some(*peer));
         }
@@ -927,7 +916,7 @@ impl Handler<PeersBeacons> for ChainManager {
         } else if pb_len < peers_needed_for_consensus {
             // Not enough outbound peers, do not unregister any peers
             log::debug!(
-                "Got {} peers after filtering but need at least {} to calculate the consensus",
+                "Got {} peers but need at least {} to calculate the consensus",
                 pb_len,
                 peers_needed_for_consensus
             );
@@ -1137,7 +1126,7 @@ impl Handler<PeersBeacons> for ChainManager {
                         // vote. We will remove those that are different from the last consensus or
                         // peers that has a block different to us
 
-                        Ok(peers_to_unregister)
+                        Ok(peers_with_no_beacon)
                     }
                 }
             }
--- a/node/src/actors/chain_manager/handlers.rs
+++ b/node/src/actors/chain_manager/handlers.rs
@@ -834,6 +834,36 @@ impl PeersBeacons {
             .collect()
     }
 
+    /// Ignore beacons that are old or has a different hash if is the same super epoch
+    pub fn ignore_no_consensus_beacons(&mut self, superbeacon: CheckpointBeacon) {
+        let mut valid_peers = vec![];
+
+        for (addr, last_beacon) in &self.pb {
+            if let Some(last_beacon) = last_beacon {
+                // Ignore beacons that would have an old superblock checkpoint
+                if last_beacon.highest_superblock_checkpoint.checkpoint < superbeacon.checkpoint
+                    // Ignore beacons that are different from the consensus superblock
+                    || (last_beacon.highest_superblock_checkpoint.checkpoint
+                        == superbeacon.checkpoint
+                        && last_beacon.highest_superblock_checkpoint != superbeacon)
+                {
+                    // Peers out of consensus there will be handle as no beacon peers
+                    log::debug!("LastBeacon from: {} was set to NO_BEACON", *addr);
+                    valid_peers.push((*addr, None));
+
+                // Do not ignore beacons with newer superblock beacon or with the same as consensus
+                } else {
+                    valid_peers.push((*addr, Some(last_beacon.clone())))
+                }
+            } else {
+                // No beacons
+                valid_peers.push((*addr, None));
+            }
+        }
+
+        self.pb = valid_peers;
+    }
+
     /// Collects the peers that have not sent us a beacon
     pub fn peers_with_no_beacon(&self) -> Vec<SocketAddr> {
         // Unregister peers which have not sent us a beacon
@@ -849,7 +879,7 @@ impl Handler<PeersBeacons> for ChainManager {
 
     // FIXME(#676): Remove clippy skip error
     #[allow(clippy::cognitive_complexity)]
-    fn handle(&mut self, peers_beacons: PeersBeacons, ctx: &mut Context<Self>) -> Self::Result {
+    fn handle(&mut self, mut peers_beacons: PeersBeacons, ctx: &mut Context<Self>) -> Self::Result {
         log::debug!(
             "PeersBeacons received while StateMachine is in state {:?}",
             self.sm_state
@@ -860,6 +890,11 @@ impl Handler<PeersBeacons> for ChainManager {
         // Activate peers beacons index to continue synced
         if !peers_beacons.pb.is_empty() {
             self.peers_beacons_received = true;
+        }
+
+        // Remove beacons that not are compatible with the last superblock consensus
+        if let Some(last_superblock_consensus) = self.last_superblock_consensus {
+            peers_beacons.ignore_no_consensus_beacons(last_superblock_consensus);
         }
 
         // Calculate the consensus, or None if there is no consensus
@@ -874,7 +909,13 @@ impl Handler<PeersBeacons> for ChainManager {
                 (usize::from(x) * consensus_threshold + 99) / 100
             })
             .unwrap_or(1);
+
         let peers_with_no_beacon = peers_beacons.peers_with_no_beacon();
+        // Ice peers with no beacon (or out of last superblock consensus)
+        for peer in &peers_with_no_beacon {
+            self.ice_peer(Some(*peer));
+        }
+
         let peers_to_unregister = if let Some((consensus, is_there_block_consensus)) =
             beacon_consensus.as_ref()
         {
@@ -886,7 +927,7 @@ impl Handler<PeersBeacons> for ChainManager {
         } else if pb_len < peers_needed_for_consensus {
             // Not enough outbound peers, do not unregister any peers
             log::debug!(
-                "Got {} peers but need at least {} to calculate the consensus",
+                "Got {} peers after filtering but need at least {} to calculate the consensus",
                 pb_len,
                 peers_needed_for_consensus
             );
@@ -1070,7 +1111,8 @@ impl Handler<PeersBeacons> for ChainManager {
 
                         // If we are synced, it does not matter what blocks have been consolidated
                         // by our outbound peers, we will stay synced until the next superblock
-                        // vote
+                        // vote. We will remove those that are different from the last consensus or
+                        // peers that did not send any beacon
 
                         Ok(peers_with_no_beacon)
                     }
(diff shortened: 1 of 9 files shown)


node/src/actors/chain_manager/mod.rs
```diff
@@ -1513,6 +1514,12 @@ impl ChainManager {
                     let sessions_manager_addr = SessionsManager::from_registry();
                     sessions_manager_addr.do_send(DropOutboundPeers {peers_to_drop: peers_to_unregister});
 
+                    // Include superblock beacon target in SessionsManager
+                    sessions_manager_addr.do_send(SetSuperBlockTargetBeacon {beacon: CheckpointBeacon{
+                        checkpoint: superblock_index,
+                        hash_prev_block: target_superblock_hash,
+                    }});
```
tmpolaczyk, line 1521: This only sets the target superbeacon when the consensus is `SuperBlockConsensus::Different`. Should we also set it in `SuperBlockConsensus::SameAsLocal`?

tmpolaczyk: And what about the other cases? When there is no superblock consensus, should we set it to `None`, or set it to the last consolidated superblock?

lrubiorod: Sure!! In the other cases it should be set to None, to avoid using old target beacons

node/src/actors/chain_manager/mod.rs
```diff
@@ -1485,6 +1486,10 @@ impl ChainManager {
                         .superblock_state
                         .set_current_superblock(superblock.clone());
 
+                    // Remove superblock beacon target in SessionsManager
+                    let sessions_manager_addr = SessionsManager::from_registry();
+                    sessions_manager_addr.do_send(SetSuperBlockTargetBeacon {beacon: None});
```
tmpolaczyk, line 1491: Why `None` here, if there is consensus?

lrubiorod: When you put None, you will use your own SuperBlock Beacon, so if your are in consensus, your beacon is the same than the consensus beacon

tmpolaczyk: True, let's add a comment explaining that

node/src/actors/chain_manager/mod.rs
```diff
@@ -1513,6 +1518,12 @@ impl ChainManager {
                     let sessions_manager_addr = SessionsManager::from_registry();
                     sessions_manager_addr.do_send(DropOutboundPeers {peers_to_drop: peers_to_unregister});
 
+                    // Include superblock beacon target in SessionsManager
+                    sessions_manager_addr.do_send(SetSuperBlockTargetBeacon {beacon: Some(CheckpointBeacon{
+                        checkpoint: superblock_index,
```
tmpolaczyk, line 1523:
```suggestion
                        checkpoint: voted_superblock_beacon.checkpoint,
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -69,6 +69,7 @@ use witnet_validations::validations::{
     validate_new_transaction, verify_signatures, VrfSlots,
 };
 
+use crate::actors::messages::SetSuperBlockTargetBeacon;
```
aesedepece, line 72: Unify imports.

node/src/actors/session/handlers.rs
```diff
@@ -179,14 +179,48 @@ impl StreamHandler<Result<BytesMut, Error>> for Session {
...
+                                    || session_type == SessionType::Outbound
+                                {
                                     let peers_manager_addr = PeersManager::from_registry();
-                                    // Ice the peer that was an error
-                                    peers_manager_addr.do_send(RemoveAddressesFromTried {
-                                        addresses: vec![self.remote_addr],
-                                        ice: true,
-                                    });
+                                    // Ice the peer that was an error (except epochs where superblock is updated)
+                                    if self.current_epoch % 10 != 0 {
```
aesedepece, line 215: I would appreciate some color on why this is needed / what exactly we want to prevent by introducing this check,

aesedepece (changes requested):

node/src/actors/chain_manager/handlers.rs
```diff
@@ -834,6 +834,35 @@ impl PeersBeacons {
             .collect()
     }
 
+    /// Remove beacons that are old or has a different hash if is the same super epoch
+    pub fn remove_no_consensus_beacons(&mut self, superbeacon: CheckpointBeacon) {
+        let mut valid_peers = vec![];
+
+        for (addr, last_beacon) in &self.pb {
```
tmpolaczyk, line 841: I think this for loop can be rewritten to mutate the elements in place:

```suggestion
        for (addr, last_beacon) in &mut self.pb {
```

The only thing that is needed is to set the last_beacon to None in some cases, so this should work:

    *last_beacon = None;

node/src/actors/chain_manager/handlers.rs
```diff
@@ -834,6 +834,35 @@ impl PeersBeacons {
             .collect()
     }
 
+    /// Remove beacons that are old or has a different hash if is the same super epoch
+    pub fn remove_no_consensus_beacons(&mut self, superbeacon: CheckpointBeacon) {
```
tmpolaczyk, line 838: Remove is a misleading name because the beacons are simply set to "None". A better word would be "ignore".

node/src/actors/chain_manager/handlers.rs
```diff
@@ -1092,9 +1133,10 @@ impl Handler<PeersBeacons> for ChainManager {
 
                         // If we are synced, it does not matter what blocks have been consolidated
                         // by our outbound peers, we will stay synced until the next superblock
-                        // vote
+                        // vote. We will remove those that are different from the last consensus or
+                        // peers that has a block different to us
 
-                        Ok(peers_with_no_beacon)
```
tmpolaczyk, line 1097: This change can have some big implications, any context on why it is needed?

node/src/actors/chain_manager/mod.rs
```diff
@@ -1485,6 +1488,23 @@ impl ChainManager {
...
 
+                    // Set last beacon in sessions manager
+                    let sessions_manager_addr = SessionsManager::from_registry();
+                    let chain_beacon = act.get_chain_beacon();
+                    sessions_manager_addr.do_send(SetLastBeacon {
+                        beacon: LastBeacon{
+                            highest_block_checkpoint: chain_beacon,
+                            highest_superblock_checkpoint: voted_superblock_beacon,
+                        },
+                    });
```
tmpolaczyk, line 1499: I opened issue #1977 to assess if this is safe

node/src/actors/session/handlers.rs
```diff
@@ -180,13 +180,33 @@ impl StreamHandler<Result<BytesMut, Error>> for Session {
...
+                                        addresses: vec![self.remote_addr],
+                                        ice: true,
+                                    });
+                                } else if session_type == SessionType::Feeler
+                                    || session_type == SessionType::Outbound
+                                {
+                                    // Ice the peer that was an error, except epochs where superblock is updated.
+                                    // During superblock updates, there will be nodes perfectly synced that they
+                                    // are in the process of updating the superblock field for handshaking, and
+                                    // icing them would be an error.
```
aesedepece, line 200:
```suggestion
                                    // Ice the peer who failed to complete a successful handshake, except in epochs
                                    // where superblock is updated.
                                    // During superblock updates, there will be nodes that are perfectly synced
                                    // yet they are in the process of updating their superblock field for handshaking,
                                    // so mistakenly icing them would be wrong.
```

node/src/actors/sessions_manager/mod.rs
```diff
@@ -50,6 +50,8 @@ pub struct SessionsManager {
     logging_messages: HashSet<String>,
     // Reference to config
     config: Option<Arc<Config>>,
+    // SuperBlock target
+    superblock_beacon_target: Option<CheckpointBeacon>,
```
aesedepece, line 54: I wonder why the members of this `enum` are not documented with `///` :thinking:

node/src/actors/chain_manager/handlers.rs
```diff
@@ -834,6 +834,35 @@ impl PeersBeacons {
             .collect()
     }
 
+    /// Remove beacons that are old or has a different hash if is the same super epoch
+    pub fn remove_no_consensus_beacons(&mut self, superbeacon: CheckpointBeacon) {
+        let mut valid_peers = vec![];
+
+        for (addr, last_beacon) in &self.pb {
+            if let Some(last_beacon) = last_beacon {
+                // Remove beacons that would have an old superblock checkpoint
```
aesedepece, line 843:
```suggestion
                // Remove beacons that point to an old superblock checkpoint
```

node/src/actors/chain_manager/handlers.rs
```diff
@@ -834,6 +834,35 @@ impl PeersBeacons {
...
+                if last_beacon.highest_superblock_checkpoint.checkpoint < superbeacon.checkpoint
+                    // Remove beacons that are different from the consensus superblock
```
aesedepece, line 845:
```suggestion
                    // Remove beacons that do not point to the right consensus superblock
```

node/src/actors/chain_manager/handlers.rs
```diff
@@ -834,6 +834,35 @@ impl PeersBeacons {
...
+                    || (last_beacon.highest_superblock_checkpoint.checkpoint
+                        == superbeacon.checkpoint
+                        && last_beacon.highest_superblock_checkpoint != superbeacon)
+                {
+                    // Peers out of consensus there will be handle as no beacon peers
```
aesedepece, line 850:
```suggestion
                    // Peers out of consensus will be treated the same peers from which we receive no beacon
```

node/src/actors/chain_manager/handlers.rs
```diff
@@ -834,6 +834,35 @@ impl PeersBeacons {
...
+                    valid_peers.push((*addr, None));
+
+                // Do not remove beacons with newer superblock beacon or with the same as consensus
```
aesedepece, line 853:
```suggestion
                // Keep the beacons that point to superblocks that are greater than the local one or the same as consensus
```

node/src/actors/chain_manager/handlers.rs
```diff
@@ -834,6 +834,35 @@ impl PeersBeacons {
...
+                } else {
+                    valid_peers.push((*addr, Some(last_beacon.clone())))
+                }
+            } else {
+                // No beacons
```
aesedepece, line 858:
```suggestion
                // No beacon received from this peer
```

node/src/actors/chain_manager/handlers.rs
```diff
@@ -874,7 +908,13 @@ impl Handler<PeersBeacons> for ChainManager {
                 (usize::from(x) * consensus_threshold + 99) / 100
             })
             .unwrap_or(1);
+
         let peers_with_no_beacon = peers_beacons.peers_with_no_beacon();
+        // Ice peers with no beacon (or out of last superblock consensus)
```
aesedepece, line 913:
```suggestion
        // Ice peers with a beacon that does not point to last consensus superblock, or provided no beacon at all
```

node/src/actors/chain_manager/handlers.rs
```diff
@@ -886,7 +926,7 @@ impl Handler<PeersBeacons> for ChainManager {
         } else if pb_len < peers_needed_for_consensus {
             // Not enough outbound peers, do not unregister any peers
             log::debug!(
-                "Got {} peers but need at least {} to calculate the consensus",
+                "Got {} peers after filtering but need at least {} to calculate the consensus",
```
aesedepece, line 929: After filtering by what criteria?

aesedepece (changes requested):

node/src/actors/chain_manager/handlers.rs
lrubiorod: When we are synced and there is no consensus, we select as peers_to_unregister those that are different to us. But later we only use the peers_with_no_beacon. This change is to use the first part of the code.

node/src/actors/chain_manager/handlers.rs
lrubiorod: Finally I remove the filter, so this comment modification is wrong

node/src/actors/chain_manager/handlers.rs
mariocao: shouldn't it be "will be treated as..."?

aesedepece: Oops, yes, I meant "treated the same as peers...."

node/src/actors/session/handlers.rs
```diff
@@ -794,31 +821,50 @@ fn check_beacon_compatibility(
         });
     }
 
-    match current_beacon
-        .highest_block_checkpoint
+    let my_beacon = if let Some(target_beacon) = target_superblock_beacon {
+        target_beacon
+    } else {
+        current_beacon.highest_superblock_checkpoint
+    };
```
aesedepece, line 828: Not a big blocker, but I'd appreciate a comment on what's the rationale behind this new argument and this statement.

node/src/actors/session/handlers.rs
```diff
@@ -794,31 +821,50 @@ fn check_beacon_compatibility(
...
-            current_beacon: current_beacon.clone(),
-            received_beacon: received_beacon.clone(),
-        }),
+        Ordering::Greater => {
+            log::debug!(
+                "Current SuperBlock beacon: {:?}",
+                current_beacon.highest_superblock_checkpoint
+            );
+            log::debug!("Target SuperBlock beacon: {:?}", my_beacon);
+
```
tmpolaczyk, line 843: Remove before merging!
```suggestion
```

node/src/actors/session/handlers.rs
tmpolaczyk: @lrubiorod you can summarize this into a comment:

Set the target superblock beacon.

When the superblock consensus is different from the superblock created by this node, the process is to revert and ask the outbound peers for new blocks. We use the target beacon to ensure that the outbound peers have the correct superblock.

Note that this target is only used if the received_beacon checkpoint is equal to the target superblock checkpoint: this means that this check is only valid during one superepoch, if the node fails to find enough peers with the target superblock beacon, it will fall back to the old behavior and allow connections to any peers, ignoring the target superblock beacon.

lrubiorod: Not really, imagine that the SuperBlock 24 is achieved with the hash "aaaa..." by more than 2/3 of the signing committee, but you are in a fork so... you will revert and your last_superblock consolidated is 23 but your superblock_target is 24. If during the next handshakings the other peers send a LastBeacon with the superblock index 23 or index 24 but with different hash to "aaaa...", they will be rejected.

It only works with the old behaviour when it doesn't have a target, and it happens in "Same as local", "Not Reached" and "Unknown"

node/src/actors/chain_manager/handlers.rs
tmpolaczyk: When we are synced and there is no consensus, all of the other peers may be different, so by dropping all the peers we will have worse connectivity... Ideally we should only drop the "bad" peers but we have no way to define "bad".

node/src/actors/session/handlers.rs
lrubiorod: Right

node/src/actors/session/actor.rs
```diff
@@ -31,6 +31,11 @@ impl Actor for Session {
                     act.remote_addr
                 );
```
tmpolaczyk, line 32: Above this line there is a "Handshake timeout" log in INFO level, it should probably be DEBUG?

mariocao (changes requested):

mariocao:
For ease of understanding the changes of this PR, in red you can see some of the changes regarding creating superblocks.

Diagram1

mariocao (approved):

node/src/actors/chain_manager/handlers.rs
aesedepece: Ping @lrubiorod

node/src/actors/session/handlers.rs
aesedepece: Ping @lrubiorod

aesedepece (changes requested):

tmpolaczyk (approved):

aesedepece (approved):

PR #1976, opened 2021-06-11
Remove threshold factor from reppoe

### LINKED ISSUE #1960: Review if maximum DR eligibility is working properly
In the commit bd03469652cfb6007917c83f38e66bd201b78510 of the `202106-recovery` branch, a new maximum data request eligibility was included. However, some logs were seen by @tmpolaczyk stating that the probability for mining a data request was 100%.

Some of the tasks that should be covered by this issue:
- [x] Code review of the DR maximum eligibility logic
- [x] Implement TODO at `node/src/actors/chain_manager/mining.rs:325` (pass difficulty as argument)
- [x] Implement some tests that check new DR maximum eligibility logic (after third hard-fork)
- [x] Identify and fix the problem of the mistmatch between the new maximum DR elelibility and the 100% prob. log

tmpolaczyk:
Implement https://github.com/witnet/WIPs/pull/58

This branch is based on the tapi2 branch, however the automated WIP activation is not handled in this PR (`active_wips.wip0016()` will always return false).

Close #1960

### DIFF
--- a/validations/src/tests.rs
+++ b/validations/src/tests.rs
@@ -2326,6 +2326,7 @@ fn test_empty_commit(c_tx: &CommitTransaction) -> Result<(), failure::Error> {
     let collateral_minimum = 1;
     let collateral_age = 1;
     let block_number = 0;
+    let minimum_reppoe_difficulty = 1;
     let utxo_diff = UtxoDiff::new(&utxo_set, block_number);
 
     validate_commit_transaction(
@@ -2340,6 +2341,7 @@ fn test_empty_commit(c_tx: &CommitTransaction) -> Result<(), failure::Error> {
         collateral_minimum,
         collateral_age,
         block_number,
+        minimum_reppoe_difficulty,
         &all_wips_active(),
     )
     .map(|_| ())
@@ -2356,6 +2358,7 @@ fn test_commit_with_dr_and_utxo_set(
     let utxo_diff = UtxoDiff::new(utxo_set, 0);
     let collateral_minimum = 1;
     let collateral_age = 1;
+    let minimum_reppoe_difficulty = 1;
 
     let mut dr_pool = DataRequestPool::default();
     let vrf_input = CheckpointVRF::default();
@@ -2392,6 +2395,7 @@ fn test_commit_with_dr_and_utxo_set(
         collateral_minimum,
         collateral_age,
         block_number,
+        minimum_reppoe_difficulty,
         &all_wips_active(),
     )?;
     verify_signatures_test(signatures_to_verify)?;
@@ -2452,6 +2456,7 @@ fn test_commit_difficult_proof() {
     let block_number = 100_000;
     let collateral_minimum = 1;
     let collateral_age = 1;
+    let minimum_reppoe_difficulty = 1;
     let utxo_diff = UtxoDiff::new(&utxo_set, 0);
     let vti = Input::new(utxo_set.iter().next().unwrap().0.clone());
 
@@ -2481,6 +2486,7 @@ fn test_commit_difficult_proof() {
         collateral_minimum,
         collateral_age,
         block_number,
+        minimum_reppoe_difficulty,
         &active_wips,
     )
     .and_then(|_| verify_signatures_test(signatures_to_verify));
@@ -2534,6 +2540,7 @@ fn test_commit_with_collateral(
     let collateral_minimum = 1;
     let collateral_age = 10;
     let utxo_diff = UtxoDiff::new(&utxo_set, 0);
+    let minimum_reppoe_difficulty = 1;
 
     let (inputs, outputs) = collateral;
     cb.collateral = inputs;
@@ -2555,6 +2562,7 @@ fn test_commit_with_collateral(
         collateral_minimum,
         collateral_age,
         block_number,
+        minimum_reppoe_difficulty,
         &all_wips_active(),
     )
     .map(|_| ())
@@ -2759,6 +2767,7 @@ fn commitment_invalid_proof() {
     let block_number = 100_000;
     let collateral_minimum = 1;
     let collateral_age = 1;
+    let minimum_reppoe_difficulty = 1;
     let utxo_diff = UtxoDiff::new(&utxo_set, 0);
     let vti = Input::new(utxo_set.iter().next().unwrap().0.clone());
 
@@ -2798,6 +2807,7 @@ fn commitment_invalid_proof() {
         collateral_minimum,
         collateral_age,
         block_number,
+        minimum_reppoe_difficulty,
         &all_wips_active(),
     )
     .and_then(|_| verify_signatures_test(signatures_to_verify));
@@ -2819,6 +2829,7 @@ fn commitment_dr_in_reveal_stage() {
     let block_number = 0;
     let collateral_minimum = 1;
     let collateral_age = 1;
+    let minimum_reppoe_difficulty = 1;
     let utxo_diff = UtxoDiff::new(&utxo_set, block_number);
 
     let mut dr_pool = DataRequestPool::default();
@@ -2872,6 +2883,7 @@ fn commitment_dr_in_reveal_stage() {
         collateral_minimum,
         collateral_age,
         block_number,
+        minimum_reppoe_difficulty,
         &all_wips_active(),
     );
     assert_eq!(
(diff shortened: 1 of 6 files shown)


validations/src/tests.rs
```diff
@@ -2326,6 +2326,7 @@ fn test_empty_commit(c_tx: &CommitTransaction) -> Result<(), failure::Error> {
     let collateral_minimum = 1;
     let collateral_age = 1;
     let block_number = 0;
+    let minimum_reppoe_difficulty = 1;
```
mariocao, line 2329: Smart to avoid chainging the tests! :smile:

tmpolaczyk: The idea is to keep old tests working exactly like before and add new tests checking the new formula, but I don't know how to do that in a maintainable way.

validations/src/validations.rs
```diff
@@ -3569,3 +3388,665 @@ mod tests {
...
+        let mut rep_engine = ReputationEngine::new(1000);
+        let id1 = PublicKeyHash::from_bytes(&[1; 20]).unwrap();
+        rep_engine
+            .trs_mut()
+            .gain(Alpha(10), vec![(id1, Reputation(50))])
+            .unwrap();
+        rep_engine.ars_mut().push_activity(vec![id1]);
+
+        // 0.05% when the total reputation is less than 2000
+        let (t00, p00) = calculate_reppoe_threshold_v2(&rep_engine, &id1, 1, 2000);
```
mariocao, line 3524:
```suggestion
        let (t00, _p00) = calculate_reppoe_threshold_v2(&rep_engine, &id1, 1, 2000);
```

validations/src/validations.rs
```diff
@@ -3569,3 +3388,665 @@ mod tests {
...
+        assert_eq!(t01, Hash::with_first_u32(0xFFFF_FFFF / 2000));
+        // 1 / 2000 = 500 / 1_000_000
+        assert_eq!((p01 * 1_000_000_f64).round() as i128, 500);
+    }
+
+    // FIXME: Allow for now, wait for https://github.com/rust-lang/rust/issues/67058 to reach stable
+    #[allow(clippy::cast_possible_truncation)]
+    #[test]
+    fn reppoe_overflow_v3() {
+        // Test the behavior of the algorithm when our node has 0 reputation
```
mariocao, line 4036:
```suggestion
        // Test the behavior of the algorithm when our node has much reputation
```

validations/src/validations.rs
```diff
@@ -3569,3 +3388,665 @@ mod tests {
...
+            .trs_mut()
+            .gain(Alpha(10), vec![(id0, Reputation(u32::max_value() - 2))])
+            .unwrap();
+
+        // Test big values that result in < 100%
+        let (t01, p01) = calculate_reppoe_threshold_v3(&rep_engine, &id0, 1, 2000);
+        assert_eq!(t01, Hash::with_first_u32(0xFFFF_FFFE));
+        assert_eq!((p01 * 100_f64).round() as i128, 100);
+    }
+}
```
mariocao, line 4052: This test was mind blowing for me because despite of being correct, it represents a network condition that is highly unlikely.

It is important to remember that the minimum difficulty is only activated in case of total active reputation is below 2000:

```
let dr_eligibity = identity_eligibility / max(total_active_reputation, minimum_difficulty) * replication_factor
```

This test shows a situation in which there is only 1 active identity with most of reputation. Therefore, the `identity_eligilibity` (aka "trapezoid") is not coming into play as there is no other active reputed identity to redistribute eligibility. IMHO, this case should be a big problem because it would be likely to happen only after several rollbacks. It will be automatically solved after a data request with a replicacion factor greater than 1.

It would be nice to have some other tests checking a situation with more than 1 active reputed identity.

tmpolaczyk: > IMHO, this case should be a big problem

I assume you mean "should not"

Actually, this test has 2 identities, but 1 of them has 0 reputation so it is not taken into account in the trapezoid. I will add more tests.

tmpolaczyk:
I want to add one test more for the worst case scenario: a data request with the maximum number of witnesses allowed by the current block size, during the last commit round (to take into account backup witnesses), and ensure that the probability for nodes with no reputation is reasonable.

validations/src/validations.rs
```diff
@@ -3569,3 +3388,1162 @@ mod tests {
...
+        let id1 = PublicKeyHash::from_bytes(&[1; 20]).unwrap();
+
+        let (_t00, p00) = calculate_reppoe_threshold_v3(
+            &rep_engine,
+            &id1,
+            max_witnesses + max_backup_witnesses,
+            minimum_difficulty,
+        );
+
+        assert_eq!((p00 * 100_f64).round() as i128, 32);
```
tmpolaczyk, line 4547: This is a test that checks the worst case reppoe probability in mainnet with an empty ARS, a data request with the maximum number of witnesses allowed by the block weight limits, and taking into account the extra commit rounds.

The result is a data request with 126 witnesses (630 after adding the backup witnesses), where a node with no reputation has a 32% chance to create a valid commitment. This 32% will only be used during one epoch, if the previous commit rounds resulted in fewer commits than expected. Here are the probabilities of the other rounds for comparison:

```
32% for round 4
19% for round 3
13% for round 2
9% for round 1
```

Also, it is not possible to include other data requests while a data request with maximum weight is being resolved, so in order to achieve a "round 4" probability in every epoch, the data request would need have around 30 witnesses, to simultaneously have 4 active requests, resulting in a 7% probability in round 4 and proportionally smaller probabilities in the other rounds.

tmpolaczyk:
Github closed this PR because I merged tapi2 into master

tmpolaczyk:
> Github closed this PR because I merged tapi2 into master

And now it thinks that I deleted the master branch:

Captura de pantalla de 2021-06-22 16-01-26

Don't panic, I didn't delete it, I deleted the tapi2 branch which was the old base branch.

validations/src/validations.rs
```diff
@@ -3569,3 +3388,1162 @@ mod tests {
         assert_eq!(a.slot(&h3), 2);
     }
 }
+
+#[cfg(test)]
+mod reppoe_tests {
+    use super::*;
+    use witnet_data_structures::chain::{Alpha, Environment};
```
lrubiorod, line 3395: fmt

validations/src/validations.rs
```diff
@@ -3569,3 +3388,1162 @@ mod tests {
...
+            .unwrap();
+    }
+
+    // FIXME: Allow for now, wait for https://github.com/rust-lang/rust/issues/67058 to reach stable
+    #[allow(clippy::cast_possible_truncation)]
+    #[test]
+    fn target_reppoe_v1() {
+        let mut rep_engine = ReputationEngine::new(1000);
+        let id1 = PublicKeyHash::from_bytes(&[1; 20]).unwrap();
+        rep_engine
```
lrubiorod, line 3489: why you dont use the auxiliary add_rep function that you create before

validations/src/validations.rs
```diff
@@ -3569,3 +3388,1162 @@ mod tests {
...
+            .gain(Alpha(10), vec![(id0, Reputation(u32::max_value() - 2))])
+            .unwrap();
+
+        // Test big values that result in < 100%
+        // Active identity with 100% of the reputation
+        let (t01, p01) = calculate_reppoe_threshold_v1(&rep_engine, &id0, 1, 2000);
+        assert_eq!(t01, Hash::with_first_u32(0xFFFF_FFFE));
+        assert_eq!((p01 * 100_f64).round() as i128, 100);
+        // Active identity with 0 reputation
+        let (t02, p02) = calculate_reppoe_threshold_v3(&rep_engine, &id1, 1, 2000);
```
lrubiorod, line 4039: should it be v1?

validations/src/validations.rs
```diff
@@ -3569,3 +3388,1162 @@ mod tests {
...
+            .unwrap();
+
+        // Test big values that result in < 100%
+        // Active identity with 100% of the reputation
+        let (t01, p01) = calculate_reppoe_threshold_v2(&rep_engine, &id0, 1, 2000);
+        assert_eq!(t01, Hash::with_first_u32(0xFFFF_FFFF / 2000));
+        // 1 / 2000 = 500 / 1_000_000
+        assert_eq!((p01 * 1_000_000_f64).round() as i128, 500);
+        // Active identity with 0 reputation
+        let (t02, p02) = calculate_reppoe_threshold_v3(&rep_engine, &id1, 1, 2000);
```
lrubiorod, line 4085: and v2 here?

validations/src/validations.rs
tmpolaczyk: :scream: 

Yes...

lrubiorod (approved):

PR #1994, opened 2021-06-24
More sincronization improvements

lrubiorod:


node/src/actors/chain_manager/mod.rs
```diff
@@ -909,8 +909,8 @@ impl ChainManager {
                         let mut rng = rand::thread_rng();
                         let checkpoints_period = act.consensus_constants().checkpoints_period;
                         let superblock_period = act.consensus_constants().superblock_period;
-                        let end_range = if superblock_period > 4 {
-                            (superblock_period - 3) * checkpoints_period
+                        let end_range = if superblock_period > 5 {
+                            (superblock_period - 5) * checkpoints_period
```
tmpolaczyk, line 913: There is a comment a few lines above this that still says

// Broadcast vote between one and ("superblock_period" - 3) epoch checkpoints later.

tmpolaczyk: According to the documentation, `rng.gen_range`

> Panics if low >= high.

So we need to ensure that `end_range` is `>` than `checkpoints_period`. After this change it can be equal, when `superblock_period = 6`.

node/src/actors/chain_manager/handlers.rs
```diff
@@ -1100,10 +1115,22 @@ impl Handler<PeersBeacons> for ChainManager {
...
-                        // vote. We will remove those that are different from the last consensus or
-                        // peers that did not send any beacon
+                        // vote. The only exception is if there are a block consensus (more than 2/3)
+                        // and we didn't consolidate any block in the last epoch, because it could
+                        // be that we lost a candidate by any reason
+                        if is_there_block_consensus
+                            && consensus_beacon.checkpoint == current_epoch - 1
+                            && our_beacon.checkpoint != current_epoch - 1
+                        {
+                            self.update_state_machine(StateMachine::WaitingConsensus);
```
tmpolaczyk, line 1125: Not sure if it's safe to move to WaitingConsensus without reverting the storage.

lrubiorod: This corner case is when you miss one block consolidation (by any reason), and your outbounds no. So in this specific case, you are only one block behind, you would not need to revert your chain

tmpolaczyk: You assume that you will be able to synchronize to a chain that includes that block, but the network may revert and that block may no longer exist, in which case you will get a "your chain is probably forked" message. And in that case the node will revert. So I guess it's not a problem, but I'm still a bit unsure.

node/src/actors/chain_manager/handlers.rs
```diff
@@ -196,7 +196,7 @@ impl Handler<EpochNotification<EveryEpochPayload>> for ChainManager {
                         self.transactions_pool.clear_commits();
 
                         // Mining
-                        if self.mining_enabled {
+                        if self.mining_enabled && self.sm_state == StateMachine::Synced {
```
tmpolaczyk, line 199: What happens if we don't check the current state?

lrubiorod: now you can mine data requests if you are in AlmostSynced or Synced, but only mining when you are Synced. The idea is that mining blocks and mining requests has the same requirement, to be in Synced state

tmpolaczyk: With this change we lose the following log lines when the node is not synced:

```text
1 data requests for this epoch
Reputation: 1, eligibility: 1, total: 1066769, active identities: 18260
```

lrubiorod: an affordable lost :P

tmpolaczyk: > now you can mine data requests if you are in AlmostSynced or Synced, but only mining blocks when you are Synced. The idea is that mining blocks and mining requests has the same requirement, to be in Synced state

I thought that the only reason why there is an `AlmostSynced` state is to avoid forks because nodes would mine a block and consolidate it instead of consolidating other candidates. There is no problem in creating commit transactions, right? The worst that could happen is that it is invalid, so they will only be broadcasted to the peers of this node, and those peers will detect it as invalid and stop broadcasting it.

Maybe we can even remove the `AlmostSynced` state?

node/src/actors/chain_manager/handlers.rs
```diff
@@ -886,6 +886,11 @@ impl Handler<PeersBeacons> for ChainManager {
             peers_beacons.ignore_no_consensus_beacons(last_superblock_consensus);
         }
 
+        if self.current_epoch.is_none() {
+            return Err(());
+        }
+        let current_epoch = self.current_epoch.unwrap();
```
mariocao, line 892: How is this possible? Because of an uninitialized the `EpochManager`?

node/src/actors/chain_manager/handlers.rs
```diff
@@ -926,25 +931,26 @@ impl Handler<PeersBeacons> for ChainManager {
             // Else, unregister all peers
             if self.sm_state == StateMachine::AlmostSynced || self.sm_state == StateMachine::Synced
             {
-                log::warn!("Lack of peer consensus while state is `AlmostSynced`: peers that do not coincide with our last beacon will be unregistered");
+                log::warn!("Lack of peer consensus while state is {:?}: peers that do not coincide with our last beacon will be unregistered", self.sm_state);
```
mariocao, line 934: A nice misleading trace! :smile:

node/src/actors/chain_manager/handlers.rs
```diff
@@ -886,6 +886,11 @@ impl Handler<PeersBeacons> for ChainManager {
...
         }
 
+        if self.current_epoch.is_none() {
+            return Err(());
+        }
+        let current_epoch = self.current_epoch.unwrap();
+
         // Calculate the consensus, or None if there is no consensus
         let consensus_threshold = self.consensus_c as usize;
         let beacon_consensus = peers_beacons.superblock_consensus(consensus_threshold);
```
mariocao, line 896: I know that it is not parat of this PR, but in the function `superblock_consensus` does not describe in the code what is the `bool` :)

node/src/actors/chain_manager/handlers.rs
```diff
@@ -1087,23 +1099,40 @@ impl Handler<PeersBeacons> for ChainManager {
...
-                        Ok(peers_with_no_beacon)
+                            Ok(peers_to_unregister)
+                        // In the rest of cases we will move to AlmostSynced to do not allow
+                        // mining and preserve network stability
+                        } else {
+                            self.update_state_machine(StateMachine::AlmostSynced);
+
+                            // We will remove those that are different from the last consensus or
+                            // peers that did not send any beacon
+
```
mariocao, line 1133:
```suggestion
```

node/src/actors/chain_manager/handlers.rs
tmpolaczyk: It's called `is_there_block_consensus`, returns true if the most common superblock has more than `consensus_threshold` % of consensus.

node/src/actors/chain_manager/handlers.rs
lrubiorod: Remove 'AlmostSynced' is not trivial

node/src/actors/chain_manager/handlers.rs
lrubiorod: yes

node/src/actors/chain_manager/handlers.rs
tmpolaczyk: Anyway, I would prefer to not include this one line.

node/src/actors/chain_manager/handlers.rs
lrubiorod: most common superblock and most common block

lrubiorod:
Synchronization_modified

node/src/actors/chain_manager/handlers.rs
mariocao: It is possible to test this somehow?

mariocao: After discussing it with @lrubiorod and @tmpolaczyk , it should not be a problem because those nodes will recover after they receive a different superblock.

mariocao (approved):

tmpolaczyk (approved):

PR #1997, opened 2021-06-25
refactor(mining): improve dr mining moving VRF proof after some valid…

### LINKED ISSUE #1991: Improve try_mine_data_request to take into account available UTXOs before creating VRFs
The `try_mine_data_request` function is used to create commits ("data request mining"):

https://github.com/witnet/witnet-rust/blob/734f5f6438873dd95f4b89f4ae8b7a1359141e47/node/src/actors/chain_manager/mining.rs#L274

The steps performed are:

* For each data request in COMMIT stage:
  * Calculate VRF proof, if invalid (does not reach target) try next data request
  * Check if the number of sources of this data request does not exceed the current limit per epoch, otherwise try next data request
  * Take UTXOs used as collateral, if there are not enough UTXOs try next data request
  * Resolve request, serialize result, create and sign reveal and commit transactions, and send commit transaction

This is not optimal, because creating the VRF proof is expensive, and it is not needed if the node does not have enough collateral. So we should change the order in which some of these checks are performed to avoid useless operations.

lrubiorod:
…ations

Close #1991

### DIFF
--- a/node/src/actors/chain_manager/mining.rs
+++ b/node/src/actors/chain_manager/mining.rs
@@ -33,7 +33,7 @@ use witnet_data_structures::{
         CommitTransaction, CommitTransactionBody, DRTransactionBody, MintTransaction,
         RevealTransaction, RevealTransactionBody, TallyTransaction, VTTransactionBody,
     },
-    transaction_factory::build_commit_collateral,
+    transaction_factory::{build_commit_collateral, check_commit_collateral},
     utxo_pool::{UnspentOutputsPool, UtxoDiff},
     vrf::{BlockEligibilityClaim, DataRequestEligibilityClaim, VrfMessage},
 };
@@ -43,7 +43,7 @@ use witnet_util::timestamp::get_timestamp;
 use witnet_validations::validations::{
     block_reward, calculate_liars_and_errors_count_from_tally, calculate_randpoe_threshold,
     calculate_reppoe_threshold, dr_transaction_fee, merkle_tree_root, radon_report_from_error,
-    tally_bytes_on_encode_error, update_utxo_diff, vt_transaction_fee,
+    tally_bytes_on_encode_error, update_utxo_diff, vt_transaction_fee, Wit,
 };
 
 use crate::{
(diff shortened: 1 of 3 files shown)


node/src/actors/chain_manager/mining.rs
```diff
@@ -317,6 +317,7 @@ impl ChainManager {
         // epoch by using a reference-counted atomic counter that can be read and updated safely.
         let current_retrieval_count = Arc::new(AtomicU16::new(0u16));
         let maximum_retrieval_count = self.data_request_max_retrievals_per_epoch;
+        let mut own_utxos_copy = self.chain_state.own_utxos.clone();
```
tmpolaczyk, line 320: This clone must go away.

This can be achieved by duplicating all the logic of `NodeUtxos` into a `NodeUtxosRef` struct, which is the same but it takes `&OwnUnspentOutputsPool` instead of `&mut OwnUnspentOutputsPool`. I tried that and it increases our line count by 30 lines so I'm not happy with it.

node/src/actors/chain_manager/mining.rs
```diff
@@ -365,6 +377,83 @@ impl ChainManager {
...
+
+                continue;
+            } else {
+                // Update `current_retrieval_count` thanks to interior mutability of the
+                // `cloned_retrieval_count` reference. This is a recursive operation so as
+                // to guarantee addition and prevent potential race conditions in a concurrent
+                // scenario.
+                loop {
+                    let internal_retrieval_count_result = cloned_retrieval_count
+                        .compare_exchange_weak(
```
tmpolaczyk, line 429: Since some recent versions of Rust we can use

https://doc.rust-lang.org/std/sync/atomic/struct.AtomicU16.html#method.fetch_update

which implements the same logic as this `loop`: try to update an atomic value.

So we can try to refactor this code, and maybe even add some tests.

tmpolaczyk: Also I think this limit is no longer correctly checked, it should be applied after checking VRF eligibility.

tmpolaczyk (approved):

node/src/actors/chain_manager/mining.rs
```diff
@@ -365,6 +376,25 @@ impl ChainManager {
...
+            if !check_commit_collateral(
+                collateral_amount,
+                &self.chain_state.own_utxos,
+                own_pkh,
+                &self.chain_state.unspent_outputs_pool,
+                timestamp,
+                // The block number must be lower than this limit
+                block_number_limit,
+            ) {
+                log::debug!("Mining data request: Insufficient collateral, the data request need {} mature wits", collateral_amount);
```
tmpolaczyk, line 394: This is nanowits but it says wits

node/src/actors/chain_manager/mining.rs
```diff
@@ -365,6 +376,25 @@ impl ChainManager {
...
+            if !check_commit_collateral(
+                collateral_amount,
+                &self.chain_state.own_utxos,
+                own_pkh,
+                &self.chain_state.unspent_outputs_pool,
+                timestamp,
+                // The block number must be lower than this limit
+                block_number_limit,
+            ) {
+                log::debug!("Mining data request: Insufficient collateral, the data request need {} mature wits", Wit::from_nanowits(collateral_amount));
```
tmpolaczyk, line 394: Also this doesn't show the dr_pointer which may be confusing if there are many data requests.

tmpolaczyk: Oh nevermind, too late, it's merged now.

PR #1999, opened 2021-06-29
chore: propose date to start TAPI voting

tmpolaczyk:


data_structures/src/mainnet_validations.rs
```diff
@@ -104,8 +104,9 @@ impl TapiEngine {
                     votes: 0,
                     period: 26880,
                     wip: "WIP0014-0016".to_string(),
-                    // TODO: decide activation date
-                    init: 999999999,
+                    // Start voting at
+                    // 29 June 2021 @ 9:00:00 UTC
```
lrubiorod, line 108: In the past!!

tmpolaczyk: It's the easiest way to ensure that the first activation check will be in two weeks. Because if we set the init date two weeks in the future, the first activation check will be 4 weeks in the future.

lrubiorod: 13 July 2021 9.00 UTC

tmpolaczyk: There are 26880 blocks in 2 weeks so I'm changing `init` to 522240.

lrubiorod (approved):

PR #2002, opened 2021-07-01
fix(bridge): retry on timeout

### LINKED ISSUE #1982: Bridge: retry on timeout
Currently if there is any error with the ethereum client, the bridge ignores that call and will probably retry later. However, if the ethereum client has disconnected, all future calls will also fail with an error. The only solution in that case is to restart the bridge.

Ideally the ethereum client should be a separate actor with the ability to detect that the client has disconnected, and reconnect automatically.

mariocao:
~~Partially~~ closes #1982

### DIFF
--- a/bridges/centralized-ethereum/src/actors/dr_sender.rs
+++ b/bridges/centralized-ethereum/src/actors/dr_sender.rs
@@ -7,7 +7,7 @@ use crate::{
 };
 use actix::prelude::*;
 use serde_json::json;
-use std::{fmt, sync::Arc, time::Duration};
+use std::{fmt, time::Duration};
 use witnet_data_structures::{
     chain::{DataRequestOutput, Hash},
     proto::ProtobufConvert,
@@ -23,7 +23,7 @@ mod tests;
 /// DrSender actor reads the new requests from DrDatabase and includes them in Witnet
 #[derive(Default)]
 pub struct DrSender {
-    witnet_client: Option<Arc<Addr<JsonRpcClient>>>,
+    witnet_client: Option<Addr<JsonRpcClient>>,
     wit_dr_sender_polling_rate_ms: u64,
     max_dr_value_nanowits: u64,
 }
@@ -53,7 +53,7 @@ impl SystemService for DrSender {}
 impl DrSender {
     /// Initialize the `DrSender` taking the configuration from a `Config` structure
     /// and a Json-RPC client connected to a Witnet node
-    pub fn from_config(config: &Config, node_client: Arc<Addr<JsonRpcClient>>) -> Self {
+    pub fn from_config(config: &Config, node_client: Addr<JsonRpcClient>) -> Self {
         let max_dr_value_nanowits = config.max_dr_value_nanowits;
         let wit_dr_sender_polling_rate_ms = config.wit_dr_sender_polling_rate_ms;
 
--- a/bridges/centralized-ethereum/src/actors/dr_sender.rs
+++ b/bridges/centralized-ethereum/src/actors/dr_sender.rs
@@ -6,11 +6,6 @@ use crate::{
     config::Config,
 };
 use actix::prelude::*;
-use async_jsonrpc_client::{
-    transports::{shared::EventLoopHandle, tcp::TcpSocket},
-    Transport,
-};
-use futures_util::compat::Compat01As03;
 use serde_json::json;
 use std::{fmt, sync::Arc, time::Duration};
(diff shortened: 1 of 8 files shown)


bridges/centralized-ethereum/src/actors/eth_poller.rs
```diff
@@ -72,7 +72,12 @@ impl EthPoller {
                     contract::Options::default(),
                     None,
                 )
-                .await;
+                .await
+                .map_err(|err| {
+                    log::error!("Fail to read dr bytes from contract: {:?}", err.to_string());
```
tmpolaczyk, line 77:
```suggestion
                    log::error!("Fail to read requestsCount from contract: {:?}", err.to_string());
```

Also, you don't need `"{:?}", err.to_string()`. You can use `"{}", err` to get the human readable error message, or `"{:?}", err` to get the debug representation.

bridges/centralized-ethereum/src/lib.rs
```diff
@@ -6,11 +6,11 @@
 #![deny(unused_mut)]
 #![deny(missing_docs)]
 
-use crate::config::Config;
 use async_jsonrpc_client::{transports::tcp::TcpSocket, Transport};
 use futures_util::compat::Compat01As03;
 use serde_json::json;
 use std::{sync::Arc, time::Duration};
+use web3::types::H160;
```
lrubiorod, line 13: fmt

mariocao: classic!

bridges/centralized-ethereum/src/actors/dr_sender.rs
```diff
@@ -27,8 +23,7 @@ mod tests;
 /// DrSender actor reads the new requests from DrDatabase and includes them in Witnet
 #[derive(Default)]
 pub struct DrSender {
-    witnet_client: Option<Arc<TcpSocket>>,
-    _handle: Option<EventLoopHandle>,
+    witnet_client: Option<Arc<Addr<JsonRpcClient>>>,
```
tmpolaczyk, line 26: `Addr` is already an `Arc`, you don't need one extra level.
```suggestion
    witnet_client: Option<Addr<JsonRpcClient>>,
```

net/src/client/tcp/actors/jsonrpc.rs
```diff
@@ -133,7 +133,18 @@ impl JsonRpcClient {
...
+        let res = tokio::time::timeout(
+            Duration::from_millis(timeout_in_millis),
+            Compat01As03::new(f),
+        )
+        .await
+        .map_err(|_err| {
+            log::error!("Failed to connect to {}, will retry later", socket.addr().to_string());
+
+            Error::RequestTimedOut(u128::from(timeout_in_millis))
+        })?;
```
tmpolaczyk, line 147: I think this change is not needed because the `impl Handler<Request> for JsonRpcClient` already has a timeout parameter.

mariocao: true....it was set to 60 seconds!

lrubiorod (approved):

bridges/centralized-ethereum/src/actors/eth_poller.rs
tmpolaczyk: @mariocao I think you missed this suggestion

bridges/centralized-ethereum/src/actors/wit_poller.rs
```diff
@@ -79,9 +69,21 @@ impl WitPoller {
...
             for (dr_id, dr_bytes, dr_tx_hash, dr_tx_creation_timestamp) in pending_drs {
-                let report = witnet_client.execute("dataRequestReport", json!([dr_tx_hash]));
+                let method = String::from("dataRequestReport");
+                let params = json!([dr_tx_hash]);
+                let req = jsonrpc::Request::method(method)
+                    .timeout(Duration::from_millis(5_000))
+                    .params(params)
+                    .expect("params failed serialization");
+                let report = witnet_client.send(req).await;
+                let report = match report {
```
clbartoli, line 79: I think the report is matched twice, in line 79 and in line 87, is it intentional?

tmpolaczyk: It's because the report is a `Result<Result<_>>`, so this match unwraps the outer error.

tmpolaczyk (approved):

PR #2003, opened 2021-07-02
feat: only move to Synced if there is a block consensus

lrubiorod:


node/src/actors/chain_manager/handlers.rs
```diff
@@ -1091,9 +1091,9 @@ impl Handler<PeersBeacons> for ChainManager {
                             highest_block_checkpoint: consensus_beacon,
                             ..
                         },
-                        _,
+                        is_there_block_consensus,
                     )) if consensus_beacon == our_beacon => {
-                        if self.sm_state == StateMachine::AlmostSynced {
+                        if self.sm_state == StateMachine::AlmostSynced && is_there_block_consensus {
```
tmpolaczyk, line 1096: What happens after a revert, when all the nodes go to `WaitingConsensus`? Is there a possibility that the network may never recover because all the nodes are in `AlmostSynced` state and no one is mining?

lrubiorod: Is no one is mining, they should achieve a consensus, and if not, they will when superblock is created

tmpolaczyk: Ok, I thought that `is_there_block_consensus` means that there was a new block in the last epoch, but it doesn't mean that. So when the peers revert to a common block, it is considered block consensus.

tmpolaczyk (approved):

PR #2004, opened 2021-07-02
feat: include reveals in all states

lrubiorod:


node/src/actors/chain_manager/mod.rs
```diff
@@ -763,11 +763,20 @@ impl ChainManager {
...
-                        let _reveals = self
+                        let reveals = self
                             .chain_state
                             .data_request_pool
                             .update_data_request_stages();
 
+                        for reveal in reveals {
+                            // Send AddTransaction message to self
+                            // And broadcast it to all of peers
+                            ctx.address().do_send(AddTransaction {
```
tmpolaczyk, line 774: If the node is not in `Synced` state, `AddTransaction` will return `ChainManagerError::NotSynced`.

node/src/actors/chain_manager/mod.rs
```diff
@@ -1085,15 +1103,20 @@ impl ChainManager {
...
-            StateMachine::Synced => {}
-            _ => {
-                return Box::pin(actix::fut::err(
-                    ChainManagerError::NotSynced {
-                        current_state: self.sm_state,
-                    }
-                    .into(),
-                ));
-            }
+            StateMachine::Synced | StateMachine::AlmostSynced => {}
```
tmpolaczyk, line 1106: This change looks important. Maybe this explains why some blocks have zero transactions?

lrubiorod: Maybe, it something that we didnt modify well when we include the AlmostSynced state

tmpolaczyk (approved):

PR #2011, opened 2021-07-07
feat(cli): add CLI method for TAPI information

### LINKED ISSUE #2010: Create a CLI method to check TAPI votes
Currently there is a JSON-RPC method to check TAPI information, but there is no CLI method for that. So it would be helpful to check it directly with CLI

lrubiorod:
Close #2010

### DIFF
--- a/src/cli/node/json_rpc_client.rs
+++ b/src/cli/node/json_rpc_client.rs
@@ -34,7 +34,7 @@ use witnet_node::actors::{
     json_rpc::json_rpc_methods::{
         AddrType, GetBalanceParams, GetBlockChainParams, GetTransactionOutput, PeersResult,
     },
-    messages::{BuildVtt, GetReputationResult},
+    messages::{BuildVtt, GetReputationResult, SignalingInfo},
 };
 use witnet_rad::types::RadonTypes;
 use witnet_util::{credentials::create_credentials_file, timestamp::pretty_print};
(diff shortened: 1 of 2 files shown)


tmpolaczyk:
Example output:

```
Current epoch: 512659

List of activated upgrades:
- Epoch 192000: WIP0008
- Epoch 376320: WIP0009-0011-0012
- Epoch 445440: THIRD_HARD_FORK

List of pending upgrades:
- WIP0014-0016 (using bit 0): Started in 502234. Votes:0/26880 (0%)
```

I think it would be interesting to provide more information on pending upgrades. Not sure on how to format it, but something like this:

```
List of pending upgrades:
- WIP0014-0016 (using bit 0): Started in epoch 502234. Next check in epoch 529114.
    blocks: 0 signaling, 16880 upcoming, 10000 non-signaling
    impossible to reach consensus ratio of 80% (current ratio: 0%)
```

For the last message, we can take inspiration from the bitcoin taproot monitoring tool [(source)](https://github.com/hsjoberg/fork-explorer/blob/d5666a2600b3d430f8b5d1f240159c8af4e488d3/backend/txt/index.ts#L56):

So there are 4 cases:

* WIP reached activation %, will activate on epoch X
* Impossible to activate in current window
* Possible to activate but the current ratio (5%) is not enough
* Possible to activate and the current ratio (99%) is enough

src/cli/node/json_rpc_client.rs
```diff
@@ -1171,6 +1171,61 @@ pub fn rewind(addr: SocketAddr, epoch: Epoch) -> Result<(), failure::Error> {
...
+                .saturating_sub(i.votes);
+            let upcoming_blocks = next_check.saturating_sub(signaling_info.epoch);
+            println!(
+                "    Blocks: {} signaling, {} non-signaling, {} upcoming",
+                signaling_blocks, non_signaling_block, upcoming_blocks
+            );
+
+            let percentage = i.votes.saturating_mul(100) / i.period;
+            let relative_percentage = i.votes.saturating_mul(100)
+                / (signaling_info.epoch.saturating_sub(i.init) % i.period);
```
tmpolaczyk, line 1216: Possible division by zero, use saturating_div :P

tmpolaczyk:
What do you think about adding one more line like this:

```rust
let percentage_target = 80;
let max_possible_votes = i.votes + upcoming_blocks;
let max_possible_percentage = max_possible_votes * 100 / i.period;
// TODO: ensure this is the same logic as the node
if percentage >= percentage_target {
    println!("    Will be activated in this period");
} else if max_possible_percentage < percentage_target {
    println!("    Will not be activated in this period");
} else if relative_percentage >= percentage_target {
    println!("    Will probably be activated in this period");
} else {
    println!("    Will probably not be activated in this period");
}
```

src/cli/node/with_node.rs
```diff
@@ -629,6 +630,12 @@ pub enum Command {
         #[structopt(short = "e", long = "epoch")]
         epoch: Epoch,
     },
+    #[structopt(name = "tapiInfo", about = "Get Information related to TAPI signaling")]
```
tmpolaczyk, line 633: I prefer the name `signalingInfo` because TAPI probably doesn't mean anything to node operators.

```suggestion
    #[structopt(name = "signalingInfo", about = "Get Information related to TAPI signaling")]
```

aesedepece:
Oh, this is much better than my one-liner xD
```
echo '{"jsonrpc":"2.0","method":"signalingInfo","id":"1"}' |  docker exec -i witnet_node witnet node raw 2>&1 | tail -1 | python -m json.tool
```

tmpolaczyk (approved):

PR #2020, opened 2021-07-20
read and select utxos to transaction creation

lrubiorod:


wallet/src/actors/app/handlers/create_vtt.rs
```diff
@@ -42,6 +43,7 @@ pub struct CreateVttRequest {
     fee_type: Option<FeeType>,
     #[serde(default)]
     utxo_strategy: UtxoSelectionStrategy,
+    selected_utxos: Vec<model::OutPtr>,
```
tmpolaczyk, line 46: To make this field optional and default to `[]`:

```suggestion
    #[serde(default)]
    selected_utxos: Vec<model::OutPtr>,
```

tmpolaczyk: Also, you can change it into a `HashSet`, so it has the same type everywhere.

wallet/src/actors/app/handlers/create_vtt.rs
```diff
@@ -9,6 +10,7 @@ use crate::{
     },
 };
 
+use witnet_data_structures::chain::OutputPointer;
```
tmpolaczyk, line 13: fmt

tmpolaczyk (approved):

PR #2023, opened 2021-07-22
Delete wallet

lrubiorod:


tmpolaczyk (approved):

wallet/src/actors/app/methods.rs
```diff
@@ -187,6 +187,27 @@ impl App {
         Box::pin(f)
     }
 
+    /// Delete wallet
```
mariocao, line 190: I think you can document this better :rofl:

wallet/src/actors/worker/methods.rs
```diff
@@ -232,6 +232,13 @@ impl Worker {
         Ok(id)
     }
 
+    /// Delete wallet
```
mariocao, line 235: Same here

wallet/src/repository/wallets/mod.rs
```diff
@@ -122,6 +122,24 @@ impl<T: Database> Wallets<T> {
         Ok(())
     }
 
+    /// Delete a wallet by its id
```
mariocao, line 125:
```suggestion
    /// Delete a wallet by its ID
```

wallet/src/repository/wallets/mod.rs
```diff
@@ -122,6 +122,24 @@ impl<T: Database> Wallets<T> {
...
+    pub fn delete(&self, wallet_id: String) -> Result<()> {
+        let mut batch = self.db.batch();
+        // FIXME: Use merge operator or a transaction when available in rocksdb crate
```
mariocao, line 128:
```suggestion
```

mariocao (changes requested):
I just discovered that `delete_wallet` does not remove the wallet session. Therefore, other API calls are still allowed (e.g. `generate_address`) which does not make much sense. :smile:

mariocao (commented):
I discovered another "strange" flow:

1. Create wallet
2. Unlock wallet -> session id
3. Delete wallet
4. Unlock wallet with `session_id` --> IT WORKS! :)

`unlock_wallet` does not check if the `wallet_id` exists in the list of known wallets.
It directly tries to decrypt the wallet from the database.

It is not an error per se, but it could feel strange for a user.

wallet/src/repository/wallets/mod.rs
```diff
@@ -122,11 +122,33 @@ impl<T: Database> Wallets<T> {
...
+
+        Ok(())
+    }
+
     /// Get a wallet's salt and IV based on its provided ID
     pub fn wallet_salt_and_iv(&self, id: &str) -> Result<(Vec<u8>, Vec<u8>)> {
-        let salt = self.db.get(&keys::wallet_id_salt(id))?;
-        let iv = self.db.get(&keys::wallet_id_iv(id))?;
+        let ids: Vec<String> = self.db.get_or_default(&keys::wallet_ids())?;
+        if !ids.contains(&id.to_string()) {
```
tmpolaczyk, line 145: Add a comment, mention that this is needed because deleted otherwise wallets can still be unlocked.

mariocao (approved):

tmpolaczyk (approved):

PR #2024, opened 2021-07-23
feat(wallet): add delete method to database api

tmpolaczyk:
In the future we may need to delete some keys from the wallet database.


tmpolaczyk:
Closing because it is not needed yet

PR #2030, opened 2021-08-06
fix(superblock): process temporal sb votes when AlmostSynced instead …

### LINKED ISSUE #1809: "Superblock consensus unknown" when constructing superblock in AlmostSynced state
When a node synchronizes to the network it cannot process the received `SUPERBLOCK_VOTE` messages, so they are stored in a temporary buffer `temp_superblock_votes`. When the node moves into `Synced` state, all those votes are processed so that they are counted for the next superblock consensus.

The problem is that sometimes the node finishes the synchronization just before the superblock consensus. It moves into `AlmostSynced` state, and tries achieve superblock consensus but fails because it has 0 votes, because the votes are only added in `Synced` state. This results in a warning "Superblock consensus unknown", and delays the synchronization for at least 2 epochs.

So we could try to process the `temp_superblock_votes` when moving to `AlmostSynced` state, or we could add a check to the superblock consensus logic: if `temp_superblock_votes` is not empty, add all those votes before calculating consensus.

lrubiorod:
…of Synced

Close #1809

### DIFF
--- a/node/src/actors/chain_manager/handlers.rs
+++ b/node/src/actors/chain_manager/handlers.rs
@@ -977,6 +977,7 @@ impl Handler<PeersBeacons> for ChainManager {
 
                             StateMachine::WaitingConsensus
                         } else if our_beacon == consensus_beacon {
+                            self.add_temp_superblock_votes(ctx);
                             StateMachine::AlmostSynced
                         } else if our_beacon.checkpoint == consensus_beacon.checkpoint
                             && our_beacon.hash_prev_block != consensus_beacon.hash_prev_block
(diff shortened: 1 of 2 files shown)


tmpolaczyk:
I'm unable to reproduce the original issue so I can't say if this fix works. However, it shouldn't break anything so we can merge it and reopen the issue if it still happens. A few comments:

* Adding a call to `add_temp_superblock_votes` on every call to `self.update_state_machine(StateMachine::AlmostSynced)` is very error prone. It would be easier to move this call inside the `update_state_machine` function so any future refactors don't break this logic.
* `add_temp_superblock_votes` runs asynchronously, so if the state change happens right before the superblock consensus check it may be possible that some votes are processed too late. This depends on how does actix handle message priority, because all the superblock vote messages are enqueued before the "construct superblock" message. So, if after this PR the issue is not fixed I would guess that it is because of this.

tmpolaczyk:
Superseded by #2040

PR #2031, opened 2021-08-06
feat(toolkit): add support for decoding data requests from .sol files

aesedepece:


toolkit/src/cli/data_requests.rs
```diff
@@ -27,8 +36,29 @@ pub(crate) fn decode_from_args(
...
+/// that is an instance of the `WitnetRequest.sol` contract, or at least implements the same
+/// interface.
+fn extract_pb_bytes_from_solidity_file(path: &Path) -> Result<Vec<u8>, Error> {
+    let mut file = File::open(path).map_err(Error::SolidityFileCantOpen)?;
+    let mut contents = String::new();
+    file.read_to_string(&mut contents)
+        .map_err(Error::SolidityFileCantRead)?;
+
+    let hex_reg_ex: Regex =
+        Regex::new(r#"\s*constructor.*public\s*Request\s*\(\s*hex"([\da-f]+)"#)?;
```
tmpolaczyk, line 55: This regex does not match the example request from here: https://docs.witnet.io/tutorials/bitcoin-price-feed/compiling/

tmpolaczyk (approved):

toolkit/src/errors.rs
```diff
@@ -7,7 +7,9 @@ use failure::Fail;
 #[derive(Debug, Fail)]
 #[fail(display = "error")]
 pub enum Error {
-    #[fail(display = "No bytes have been provided. Please use the --hex argument.")]
+    #[fail(
+        display = "No bytes have been provided. Please use the --hex or --from-solidity argument"
+    )]
     DataRequestNoBytes,
     #[fail(
         display = "The string provided in the --hex field is not a valid hexadecimal byte string: {}",
```
tmpolaczyk, line 15: Now this error can happen without using the `--hex` field, when using the `from-solidity` option

toolkit/src/cli/data_requests.rs
aesedepece: True. `public` is not used in constructors anymore. I adjusted the regex to account for either presence or absence of such modifier.

toolkit/src/errors.rs
aesedepece: This change was already covered in this PR, I don't know why GitHub didn't show it to you :shrug:

toolkit/src/cli/data_requests.rs
tmpolaczyk: Now it does match the request but the bytes are not valid protobuf. So we should probably update that tutorial.

aesedepece: Yep, it seems those are totally outdated or otherwise simply broken.

tmpolaczyk (approved):

PR #2038, opened 2021-08-13
feat: improve rocksdb error message

tmpolaczyk:
Old error:

```
Error when writing blocks to storage: RocksDB error
```

New error:

```
Error when writing blocks to storage: RocksDB error: IO error: No space left on deviceWhile appending to file: .witnet/000009.log: No space left on device
```

### DIFF
--- a/storage/src/backends/rocksdb.rs
+++ b/storage/src/backends/rocksdb.rs
@@ -11,7 +11,7 @@ use crate::storage::{Result, Storage};
 pub type Backend = rocksdb::DB;
 
 #[derive(Debug, Fail)]
-#[fail(display = "RocksDB error")]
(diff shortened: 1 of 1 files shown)


tmpolaczyk:
Steps to test:

```
# Create 5MB file and mount it as the witnet storage
fallocate -l 5M small.ext4
mkfs.ext4 small.ext4
sudo mount -o loop,rw small.ext4 .witnet/

# Start node in another terminal

# Fill partition with random bytes
cat /dev/urandom > .witnet/random
```

After a few blocks when rocksdb decides to persist its cache to disk, it will show the RocksDB error.

mariocao (approved):

PR #2029, opened 2021-08-05
Total supply cli

lrubiorod:
Close #1989

PR #2039, opened 2021-08-16
Total supply cli +

### LINKED ISSUE #1989: Implement JSON-RPC method to calculate supply (current, locked, total)
PR #1721 almost implemented this, but it was incompatible with the storage and we don't have any simple way to add new fields to existing data structures.

The main impediment is that we are missing a method to calculate the total "value" of the `DataRequestPool`: the sum of the unpaid rewards, fees, and collateral of the current data requests. Without this value, the `getSupplyInfo` method will be inaccurate. This may not be a problem for some use cases but it may mislead people to think that there is a bug in the protocol because the supply sometimes goes down. I tried to implement that here: https://github.com/tmpolaczyk/witnet-rust/commit/45db5bb60fab0e1ffec4a5732876816bad472c03 but it is not 100% accurate, so we need to extensively test the implementation.

The expected API should be similar to this:

```rust
/// Information about the total supply
#[derive(Clone, Debug, Serialize, Deserialize)]
pub struct SupplyInfo {
    /// Current epoch
    pub epoch: u32,
    /// Current time
    pub current_time: u64,
    /// Number of blocks minted
    pub blocks_minted: u32,
    /// WIT minted through block creation
    pub blocks_minted_reward: u64,
    /// Number of blocks missing
    pub blocks_missing: u32,
    /// WIT missing because a block was not created
    pub blocks_missing_reward: u64,
    /// Amount of in-flight data requests
    pub collateralized_data_requests: u32,
    /// Supply currently locked in data requests
    pub collateral_locked: u64,
    /// Current unlocked supply
    pub current_unlocked_supply: u64,
    /// Current locked supply
    pub current_locked_supply: u64,
    /// Total supply
    pub total_supply: u64,
}
```

tmpolaczyk:
Supersedes #2029 

Close #1989

Output:

```
[2021-08-17T14:21:37Z INFO  witnet::cli::node::json_rpc_client] Connecting to JSON-RPC server at 127.0.0.1:21338
[2021-08-17T14:21:37Z INFO  witnet::cli::node::json_rpc_client] SupplyInfo { epoch: 589868, current_time: 1629210097, blocks_minted: 540691, blocks_minted_reward: 135172750000000000, blocks_missing: 49176, blocks_missing_reward: 12294000000000000, in_flight_requests: 5, locked_wits_by_requests: 2350200003401, current_unlocked_supply: 730394955285531143, current_locked_supply: 154775444522133176, maximum_supply: 2499999999962167720 }

Supply info at 2021-08-17 14:21:37 UTC (epoch 589868):

 Supply type                         |    Total WITs 
-------------------------------------+---------------
 Temporarily locked in data requests |         2,350 
 Unlocked supply                     |   730,397,305 
 Locked supply                       |   154,775,444 
 Circulating supply                  |   885,172,749 
 Actual maximum supply               | 2,487,705,999 
 Expected maximum supply             | 2,499,999,999 

 Blocks   |  Amount |  Total WITs 
----------+---------+-------------
 Minted   | 540,691 | 135,172,750 
 Reverted |  49,176 |  12,294,000 
 Expected | 589,867 | 147,466,750 

17% of circulating supply is locked.
8% of all blocks so far have been reverted.
For more information about block rewards and halvings, see:
https://github.com/witnet/WIPs/blob/master/wip-0003.md
```

### DIFF
--- a/data_structures/src/data_request.rs
+++ b/data_structures/src/data_request.rs
@@ -349,10 +349,9 @@ impl DataRequestPool {
         std::mem::take(&mut self.to_be_stored)
     }
 
-    /// Return the sum of all the collateral that is currently being used to resolve data requests
-    pub fn collateral_locked(&self, collateral_minimum: u64) -> u64 {
+    /// Return the sum of all the wits that is currently being used to resolve data requests
+    pub fn locked_wits_by_requests(&self, collateral_minimum: u64) -> u64 {
         let mut total = 0;
-
         for dr_state in self.data_request_pool.values() {
             let dr_collateral = dr_state.data_request.collateral;
             let dr_collateral = if dr_collateral == 0 {
@@ -361,7 +360,33 @@ impl DataRequestPool {
                 dr_collateral
             };
 
-            total += dr_collateral * u64::try_from(dr_state.info.commits.len()).unwrap();
+            let locked_wits = match dr_state.stage {
+                DataRequestStage::COMMIT => {
+                    // In commit stage, there are not collateralized wits yet,
+                    // but the requester has locked wits to pay fees and rewards
+                    dr_state.data_request.checked_total_value().unwrap()
+                }
+                DataRequestStage::REVEAL | DataRequestStage::TALLY => {
+                    // In reveal and tally stage, there are some wits collateralized by the witnesses,
+                    // and there are some fees that were already paid
+                    let dr_total_value = dr_state.data_request.checked_total_value().unwrap();
+                    let n_commits = u64::try_from(dr_state.info.commits.len()).unwrap();
+                    let n_reveals = u64::try_from(dr_state.info.reveals.len()).unwrap();
+
+                    let collateralized_wits = n_commits.saturating_mul(dr_collateral);
+                    let commit_fees =
+                        n_commits.saturating_mul(dr_state.data_request.commit_and_reveal_fee);
+                    let reveal_fees =
+                        n_reveals.saturating_mul(dr_state.data_request.commit_and_reveal_fee);
+
+                    dr_total_value
+                        .saturating_add(collateralized_wits)
+                        .saturating_sub(commit_fees)
+                        .saturating_sub(reveal_fees)
+                }
+            };
+
+            total += locked_wits;
         }
 
         total
--- a/data_structures/src/data_request.rs
+++ b/data_structures/src/data_request.rs
@@ -348,6 +348,24 @@ impl DataRequestPool {
     pub fn finished_data_requests(&mut self) -> Vec<DataRequestInfo> {
         std::mem::take(&mut self.to_be_stored)
     }
+
+    /// Return the sum of all the collateral that is currently being used to resolve data requests
+    pub fn collateral_locked(&self, collateral_minimum: u64) -> u64 {
+        let mut total = 0;
+
+        for dr_state in self.data_request_pool.values() {
+            let dr_collateral = dr_state.data_request.collateral;
+            let dr_collateral = if dr_collateral == 0 {
+                collateral_minimum
+            } else {
+                dr_collateral
+            };
+
+            total += dr_collateral * u64::try_from(dr_state.info.commits.len()).unwrap();
+        }
+
+        total
+    }
 }
 
 /// Return the change that should be returned to the creator of the data request if
--- a/node/src/actors/chain_manager/handlers.rs
+++ b/node/src/actors/chain_manager/handlers.rs
@@ -5,7 +5,7 @@ use std::{
     convert::{TryFrom, TryInto},
     future,
     net::SocketAddr,
-    time::{Duration, SystemTime, UNIX_EPOCH},
+    time::Duration,
 };
 
 use witnet_data_structures::{
@@ -1427,10 +1427,7 @@ impl Handler<GetSupplyInfo> for ChainManager {
         let collateral_minimum = chain_info.consensus_constants.collateral_minimum;
 
         let epoch = self.current_epoch.unwrap();
-        let current_time = SystemTime::now()
-            .duration_since(UNIX_EPOCH)
-            .expect("Time went backwards")
-            .as_secs();
+        let current_time = u64::try_from(get_timestamp()).unwrap();
 
         let mut current_unlocked_supply = 0;
         let mut current_locked_supply = 0;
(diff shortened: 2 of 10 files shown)


aesedepece:
Does "in flight requests" also include VTTs?

tmpolaczyk:
> Does "in flight requests" also include VTTs? What is it exactly?

It's the sum of "UTXOs that are not in the UTXO set" because they are in the DataRequestPool. Essentially the outputs of a tally transaction: witness reward, collateral refund, change for the data request creator. They are removed from the UTXO set when creating a data request, and they are inserted again after the data request finishes. It is not anything about pending transactions from the mempool.

drcpu-github:
I like the split into two tables, that does improve the overall readability.

I'm probably to blame for the name, but maybe instead of `Missing` we should have `Reverted`? I'm a bit on the fence of whether `Missing` really articulates that those blocks are gone forever.

Maybe instead of `In flight requests`, a more clear description would be `Locked in requests`?

aesedepece:
For the "in-flight" value, I'm supportive of something like @drcpu-github's suggestion: `Temporarily locked in data requests`.

Regarding `Missing` vs. `Reverted`, probably `Reverted` is more descriptive of what happened.

tmpolaczyk:
`Temporarily locked in data requests` is long, will see how it fits in the table. 

I had considered `Reverted`, but in theory it could also be that blocks have not been mined because there was no valid VRF proof so I decided to keep it as `Missing`. Other options that come to mind are `Skipped` or simply `Not mined`. But I will change it to `Reverted` because that's currently the most probable reason to have missing blocks.

node/src/actors/chain_manager/handlers.rs
```diff
@@ -1409,6 +1408,81 @@ impl Handler<GetBalance> for ChainManager {
...
+            .unwrap();
+        let locked_wits_by_requests = self
+            .chain_state
+            .data_request_pool
+            .locked_wits_by_requests(collateral_minimum);
+
+        let (mut blocks_minted, mut blocks_minted_reward) = (0, 0);
+        let (mut blocks_missing, mut blocks_missing_reward) = (0, 0);
+        for e in 1..epoch {
+            let block_reward = block_reward(epoch, initial_block_reward, halving_period);
```
aesedepece, line 1458: I think there is a bug here, although probably we wouldn't have noticed before first halving: 
```suggestion
            let block_reward = block_reward(e, initial_block_reward, halving_period);
```

Let me suggest though to rename `epoch` to `current_epoch` and `e` to `epoch` so the iterator clearly reads `for epoch in 1..current_epoch`, which more clearly states what's going on and avoids this kind of confusion.

node/src/actors/json_rpc/json_rpc_methods.rs
```diff
@@ -1180,6 +1184,28 @@ pub async fn get_balance(params: Params) -> JsonRpcResult {
...
+                .and_then(|supply_info| match supply_info {
+                    Ok(x) => match serde_json::to_value(&x) {
+                        Ok(x) => Ok(x),
+                        Err(e) => {
+                            let err = internal_error_s(e);
+                            Err(err)
+                        }
+                    },
+                    Err(e) => Err(internal_error_s(e)),
+                })
```
aesedepece, line 1204: All this block is ugly AF, but I hope this makes things a little more readable:
```suggestion
                .and_then(|supply_info| match supply_info {
                    Ok(x) => serde_json::to_value(&x).map_err(internal_error_s),
                    Err(e) => Err(internal_error_s(e)),
                })
```

Or maybe this could work:
```suggestion
                .and_then(|supply_info| supply_info
                    .and_then(|x| serde_json::to_value(&x).map_err(internal_error_s))
                    .map_err(internal_error_s)
                )
```

aesedepece (changes requested):

node/src/actors/json_rpc/json_rpc_methods.rs
tmpolaczyk: This is copy-pasted from the other JSON-RPC handlers, the solution is to abstract this logic into a function and just do something like

```suggestion
                .and_then(|supply_info| to_json_result(supply_info))
```

aesedepece: Loving `.and_then(to_json_result)` :stuck_out_tongue:

node/src/actors/chain_manager/handlers.rs
```diff
@@ -1409,6 +1408,81 @@ impl Handler<GetBalance> for ChainManager {
...
+
+    fn handle(&mut self, _msg: GetSupplyInfo, _ctx: &mut Self::Context) -> Self::Result {
+        if self.sm_state != StateMachine::Synced {
+            return Err(ChainManagerError::NotSynced {
+                current_state: self.sm_state,
+            }
+            .into());
+        }
+
+        let total_supply = 2_500_000_000_000_000_000;
```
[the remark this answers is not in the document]
tmpolaczyk: It's not available as a constant but it can be calculated on the fly by using the other fields. The value turns out to be `2499999999962167720`.

data_structures/src/data_request.rs
```diff
@@ -348,6 +348,49 @@ impl DataRequestPool {
...
+    /// Return the sum of all the wits that is currently being used to resolve data requests
+    pub fn locked_wits_by_requests(&self, collateral_minimum: u64) -> u64 {
+        let mut total = 0;
+        for dr_state in self.data_request_pool.values() {
+            let dr_collateral = dr_state.data_request.collateral;
+            let dr_collateral = if dr_collateral == 0 {
+                collateral_minimum
+            } else {
+                dr_collateral
+            };
```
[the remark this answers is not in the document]
tmpolaczyk: Other parts of the code already use this form, so I prefer to keep it consistent.

node/src/actors/json_rpc/json_rpc_methods.rs
tmpolaczyk: But in a different PR

node/src/actors/chain_manager/handlers.rs
```diff
@@ -1409,6 +1408,86 @@ impl Handler<GetBalance> for ChainManager {
...
+                blocks_missing_reward += block_reward;
+            }
+        }
+
+        let genesis_amount =
+            current_locked_supply + current_unlocked_supply + locked_wits_by_requests
+                - blocks_minted_reward;
+        let maximum_block_reward = total_block_reward(initial_block_reward, halving_period);
+        let maximum_supply = genesis_amount + maximum_block_reward;
+        //let total_supply_minus_missing_blocks = total_supply - blocks_missing_reward;
```
aesedepece, line 1473:
```suggestion
```

aesedepece (changes requested):

aesedepece (approved):

aesedepece:
This is RTM

PR #2040, opened 2021-08-18
fix(superblock): process temporal sb votes when AlmostSynced instead of Synced 2

### LINKED ISSUE #1809: "Superblock consensus unknown" when constructing superblock in AlmostSynced state

tmpolaczyk:
Close #1809

Supersedes #2030

### DIFF
--- a/node/src/actors/chain_manager/handlers.rs
+++ b/node/src/actors/chain_manager/handlers.rs
@@ -976,6 +976,7 @@ impl Handler<PeersBeacons> for ChainManager {
 
                             StateMachine::WaitingConsensus
                         } else if our_beacon == consensus_beacon {
+                            self.add_temp_superblock_votes(ctx);
                             StateMachine::AlmostSynced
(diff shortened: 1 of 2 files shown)


aesedepece (approved):
Looks fine.

PR #2049, opened 2021-08-30
Include median

### LINKED ISSUE #2047: Add MEDIAN reducer implementation
Implement the Median reducer taking into account that it should be activated via TAPI.

lrubiorod:
Close #2047 and #2054


rad/src/reducers/median.rs
```diff
@@ -0,0 +1,287 @@
...
+            for item in value {
+                match item {
+                    RadonTypes::Float(f64_value) => {
+                        if let Ok(not_nan) = NotNan::new(f64_value.value()) {
+                            input_not_nan.push(not_nan)
+                        }
+                    }
+                    _ => {
+                        return Err(RadError::MismatchingTypes {
+                            method: RadonReducers::AverageMean.to_string(),
```
lrubiorod, line 35:
```suggestion
                            method: RadonReducers::AverageMedian.to_string(),
```

rad/src/reducers/median.rs
```diff
@@ -0,0 +1,287 @@
...
+                let left_pos = right_pos - 1;
+
+                // Create new array to be able to use average::mean reducer
+                let rl = RadonArray::from(vec![
+                    sorted_input[left_pos].clone(),
+                    sorted_input[right_pos].clone(),
+                ]);
+                // RoundToInteger means that when the average is not an integer, it will be rounded to an
+                // integer. For example, the average of 1 and 2, which is 1.5, will be rounded to 1.
+                mean(&rl, MeanReturnPolicy::RoundToInteger)
```
lrubiorod, line 98: If it is a Float, it is neccessary to round to Integer?

tmpolaczyk: According to the documentation of `MeanReturnPolicy`, `RoundToInteger` actually means "keep same type as the input":

```rust
/// Different available policies regarding what to do with the resulting Float after applying the
/// average mean.
#[derive(Clone, Copy, PartialEq, Eq)]
pub enum MeanReturnPolicy {
    /// Pure `Fn(Array<T>) -> Float` behavior.
    ReturnFloat,
    /// Enforces `Fn(Array<T>) -> T` behavior (if needed).
    RoundToInteger,
}
```

rad/src/reducers/median.rs
```diff
@@ -0,0 +1,287 @@
...
+                let left_pos = right_pos - 1;
+                let left_elem = input_not_nan[left_pos].into_inner();
+
+                // Create new array to be able to use average::mean reducer
+                let rl = RadonArray::from(vec![
+                    RadonTypes::Float(RadonFloat::from(left_elem)),
+                    RadonTypes::Float(RadonFloat::from(right_elem)),
+                ]);
+                // MeanReturnPolicy only applies to integers, so this will actually return a float
+                mean(&rl, MeanReturnPolicy::RoundToInteger)
```
lrubiorod, line 67: If it is a Float, it is necessary to round to Integer?

rad/src/reducers/median.rs
```diff
@@ -0,0 +1,287 @@
...
+        let expected_error = RadError::UnsupportedReducer {
+            array: input,
+            reducer: "RadonReducers::AverageMedian".to_string(),
+        };
+        assert_eq!(output, expected_error);
+    }
+
+    #[test]
+    fn test_operate_reduce_median_str_even() {
+        // When the number of elements is even, the median is element at index (N/2 - 1)
```
lrubiorod, line 227: Is this comment necessary?

tmpolaczyk: No, and it's wrong.

node/src/actors/chain_manager/mod.rs
```diff
@@ -2171,20 +2171,23 @@ impl ChainManager {
...
-        if !oppose_wip0014
-            && !oppose_wip0016
+        // Bit 1
+        let bit = 1;
+        if !oppose_wip0017
+            && !self
+                .chain_state
+                .tapi_engine
+                .wip_activation
+                .contains_key("WIP0017")
```
tmpolaczyk, line 2184: This condition is new: it means that miners will stop signaling when the WIP is activated. Not sure if this is the desired behavior, because this field is also used as a census to see the percentage of miners that are on the latest version. Also, bit 0 is set to 0 in this PR while currently in mainnet it is set to 1. That doesn't break anything consensus-wise but it may be unexpected.

lrubiorod: In my opinion, this bit only should be used to signaling, and after achieve the necessary votes, it is not longer required. We should remove the activation of this bit, to avoid overlap bit problems in the future

rad/src/reducers/mod.rs
```diff
@@ -51,6 +57,10 @@ pub fn reduce(input: &RadonArray, reducer_code: RadonReducers) -> Result<RadonTy
             }
             RadonReducers::Mode => mode::mode(input),
             RadonReducers::DeviationStandard => deviation::standard(input),
+            RadonReducers::AverageMedian => match &context.active_wips {
+                Some(active_wips) if active_wips.wip0017() => median::median(input),
+                _ => error(),
+            },
```
tmpolaczyk, line 63: Here if `context.active_wips` is `None`, the WIP is assumed to be not activated. This is the default for external tools that use the radon library, because they may not have access to a witnet node so they don't know the active wips. However we need to ensure that the `None` case means "enabled" after this WIP is activated.

tmpolaczyk: What do you think about changing the `None` case to mean enabled even if the WIP is not activated yet? This way external tools can create and run requests that use the new operators, but if they deploy the request before the WIP is active they will get an `UnsupportedReducer` error.

lrubiorod: But if they create a request, they try it with this feature and obtain a proper result, but when send to mainnet, it fails... It wasnt a good test to try before...

rad/src/reducers/median.rs
```diff
@@ -0,0 +1,318 @@
...
+        types::{float::RadonFloat, integer::RadonInteger, string::RadonString},
+        RadError::ModeEmpty,
+    };
+    use serde_cbor::Value;
+
+    #[test]
+    fn test_operate_reduce_median_empty() {
+        let input = RadonArray::from(vec![]);
+        let output = median(&input).unwrap_err();
+        let expected_error = ModeEmpty;
```
tmpolaczyk, line 124: TODO: we use `ModeEmpty` as the error for when the median reducer is applied over an empty array. Should we create a new error `MedianEmpty`, or rename `ModeEmpty` to a more generic `EmptyArray` error?

lrubiorod: I think EmptyArray sounds better

node/src/actors/chain_manager/mod.rs
aesedepece: I have mixed opinions. WIP-0014 does not define what to do because it is consensus neutral. It is true that setting bit position 0 to value 0 at this point may come as a surprise, but also we know for certain that we'll need to do so at some point in the coming years. So maybe it is time to define whether we are OK with setting them to 0 upon next bit position being assigned, or we make a commitment to set all bits to 0 when we assign bit 31 (although that removes the ability of signaling 30 and 31 at the same time).

rad/src/reducers/median.rs
aesedepece: Let's take this into account for the eventual WIP, sirs.

tmpolaczyk: So it looks like this error cannot be easily renamed because it appears in the `UnhandledIntercept` error message, so changing its name would make some tallies invalid. However we could maybe avoid the `UnhandledIntercept` by implementing the conversion from `RadError` to `RadonError` and assigning an error code to this error. The currently assigned error codes can be found [here in the bridge](https://github.com/witnet/witnet-ethereum-bridge/blob/b18af3ca43454316f6034d5e131691344687db87/contracts/libs/Witnet.sol#L69) and [here in the node](https://github.com/witnet/witnet-rust/blob/44522fc1d5253da2dcd8d42a3d614aa60ac1772e/data_structures/src/radon_error.rs#L14).

lrubiorod: I could add a TODO or FIXME and create a new issue, because I think it is out of the scope of this issue. What do you think?

lrubiorod: We can use EmptyArray only in the median by the moment

node/src/actors/chain_manager/mod.rs
drcpu-github: I think being able to query which (ARS) nodes have signalled support for a certain WIP after the WIP is passed or defeated is quite a useful feature. Yes, in the (distant) future, bits will be reused and at that point this'll become a problem, but maybe that is not something that should be solved right now.

rad/src/reducers/median.rs
mariocao: Are you sure that the round of 1.5 is 1?
Checking the documentation it seems it rounds away from zero:

```
    /// Returns the nearest integer to an `f64`. Rounds half-way cases away from zero.
    ///
    /// The stabilized version of this intrinsic is
    /// [`f64::round`](../../std/primitive.f64.html#method.round)
    pub fn roundf64(x: f64) -> **f64;**
```

rad/src/reducers/median.rs
```diff
@@ -0,0 +1,288 @@
...
+                let right_pos = sorted_input.len() / 2;
+                let left_pos = right_pos - 1;
+
+                // Create new array to be able to use average::mean reducer
+                let rl = RadonArray::from(vec![
+                    sorted_input[left_pos].clone(),
+                    sorted_input[right_pos].clone(),
+                ]);
+                // RoundToInteger means that when the average is not an integer, it will be rounded to an
+                // integer. For example, the average of 1 and 2, which is 1.5, will be rounded to 1.
```
mariocao, line 97:
```suggestion
                // integer. For example, the average of 1 and 2, which is 1.5, will be rounded to 2.
```

node/src/actors/chain_manager/mod.rs
lrubiorod: Could be an option to leave as before and open an issue with a discussion about that

tmpolaczyk: Yes, let's just leave it like it was before.

node/src/actors/chain_manager/mod.rs
```diff
@@ -2171,20 +2171,18 @@ impl ChainManager {
 
     /// Return the value of the version field for a block in this epoch
     fn tapi_signals_mask(&self, epoch: Epoch) -> u32 {
-        let Tapi {
-            oppose_wip0014,
-            oppose_wip0016,
-        } = &self.tapi;
+        let Tapi { oppose_wip0017 } = &self.tapi;
 
         let mut v = 0;
```
tmpolaczyk, line 2176: This should be `v = 1`, to keep bit 0 enabled?

tmpolaczyk:
Let's merge this pull request to a new branch, since we don't know the final activation date yet and the WIP is not final yet.

PR #2053, opened 2021-09-02
Random Number Generation

### LINKED ISSUE #2056: Implement and test prototype of RNG functionality

lrubiorod:
Close #2056


data_structures/src/data_request.rs
```diff
@@ -83,8 +86,14 @@ impl DataRequestPool {
...
-                            // TODO: Hash(pkh+dr)
-                            PublicKeyHash::default()
+                            let mut bytes_to_hash = vec![];
+                            bytes_to_hash.extend(reveal.body.pkh.hash);
+                            bytes_to_hash.extend(dr_pointer.as_ref());
+
+                            let hashed_bytes = calculate_sha256(bytes_to_hash.as_ref()).0;
+
+                            // We are forcing 20 bytes, so we can not obtain an error here
+                            PublicKeyHash::from_bytes(&hashed_bytes[..20]).unwrap()
```
tmpolaczyk, line 96: You can avoid this hack by moving the `if active_wips.wip0019()` outside, so that `sort_unstable_by_key` is inside the if.

rad/src/reducers/hash_concatenate.rs
```diff
@@ -0,0 +1,62 @@
...
+};
+use serde_cbor::value::Value;
+
+/// Reducer that concatenate all the RadonTypes and hash
+#[allow(clippy::cast_precision_loss)]
+pub fn hash_concatenate(input: &RadonArray) -> Result<RadonTypes, RadError> {
+    let value = input.value();
+
+    match value.first() {
+        None => Ok(RadonTypes::from(RadonInteger::from(0))),
```
tmpolaczyk, line 16: Empty array case, return integer is wrong here. Either return error or return `sha256` of empty string, which is defined as `EMPTY_SHA256` in `witnet_crypto` crate.

rad/src/reducers/hash_concatenate.rs
```diff
@@ -0,0 +1,62 @@
...
+                            Ok(bytes)
+                        }
+                        _ => Err(RadError::MismatchingTypes {
+                            method: RadonReducers::HashConcatenate.to_string(),
+                            expected: RadonBytes::radon_type_name(),
+                            found: item.clone().radon_type_name(),
+                        }),
+                    })?;
+
+            let hash_function = [Value::from(RadonHashFunctions::SHA2_256 as u8)];
```
tmpolaczyk, line 50: To avoid the clippy warning, we could use `u8::from`:

```suggestion
            let hash_function = [Value::from(u8::from(RadonHashFunctions::SHA2_256))];
```

But that requires adding `#[derive(IntoPrimitive)]` to the definition of `pub enum RadonHashFunctions`.

tmpolaczyk: Or we could avoid creating this `Value` by calling `hash_functions::hash` instead of `bytes_operators::hash`.

validations/src/validations.rs
```diff
@@ -1171,6 +1197,11 @@ pub fn validate_tally_transaction<'a>(
         .into());
     }
 
+    #[cfg(test)]
+    println!(
+        "expected_ta_tx.tally: {}",
+        hex::encode(&expected_ta_tx.tally)
+    );
```
tmpolaczyk, line 1204: This should be removed

data_structures/src/data_request.rs
```diff
@@ -71,14 +74,30 @@ impl DataRequestPool {
     }
 
     /// Get all the reveals
-    pub fn get_all_reveals(&self) -> HashMap<Hash, Vec<RevealTransaction>> {
+    pub fn get_all_reveals(
```
tmpolaczyk, line 77: Aside from this function, there is also a `get_reveals` function which still uses the old sort by pkh method.

validations/src/validations.rs
lrubiorod: why?

tmpolaczyk: I added it to help me write the tests, if the tests are done we don't need it.

rad/src/lib.rs
```diff
@@ -245,11 +258,20 @@ pub fn run_aggregation_with_context_report(
...
-    let radon_script = create_radon_script_from_filters_and_reducer(filters, reducer)?;
+    let is_rng = reducer == 0x11;
 
-    let items_to_aggregate = RadonTypes::from(RadonArray::from(radon_types_vec));
+    if is_rng {
+        Ok(RadonReport::from_result(
+            Ok(radon_types_vec[0].clone()),
+            context,
+        ))
+    } else {
```
aesedepece, line 268: What is this doing?

lrubiorod: unwrap() :P

rad/src/reducers/hash_concatenate.rs
```diff
@@ -0,0 +1,197 @@
...
+                    RadonTypes::Bytes(rad_bytes) => {
+                        let new_bytes = rad_bytes.value();
+
+                        // Take up to 32 bytes from new_bytes
+                        let new_bytes_len = std::cmp::min(new_bytes.len(), 32);
+                        bytes.extend_from_slice(&new_bytes[..new_bytes_len]);
+                        // In case of less than 32 bytes, it will be zero-padded
+                        for _i in new_bytes_len..32 {
+                            bytes.push(0);
+                        }
```
aesedepece, line 32: This is LSB padding, which changes the Big-Endian value of the bytes. I'd prefer MSB padding (i.e. pushing the zeroes first).

validations/src/validations.rs
```diff
@@ -244,23 +247,50 @@ pub fn validate_rad_request(rad_request: &RADRequest) -> Result<(), failure::Err
...
     }
+    let mut is_rng = false;
     for path in retrieval_paths {
+        if path.kind == RADType::Rng {
+            if path.url.is_empty() || path.script.is_empty() {
+                is_rng = true;
+                break;
+            } else {
+                return Err(DataRequestError::InvalidRngRequest.into());
+            }
```
aesedepece, line 258: I don't really follow the logic here hehehe.

- I can see why `url` must be empty (won't be used).
- I have doubts on whether `script` should be allowed or not.
- By reading this code, I understand that as long as `url` is empty, `script` can be non-empty, but will it be used anyway?

aesedepece: Some general but related comments:

- `RADTypes` only exist at the data source / retrieval level. We should avoid handling a data request differently only because one of the data sources is of a special kind.
- There won't exist such thing as RNG requests, only requests with RNG sources.
- RNG and HttpGet sources should coexist in the same data request as long as they have the same output type.

validations/src/validations.rs
```diff
@@ -244,23 +247,50 @@ pub fn validate_rad_request(rad_request: &RADRequest) -> Result<(), failure::Err
...
+        }
+
         // If the sources are empty the data request is set as invalid
         if path.url.is_empty() {
             return Err(DataRequestError::NoRetrievalSources.into());
         }
```
aesedepece, line 264: Won't this cause an early return for any retrieval path / data source of RADType::Rng that has both `url` and `script` being empty?

validations/src/validations.rs
```diff
@@ -244,23 +247,50 @@ pub fn validate_rad_request(rad_request: &RADRequest) -> Result<(), failure::Err
...
         unpack_radon_script(path.script.as_slice())?;
     }
 
+    if is_rng && retrieval_paths.len() > 1 {
+        return Err(DataRequestError::InvalidRngRequest.into());
+    }
```
aesedepece, line 270: As per my general comment above, this goes against the three points.

validations/src/validations.rs
aesedepece: E.g. in the future we may have a data source type that performs general computation by running WASM bytecode in a sandbox.

And you may want to use that side by side with a classic HttpGet data source that triggers a similar computation on some cloud, or fetches similar outputs from some API.

My point is that you should perfectly be able to do so, and this part of the code should only care about routing the data source definitions to their "resolvers" (the resolver of HttpGet is our wrapper of the Surf library, the resolver of Rng is the RNG and the resolver of the hypothetical Wasm type would be a WASM virtual machine).

validations/src/validations.rs
tmpolaczyk: In the rng case it does never reach this line because there is a `break` in the previous condition. But that limits rng requests to one retrieval source, so we need to change this logic.

validations/src/validations.rs
tmpolaczyk: Agree that `is_rng` is a hack and we should find a way to avoid it.

rad/src/lib.rs
tmpolaczyk: In the rng case the request will usually have only one retrieval source. So this way we avoid the aggregation step, which is not needed. However I think that could apply to all kinds of requests, because you don't need aggregation if there is only one source. This would be a breaking change, but we can implement it with TAPI.

validations/src/validations.rs
lrubiorod: It should be and AND

validations/src/validations.rs
lrubiorod: If it is a RADType::Rng, it will always return earlier in the previous branchs (break or DataRequestError::InvalidRngRequest)

validations/src/validations.rs
lrubiorod: This only check that if it is RNG request, it should not have more sources

rad/src/reducers/hash_concatenate.rs
tmpolaczyk: This is fixed, now extend is called after the zeros.

node/tests/data_request_examples.rs
```diff
@@ -417,4 +436,36 @@ mod examples {
             fee: 0,
         }
     }
+
+    pub fn random_bytes() -> BuildDrt {
+        let url_0 = "";
+        let r0_script = vec![];
```
tmpolaczyk, line 442: This results in an error because an empty script must be represented as an empty CBOR array:

```suggestion
        let r0_script = vec![128];
```

This test does not detect that error because it doesn't validate the request, I will try to fix that and open a PR.

tmpolaczyk: Opened #2062

rad/src/script.rs
```diff
@@ -252,7 +252,11 @@ pub fn create_radon_script_from_filters_and_reducer(
     )
     .map_err(|_| unknown_reducer(i128::from(reducer)))?;
     match rad_reducer {
-        RadonReducers::AverageMean | RadonReducers::Mode | RadonReducers::AverageMedian => {}
+        RadonReducers::AverageMean
+        | RadonReducers::Mode
+        | RadonReducers::AverageMedian
+        | RadonReducers::HashConcatenate
+        | RadonReducers::Unwrap => {}
```
tmpolaczyk, line 259: This function needs a context, the new operators can only be considered valid if the WIP is active.

lrubiorod: It is fixed in #2063, because it also affect to the median

rad/src/reducers/mod.rs
```diff
@@ -10,17 +10,20 @@ use witnet_data_structures::radon_report::ReportContext;
...
+#[derive(Debug, PartialEq, TryFromPrimitive, IntoPrimitive)]
 #[repr(u8)]
 pub enum RadonReducers {
     // Implemented
     Mode = 0x02,
     AverageMean = 0x03,
     AverageMedian = 0x05,
     DeviationStandard = 0x07,
+    HashConcatenate = 0x0b,
+    Unwrap = 0x0c,
```
aesedepece, line 26: Interesting. Where did this come from?

aesedepece (changes requested):

rad/src/lib.rs
```diff
@@ -208,6 +215,12 @@ pub async fn run_retrieval_report(
 
             result
         }
+        RADType::Rng => {
+            let random_bytes: [u8; 32] = rand::random();
+            let random_bytes = RadonTypes::from(RadonBytes::from(random_bytes.to_vec()));
+
+            Ok(RadonReport::from_result(Ok(random_bytes), context))
```
tmpolaczyk, line 222: This ignores the retrieval script, according to the WIP the script should be executed.

tmpolaczyk: Well actually I'm not sure if the retrieval script should be executed or not.

aesedepece: Let's discuss here:
https://github.com/witnet/WIPs/pull/70#discussion_r706076131

rad/src/reducers/mod.rs
lrubiorod: We dont have a reducer to convert an array of 1 element in that element

tmpolaczyk:
Using commit 0c4ffecfbb26b167c8e2c514801eb7717ae5826f and this data request:

```
{"jsonrpc":"2.0","method":"sendRequest","id":"1","params":{"dro":{"data_request":{"time_lock":0,"retrieve":[{"kind":"RNG","url":"","script":[128]}],"aggregate":{"filters":[],"reducer":3},"tally":{"filters":[],"reducer":3}},"witness_reward":1000,"witnesses":5,"commit_and_reveal_fee":10,"min_consensus_percentage":51,"collateral":1000000000},"fee":0}}
```

An updated node accepts that as a valid request, while a not-updated node does not, leading to a fork.

data_structures/src/chain.rs
```diff
@@ -4006,7 +4009,7 @@ mod tests {
     #[test]
     fn test_transaction_hashable_trait() {
         let transaction = transaction_example();
-        let expected = "7b4001b4a43b3e3dccec642791031d8094ea52164d89c2a4732d0be79ed1af83";
+        let expected = "9221b651aaef0f6d75a175958825775ed0add54102a431a27e485052478e0f9b";
```
tmpolaczyk, line 4012: The hash of the `transaction_example()` changes because the `RADType` changes from 0 to 1.

validations/src/validations.rs
lrubiorod: I had to modify several times the rng tests, so I think it could be useful in the future too

rad/src/lib.rs
lrubiorod: I removed it and use the Mode instead, because the mode of a vector of one element is that element. To avoid strange cases

rad/src/reducers/mod.rs
lrubiorod: Deleted unwrap, using Mode instead of

rad/src/lib.rs
lrubiorod: What was the conclusion?

data_structures/src/chain.rs
```diff
@@ -1584,14 +1584,21 @@ impl Hashable for Bn256Signature {
...
+    /// Random number generation
+    #[serde(rename = "RNG")]
+    Rng,
 }
 
 impl Default for RADType {
     fn default() -> Self {
-        RADType::HttpGet
+        // TODO: Use RadType::HttpGet after WIP-0019 activation
+        RADType::Unknown
```
tmpolaczyk, line 1601: I think `RADType::Unknown` is a good default value, the problem is that this is used in tests so any tests that want to create a "default HttpGet request" can use a helper function.
```suggestion
        RADType::Unknown
```

lrubiorod: I think we could do a refactor in tests after WIP19 activation

node/src/actors/chain_manager/mining.rs
```diff
@@ -930,6 +934,11 @@ pub fn build_block(
             }
         };
 
+        if let Err(e) = validate_rad_request(&dr_tx.body.dr_output.data_request, active_wips) {
+            log::warn!("DataRequest Transaction invalid: {}", e);
+            continue;
+        }
```
tmpolaczyk, line 940: We can add a comment that says that this can be removed after the WIP-0019 is activated

bridges/centralized-ethereum/Cargo.toml
```diff
@@ -22,5 +22,6 @@ witnet_data_structures = { path = "../../data_structures" }
 witnet_config = { path = "../../config" }
 witnet_net = { path = "../../net" }
 witnet_node = { path = "../../node" }
+witnet_rad = { path = "../../rad" }
```
mariocao, line 25: Why is the `current_active_wips` under the rad module?
Maybe it should be under `validations` :)

bridges/ethereum/Cargo.toml
```diff
@@ -21,5 +21,6 @@ tokio-core = "0.1.17"
 web3 = "0.10.0"
 witnet_data_structures = { path = "../../data_structures" }
 witnet_crypto = { path = "../../crypto" }
+witnet_rad = { path = "../../rad" }
```
mariocao, line 24: Same as before :)

data_structures/src/data_request.rs
```diff
@@ -56,10 +59,24 @@ impl DataRequestPool {
...
+    pub fn get_reveals(
+        &self,
+        dr_pointer: &Hash,
+        active_wips: &ActiveWips,
+    ) -> Option<Vec<&RevealTransaction>> {
         self.data_request_pool.get(dr_pointer).map(|dr_state| {
             let mut reveals: Vec<&RevealTransaction> = dr_state.info.reveals.values().collect();
-            reveals.sort_unstable_by_key(|reveal| reveal.body.pkh);
+            if active_wips.wip0019() {
+                reveals.sort_unstable_by_key(|reveal| {
```
mariocao, line 70:
```suggestion
                // As specified in (7) in WIP-0019
                reveals.sort_unstable_by_key(|reveal| {
```

data_structures/src/data_request.rs
```diff
@@ -71,14 +88,28 @@ impl DataRequestPool {
...
                         dr_state.info.reveals.values().cloned().collect();
-                    reveals.sort_unstable_by_key(|reveal| reveal.body.pkh);
+
+                    if active_wips.wip0019() {
+                        reveals.sort_unstable_by_key(|reveal| {
+                            let mut bytes_to_hash = vec![];
+                            bytes_to_hash.extend(reveal.body.pkh.hash);
+                            bytes_to_hash.extend(dr_pointer.as_ref());
+
+                            calculate_sha256(bytes_to_hash.as_ref()).0
```
mariocao, line 108: refactorizable with code above?

node/src/actors/chain_manager/handlers.rs
```diff
@@ -1277,13 +1277,26 @@ impl Handler<BuildDrt> for ChainManager {
             block_epoch: self.current_epoch.unwrap(),
         };
 
-        if let Err(e) = validate_rad_request(&msg.dro.data_request, Some(&active_wips)) {
+        let mut dro = msg.dro;
```
mariocao, line 1280:
```suggestion
        let mut drOutput = msg.dro;
```

node/src/actors/chain_manager/mining.rs
```diff
@@ -143,6 +143,8 @@ impl ChainManager {
             active_wips: self.chain_state.tapi_engine.wip_activation.clone(),
             block_epoch: current_epoch,
         };
+        let active_wips = Arc::new(active_wips);
+        let active_wips2 = active_wips.clone();
```
mariocao, line 147: Add TODO here explaining why and when to remove :)

rad/src/hash_functions/mod.rs
```diff
@@ -1,13 +1,13 @@
...
-use num_enum::TryFromPrimitive;
+use num_enum::{IntoPrimitive, TryFromPrimitive};
 
 use crate::error::RadError;
 use crate::hash_functions::sha2::sha2_256;
 
 mod sha2;
 
-#[derive(Debug, PartialEq, TryFromPrimitive)]
+#[derive(Debug, PartialEq, TryFromPrimitive, IntoPrimitive)]
```
mariocao, line 10:
```suggestion
#[derive(Debug, PartialEq, IntoPrimitive, TryFromPrimitive)]
```

rad/src/lib.rs
```diff
@@ -51,6 +51,15 @@ pub fn current_active_wips() -> ActiveWips {
     }
 }
 
+/// Auxiliary function that returns the current active wips and the WIPs in voting process as actived
+/// It is only used for testing
+pub fn all_wips_active() -> ActiveWips {
+    let mut h = current_active_wips();
```
mariocao, line 57: Why `h`? :rofl:

rad/src/lib.rs
```diff
@@ -151,63 +186,84 @@ pub fn run_retrieval_with_data(
...
-                }
-                Err(e) => log::debug!("Failed result for source {}: {:?}", retrieve.url, e),
-            }
-
-            result
+    if wip_0019_active {
+        match retrieve.kind {
+            RADType::Unknown => Err(RadError::UnknownRetrieval),
+            RADType::HttpGet => http_get_response(retrieve, context, settings).await,
+            RADType::Rng => rng_response(context).await,
```
mariocao, line 245: Love it!

rad/src/lib.rs
tmpolaczyk: It stands for `HashMap`, because the first version of `ActiveWips` was just a `HashMap`.

rad/src/hash_functions/mod.rs
lrubiorod: I think is better to open an issue and refactor all of them

mariocao (approved):

tmpolaczyk (approved):

PR #2063, opened 2021-09-09
fix(radon): include tapi in validate_rad_request to avoid hard forks

lrubiorod:


rad/src/script.rs
```diff
@@ -211,10 +213,17 @@ fn errorify(kind: RadError) -> RadError {
 pub fn create_radon_script_from_filters_and_reducer(
     filters: &[RADFilter],
     reducer: u32,
+    active_wips: &Option<ActiveWips>,
```
tmpolaczyk, line 216: Clippy should detect this: prefer using `Option<&ActiveWips>` here, and in the callsite use `.as_ref()` to perform the conversion.

validations/src/validations.rs
```diff
@@ -238,7 +238,10 @@ pub fn validate_mint_transaction(
 }
 
 /// Function to validate a rad request
-pub fn validate_rad_request(rad_request: &RADRequest) -> Result<(), failure::Error> {
+pub fn validate_rad_request(
+    rad_request: &RADRequest,
+    active_wips: Option<ActiveWips>,
```
tmpolaczyk, line 243:
```suggestion
    active_wips: Option<&ActiveWips>,
```

rad/src/script.rs
```diff
@@ -252,7 +261,16 @@ pub fn create_radon_script_from_filters_and_reducer(
     )
     .map_err(|_| unknown_reducer(i128::from(reducer)))?;
     match rad_reducer {
-        RadonReducers::AverageMean | RadonReducers::Mode | RadonReducers::AverageMedian => {}
+        RadonReducers::AverageMean | RadonReducers::Mode => {}
+        RadonReducers::AverageMedian => {
+            if active_wips.wip0017() {
+            } else {
```
tmpolaczyk, line 267:
```suggestion
            if !active_wips.wip0017() {
```

tmpolaczyk (approved):

PR #2075, opened 2021-09-23
refactor(tapi): update tapi and to have WIP-17-18-19 activated

lrubiorod:


bridges/centralized-ethereum/src/actors/dr_sender/tests.rs
```diff
@@ -53,7 +53,7 @@ fn deserialize_dr_high_value() {
     // Setting the maximum allowed value to 1 nanowit below that will result in an error 0xE1:
     // BridgePoorIncentives
     let err = deserialize_and_validate_dr_bytes(&dro_bytes, total_value - 1).unwrap_err();
-    assert_eq!(err.encode_cbor(), vec![216, 39, 129, 24, 225]);
```
tmpolaczyk, line 56: This shouldn't change, I think it's because the request created above is no longer valid. Maybe `example_request` is setting the kind to unknown instead of http-get?

node/src/actors/json_rpc/json_rpc_methods.rs
```diff
@@ -1869,7 +1869,7 @@ mod tests {
         let inv_elem = InventoryItem::Block(block);
         let s = serde_json::to_string(&inv_elem).unwrap();
         // TODO: Use RadType::HttpGet after WIP-0019 activation
```
tmpolaczyk, line 1871:
```suggestion
```

rad/src/reducers/mod.rs
```diff
@@ -125,29 +124,15 @@ mod tests {
     }
 
     #[test]
-    fn test_reduce_average_median_tapi_activation() {
```
tmpolaczyk, line 128: Maybe it can be useful to keep a test that checks tapi activation? By doing something like:

```rust
active_wips
            .active_wips
            .remove(&"WIP0017-0018-0019".to_string());
```

Although I think this is already tested in the validations crate.

validations/src/tests/mod.rs
```diff
@@ -8149,90 +8330,6 @@ fn block_duplicated_tallies() {
     );
 }
 
-#[test]
-fn block_before_and_after_hard_fork() {
```
tmpolaczyk, line 8153: Why remove this test?

tmpolaczyk (approved):

PR #2082, opened 2021-09-28
feat(rad): add a maximum for retrieval timeout

### LINKED ISSUE #2081: Add a maximum value to "data_request_timeout_milliseconds" configurable parameter
Currently we are detecting that some requests that should be solved with a RequestTimeoutError are not being committed. It could be due to some miners increase this field more than half epoch, and after this time a commit will be considered invalid. To avoid this situation, a maximum value could be added.

Half epoch is 22,5 seconds, but we should leave time to broadcast it. In my opinion, 10 seconds is enough time to report a source as a RetrievalTimeoutError

lrubiorod:
Close #2081

### DIFF
--- a/node/src/actors/rad_manager/handlers.rs
+++ b/node/src/actors/rad_manager/handlers.rs
@@ -1,6 +1,7 @@
 //! Message handlers for `RadManager`
 
 use actix::{Handler, ResponseFuture};
+use std::time::Duration;
 use witnet_data_structures::radon_report::{RadonReport, ReportContext};
 use witnet_rad::{error::RadError, script::RadonScriptExecutionSettings, types::RadonTypes};
 use witnet_validations::validations::{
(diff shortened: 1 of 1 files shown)


node/src/actors/rad_manager/handlers.rs
```diff
@@ -12,6 +13,8 @@ use crate::actors::messages::{ResolveRA, RunTally};
 use super::RadManager;
 use futures::FutureExt;
 
+const MAX_RETRIEVAL_TIMEOUT: Duration = Duration::from_millis(10000);
```
tmpolaczyk, line 16: We should document this: why it is needed, why this default value.

node/src/actors/rad_manager/handlers.rs
```diff
@@ -69,26 +72,27 @@ impl Handler<ResolveRA> for RadManager {
...
-            )
-        } else {
-            Box::pin(fut)
-        }
+        let timeout = match timeout {
+            None => MAX_RETRIEVAL_TIMEOUT,
+            Some(timeout_from_config) => std::cmp::min(timeout_from_config, MAX_RETRIEVAL_TIMEOUT),
+        };
+
+        // Add timeout, if there is one
```
tmpolaczyk, line 80: Now the timeout is not optional.

```suggestion
        // Add timeout
```

tmpolaczyk: In which cases was this timeout set to `None`?

lrubiorod: It is set to None when the miner set 0 as a timeout

node/src/actors/rad_manager/handlers.rs
tmpolaczyk: Also add that this must be lower than half an epoch, and the commit transactions must have enough time to propagate through the entire network before block mining begins.

tmpolaczyk (approved):

PR #2084, opened 2021-09-29
chore: update list of user agents

tmpolaczyk:


aesedepece (approved):
Thanks, this was long due!

PR #2085, opened 2021-10-01
chore: update surf version

### LINKED ISSUE #1820: Update version of HTTP client "surf"
Currently we are using [surf](https://crates.io/crates/surf) 1.0.3 which was released on November 7, 2019. The latest version is surf 2.1 but there are some unknown problems when trying to actually use it, see PR https://github.com/witnet/witnet-rust/pull/1812

So if surf releases a new version, we should check if those problems are fixed.

The main drawback of keeping the old version is that it uses old versions of some `futures` libraries, which means more dependencies and longer compile time for the node. But we still use other dependencies that use that versions, so this may not be high priority right now.

Alternatively, we could try to migrate to the lower level crate [isahc](https://crates.io/crates/isahc), which is what surf uses internally.

tmpolaczyk:
Update version of surf, the http client used to resolve data requests.

Last time this was attempted some sources broke. See #1812

Fix #1820

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11,7 +11,7 @@ dependencies = [
  "actix-rt",
  "bitflags 1.2.1",
  "bytes 1.0.0",
- "crossbeam-channel 0.5.0",
+ "crossbeam-channel",
  "futures-core",
  "futures-sink",
  "futures-task",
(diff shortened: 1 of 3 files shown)


aesedepece:
This time it's working, I guess?

tmpolaczyk:
> This time it's working, I guess?

I hope so, but I wouldn't merge this yet, just in case.

lrubiorod (approved):

PR #2088, opened 2021-10-05
update user agent version and remove corner case

lrubiorod:


data_structures/src/builders.rs
```diff
@@ -27,7 +27,7 @@ pub const PROTOCOL_VERSION: u32 = 0x0000_0001;
 pub const CAPABILITIES: u64 = 0x0000_0000_0000_0001;
 
 /// User agent
-pub const USER_AGENT: &str = "full-node-desktop-edition";
+pub const USER_AGENT: &str = "full-node-desktop-edition-v1_4_1";
```
tmpolaczyk, line 30:
```suggestion
pub const USER_AGENT: &str = "full-node-desktop-edition-v1.4.1";
```

data_structures/src/builders.rs
```diff
@@ -27,7 +27,7 @@ pub const PROTOCOL_VERSION: u32 = 0x0000_0001;
 pub const CAPABILITIES: u64 = 0x0000_0000_0000_0001;
 
 /// User agent
-pub const USER_AGENT: &str = "full-node-desktop-edition";
+pub const USER_AGENT: &str = "full-node-desktop-edition-v1.4.1";
```
aesedepece, line 30: Let's construct this string dinamically with `witnet-rust`, plus arch, plus OS, plus version!

node/src/actors/chain_manager/handlers.rs
```diff
@@ -1114,26 +1113,13 @@ impl Handler<PeersBeacons> for ChainManager {
...
-                        // mining and preserve network stability
-                        } else {
-                            self.update_state_machine(StateMachine::AlmostSynced, ctx);
+                        // We will move to AlmostSynced to do not allow mining and preserve network stability
+                        self.update_state_machine(StateMachine::AlmostSynced, ctx);
 
-                            // We will remove those that are different from the last consensus or
-                            // peers that did not send any beacon
+                        // We will remove those that are different from the last consensus or
+                        // peers that did not send any beacon
```
aesedepece, line 1120:
```suggestion
                        // We will remove those peers that either signaled beacons differing from the last consensus or
                        // did not send any beacon
```

aesedepece (changes requested):

aesedepece (approved):

tmpolaczyk (approved):

PR #2089, opened 2021-10-05
fix(node): ice real sender address

tmpolaczyk:
Instead of the self-reported public_addr field


lrubiorod (approved):

aesedepece (approved):
Seems to be a reasonable and sensible change in this context, given that announced addresses are self-declarative.

PR #2105, opened 2021-10-28
feat(node): add method to fix missing blocks in storage

### LINKED ISSUE #2102: Missing blocks in storage
I noticed some errors about missing blocks in one node. It is not clear if that's a problem with witnet-rust, or a problem with rocksdb, so I think the best solution for now would be to assume that this can happen and perform automatic database integrity checks or something similar.

I implemented a JSON-RPC command to check for missing blocks, available in [this branch](https://github.com/tmpolaczyk/witnet-rust/tree/check-block-chain). Can be run as:

    echo '{"jsonrpc":"2.0","id":"1","method":"checkBlockChain","params":null}' | ./target/release/witnet -c witnet.toml node raw

And the result is this list of `(epoch, block_hash)` that is missing from that one node:

```json
[[681700,"c2674475ed7cb14f83d782ca6d40d27908b1393df779445e07253667e690983c"],[681701,"661fae380e01a03acd7332f0430d735c1b98eb615bb8ef00c58183b72317a173"],[681702,"1162fdfc976212e6571b848a59d4288496333bf5a4e87413093df937bd9e37e4"],[681703,"e004ef4925350f4bb4b916fc72bb14a3043cb6bc926a646465d9d180978fa9bf"],[681704,"dadabbbb1e9d99ed5c444cfceacc71791ab657931306feaac6f6e6e9edc7ef38"],[681705,"ae36350715c1f68df72377aa07590c3c4b0e8d78985881aba9b2658e1c9624d8"],[681706,"218f41843fcdc5f67e2e0e798fbc49c7293b0df51ac9dd05a3bf77324b17857c"],[681707,"07995e9b4857265fa0f6df95f60357f67ce7e5256f62a017b896992c69bdd6c4"],[681708,"7adcc6fd6df0fe2f205adf2d7cd44db63d253b0f2253b224c1496f4bdd690dd2"],[681709,"b995950291409753dd4c91184a5edbf703210284069d500d2d3401536bf3cb60"]]
```

Currently this kind of errors can be solved by doing a `rewind`, which will process all the blocks until the first missing block, and then continue doing a normal (slow) synchronization. So the next step would be to implement some functionality to retrieve those missing blocks from other peers automatically, without getting the user to run the rewind command.

tmpolaczyk:
Fix #2102

### DIFF
--- a/node/src/actors/chain_manager/handlers.rs
+++ b/node/src/actors/chain_manager/handlers.rs
@@ -10,8 +10,8 @@ use std::{
 
 use witnet_data_structures::{
     chain::{
-        Block, ChainState, CheckpointBeacon, DataRequestInfo, Epoch, Hash, Hashable, NodeStats,
-        SuperBlockVote, SupplyInfo,
+        Block, BlockMerkleRoots, ChainState, CheckpointBeacon, DataRequestInfo, Epoch, Hash,
+        Hashable, NodeStats, SuperBlockVote, SupplyInfo,
     },
     error::{ChainInfoError, TransactionError::DataRequestNotFound},
     mainnet_validations::ActiveWips,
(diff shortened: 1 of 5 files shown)


tmpolaczyk:
TODO: check if this method persists the data request (for dataRequestReport command)

tmpolaczyk:
> TODO: check if this method persists the data request (for dataRequestReport command)

It does not, so that's missing: persist data request report

aesedepece:
@tmpolaczyk what should we do with this?

tmpolaczyk:
So I took a look at this issue because I have been affected again, but this time it was only the transactions from the block that were missing, not the block. The missing transaction hash is 02bfa7550c30380f667dff1f51061ad0cf091f54c96e602fb78545e6d81f6b3f and the block hash is c92daffc61cc67b471b82804a9b01d6bbaf16a8a47546c7fe7e1dfa6abdbfa99 which is from 2022-07-01, so most probably the issue is still present in the master branch.

The approach presented in this draft, which consists of iterating over all the blocks and checking that they exist in the storage, does not work in the case above. We would need to iterate over all the transactions as well, and that would make this much slower.

Therefore I think it is better to use the another solution to fix this problem. In the past I used the rewind method and that fixes the missing block case, but it is slow. It currently does not work in the missing transaction case but that can be easily fixed by always writing all the transactions again. This also solves the "TODO: persist data request report" from this draft.

So improving the rewind method would be a better workaround, I will look into that.

PR #2108, opened 2021-11-02
chore: update rocksdb crate to 0.17.0

tmpolaczyk:


tmpolaczyk:
I think CI fails because the rocksdb version installed in github actions is not compatible with the rocksdb crate. So this update may break the setup of users that compile the node from source.

tmpolaczyk:
Blocked, see #2175

PR #2115, opened 2021-11-11
fix(rad): fix try_data_request function

lrubiorod:


aesedepece:
Can we add a test for `try_data_request` covering RNG?

wallet/tests/data_requests.rs
```diff
@@ -154,3 +155,55 @@ fn test_data_request_report_json_serialization() {
         report.tally.partial_results.unwrap().last().unwrap()
     );
 }
+
+/// This is a rather end-2-end test that use a RNG input and checks whether the
+/// final `RadonReport` complies with the Witnet Wallet API.
+#[test]
+fn test_try_data_request_rng() {
```
tmpolaczyk, line 162: This test does not fail if I remove the other changes of this pull request.

You can add the following line:

```rust
    assert_eq!(report.tally.result, RadonTypes::Bytes(RadonBytes::from(vec![254, 69, 239, 98, 10, 203, 40, 254, 158, 109, 130, 121, 108, 158, 200, 110, 49, 9, 152, 250, 37, 113, 2, 99, 125, 65, 249, 61, 118, 176, 184, 132])));
```

Also, this test is in the `wallet` crate, which does not make much sense. Let's move it to somewhere inside the `rad` crate. And remove all the partial_results asserts, because they are already checked in another test.

node/src/actors/rad_manager/handlers.rs
```diff
@@ -74,12 +74,14 @@ impl Handler<ResolveRA> for RadManager {
                     // Perform aggregation on the values that made it to the output vector after applying the
                     // source scripts (aka _normalization scripts_ in the original whitepaper) and filtering out
                     // failures.
-                    witnet_rad::run_aggregation_report(
+                    let (res, _ctx) = witnet_rad::run_aggregation_report(
```
tmpolaczyk, line 77: `_ctx` already means actor context, it may be confusing to give it the same name
```suggestion
                    let (res, _) = witnet_rad::run_aggregation_report(
```

rad/src/lib.rs
```diff
@@ -267,11 +269,14 @@ pub fn run_aggregation_report(
     aggregate: &RADAggregate,
     settings: RadonScriptExecutionSettings,
     active_wips: ActiveWips,
-) -> Result<RadonReport<RadonTypes>> {
-    let context = &mut ReportContext::from_stage(Stage::Aggregation);
+) -> (Result<RadonReport<RadonTypes>>, ReportContext<RadonTypes>) {
```
tmpolaczyk, line 272: The documentation of this function has not been updated to mention that now it also returns a `ReportContext`. Same in `run_tally_report`.

tmpolaczyk (approved):

PR #2120, opened 2021-11-17
[PoC] include a delay in block broadcasting to mitigate late blocks

### LINKED ISSUE #2117: PoC: late blocks retention delay

lrubiorod:
Close #2117

### DIFF
--- a/node/src/actors/chain_manager/mod.rs
+++ b/node/src/actors/chain_manager/mod.rs
@@ -2230,8 +2230,8 @@ impl ChainManager {
 // Auxiliary function that converts one delay in another
 fn delay_function(initial_delay: Duration) -> Duration {
     // TODO: Apply a right delay function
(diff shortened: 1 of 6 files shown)


node/src/actors/chain_manager/mod.rs
```diff
@@ -3528,19 +3593,19 @@ mod tests {
             assert_ne!(block_1, block_mal_1);
 
             // Process the modified candidate first
-            chain_manager.process_candidate(block_mal_1);
+            chain_manager.process_candidate(&mut Context::new(), block_mal_1, (0, 0));
```
tmpolaczyk, line 3596: I didn't know it was possible to call `Context::new()`... But maybe it's better to send an `AddCandidates` message? This is inside an async block so we can just call `send().await`.

PR #2124, opened 2021-11-24
chore: update secp256k1 dependency

tmpolaczyk:
And use re-exported version from witnet_crypto in other crates.

Also remove `SignatureError` which was not used anywhere, preferring `secp256k1::Error` instead.


lrubiorod (approved):

aesedepece:
Impediments to merging?

tmpolaczyk:
Nope, I just forgot

PR #2136, opened 2021-12-02
WIP HTTP-POST

### LINKED ISSUE #2130: Add HTTP-POST as witnet capability
Currently, the Currently, witnet-rust only supports HTTP-GET as a source type. To allow data request creators to choose more sources, we have to add support for HTTP-POST

tmpolaczyk:
Close #2130

See also https://github.com/witnet/WIPs/pull/76/files

Warning: this PR makes breaking changes to the storage, so backup before testing.


rad/src/lib.rs
```diff
@@ -4,6 +4,9 @@ use futures::{executor::block_on, future::join_all};
 use serde::Serialize;
 pub use serde_cbor::to_vec as cbor_to_vec;
 pub use serde_cbor::Value as CborValue;
+use std::str::FromStr;
+use surf::http::headers::{HeaderName, HeaderValues, ToHeaderValues};
```
lrubiorod, line 8: fmt

rad/src/lib.rs
```diff
@@ -152,6 +168,13 @@ pub fn run_retrieval_with_data_report(
                 http_get_response_with_data_report(retrieve, response, context, settings)
             }
             RADType::Rng => rng_response_with_data_report(response, context),
+            RADType::HttpPost => {
+                if context.active_wips.as_ref().map(|w| w.wiphttppost()) == Some(true) {
+                    http_post_response_with_data_report(retrieve, response, context, settings)
+                } else {
+                    panic!("Unsupported retrieval type: HTTP-POST");
```
lrubiorod, line 175: should be -> Err(RadError::UnknownRetrieval)?

rad/src/lib.rs
```diff
@@ -138,6 +141,19 @@ fn rng_response_with_data_report(
...
+    response: &str,
+    context: &mut ReportContext<RadonTypes>,
+    settings: RadonScriptExecutionSettings,
+) -> Result<RadonReport<RadonTypes>> {
+    let input = RadonTypes::from(RadonString::from(response));
+    let radon_script = unpack_radon_script(&retrieve.script)?;
+
+    execute_radon_script(input, &radon_script, context, settings)
+}
+
```
lrubiorod, line 156: should we rename to http_response_with_data_report and do not have the same function with 2 names?

rad/src/lib.rs
```diff
@@ -242,6 +264,106 @@ async fn rng_response(
     Ok(RadonReport::from_result(Ok(random_bytes), context))
 }
 
+/// Handle HTTP-POST response
+async fn http_post_response(
```
lrubiorod, line 268: also here, could be combined with the http_get?

tmpolaczyk: This one has a different line:

    let request = surf::post(&retrieve.url).body(retrieve.body.clone());

I guess we can add a boolean flag...

data_structures/tests/transaction_weight.rs
```diff
@@ -0,0 +1,108 @@
+use witnet_data_structures::chain::{transaction_example, RADRequest};
+use witnet_data_structures::proto::ProtobufConvert;
+use witnet_data_structures::transaction::Transaction;
```
lrubiorod, line 3: fmt

rad/src/lib.rs
```diff
@@ -266,25 +285,79 @@ async fn rng_response(
...
+) -> Result<RequestBuilder> {
+    // Add random user agent
+    request = request.header("User-Agent", UserAgent::random());
+
+    // Only set extra headers if wip-http-post is active
+    if context.active_wips.as_ref().map(|w| w.wiphttppost()) != Some(true) {
+        return Ok(request);
+    }
+
+    // Add extra_headers from retrieve.headers
```
tmpolaczyk, line 302: TODO: also check what happens when a HTTP-GET request with headers is processed before, during, and after the hard fork.

tmpolaczyk:
Try to add this graphql source as an example: https://github.com/witnet/witnet-price-feed-examples/issues/59

data_structures/src/mainnet_validations.rs
```diff
@@ -47,6 +47,9 @@ pub fn wip_info() -> HashMap<String, Epoch> {
     active_wips.insert("THIRD_HARD_FORK".to_string(), THIRD_HARD_FORK);
     active_wips.insert("WIP0014-0016".to_string(), 549141);
     active_wips.insert("WIP0017-0018-0019".to_string(), 683541);
+    // TODO: this is for testing HTTP-POST requests in rad/src/lib.rs
+    // Remove before merging
+    active_wips.insert("WIP0020".to_string(), 0);
```
lrubiorod, line 52: remove before merging :P

data_structures/src/mainnet_validations.rs
```diff
@@ -678,6 +686,7 @@ mod tests {
     }
 
     #[test]
+    #[ignore]
```
lrubiorod, line 689: remove this

validations/src/validations.rs
```diff
@@ -247,21 +247,26 @@ pub fn validate_rad_request(
...
-        let is_rng = active_wips.wip0019() && path.kind == RADType::Rng;
-        // If the sources are empty the data request is set as invalid
-        if path.url.is_empty() && !is_rng {
-            return Err(DataRequestError::NoRetrievalSources.into());
+            let is_rng = active_wips.wip0019() && path.kind == RADType::Rng;
+            // If the sources are empty the data request is set as invalid
+            if path.url.is_empty() && !is_rng {
+                return Err(DataRequestError::NoRetrievalSources.into());
+            }
+            unpack_radon_script(path.script.as_slice())?;
```
lrubiorod, line 268: could be refactored as an only if-elseif-else clause?

tmpolaczyk: I tried to avoid touching the "legacy code". Is it better after the latest change?

tmpolaczyk:
So it turns out that the changes to the `RADRetrieve` struct are actually breaking changes in
the storage:

```diff
pub struct RADRetrieve {
    /// Kind of retrieval
    pub kind: RADType,
    /// URL
    pub url: String,
    /// Serialized RADON script
    pub script: Vec<u8>,
+    /// Body of a HTTP-POST request
+    pub body: Vec<u8>,
+    /// Extra headers of a HTTP-GET or HTTP-POST request
+    pub headers: Vec<(String, String)>,
}
```

Adding new fields is backwards compatible when using protocol buffers because the missing
fields will just have the default value, so at the network level the change seems to be working fine.
However, the storage uses bincode to serialize this data type, and bincode does not support adding
new fields because it is an untagged binary format.

One possible way to fix and avoid this breaking change would be to use protocol buffers
in the storage, at least for data types that already have a protobuf schema, like the `RADRetrieve` type.
However that would need a migration strategy for the existing types.

The reason this time we cannot do the same as we did to change the ChainState is because the affected
struct (`RADRetrieve`) is used inside other data structures, so we would need to implement some logic
to handle mutating bytes in nested fields in bincode format. Or, alternatively, we could duplicate
all of the affected structs (save a copy of the definition of the old version), and perform a one
time migration by reading all the keys of the database as the old version, and saving them as the
new version. Both of these approches require considerably more effort to implement, but they would
hopefully be generic enough so that breaking the storage stops being a big problem like it is
currently.

Affected structs: `ChainState` when there are some pending data requests, and all the `Blocks` that have at least one data request.
Essentially anything that uses one of these types:

```
ChainState / DataRequestPool / HashMap<Hash, DataRequestState> / DataRequestState / DataRequestOutput / RADRequest / Vec<RADRetrieve> / RADRetrieve

Block / BlockTransactions / Vec<DRTransaction> / DRTransaction / DRTransactionBody / DataRequestOutput / RADRequest / Vec<RADRetrieve> / RADRetrieve
```

Here is a summary of the storage migration strategies we have implemented so far.

# witnet_node storage migrations

### ChainState hack:

https://github.com/witnet/witnet-rust/blob/59e90da85120d21586be803cf2b34077c1978d13/node/src/storage_mngr/node_migrations.rs

- Works by using raw byte manipulation

Pros:
- Migration code is in witnet_node (does not affect other dependencies, only affects the storage manager)
- Easy to add new fields to complex data types

Cons:
- Only affects select keys from the database (like "CHAIN_STATE", forgetting about one key will break that key)
- Does not work with nested types: `Option<ChainState>` or `Vec<ChainState>` will not be backwards compatible


### RADRetrieve hack:

https://github.com/witnet/witnet-rust/pull/2136

- Works by implementing a custom serde deserialization visitor

Pros:
- Works with nested types: `Vec<RADRetrieve>` is handled correctly
- Affects all the keys of the database (no danger of forgetting about some key and accidentally breaking that functionality)

Cons:
- Migration code is in witnet_data_structures (affects other dependencies)
- Migration code is in the Deserialize implementation (affects other serialization formats, but only if they are binary: JSON will not be affected, but MessagePack will)
- Version field must be stored alongside every instance of RADRetrieve. It shouldn't be needed because RADRetrieve is always inside a DataRequestOutput, so we could store one db_version in DataRequestOutput. Or even one db_version per database key, because all the transactions inside a block will have the same database version.

tmpolaczyk:
I sometimes get this error when running the `bitcoin_last_hash.json` example:

```
[2021-12-31T08:10:45Z WARN  isahc::handler] request completed with error: ResponseBodyError(None): unknown error 
```

tmpolaczyk:
I merged this alongside #2138 into the https://github.com/witnet/witnet-rust/tree/witnet-1.5 branch.

aesedepece:
> I merged this alongside #2138 into the https://github.com/witnet/witnet-rust/tree/witnet-1.5 branch.

I'll close then. Thanks a lot @tmpolaczyk

tmpolaczyk:
Ok, I opened #2148 if there are any further comments.

PR #2137, opened 2021-12-02
rust 1.57

tmpolaczyk:


src/cli/node/json_rpc_client.rs
```diff
@@ -1518,15 +1518,12 @@ type ResponseBlockChain<'a> = Vec<(u32, &'a str)>;
...
     jsonrpc: &'a str,
-    id: Id<'a>,
     result: T,
 }
 
 /// A failed request returns an error with code and message
 #[derive(Debug, Deserialize)]
-struct JsonRpcError<'a> {
-    jsonrpc: &'a str,
-    id: Id<'a>,
```
tmpolaczyk, line 1529: Clippy complains that this fields are never read. Removing them means that we no longer validate that the JSON-RPC error has a `{"jsonrpc":"2.0"}` field. But this is a hand-written JSON-RPC client and we are missing many error handling anyway, so I just removed it.

lrubiorod (approved):

PR #2138, opened 2021-12-03
feat(rad): create function to parse xml to RadonMap

### LINKED ISSUE #2129: Implement StringtoXml radon operator
Implement radon operator `String.toXml` which parses a string containing an XML to a Map

lrubiorod:
Implement radon operator `String.toXml` which parses a string containing an XML to a Map

Close #2129


rad/src/operators/string.rs
```diff
@@ -35,6 +36,89 @@ pub fn parse_json_array(input: &RadonString) -> Result<RadonArray, RadError> {
...
+    let mut a = vec![];
+    for children in input.nodes() {
+        match children {
+            minidom::Node::Element(elem) => {
+                let radon_map = parse_element_map(elem.clone(), false);
+                a.push(RadonTypes::from(radon_map));
+            }
+            minidom::Node::Text(text) => {
+                let text_var = text.trim().clone().to_string();
+                if !text_var.is_empty() {
```
tmpolaczyk, line 94: Can you add a test to show the purpose of this check?

tmpolaczyk: Or a comment

tmpolaczyk:
The current implementation fails to parse elements that are missing a `xmlns` key:

```rust
    #[test]
    fn test_parse_xml_map_3() {
        let xml_map = RadonString::from(
            r#"<?xml version="1.0" encoding="UTF-8"?>
<note>
  <to>Tove</to>
  <from>Jani</from>
  <heading>Reminder</heading>
  <body>Don't forget me this weekend!</body>
</note>
        "#,
        );
        let output = parse_xml_map(&xml_map).unwrap();
}
```

Outputs:

```
thread 'operators::string::tests::test_parse_xml_map_3' panicked at 'called `Result::unwrap()` on an `Err` value: XmlParse { description: "the XML element is missing a namespace" }', rad/src/operators/string.rs:481:46
```

rad/src/operators/string.rs
```diff
@@ -35,6 +36,103 @@ pub fn parse_json_array(input: &RadonString) -> Result<RadonArray, RadError> {
...
+        _ => unreachable!("XML parsing error"),
+    };
+
+    match map.remove("_children") {
+        Some(RadonTypes::Array(rad_array)) => {
+            let children = rad_array.value();
+            let children_len = children.len();
+
+            match (map.is_empty(), children_len) {
+                (true, x) if x == 0 => {
```
tmpolaczyk, line 52:
```suggestion
                (true, 0) => {
```

Also it would be nice to have some examples of each branch, perhaps as tests.

rad/src/operators/string.rs
```diff
@@ -35,6 +36,103 @@ pub fn parse_json_array(input: &RadonString) -> Result<RadonArray, RadError> {
...
+                    map.insert(label, RadonString::from("".to_string()).into());
+                }
+                (true, x) if x == 1 => {
+                    map.insert(label, children.first().unwrap().clone());
```
tmpolaczyk, line 56: I think this way you can avoid the clone

```suggestion
                    map.insert(label, children.into_iter().next().unwrap());
```

rad/src/operators/string.rs
```diff
@@ -35,6 +36,103 @@ pub fn parse_json_array(input: &RadonString) -> Result<RadonArray, RadError> {
...
+    let mut a = vec![];
+    for children in input.nodes() {
+        match children {
+            minidom::Node::Element(elem) => {
+                let radon_map = parse_element_map(elem.clone(), false);
+                a.push(radon_map.into());
+            }
+            minidom::Node::Text(text) => {
+                // This check is to avoid blank spaces in xml would be included in the RadonMap
+                let text_var = text.clone().trim().to_string();
```
tmpolaczyk, line 112:
```suggestion
                let text_var = text.trim().to_string();
```

rad/src/operators/string.rs
```diff
@@ -35,6 +36,103 @@ pub fn parse_json_array(input: &RadonString) -> Result<RadonArray, RadError> {
...
+    }
+    for (k, v) in input.attrs() {
+        map.insert(format!("@{}", k), RadonString::from(v).into());
+    }
+
+    let mut a = vec![];
+    for children in input.nodes() {
+        match children {
+            minidom::Node::Element(elem) => {
+                let radon_map = parse_element_map(elem.clone(), false);
```
tmpolaczyk, line 107: You can avoid this clone by making `parse_element_map` take a `&minidom::Element`.

lrubiorod:
> 

The library minidom assumes that a correct xml has a namespace

lrubiorod:
> The current implementation fails to parse elements that are missing a `xmlns` key:
> 
> ```rust
>     #[test]
>     fn test_parse_xml_map_3() {
>         let xml_map = RadonString::from(
>             r#"<?xml version="1.0" encoding="UTF-8"?>
> <note>
>   <to>Tove</to>
>   <from>Jani</from>
>   <heading>Reminder</heading>
>   <body>Don't forget me this weekend!</body>
> </note>
>         "#,
>         );
>         let output = parse_xml_map(&xml_map).unwrap();
> }
> ```
> 
> Outputs:
> 
> ```
> thread 'operators::string::tests::test_parse_xml_map_3' panicked at 'called `Result::unwrap()` on an `Err` value: XmlParse { description: "the XML element is missing a namespace" }', rad/src/operators/string.rs:481:46
> ```

The library minidom assumes that a correct xml has a namespace

tmpolaczyk:
> The library minidom assumes that a correct xml has a namespace

So, fork? :P

lrubiorod:
> > The library minidom assumes that a correct xml has a namespace
> 
> So, fork? :P

XML format should have namespace, so if we do not a specific demand of XML APIs without namespace, I would not fork

tmpolaczyk:
The use of recursion in `parse_element_map` can lead to a stack overflow with deeply nested children:

```rust
    #[test]
    fn test_parse_xml_map_stack_overflow() {
        let n = 1000;
        let xml_map = RadonString::from(format!(
            r#"<?xml version="1.0"?>
            <Tests xmlns="https://witnet.io/">
            {}{}
            </Tests>
        "#,
            "<A>".repeat(n),
            "</A>".repeat(n)
        ));
        let output = parse_xml_map(&xml_map).unwrap();
    }
```

rad/src/operators/string.rs
```diff
@@ -35,6 +38,117 @@ pub fn parse_json_array(input: &RadonString) -> Result<RadonArray, RadError> {
     item.try_into()
 }
 
+fn magic_map_cleaning(input: BTreeMap<String, RadonTypes>) -> RadonTypes {
+    let map_lenght = input.len();
```
tmpolaczyk, line 42:
```suggestion
    let map_length = input.len();
```

rad/src/operators/string.rs
```diff
@@ -35,6 +38,117 @@ pub fn parse_json_array(input: &RadonString) -> Result<RadonArray, RadError> {
...
+    let mut map: BTreeMap<String, RadonTypes> = BTreeMap::new();
+    if use_namespace {
+        map.insert("_xmlns".to_string(), RadonString::from(input.ns()).into());
+    }
+    for (k, v) in input.attrs() {
+        map.insert(format!("@{}", k), RadonString::from(v).into());
+    }
+
+    for child in input.nodes() {
+        let (ignore, key, value) = match child {
```
tmpolaczyk, line 96: Instead of `(bool, K, V)` you can use `Option<(K, V)>` to better represent this condition.

rad/src/operators/string.rs
```diff
@@ -35,6 +38,117 @@ pub fn parse_json_array(input: &RadonString) -> Result<RadonArray, RadError> {
...
+                        "_text".to_string(),
+                        RadonString::from(text_var).into(),
+                    )
+                } else {
+                    (true, "".to_string(), RadonString::from("").into())
+                }
+            }
+        };
+
+        match (ignore, map.remove(&key)) {
```
tmpolaczyk, line 118: remove + insert can be avoided by using the entry api:

```rust
map.entry(key).or_default().push(value)
```

rad/src/operators/string.rs
```diff
@@ -35,6 +38,117 @@ pub fn parse_json_array(input: &RadonString) -> Result<RadonArray, RadError> {
...
+                    None => unreachable!("Invalid map"),
+                }
+            }
+
+            RadonMap::from(new_map).into()
+        }
+    }
+}
+
+pub fn parse_element_map(
```
tmpolaczyk, line 78: This looks like it should be a private function

```suggestion
fn parse_element_map(
```

rad/src/operators/string.rs
```diff
@@ -35,6 +38,117 @@ pub fn parse_json_array(input: &RadonString) -> Result<RadonArray, RadError> {
...
+                map.insert(key, RadonArray::from(new_array).into());
+            }
+            _ => unreachable!("Invalid XML parsing"),
+        }
+    }
+
+    Ok(magic_map_cleaning(map))
+}
+
+pub fn parse_xml_map(input: &RadonString) -> Result<RadonMap, RadError> {
```
tmpolaczyk, line 135: Maybe it's worth to add a link to the WIP, because it explains the parsing process?

rad/src/operators/string.rs
```diff
@@ -35,6 +38,117 @@ pub fn parse_json_array(input: &RadonString) -> Result<RadonArray, RadError> {
...
+            // <tag></tag>
+            RadonString::from("".to_string()).into()
+        }
+        (1, Some(RadonTypes::Array(rad_array))) => {
+            if rad_array.value().len() == 1 {
+                // <tag>Hello</tag>
+                rad_array.value().into_iter().next().unwrap()
+            } else {
+                // <tag>Hello\nWorld</tag>
+                rad_array.clone().into()
```
tmpolaczyk, line 55: This branch does not have any tests

rad/src/operators/string.rs
```diff
@@ -35,6 +38,113 @@ pub fn parse_json_array(input: &RadonString) -> Result<RadonArray, RadError> {
...
+        }
+    }
+
+    let only_text_children = element_children.is_empty() && map.is_empty();
+
+    if only_text_children && text_children.is_empty() {
+        let text_value = RadonString::from("").into();
+        Ok(text_value)
+    } else if only_text_children {
+        let text_value = text_children.into_iter().next().unwrap();
```
tmpolaczyk, line 122: This branch has an implicit assertion that when `only_text_children` is true, there will only be 1 element in `text_children` vector. How about we make it explicit? Or handle that case somehow.

```suggestion
        assert_eq!(text_children.len(), 1);
        let text_value = text_children.into_iter().next().unwrap();
```

tmpolaczyk (approved):
I merged this into the https://github.com/witnet/witnet-rust/tree/witnet-1.5 branch

tmpolaczyk:
Closing, see #2148 if there are any further comments.

PR #2139, opened 2021-12-15
Fix/toolkit/errors when tallying

aesedepece:
This allows `witnet_toolkit` and other components reliant on the `try_data_request` function to run data requests in a more similar way to how `RadManager` does (0.2 tally pre-condition before aggregation stage).


tmpolaczyk:
Clippy is failing because the tests are failing to compile because of the refactor.

`validations/src/tests/tally_precondition.rs` can be moved to `rad/tests/tally_precondition.rs` and the `test_counter` test can be moved to `rad/src/conditions.rs`

tmpolaczyk:
Also you can add this test to `rad/src/lib.rs` to ensure that the precondition works as expected. In master this fails with the same error as https://github.com/witnet/witnet-price-feed-examples/pull/58#discussion_r768797910:

```
  left: `RadonError(RadonError { inner: UnhandledIntercept { inner: Some(UnsupportedOpNonHomogeneous { operator: "RadonReducers::Mode" }), message: None } })`,
```

```rust
    /// Ensure that `try_data_request` filters errors before calling `run_aggregation`.
    #[test]
    fn test_try_data_request_filters_aggregation_errors() {
        let script = cbor_to_vec(&Value::Array(vec![Value::Integer(
            RadonOpCodes::StringAsInteger as i128,
        )]))
        .unwrap();
        let request = RADRequest {
            time_lock: 0,
            retrieve: vec![
                RADRetrieve {
                    kind: RADType::HttpGet,
                    url: String::from(""),
                    script: script.clone(),
                },
                RADRetrieve {
                    kind: RADType::HttpGet,
                    url: String::from(""),
                    script: script.clone(),
                },
                RADRetrieve {
                    kind: RADType::HttpGet,
                    url: String::from(""),
                    script,
                },
            ],
            aggregate: RADAggregate {
                filters: vec![],
                reducer: RadonReducers::Mode as u32,
            },
            tally: RADTally {
                filters: vec![],
                reducer: RadonReducers::Mode as u32,
            },
        };
        let report = try_data_request(
            &request,
            RadonScriptExecutionSettings::enable_all(),
            Some(&["1", "1", "error"]),
        );
        let tally_result = report.tally.into_inner();

        assert_eq!(tally_result, RadonTypes::Integer(RadonInteger::from(1)));
    }
```

rad/src/lib.rs
```diff
@@ -81,19 +85,40 @@ pub fn try_data_request(
...
-        .iter()
-        .map(|report| report.result.clone())
-        .collect();
 
-    let (aggregation_result, aggregation_context) = run_aggregation_report(
-        retrieval_values,
-        &request.aggregate,
-        settings,
-        current_active_wips(),
+    // Evaluate aggregation pre-condition by using the same logic than for tally pre-condition,
```
tmpolaczyk, line 89: This is duplicated in `rad/src/lib.rs` and in `node/src/actors/rad_manager/handlers.rs`. It would be nice to move this logic into a new function, similar to `run_tally` which handles the precondition internally.

tmpolaczyk: Actually I just noticed that there are two different `run_tally` functions, one in `rad/src/lib.rs` and one in `validations/src/validations.rs`. The one from validations handles the preconditions and calls the one from rad...

So this `try_data_request` function is missing the tally precondition check, but it may be not needed because there should always be only one "reveal", so it's always in consensus with itself, right?

aesedepece:
All fixes and suggested have been incorporated in b1d1875d9e3ec225d47d52fb596293d218393fe1. Thanks @tmpolaczyk

tmpolaczyk (approved):

node/src/actors/chain_manager/mining.rs
```diff
@@ -56,6 +55,7 @@ use crate::{
     },
     signature_mngr,
 };
+use witnet_rad::conditions::radon_report_from_error;
```
lrubiorod, line 58: fmt

rad/src/lib.rs
```diff
@@ -5,7 +5,15 @@ use serde::Serialize;
 pub use serde_cbor::to_vec as cbor_to_vec;
 pub use serde_cbor::Value as CborValue;
 
+use witnet_data_structures::{
+    chain::{RADAggregate, RADRequest, RADRetrieve, RADTally, RADType},
+    mainnet_validations::{current_active_wips, ActiveWips},
+    radon_report::{RadonReport, ReportContext, RetrievalMetadata, Stage, TallyMetaData},
+};
+
+use crate::conditions::TallyPreconditionClauseResult;
```
lrubiorod, line 14: fmt

Issue #2094, opened 2021-10-08
Check malleability attack vectors


tmpolaczyk:
In some parts of the node we use a set of seen hashes to avoid validating the same item more than once. For example when validating block candidates we first check if the new hash is already in the `seen_candidates` hashset, and only new hashes are further validated. The problem is that invalid blocks are also inserted into `seen_candidates`, and this introduces an attack vector:

* Attacker mines block candidate A
* Attacker sends block candidate A to node C
* Node B mines block candidate B, which is a better candidate than A
* Attacker sees block candidate B, changes the block signature making it invalid, and sends the modified candidate B to node C
* Node C receives modified candidate B and inserts `hash(B)` into `seen_candidates` but since candidate B has an invalid signature it not considered better candidate than block candidate A
* Node C receives candidate B with valid signature from another peer, but since `hash(B)` is already in `seen_candidates`, this block is not validated
* Node C consolidates block candidate A instead of the better block candidate B

In pull request #2092 we fix this bug for the case of block candidates, but other messages (transactions, superblocks) could also be affected.

- [x] Block candidates #2092 
- [x] Value transfer transactions #2144
- [x] Other transactions
- [x] Superblocks

tmpolaczyk:
Superblock votes are not affected because we already assume that anyone can tamper with them, so if the secp signature is not valid the votes are ignored. And also we already need to check more than one vote per identity to detect double voting, so the list of received votes is a `HashSet<SuperBlockVote>`, which includes the signature so any change to the signature will be handled as a different vote. And two votes to the same superblock with a different but valid signature are handled as one vote, so everything works as expected in that case.

Tommytrg:
This issue is no longer relevant. Besides being resolved, the new candidate selection mechanism now follows a deterministic order of eligibility. This means we can anticipate which validator will produce the best possible candidate, making the rest irrelevant unless the top candidate is unavailable and so on in a cascading manner.

PR #2144, opened 2021-12-22
Fix potential transaction malleability attack vectors

tmpolaczyk:
Part of #2094

The worst that could happen is that a node misses some transactions, which does not sound very dangerous. However if a large portion of the network is affected by this bug, they could create empty blocks that reduce the throughput of the network.

The solution is the same as in #2092

```diff
-// A hashset of recently received transactions hashes
-// Used to avoid validating the same transaction more than once
-pending_transactions: HashSet<Hash>,
+/// Set that stores all the recently received transactions
+seen_transactions: HashSet<Transaction>,
```


aesedepece (approved):
Looks like the right approach to me.

PR #2158, opened 2022-03-02
chore: update Cargo.lock dependencies

lrubiorod:


tmpolaczyk:
Not sure if this can be called a refactor because updating dependencies can fix some bugs or introduce new ones, I usually call it "chore" because creating this pull request is equivalent to running one command.

tmpolaczyk (approved):

PR #2159, opened 2022-03-03
WIP store utxo set in the database and not in ram 2

tmpolaczyk:


data_structures/src/utxo_pool.rs
```diff
@@ -5,35 +5,64 @@ use crate::{
 use itertools::Itertools;
 use serde::{Deserialize, Serialize};
 use std::collections::{HashMap, HashSet};
+use std::fmt;
+use std::str::FromStr;
+use std::sync::Arc;
```
lrubiorod, line 10: fmt

data_structures/src/utxo_pool.rs
```diff
@@ -5,35 +5,64 @@ use crate::{
...
-    /// Map of output pointer to a tuple of:
-    /// * Value transfer output
-    /// * The number of the block that included the transaction
-    ///   (how many blocks were consolidated before this one).
-    map: HashMap<OutputPointer, (ValueTransferOutput, u32)>,
+    // Map of output pointer to a tuple of:
+    // * Value transfer output
+    // * The number of the block that included the transaction
+    //   (how many blocks were consolidated before this one).
+    //map: HashMap<OutputPointer, (ValueTransferOutput, u32)>,
```
lrubiorod, line 21: it should be removed before merging

data_structures/src/utxo_pool.rs
```diff
@@ -43,33 +72,160 @@ impl UnspentOutputsPool {
...
+        v: ValueTransferOutput,
+        block_number: u32,
+    ) -> Option<(ValueTransferOutput, u32)> {
+        // TODO: this get before insert does not make much sense because UTXOs are immutable
+        // Maybe it can be used as a sanity check that UTXOs are only written once, but it is not
+        // very useful
+        let old_vto = self.get_map(&k);
+
+        let key_string = format!("UTXO-{}", k);
+        log::debug!("PUT {}", key_string);
```
lrubiorod, line 107: dont forget remove these logs

data_structures/src/utxo_pool.rs
```diff
@@ -470,10 +614,10 @@ impl<'a> UtxoDiff<'a> {
         output: ValueTransferOutput,
         block_number: Option<u32>,
     ) {
+        // TODO: this assumes that we never insert-remove-insert again
```
tmpolaczyk, line 617:
```suggestion
```

That operation would now panic with error "UTXO did already exist"

node/src/actors/chain_manager/actor.rs
```diff
@@ -225,6 +225,60 @@ impl ChainManager {
...
+                    .into_actor(act)
+                    .map_err(|err, _act, _ctx| {
+                        log::error!("Failed to get storage backend: {}", err);
+                    })
+                    .and_then(|backend, act, _ctx| {
+                        chain_state.unspent_outputs_pool.db = Some(backend);
+
+                        let fut: Pin<Box<dyn ActorFuture<Self, Output = Result<ChainState, ()>>>> = if !chain_state.unspent_outputs_pool_old_migration_db.is_empty() {
+                            // TODO: this fails because the backend is not yet initialized here, need to
+                            // mode that and_then (the last one of this function) before this one
```
tmpolaczyk, line 243:
```suggestion
```

This is fixed

lrubiorod:
I think we should have a discussion before merge this PR due to being a non retrocompatible change for the data base

tmpolaczyk:
There are some problems when the synchronization process is interrupted, it is possible for the node to become forked and the only way to recover is to use the rewind command. I think it is related to #2102, will keep investigating.

node/src/actors/chain_manager/actor.rs
```diff
@@ -305,6 +385,57 @@ impl ChainManager {
...
+                    .map_err(|err, _act, _ctx| {
+                        log::error!("Failed to get storage backend: {}", err);
+                    })
+                    .map_ok(move |backend, _act, _ctx| (config, backend))
+            })
+            .map_ok(|(config, backend), _act, _ctx| {
+                let magic = config.consensus_constants.get_magic();
+
+                let all_chain_states: Vec<_> = backend
+                    .prefix_iterator(b"chain-")
```
tmpolaczyk, line 410: This doesn't work because the chain state key is serialized using bincode, and the prefix includes the key length as a little endian u64. So we would need to check all the possible key lengths, from "chain-0" to "chain-65536".

tmpolaczyk, line 410: This doesn't work because the chain state key is serialized using bincode, and the prefix includes the key length as a little endian u64. So we would need to check all the possible key lengths, from "chain-0" to "chain-65536".

tmpolaczyk:
> There are some problems when the synchronization process is interrupted, it is possible for the node to become forked and the only way to recover is to use the rewind command. I think it is related to #2102, will keep investigating.

This is fixed now

node/src/actors/chain_manager/actor.rs
tmpolaczyk: Fixed

data_structures/src/utxo_pool.rs
tmpolaczyk: Done

tmpolaczyk:
> This is still unresolved but I think I will just add a check to ensure that there is only one chain state, and panic otherwise.

This is also implemented now

node/src/storage_mngr.rs
```diff
@@ -19,9 +19,10 @@ use crate::{
 };
 use witnet_config::{config, config::Config};
 use witnet_data_structures::chain::ChainState;
-use witnet_storage::{backends, storage};
+use witnet_storage::{backends, storage, storage::Storage};
 
 pub use node_migrations::*;
+use witnet_storage::storage::WriteBatch;
```
lrubiorod, line 25: fmt

storage/src/backends/hashmap.rs
```diff
@@ -3,38 +3,97 @@
 //! Storage backend that keeps data in a heap-allocated HashMap.
 use std::collections::HashMap;
 
-use crate::storage::{Result, Storage};
+use crate::storage::{Result, Storage, StorageIterator, WriteBatch, WriteBatchItem};
+use std::sync::{RwLock, RwLockReadGuard};
```
lrubiorod, line 7: fmt

lrubiorod (approved):

PR #2171, opened 2022-04-06
Storage: implement forward and reverse prefix iterators

tmpolaczyk:


aesedepece:
This is rel: #2177 , right?

tmpolaczyk:
> This is rel: #2177 , right?

Yes

aesedepece (approved):

storage/src/storage.rs
```diff
@@ -21,9 +21,23 @@ pub trait Storage {
...
+        prefix: &'b [u8],
+    ) -> Result<StorageIterator<'a>>;
+
+    /// Create an iterator over all the keys that start with the given prefix, in reverse, starting
+    /// from the last key.
+    fn prefix_iterator_reverse<'a, 'b: 'a>(
+        &'a self,
+        prefix: &'b [u8],
+    ) -> Result<StorageIterator<'a>>;
+
```
tmpolaczyk, line 40: Actually, it may be better to only have a single method in this API, something like `iterate_range(start: &[u8], end: &[u8]) -> DoubleEndedIterator`

tmpolaczyk:
Closing for now, I will probably open a new PR that also includes the block_chain in database feature, when implemented.

PR #2193, opened 2022-04-26
fix(node): do not consolidate genesis block twice during resync

tmpolaczyk:
Currently the rewind command is broken because the genesis block is consolidated twice, resulting in error "UTXO did already exist". This pull request fixes it.


aesedepece (approved):
utACK

PR #2203, opened 2022-05-25
MultiSig PoC using Scripting

### LINKED ISSUE #2196: Scripting: Validate signatures in validation.rs from witness field

### LINKED ISSUE #2197: Scripting: Implement CheckMultiSig OPCODE

### LINKED ISSUE #2198: Scripting: Validate redeem script

lrubiorod:
Close #2196 
Close #2197 
Close #2198


data_structures/src/chain.rs
```diff
@@ -61,7 +60,7 @@ pub struct ChainInfo {
     /// Blockchain valid environment
     pub environment: Environment,
 
-    /// Blockchain Protocol constants
+    /// Blockchain Protocol constantsq
```
tmpolaczyk, line 63:
```suggestion
    /// Blockchain Protocol constants
```

data_structures/src/chain.rs
```diff
@@ -2398,22 +2516,32 @@ impl TransactionsPool {
...
+        }
     }
 
     /// Remove a data request transaction from the pool but do not remove other transactions that
     /// may try to spend the same UTXOs.
     /// This should be used to remove transactions that did not get included in a consolidated
     /// block.
-    /// If the transaction did get included in a consolidated block, use `dr_remove` instead.
-    fn dr_remove_inner(&mut self, key: &Hash, consolidated: bool) -> Option<DRTransaction> {
+    /// If the transaction did get included in a consolidated block, use `dr_rTemove` instead.
```
tmpolaczyk, line 2529:
```suggestion
    /// If the transaction did get included in a consolidated block, use `dr_remove` instead.
```

data_structures/src/chain.rs
```diff
@@ -2319,11 +2418,14 @@ impl TransactionsPool {
...
         let key = tx.hash();
-        let transaction = self.vt_remove_inner(&key, true);
+        let transaction = self.vt_and_sh_remove_inner(&key, true);
 
         self.remove_inputs(&tx.body.inputs);
 
-        transaction
+        match transaction {
+            Some(Transaction::ValueTransfer(vt_tx)) => Some(vt_tx),
+            _ => None,
```
tmpolaczyk, line 2427: Maybe panic instead of returning None?

data_structures/src/chain.rs
```diff
@@ -2722,21 +2894,21 @@ impl TransactionsPool {
...
             .iter()
             .rev()
-            .filter_map(move |(_, h)| self.vt_transactions.get(h).map(|(_, t)| t))
+            .filter_map(move |(_, h)| self.vt_and_sh_transactions.get(h).map(|(_, t)| t))
     }
 
     /// An iterator visiting all the data request transactions
     /// in the pool
-    pub fn dr_iter(&self) -> impl Iterator<Item = &DRTransaction> {
+    pub fn dr_iter(&self) -> impl Iterator<Item = &Transaction> {
```
tmpolaczyk, line 2911: Maybe it's better to leave this as a `DRTransaction` iterator:

```suggestion
    pub fn dr_iter(&self) -> impl Iterator<Item = &DRTransaction> {
```

This can be done using a simple `match t { DR(x) => x, _ => unreachable!() }` in the `.map` below

validations/src/validations.rs
```diff
@@ -45,45 +49,51 @@ use witnet_rad::{
...
+        Transaction::DataRequest(dr_tx) => {
+            let in_value =
+                transaction_inputs_sum(&dr_tx.body.inputs, utxo_diff, epoch, epoch_constants)?;
+            let out_value = transaction_outputs_sum(&dr_tx.body.outputs)?
+                .checked_add(dr_tx.body.dr_output.checked_total_value()?)
+                .ok_or(TransactionError::OutputValueOverflow)?;
+
+            (in_value, out_value)
+        }
+        _ => unreachable!("Only VT, DR and script transactions allowed"),
```
tmpolaczyk, line 95:
```suggestion
        _ => panic!("Only VT, DR and script transactions allowed"),
```

Because there is nothing preventing users from calling `transaction_fee` with other kinds of transactions.

validations/src/validations.rs
```diff
@@ -1092,6 +1187,80 @@ pub fn validate_pkh_signature(
             }
             .into());
         }
+    } else {
+        log::error!("Normal Input can not spent a ScriptOutput");
```
tmpolaczyk, line 1191: This branch will also be entered when the output is `None`, which can happen in some cases when trying to do a double spend. So convert the `if let` into a `match`, where the `None` case is just an empty `{}`.

data_structures/src/chain.rs
lrubiorod: Nop, remove returns None when it doesnt find the key specified

data_structures/src/chain.rs
lrubiorod: Regarding transaction_fee function now needs a Transaction as an argument, it could be easier on this way

data_structures/src/chain.rs
tmpolaczyk: I see, I was thinking about the case when the transaction is `Some(Transaction::Script(_))`

tmpolaczyk (approved):

PR #2208, opened 2022-05-30
feat: implement script transaction weight calculation

### LINKED ISSUE #2205: Scripting: support multiple ScriptTransaction weight

tmpolaczyk:
Close #2205

### DIFF
--- a/data_structures/src/transaction.rs
+++ b/data_structures/src/transaction.rs
@@ -296,7 +296,18 @@ impl ScriptTransaction {
     /// This is the weight that will be used to calculate
     /// how many transactions can fit inside one block
     pub fn weight(&self) -> u32 {
-        self.body.weight()
+        // TODO: should witness field affect weight?
+        // INPUT_SIZE does already take into account the weight of a keyed signature, so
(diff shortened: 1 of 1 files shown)


data_structures/src/transaction.rs
```diff
@@ -324,11 +335,28 @@ impl ScriptTransactionBody {
     /// Script transaction weight. It is calculated as:
     ///
     /// ```text
-    /// ST_weight = N*INPUT_SIZE + M*OUTPUT_SIZE*gamma
+    /// ST_weight = N*INPUT_SIZE + M*OUTPUT_SIZE*gamma + REDEEM_SCRIPTS_SIZE
```
lrubiorod, line 338: Should spent the same weight a VT output and a ScriptOutput?

tmpolaczyk: I assumed they can be the same because they are roughly the same size, the only difference is that the `ScriptOutput` does not have a timelock field, but that's only 8 bytes so I believe it can be ignored.

lrubiorod (approved):

PR #2210, opened 2022-06-02
Script change output

lrubiorod:


node/src/actors/messages.rs
```diff
@@ -220,12 +220,12 @@ pub struct BuildScriptTransaction {
     pub utxo_strategy: UtxoSelectionStrategy,
     /// Extra script inputs
     pub script_inputs: Vec<ScriptInput>,
-    /// List of bytes that will allow spending each script input
-    pub script_witnesses: Vec<Vec<u8>>,
+    /// Flag to send transaction
+    pub send_flag: bool,
```
tmpolaczyk, line 224: Can we add a new field `change_address: Option<ScriptHash>`?

lrubiorod: should be added in a new pr?

PR #2217, opened 2022-06-07
feat(stack): implement control flow and checktimelock

tmpolaczyk:


stack/src/lib.rs
```diff
@@ -137,11 +143,158 @@ fn check_multi_sig(bytes_pkhs: Vec<MyValue>, bytes_keyed_signatures: Vec<MyValue
...
+                // false value to mark the script execution as failed, otherwise it may be marked as
+                // success
+                stack.push(top);
+                return MyControlFlow::Break;
+            }
+        }
+        MyOperator::If => {
+            let top = stack.pop();
+            if let MyValue::Boolean(bool) = top {
+                if_stack.push_back(bool);
```
aesedepece, line 191: Lol you shadowed the reserved `bool` word here.

aesedepece (changes requested):

aesedepece:
This is looking great. Let's address the TODOs and go for it.

node/src/actors/chain_manager/handlers.rs
```diff
@@ -1310,7 +1310,7 @@ impl Handler<BuildScriptTransaction> for ChainManager {
                     .into_actor(self)
```
lrubiorod, line 1310: If you are not going to use the signature, you can remove the call:

```rust
            Ok(vtt) => {
                let multi_sig_witness = witnet_stack::encode(vec![]).unwrap();
                let num_inputs = vtt.inputs.len();
                let transaction = Transaction::ValueTransfer(VTTransaction {
                    body: vtt,
                    witness: vec![multi_sig_witness; num_inputs],
                });

                Box::pin(actix::fut::result(Ok(transaction)))
            }
```

stack/src/lib.rs
```diff
@@ -872,4 +872,122 @@ mod tests {
...
+            Item::Value(MyValue::Bytes(hash_secret.as_ref().to_vec())),
+            Item::Operator(MyOperator::Equal),
+            Item::Operator(MyOperator::Verify),
+            Item::Value(MyValue::Bytes(pk_2.pkh().bytes().to_vec())),
+            Item::Operator(MyOperator::CheckSig),
+            Item::Operator(MyOperator::Verify),
+            Item::Operator(MyOperator::EndIf),
+        ];
+        assert!(!execute_script(s, &ScriptContext { block_timestamp: 0 }));
+    }
```
aesedepece, line 992: Should we check that 2 cannot spend after timelock even if the right secret is provided? :thinking:

aesedepece: I think that, ideally, they shouldn't. It is a sort of assumption for an atomic swap that only 1 party can spend at a time.

aesedepece (changes requested):

tmpolaczyk:
Now the tests are failing because of a `stack.pop()` on an empty stack, this is fixed by https://github.com/aesedepece/scriptful/pull/6

tmpolaczyk:
Closing, will continue in #2240.

PR #2218, opened 2022-06-08
feat(nodestats): add counters to monitor failed data requests

drcpu-github:
This is a first push towards adding some extra counters that can be used to monitor why your node fails to solve data requests. Currently this info is printed to the standard output, but that's not too useful for the average node operator. This was discussed [here](https://discord.com/channels/492453503390842880/492453503390842882/979090799264432198) at the moment of the Witnet node monitor demo on Discord.

This PR will fail the tests because it modifies NodeStats and thus the ChainState. However, I don't quite know how to create migrations for the ChainState, but I still wanted to get this out for comments. Any hints?

### DIFF
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -3195,8 +3195,12 @@ pub struct NodeStats {
     pub block_proposed_count: u32,
     /// Number of blocks included in the block chain
     pub block_mined_count: u32,
+    /// Number of times we could not solve a data request because the collateral requirement was too high
+    pub dr_insufficient_collateral_count: u32,
     /// Number of times we were eligible to participate in a Data Request
     pub dr_eligibility_count: u32,
+    /// Number of times we were eligible to participate in a Data Request but all our collateral was locked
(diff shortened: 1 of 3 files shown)


tmpolaczyk:
Yes, migrations are hard because we are using bincode. The ChainState struct has support for adding new fields, however these fields must always be at the end of the list of fields, old fields cannot be removed or modified, and it requires a bit of code to set it up properly. See PR #2159 for an example with `unspent_outputs_pool_old_migration_db`.

Since the NodeStats struct is pretty small and it's likely to change (with more fields and useful stats), I think it may be worth it to change its serialization format to something that allows adding or removing fields easily. I made a branch with a proof of concept: https://github.com/tmpolaczyk/witnet-rust/tree/node-stats-json which uses JSON to store the NodeStats field, so it's JSON inside bincode. If you rebase this pull request it should just work, but not sure yet if that will be the best approach. Probably the final version will be using something else instead of JSON, or storing the NodeStats in a separate key in the database (to avoid the bincode step).

As per the content of this PR, I think it's a great addition, the more useful stats the better.

PR #2219, opened 2022-06-08
feat(jsonrpc): add block weight as a value to getBlock

drcpu-github:
As requested in drcpu-github/witnet-explorer-frontend#2, it would be a nice feature to show block weight in the block overview of the explorer. Adding this to the witnet-rust node allows for having one single source of truth.

### DIFF
--- a/data_structures/src/chain.rs
+++ b/data_structures/src/chain.rs
@@ -489,6 +489,26 @@ impl Block {
             txns,
         )
     }
+
+    pub fn dr_weight(&self) -> u32 {
+        let mut dr_weight = 0;
+        for dr_txn in self.txns.data_request_txns.iter() {
+            dr_weight += dr_txn.weight();
+        }
+        dr_weight
+    }
+
+    pub fn vt_weight(&self) -> u32 {
+        let mut vt_weight = 0;
+        for vt_txn in self.txns.value_transfer_txns.iter() {
+            vt_weight += vt_txn.weight();
+        }
+        vt_weight
+    }
+
+    pub fn weight(&self) -> u32 {
+        self.dr_weight() + self.vt_weight()
+    }
 }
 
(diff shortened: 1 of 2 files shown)


tmpolaczyk:
Good idea! Maybe it would be better to keep the two kind of weights separate? So, instead of a `weight` field having `dr_weight` and `vt_weight`. This way we could detect if blocks are full because of too many data requests, or if they are full because of too many value transfer transactions.

lrubiorod:
> Good idea! Maybe it would be better to keep the two kind of weights separate? So, instead of a `weight` field having `dr_weight` and `vt_weight`. This way we could detect if blocks are full because of too many data requests, or if they are full because of too many value transfer transactions.

Agree

drcpu-github:
I can certainly do that. Both weights are checked independently though, no? A block can still accept VT's even when it's filled with DR's.

drcpu-github:
> Yes, that's correct

Okay, I was a bit confused as it seemed your comment suggested otherwise.

I pushed the changes. I added both `dr_weight`, `vt_weight` and `block_weight` + the necessary functions. Obviously the latter is technically redundant, but it does not hurt to include (I think).

tmpolaczyk (approved):
Works in a local testnet and it only adds new fields, so it shouldn't break any clients.

drcpu-github:
> Well the way it is implemented it's calculating the weights twice, but let's hope the compiler optimizes that away.

Agreed. My reasoning was that the definition of `block_weight` may change in the future independently of `dr_weight` or `vt_weight` (for example by adding a new transaction type) and because of the way it is currently implemented one does not have to worry about the `json_rpc_methods.rs` implementation. Though I'll admit that is a pretty insignificant issue.

lrubiorod (approved):

PR #2220, opened 2022-06-09
fix(wallet): fix broken wallet tests

lrubiorod:


tmpolaczyk:
These tests are used to detect breaking changes in the wallet, fixing the tests does not fix the breaking changes: new versions of the wallet will be incompatible with old versions of the node. So I think fixing these tests does not provide any value over just disabling them.

lrubiorod:
> These tests are used to detect breaking changes in the wallet, fixing the tests does not fix the breaking changes: new versions of the wallet will be incompatible with old versions of the node. So I think fixing these tests does not provide any value over just disabling them.

If you disable those tests, most probably forgot to enable them again

tmpolaczyk:
So let's remember to enable them during the scripting migration phase, to ensure that the wallet keeps working as always.

PR #2229, opened 2022-06-28
WIP: minimum data request reward to collateral percentage

drcpu-github:
Following [this discussion](https://github.com/witnet/witnet-rust/discussions/2168), I started with a PoC implementation for the minimum reward to collateral mechanism. In summary, there are two main changes to the code:

1) Introduction of a `minimum_reward_collateral_percentage` configuration variable which can freely be set by miners (as long as the value is higher than the related consensus constant). Essentially, this variable will dictate if a miner inserts a newly received data request into his local transaction pool.

2) Introduction of a `required_reward_collateral_percentage` consensus constant which is used for data request validation whenever a block is received or a new data request is created.

I currently am facing three issues and could use some insight into how to tackle them:

1) I doubt I can just add a new consensus constant given that it is used everywhere (e.g., in the communication between nodes). I did this anyway to make the PoC, but for the actual deployment I probably need to change this? How would adding a consensus constant need to be handled? Could we just make this a non-consensus, but fixed, parameter somewhere? People can obviously still change it and recompile the node, but as long as the majority of node operators does not, it's probably safe to not be a consensus constant.

2) Provided we can add a new consensus constant, in the Default implementation of TransactionsPool, I would like to set the `required_reward_collateral_percentage` using the consensus parameter, but I don't quite know how to achieve that? Granted, there are other parts of the code where consensus constants are hard coded, but if possible, it seems sensible to minimize hard coding them.

3) This patch essentially breaks every test where a data request is involved. I started changing some of them, but in total there are probably a couple hundred of tests that require changing, so I first wanted to check what the best way forward is? The only thing I currently changed in this patch is making sure the test code correctly compiles.


tmpolaczyk:
Hi, so we haven't made any changes to the `ConsensusConstants` struct since before mainnet, so I'm not sure how would that work. Most probably we can use TAPI to change it, and then if the magic number changes and nodes are not able to connect to the mainnet, we should be able to find some workaround.

> Could we just make this a non-consensus, but fixed, parameter somewhere? 

We could, but the main benefit of having it in `ConsensusConstants` is that we can easily launch a testnet with a different parameter just by changing one line in the witnet.toml, so we should try to keep any configurable parameters there.

Regarding the TransactionsPool::default, you cannot use the consensus constants to set the default value because consensus constants are loaded at runtime from the witnet.toml file. A simple workaround is to do the same as the `set_minimum_vtt_fee` or `set_total_weight_limit` functions, which set the value after creating the TransactionsPool. There shouldn't be any problem with setting the default value to 0%, because transactions are only inserted in the TransactionsPool after being validated, so any transactions that have less than 1% reward percentage will not be inserted.

Regarding tests I think most of them will fail because the collateral and value fields are set to dummy values, so the fix is to increase the value. Maybe you can set `REQUIRED_REWARD_COLLATERAL_PERCENTAGE = 0` to ensure that old tests keep working, and then create new tests where that value is set to 1%.

The `minimum_reward_collateral_percentage` config variable looks useful and it is not a breaking change, so if you want you can open another pull request with only that, and it will be more likely to be merged.

drcpu-github:
> Hi, so we haven't made any changes to the `ConsensusConstants` struct since before mainnet, so I'm not sure how would that work. Most probably we can use TAPI to change it, and then if the magic number changes and nodes are not able to connect to the mainnet, we should be able to find some workaround.
> 
Yes, I was also thinking that TAPI may work for this. The magic number that is broadcasted and checked could probably be guarded in some `is_wip_active` check.

> Regarding the TransactionsPool::default, you cannot use the consensus constants to set the default value because consensus constants are loaded at runtime from the witnet.toml file. A simple workaround is to do the same as the `set_minimum_vtt_fee` or `set_total_weight_limit` functions, which set the value after creating the TransactionsPool. There shouldn't be any problem with setting the default value to 0%, because transactions are only inserted in the TransactionsPool after being validated, so any transactions that have less than 1% reward percentage will not be inserted.
>
I assumed that was the solution and it's basically how I implemented it, albeit at 1%. I was just wondering if it could be set by `ConsensusConstants` because that felt cleaner.

> Regarding tests I think most of them will fail because the collateral and value fields are set to dummy values, so the fix is to increase the value. Maybe you can set `REQUIRED_REWARD_COLLATERAL_PERCENTAGE = 0` to ensure that old tests keep working, and then create new tests where that value is set to 1%.
>
Yeah, basically all tests initially fail because transactions don't get inserted in the `TransactionsPool` or because the data request is not valid or ... I could of course delete the line in `set_minimum_reward_collateral_percentage` that checks versus the `ConsensusConstant` and use the old values by calling that function with a 0 value. Not sure yet. But the way forward is likely indeed to somehow make sure the old tests are still valid rather than modifying all of them.

> The `minimum_reward_collateral_percentage` config variable looks useful and it is not a breaking change, so if you want you can open another pull request with only that, and it will be more likely to be merged.
>
It is certainly possible to do this in a separate patch. However, for your average miner it does not make sense to have a non-zero `minimum_reward_collateral_percentage` value as that will result in dropped data requests and losing reputation if others do not also use a non-zero setting. Furthermore, if the majority of miners set this to a non-zero value, the existence of this setting could be abused. The complete patch and the introduction of a `required_reward_collateral_percentage` offers a way to establish a baseline reward while also minimizing abuse as per the initial discussion. I don't think it makes sense to split it up.

drcpu-github:
Implemented WIP0022-guards around the relevant code.

Made sure that all tests pass when the ratio is not activated by setting the accepted ratio to `u64::MAX`.

All tests can be run with WIP0022 activated by uncommenting the active_wips insertions at lines 51 and 65 in `mainnet_validations.rs`.

I am not quite sure how I can modify the `get_magic` function in `chain.rs` to ignore the `required_reward_collateral_ratio` field until after WIP0022 is activated. The best option I can think of right now is to hard-code the current magic number in an if-else construct which seems rather dirty.

wallet/src/actors/app/handlers/create_data_req.rs
```diff
@@ -31,6 +31,7 @@ pub struct CreateDataReqRequest {
     )]
     fee: u64,
     fee_type: Option<FeeType>,
+    required_reward_collateral_ratio: u64,
```
tmpolaczyk, line 34: Adding this field here would require the client to pass the ratio. Since this should be part of `ConsensusConstants`,  the wallet should already know the value from somewhere else.

```suggestion
```

validations/src/validations.rs
```diff
@@ -499,6 +508,16 @@ pub fn validate_data_request_output(request: &DataRequestOutput) -> Result<(), T
         });
     }
 
+    if active_wips.wip0022() {
+        let reward_collateral_ratio = request.collateral / request.witness_reward;
```
tmpolaczyk, line 512: Not sure if this handles the case where `collateral` is set to 0 correctly, it should be interpreted as `collateral_minimum`.

drcpu-github: It indeed does not, I assumed you could never receive one with that setting here since other parts of the code would already have set it to `collateral_minimum`. Not a big problem to add it though, just to play if safe.

wallet/src/actors/app/handlers/create_data_req.rs
drcpu-github: I could not find a reasonable way of getting the consensus constants there, but it's likely I am missing something.

drcpu-github: Yeah, I found those too and realized I could not really use them. As you suggested, I just added it as a TODO (note that I also had to add the `collateral_minimum` constant).

validations/src/validations.rs
drcpu-github: Done.

validations/src/validations.rs
```diff
@@ -509,7 +511,16 @@ pub fn validate_data_request_output(
     }
 
     if active_wips.wip0022() {
-        let reward_collateral_ratio = request.collateral / request.witness_reward;
+        let collateral = if request.collateral == 0 {
+            collateral_minimum
+        } else {
+            request.collateral
+        };
```
aesedepece, line 518: This is a bit strange, because there's a gap from 0 to collateral_minimum, so I assume you actually wanted to cap it like this:

```suggestion
        let collateral = std::cmp::max(request.collateral, collateral_minimum);
```

validations/src/validations.rs
```diff
@@ -509,7 +511,16 @@ pub fn validate_data_request_output(
...
+        let reward_collateral_ratio = if request.witness_reward > 0 {
+            collateral / request.witness_reward
+        } else {
+            u64::MAX
+        };
```
aesedepece, line 523: Code repetition. I'd try to have a pure function with a `Fn(u64, u64) -> u64` contract that derives this, cover it with some tests, and use it here and anywhere else.

aesedepece (changes requested):

data_structures/src/chain.rs
```diff
@@ -2716,12 +2726,19 @@ impl TransactionsPool {
                 }
             }
             Transaction::DataRequest(dr_tx) => {
+                let collateral = if dr_tx.body.dr_output.collateral == 0 {
+                    self.collateral_minimum
+                } else {
+                    dr_tx.body.dr_output.collateral
+                };
```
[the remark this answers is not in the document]
drcpu-github: Makes sense here because the transaction has been validated already.

validations/src/validations.rs
drcpu-github: This is actually on purpose. Your suggestion changes the validation behavior.

If a data request with a 100 mWIT collateral requirement and 1 mWIT reward is sent into the network it is valid from the perspective of the `TransactionError::RewardTooLow` error (assuming a 100 to 1 ratio). Changing the collateral to `collateral_minimum` would result in the error being (incorrectly) thrown.

What I actually want to see is that the `TransactionError::InvalidCollateral` error is being thrown in the code following the data request output validation given that this is more informative to the end-user building the data request. AFAIK errors will act as a short-circuit and thus adopting your suggestion would prevent the 'correct' error from being thrown.

validations/src/validations.rs
drcpu-github: Feels like a bit of overkill for an if-statement which occurs only twice, but sure.

validations/src/validations.rs
aesedepece: > If a data request with a 100 mWIT collateral requirement and 1 mWIT reward is sent into the network it is valid from the perspective of the `TransactionError::RewardTooLow` error (assuming a 100 to 1 ratio). Changing the collateral to `collateral_minimum` would result in the error being (incorrectly) thrown.

Indeed I'm totally expecting a `TransactionError::InvalidCollateral` in this case because 1 mWit < `collateral_minimum`.

validations/src/validations.rs
aesedepece: My personal take is that the more consensus critical logic we can have in "library code" (i.e. not tied to specific data structures or error types), the better for unit testing and for reusing the same implementation across different components.

drcpu-github: Done. Moved it into a function at `data_request.rs`.

tmpolaczyk:
I rebased this PR and fixed the missing parts here: #2337 

Let me know if I missed anything, closing.

PR #2230, opened 2022-06-28
ci: ✨  CD for witnet-rust

aeweda:
Visualization:
image

Entire workflow is triggered by pushing a new tag

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -1,0 +1,279 @@
+name: Release Tag
+
+on:
+  push:
+    tags:
+      - '*'
+
+jobs:
+#?####################################################################################################?#
+#?                                                                                                    ?#
+#?                                        Build Helper Images                                         ?#
+#?                                                                                                    ?#
+#?####################################################################################################?#
+  aarch64:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3  
+      - uses: ./.github/actions/build-helper
+        name: Building Docker Image
+        with:
+          imagename: aarch64-unknown-linux-gnu
+
+  armv7:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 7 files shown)


.github/workflows/master.yml
```diff
@@ -0,0 +1,279 @@
...
+
+      # Publish new github release
+      - uses: ncipollo/release-action@v1
+        with:
+          name: "Witnet-rust ${{github.ref_name}}"
+          artifacts: "./release/*"
+          bodyFile: "RELEASE.md"
+          artifactErrorsFailBuild: true
+    
+  Publish:
```
aesedepece, line 258: Is there a way we can publish it as a pre-release?

Then we only need to fill in the release notes and manually turn it into a proper release once all that info is in place.

Otherwise someone could see the notification for the new release, go to the release page and find that there's no info.

aesedepece (changes requested):

.github/workflows/master.yml
aeweda: we can, problem is, if it's a pre-release and i fetch "latest binaries" in the following stage for building the docker images will they be outdated or will it fetch from a pre-release

aesedepece: They will be outdated. Ok, let's leave it as is!

aesedepece (changes requested):
This is looking great!

Now I need you to sign all your commits, not just the last one.

You can either squash all the commits into one, or sign them all at once with `git rebase --exec 'git commit --amend --no-edit -n -S' -i upstream/master`

Then, regarding commit messages, you need to follow [this rule](https://github.com/witnet/witnet-rust/blob/master/CONTRIBUTING.md#sending-a-pull-request).

aesedepece:
@aeweda you have the green light to merging

tmpolaczyk (approved):

PR #2235, opened 2022-07-27
Rad replace separators

### LINKED ISSUE #1133: Error when parsing floats with commas

### LINKED ISSUE #2225: Plan for a eventual fix to #1133
See #1133, #1610, and #1784. 

PR #1784 implemented a fix, but it got postponed because it was unsafe to release before TAPI.

Now that we have TAPI, we should be fine adopting something like that.

We need to:

1. Agree on a fix
2. Discuss a plan for deploying the fix
3. Write a WIP
4. Make a PR

DoD for this issue covers 1 and 2.

aesedepece:
Follow-up to #1784
Fix #1133
Fix #2225

### DIFF
--- a/rad/src/operators/array.rs
+++ b/rad/src/operators/array.rs
@@ -370,17 +370,17 @@ pub fn transpose(input: &RadonArray) -> Result<RadonArray, RadError> {
     Ok(RadonArray::from(v))
 }
 
-/// This module was introduced for encapsulating the interim legacy logic before WIP-0022 is
+/// This module was introduced for encapsulating the interim legacy logic before WIP-0024 is
 /// introduced, for the sake of maintainability.
 ///
 /// Because RADON scripts are never evaluated for old blocks (e.g. during synchronization), this
-/// module can theoretically be removed altogether once WIP-0022 is activated.
+/// module can theoretically be removed altogether once WIP-0024 is activated.
 pub mod legacy {
     use super::*;
     use crate::types::float::RadonFloat;
 
-    /// Legacy (pre-WIP0022) version of `get::<RadonFloat, _>`.
-    pub fn get_float_before_wip0022(
+    /// Legacy (pre-WIP0024) version of `get::<RadonFloat, _>`.
+    pub fn get_float_before_wip0024(
         input: &RadonArray,
         args: &[Value],
     ) -> Result<RadonFloat, RadError> {
@@ -390,8 +390,8 @@ pub mod legacy {
         })
     }
 
-    /// Legacy (pre-WIP0022) version of `get::<RadonInteger, _>`.
-    pub fn get_integer_before_wip0022(
+    /// Legacy (pre-WIP0024) version of `get::<RadonInteger, _>`.
+    pub fn get_integer_before_wip0024(
         input: &RadonArray,
         args: &[Value],
     ) -> Result<RadonInteger, RadError> {
--- a/rad/src/operators/array.rs
+++ b/rad/src/operators/array.rs
@@ -88,13 +88,13 @@ pub fn get_bytes(input: &RadonArray, args: &[Value]) -> Result<RadonBytes, RadEr
 /// Try to get a `RadonFloat` from a position in the input `RadonArray`, as specified by the first
 /// argument, which is used as the positional index.
 pub fn get_float(input: &RadonArray, args: &[Value]) -> Result<RadonFloat, RadError> {
-    get_number(input, args)?.try_into()
+    get_numeric_string(input, args)?.try_into()
 }
 
 /// Try to get a `RadonInteger` from a position in the input `RadonArray`, as specified by the first
 /// argument, which is used as the positional index.
 pub fn get_integer(input: &RadonArray, args: &[Value]) -> Result<RadonInteger, RadError> {
-    get_number(input, args)?.try_into()
+    get_numeric_string(input, args)?.try_into()
 }
 
 pub fn get_map(input: &RadonArray, args: &[Value]) -> Result<RadonMap, RadError> {
@@ -107,7 +107,7 @@ pub fn get_map(input: &RadonArray, args: &[Value]) -> Result<RadonMap, RadError>
 ///
 /// This simply assumes that the element in that position is a number (i.e., `RadonFloat` or
 /// `RadonInteger`). If it is not, it will fail with a `RadError` because of `replace_separators`.
-fn get_number(input: &RadonArray, args: &[Value]) -> Result<RadonTypes, RadError> {
+fn get_numeric_string(input: &RadonArray, args: &[Value]) -> Result<RadonTypes, RadError> {
     let item = get(input, args)?;
 
     if args.len() == 3 {
--- a/rad/src/operators/array.rs
+++ b/rad/src/operators/array.rs
@@ -85,10 +85,14 @@ pub fn get_bytes(input: &RadonArray, args: &[Value]) -> Result<RadonBytes, RadEr
     item.try_into()
 }
 
+/// Try to get a `RadonFloat` from a position in the input `RadonArray`, as specified by the first
+/// argument, which is used as the positional index.
 pub fn get_float(input: &RadonArray, args: &[Value]) -> Result<RadonFloat, RadError> {
     get_number(input, args)?.try_into()
 }
 
+/// Try to get a `RadonInteger` from a position in the input `RadonArray`, as specified by the first
+/// argument, which is used as the positional index.
 pub fn get_integer(input: &RadonArray, args: &[Value]) -> Result<RadonInteger, RadError> {
     get_number(input, args)?.try_into()
 }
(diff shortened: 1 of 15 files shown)


rad/src/operators/array.rs
```diff
@@ -84,18 +83,37 @@ pub fn get_bytes(input: &RadonArray, args: &[Value]) -> Result<RadonBytes, RadEr
...
+
+/// Try to get a `RadonTypes` from a position in the input `RadonArray`, as specified by the first
+/// argument, which is used as the positional index.
+///
+/// This simply assumes that the element in that position is a number (i.e., `RadonFloat` or
+/// `RadonInteger`). If it is not, it will fail with a `RadError` because of `replace_separators`.
+fn get_numeric_string(input: &RadonArray, args: &[Value]) -> Result<RadonTypes, RadError> {
+    let item = get(input, &args[..1])?;
+
+    let (thousands_separator, decimal_separator) = string::read_separators_from_args(args, 1);
```
tmpolaczyk, line 112: The skip argument seems redundant because you can pass a sub-slice, so this should be equivalent:

```suggestion
    let (thousands_separator, decimal_separator) = string::read_separators_from_args(&args[1..], 0);
```

rad/src/operators/string.rs
```diff
@@ -264,12 +280,62 @@ fn json_to_cbor(value: &json::JsonValue) -> Value {
...
+
+    Ok(RadonTypes::from(RadonString::from(_replace_separators(
+        rad_str_value.value(),
+        thousands_separator,
+        decimal_separator,
+    ))))
+}
+
+/// Replace thousands and decimals separators in a `String`.
+pub fn _replace_separators(
```
tmpolaczyk, line 302: A public function starting with `_` looks weird, I would call it `replace_separators_string` or `replace_separators_inner` if it's private.

rad/src/operators/string.rs
```diff
@@ -264,12 +280,62 @@ fn json_to_cbor(value: &json::JsonValue) -> Value {
...
+    let decimal = from_value::<String>(decimal_separator).unwrap_or_else(|_| ".".to_string());
+
+    value.replace(&thousands, "").replace(&decimal, ".")
+}
+
+/// Read separators from RAD call arguments, and fall back to the default ones if not provided.
+pub fn read_separators_from_args(
+    args: &[serde_cbor::Value],
+    skip: usize,
+) -> (serde_cbor::Value, serde_cbor::Value) {
```
tmpolaczyk, line 317: This function could return a pair of strings, because later in `_replace_separators` the `Value`s are converted to `Strings` anyway, and in case of error the result is set to the same default value as here.

```suggestion
) -> (String, String) {
```

Or a `Result<(String, String)>` if we wanted to handle errors, but I guess returning the default separators in case of error is fine.

rad/src/operators/string.rs
```diff
@@ -148,25 +149,40 @@ pub fn radon_trim(input: &RadonString) -> String {
...
-    f64::from_str(&str_value)
+    bool::from_str(&str_value)
+        .map(RadonBoolean::from)
+        .map_err(Into::into)
+}
+
+/// Converts a `RadonString` into a `RadonFloat`, provided that the input string actually represents
+/// a valid floating point number.
+pub fn as_float(input: &RadonString, args: &Option<Vec<Value>>) -> Result<RadonFloat, RadError> {
+    f64::from_str(&as_numeric_string(input, &args.clone().unwrap_or_default()))
```
tmpolaczyk, line 162: You can use `Option::as_deref` to convert `&Option<Vec<T>>` into `&Option<&[T]>`:

```suggestion
    f64::from_str(&as_numeric_string(input, args.as_deref().unwrap_or_default()))
```

Same below in as_integer.

aesedepece:
@tmpolaczyk thanks a lot for your suggestions. Those helped make the code much cleaner.

aesedepece:
I'm fixing something really stupid that breaks the end-to-end tests under `/examples`. Happy that we have those.

tmpolaczyk:
> I'm fixing something really stupid that breaks the end-to-end tests under /examples. Happy that we have those.

Oh, I didn't notice the test failure. Funnily enough, I was considering removing those examples because they didn't seem useful, and they are missing many data requests from the data feeds app. Anyway, if you find the cause please also add a unit test somewhere!

aesedepece:
The issue was with trying to read JSON native numbers as strings. As integers and floats are now internally handled first as strings, pre-processed, and then converted to their right type, I was having problems with one of the examples that has a data source with a raw float (not encoded into a string).

With the latest changes, those wild floats can be read either as a float or as a string.

Also took the chance to simplify how RadonType conversions work, and to improve how the decoding errors are managed.

rad/src/operators/map.rs
```diff
@@ -26,33 +25,51 @@ pub fn get(input: &RadonMap, args: &[Value]) -> Result<RadonTypes, RadError> {
...
+/// `RadonInteger`). If it is not, it will fail with a `RadError` because of `replace_separators`.
+fn get_numeric_string(input: &RadonMap, args: &[Value]) -> Result<RadonString, RadError> {
+    let item = get::<RadonString, _>(input, &args[..1])?.value();
+    let (decimal_separator, thousands_separator) = string::read_separators_from_args(&args[1..]);
+
+    Ok(RadonString::from(string::replace_separators(
+        item,
+        decimal_separator,
+        thousands_separator,
+    )))
```
tmpolaczyk, line 72: These names are swapped, the tests pass because they are swapped twice.

```suggestion
    let (thousands_separator, decimal_separator) = string::read_separators_from_args(&args[1..]);

    Ok(RadonString::from(string::replace_separators(
        item,
        thousands_separator,
        decimal_separator,
    )))
```

aesedepece: Lol. I blame it on CLion's automatic name refactors.

aesedepece:
@tmpolaczyk do you think it's worth it if I guard these changes behind a TAPI conditional in this same PR? Or shall I merge it into a new branch and make another PR there for the guards?

tmpolaczyk:
The changes in `impl TryFrom<RadonTypes> for RadonString` are problematic because they cannot be made conditional using TAPI (there is no way to pass an extra argument to try_from), and it's hard to find all the usages of that function to ensure that it doesn't cause any accidental breaking changes. I haven't found any potential breaking changes in the tally creation, which are the dangerous ones, but I can't guarantee that they don't exist yet.

So I think it's better to solve the TAPI issues before merging.
The changes in operators can be easily guarded because `operate_in_context` has access to the list of active wips, so it's just a matter of moving the operator calls from `operate` to `operate_in_context`. And it doesn't need to have a custom wip name with an activation date, something like `if wip_xxxx` should be fine. Also we need to create a `witnet-1.6` branch.

aesedepece:
> The changes in `impl TryFrom<RadonTypes> for RadonString` are problematic because they cannot be made conditional using TAPI (there is no way to pass an extra argument to try_from)

This shouldn't be a problem anymore as I have reverted most of those changes and covered them with exhaustive tests that verify that the behavior before the activation of the WIP remains unchanged.

> The changes in operators can be easily guarded because operate_in_context has access to the list of active wips, so it's just a matter of moving the operator calls from operate to operate_in_context.

Great suggestion. That is exactly what I have done.

tmpolaczyk:
Looking great! I found some unexpected changes in the StringMatch operator: basically if the default value is a string, the result of the match is now converted to a string. I was able to find a quick workaround so that's fixed in [this branch](https://github.com/tmpolaczyk/witnet-rust/tree/rad-replace-separators), with tests. I disabled the new behavior, but we could enable it with TAPI if we find it more useful, and that would allow us to remove the workaround after the WIP is enabled.

rad/src/types/array.rs
```diff
@@ -118,77 +121,69 @@ impl fmt::Display for RadonArray {
...
+    fn operate_in_context(
+        &self,
+        call: &RadonCall,
+        context: &mut ReportContext<RadonTypes>,
+    ) -> Result<RadonTypes, RadError> {
+        let wip0022 = context
+            .active_wips
+            .as_ref()
+            .map(ActiveWips::wip0022)
+            .unwrap_or_else(|| true);
```
tmpolaczyk, line 136:
```suggestion
            .unwrap_or(true);
```

Clippy should have detected this

aesedepece: Fixed :wink:

aesedepece:
I have merged in the changes from @tmpolaczyk, and updated the TAPI guards so they have the right WIP number ([0024](https://github.com/witnet/WIPs/blob/master/wip-0024.md)). This should be ready to review and merge.

tmpolaczyk (approved):
We can merge this. Regarding TAPI, it is missing the `oppose_wip0024` config flag and the voting bit, but that's already implemented in #2337 so I can add it in that pull request and make both WIPs use the same voting bit.

drcpu-github:
> make both WIPs use the same voting bit

Please no 😄

PR #2236, opened 2022-07-28
feat: allow keeping an index of utxos in memory 

### LINKED ISSUE #2173: Fix performance of getBalance and getUtxoInfo methods
The getBalance method is implemented by iterating over all the existing Unspent Transaction Outputs (UTXOs). This was good enough when the UTXO set was stored in memory, but with the recent changes from #2159 the UTXO set is now not in memory, so iterating the UTXO set is slow. The same happens with the getUtxoInfo method, which needs to iterate the UTXO set to find UTXOs that belong to the provided address.

This is a table with the approximate time it takes to execute each method, in milliseconds:

| Method \ Witnet version          | 1.5.0 | 1.5.1 |
|----------------------------------|-------|-------|
| getBalance                       | 4300  | 4100  |
| getBalance (simple)              | 80    | 3950  |
| getBalance (simple, own address) | 80    | 15    |
| getUtxoInfo                      | 100   | 4000  |
| getUtxoInfo (own address)        | 15    | 15    |

The reason why querying info about the node's "own address", is because there is a data structure called `OwnUnspentOutputsPool` which serves as a cache of the UTXOs that belong to this node. This data structure is needed to implement the commit collateral creation, as well as the methods to create value transfer and data request transactions.

So one way of fixing the performance of this methods would be to implement something similar to the `OwnUnspentOutputsPool`, and have a map of "address" to "list of UTXOs". This would make all the methods take around 15ms but at the cost of consuming a few GB of memory, and making every block consolidation slightly slower. So if implemented that way it should be an optional feature.

tmpolaczyk:
Fix #2173

This PR adds a new option in witnet.toml `storage.utxos_in_memory` that defaults to `false`. When enabled, the node will keep a map of address to list of UTXOs in memory, which allows to implement fast `getBalance` and `getUtxoInfo` methods. With this option enabled a call to `getBalance` takes around 5ms, while the same call with this option disabled may take up to 5000ms. The expected memory usage in mainnet is around 500 MB.

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -283,6 +283,8 @@ pub struct Storage {
     #[partial_struct(skip)]
     #[partial_struct(serde(default))]
     pub master_key_import_path: Option<PathBuf>,
+    /// Keep a map of "address" to "list of UTXOs" in memory, to speed up getBalance and getUtxoInfo methods
+    pub utxos_in_memory: bool,
 }
 
 /// JsonRPC API configuration
@@ -703,6 +705,10 @@ impl Storage {
                 .to_owned()
                 .unwrap_or_else(|| defaults.storage_db_path()),
             master_key_import_path: config.master_key_import_path.clone(),
+            utxos_in_memory: config
+                .utxos_in_memory
+                .to_owned()
+                .unwrap_or_else(|| defaults.storage_utxos_in_memory()),
         }
     }
 
@@ -711,6 +717,7 @@ impl Storage {
             backend: self.backend.clone(),
             db_path: Some(self.db_path.clone()),
             master_key_import_path: self.master_key_import_path.clone(),
+            utxos_in_memory: Some(self.utxos_in_memory),
         }
     }
 }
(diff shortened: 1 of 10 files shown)


aesedepece (commented):
Seems to work just fine :+1: 
Would appreciate feedback from @drcpu-github, who I suspect that will be the main user for this feature for the time being!

drcpu-github:
> Seems to work just fine 👍 Would appreciate feedback from @drcpu-github, who I suspect that will be the main user for this feature for the time being!

I'll try and have a look by integrating it in the test version of my explorer.

drcpu-github:
@tmpolaczyk I have been testing this branch in the test version of my explorer and it works like a charm! I don't really notice any slowdown in either address lookups in the explorer or the building of the list of all accounts with a non-zero balance.

aesedepece (approved):

aesedepece:
Merged in as 2d497aad9e3c8fb27d431378ffa4917c351f5603

PR #2241, opened 2022-07-29
Rehash retrieval proxies / multiple transports / inconsistent source attack protection (aka "smart retrievals")

### LINKED ISSUE #1274: Assesment: Smart retrievals for preventing Inconsistence source attack

### LINKED ISSUE #1511: Discussion: How to avoid "malicious" requests? (Inconsistent source attack)

aesedepece:
This is a rehash of #1523 

Enables using proxies (HTTP, HTTPS, SOCKS4 and SOCKS5) when performing data retrieval. This allows retrieving data sources through different transports so as to ensure that the data sources are consistent and we are taking as small of a risk as
possible when committing to specially crafted data requests that may be potentially ill-intended.

Just add this to the `[connections]` section of `witnet.toml` to proxy your retrievals through 3 different open proxies:
```toml
retrieval_proxies = [
        "socks5://181.102.35.195:1080",
        "socks5://8.210.60.118:1080",
        "socks5://47.114.8.133:1080"
]
```

These addresses have been taken from [this public list of proxies][proxies].

Close #1274 
Close #1511 

[proxies]: https://spys.one/en/socks-proxy-list/


aesedepece:
This is still a draft because tests need yet to be fixed, and I want to improve some other parts.

rad/src/lib.rs
```diff
@@ -349,40 +357,118 @@ fn add_http_headers(
...
 
+/// Run retrieval using multiple transports, and only produce a positive result if the retrieved
+/// values pass the filter function from the tally stage.
+///
+/// The idea behind this is to avoid producing commitments for data requests with sources that act
+/// in an inconsistent way, i.e. they return very different values when queried through different
+/// HTTP transports at once.
+pub async fn run_paranoid_retrieval(
+    retrieve: &RADRetrieve,
+    tally: &RADTally,
```
tmpolaczyk, line 395: It looks strange that this uses the retrieve stage and the tally stage, but does not use the aggregation stage. Maybe it would be better to run all retrievals + aggregation for each proxy, and then apply the paranoid logic to the vector of aggregated results. Then this PR would not be detecting "inconsistent sources" but "inconsistent requests", but I guess that's fine?

node/src/actors/rad_manager/mod.rs
```diff
@@ -8,12 +8,45 @@
...
+
 mod actor;
 mod handlers;
 
 /// RadManager actor
 #[derive(Default)]
-pub struct RadManager;
+pub struct RadManager {
+    /// Addresses of proxies to be used as extra transports when performing data retrieval.
+    proxies: Vec<String>,
```
tmpolaczyk, line 21: Proxy addresses are only validated when executing a data request, so if the user inputs an invalid proxy they have no way to detect it until their node can resolve a data request.

So it would be nice to have either a way to validate the proxies on startup, or a JSON-RPC command to test the execution of a data request using proxies. If the old "try data request" command uses these proxies we can refer to it in the documentation as the expected way to ensure that the node is configured correctly.

tmpolaczyk (commented):
Some comments.

witnet.toml
```diff
@@ -56,6 +56,10 @@ bootstrap_peers_period_seconds = 1
 # Reject (tarpit) inbound connections coming from addresses that are alike (i.e. by default having the first 18 bits equal),
 # so as to prevent sybil peers from monopolizing our inbound capacity.
 reject_sybil_inbounds = true
+# Add addresses here to be used as proxies when performing data retrieval. This allows retrieving data sources through
+# different transports so as to ensure that the data sources are consistent and we are taking as small of a risk as
+# possible when committing to specially crafted data requests that may be potentially ill-intended.
+retrieval_proxies = []
```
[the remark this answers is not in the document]
aesedepece: Some examples can be found in the first comment of the PR.

rad/src/lib.rs
aesedepece: I had the same doubt when I was revisiting this.

aesedepece:
Thanks @tmpolaczyk for all your suggestions. I'll try to address them all today.

node/src/actors/rad_manager/mod.rs
```diff
@@ -8,12 +8,45 @@
...
+    /// and we are taking as small of a risk as possible when committing to specially crafted data
+    /// requests that may be potentially ill-intended.
+    pub fn add_proxy(&mut self, proxy_address: String) {
+        self.proxies.push(proxy_address)
+    }
+
+    /// Derive HTTP transports from registered proxies.
+    ///
+    /// This internally injects a `None` at the beginning, standing for the base "clearnet"
+    /// transport (no proxy).
```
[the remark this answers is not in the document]
aesedepece: I have now added such possibility through the `connections.unproxied_retrieval` setting, which defaults to `true` standing for enabled unproxied retrievals. `false` will disable them. In that case, proxy addresses need to be provided, otherwise it will  panic at start. This logic is covered by tests.

rad/src/lib.rs
```diff
@@ -349,40 +357,118 @@ fn add_http_headers(
...
+/// Evaluate whether the values obtained when retrieving a data source through multiple transports
+/// are consistent, i.e. they all pass the filters from the tally stage.
+fn evaluate_paranoid_retrieval(
+    data: Vec<Result<RadonReport<RadonTypes>>>,
+    tally: &RADTally,
+    settings: RadonScriptExecutionSettings,
+) -> Result<RadonTypes> {
+    let mut values = vec![];
+
+    // Turn the reports into values, short-circuiting if any retrieval inevitably failed.
```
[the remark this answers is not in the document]
aesedepece: This is now customizable through `connections.witnessing_paranoid_percentage`, where a value of `51` will require a "half plus one" majority of data sources to be in consensus before committing to it.

rad/src/lib.rs
```diff
@@ -349,40 +357,118 @@ fn add_http_headers(
...
+    // If there was only one retrieved value, there's no actual need to run the tally, as this means
+    // that only one transport was used and therefore the node is not in paranoid mode.
+    if values.len() > 1 {
+        let mut context = ReportContext::from_stage(Stage::Tally(TallyMetaData::default()));
+        let _tally = run_tally_with_context_report(values.clone(), tally, &mut context, settings);
+
+        // If at least one of the retrieved values is marked as an outlier, resolve to the
+        // `InconsistentSource` error.
+        if let Stage::Tally(TallyMetaData { liars, .. }) = context.stage {
+            if !liars.is_empty() {
```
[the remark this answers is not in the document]
aesedepece: Now I'm not using the liars vector but directly the consensus metric `c` that we get back from the tally evaluator, and compare that with the paranoid percentage as configured by the node operator.

witnet.toml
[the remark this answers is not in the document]
aesedepece: I have added more detail to this explanation in `witnet.toml`

rad/src/lib.rs
aesedepece: This was utterly wrong.

I was using the tally executor for a good reason (it accumulates much more metadata that we need for evaluating consistency), but I was wrong passing the tally instead of the aggregation.

Now I'm taking the aggregation and synthetically turning it into a tally.

In the future, if we think that accumulating the metadata for aggregations is useful for debugging or whatever, we can refactor this part and simply use the aggregator executor.

net/src/client/http/mod.rs
```diff
@@ -0,0 +1,346 @@
...
+    }
+}
+
+impl TryFrom<&mut surf::http::Request> for WitnetHttpRequest {
+    type Error = WitnetHttpError;
+
+    fn try_from(req: &mut surf::http::Request) -> Result<Self, Self::Error> {
+        let method = isahc::http::Method::from(WitnetHttpMethod::try_from(req.method())?);
+        let version = isahc::http::version::Version::from(WitnetHttpVersion::try_from(
+            req.version().unwrap_or(surf::http::Version::Http3_0),
```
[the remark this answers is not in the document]
aesedepece: I changed it to let `isahc` apply its own default logic (which seems to use HTTP/1.1)

rad/src/lib.rs
```diff
@@ -349,40 +357,118 @@ fn add_http_headers(
...
+        .clone())
+}
+
 /// Run aggregate stage of a data request, return a tuple of `Result<RadonReport>` and `ReportContext`
 pub fn run_aggregation_report(
     radon_types_vec: Vec<RadonTypes>,
     aggregate: &RADAggregate,
     settings: RadonScriptExecutionSettings,
-    active_wips: ActiveWips,
+    active_wips: &ActiveWips,
```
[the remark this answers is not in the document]
aesedepece: Gotcha. It is always a best practice to let the caller decide when to clone. Removing the `&` here will essentially displace the cloning responsibility to `<RadManager as Handler<ResolveRA>>::handle` (as the `Aggregate` structure is still needed further down the line, beyond the lifetime of `run_paranoid_retrieval`, namely at `run_aggregation_report`).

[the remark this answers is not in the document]
aesedepece: When I see stuff like this happen, I feel the urge to wrap every fricking property of a `struct` in an `Option`, then do `.take()` on the inner value to avoid any cloning :rofl:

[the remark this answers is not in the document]
aesedepece: At a second thought... maybe the compiler is smart enough to realize that the value will be dropped shortly after the borrow and will actually skip the cloning for us? Or am I expecting just too much from it? :thinking:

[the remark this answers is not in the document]
aesedepece: In any case, I have reverted these changes so `RADAggregate` are borrowed and the cloning happens in the outermost unit of business logic (`<RadManager as Handler<ResolveRA>>::handle`).

node/src/actors/rad_manager/mod.rs
aesedepece: > So it would be nice to have either a way to validate the proxies on startup

Let me give this a further thought on how to approach it properly.

> or a JSON-RPC command to test the execution of a data request using proxies. If the old "try data request" command uses these proxies we can refer to it in the documentation as the expected way to ensure that the node is configured correctly.

I think I prefer a more automated way :thinking: 

In any case, `try_data_request` is not currently using the proxies (as it implements its own retrieval logic that is not unified with that of `RadManager`). I guess it would be nice to have it use the proxies though, for it to be consistent with the actual witnessing behavior of the node and, as you pointed out, as a funny way to test the proxies functionality.

rad/src/lib.rs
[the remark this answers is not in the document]
tmpolaczyk: > maybe the compiler is smart enough to realize that the value will be dropped without further usage shortly after the move of the clone and will actually skip the cloning for us?

Probably not because `ActiveWips` has a `HashMap` field, in theory we should be able to change it into an `Arc<HashMap>` and make the clone cheap, but not now.

node/src/actors/chain_manager/mining.rs
```diff
@@ -548,7 +551,16 @@ impl ChainManager {
                         })
                         .map(move |res|
                             res.map(move |result| match result {
-                                    Ok(value) => Ok((vrf_proof, collateral, value)),
+                                    Ok(value) => {
+                                if let RadonTypes::RadonError(error) = &value.result {
+                                    if error.inner() == &RadError::InconsistentSource {
+                                        log::warn!("Refraining not to commit to data request {} because the sources are apparently inconsistent", dr_pointer);
+                                        return Err(())
```
tmpolaczyk, line 558: I believe in this case it would be better for the network safety to commit the `RadError::InconsistentSource` than to refrain from commiting.

In https://github.com/witnet/witnet-rust/discussions/2050 we discuss possible attacks that arise from data requests that can only be solved by a small fraction of all the nodes. I believe the same arguments apply to the inconsistent source attack, because nodes that enable this inconsistent source attack protection will not participate in that requests.

Imagine an attacker that creates a data request that returns a bad value for some nodes in order to penalize them, the classical inconsistent source attack. If some nodes refrain from participating in that request, then the attacker will control a higher percentage of the quorum, leading to the attacker gaining more reputation. This is true until a majority of nodes enable inconsistent source attack protection, then the request will be resolved with a `RadError::InsufficientCommits`.

On the other hand, if the eligible nodes commit an error, then all the nodes will participate in solving this request, so it will be equally hard for the attacker to create an eligibility proof for this request than for any other request. If the majority of nodes have enabled inconsistent source attack protection, it is possible that the data request is solved with a radon error as a result, and in that case nobody gains reputation. And in the case where only a minority have enabled inconsistent source attack protection, there will be a few error commits in the data request, so the attack should still work, with a slightly lower profit.

aesedepece: Hmmm... let me give that a thought. This is not a trivial point because what you are suggesting actually requires a protocol change to assign an identifier for `InconsistentSource` in `RadonErrors`.

aesedepece: Of course I had already created such identifier, but this was a non-consensus-breaking change only because we are never putting that error into a commitment.

tmpolaczyk: True, in that case we could commit a `RadError::Unknown`, and in a future pull request add the logic to enable a new RadonError using TAPI.

aesedepece: I have solved it like this:
```rust
.map_err(
    // The `InconsistentSource` error is mapped here for the sake of backwards
    // compatibility. Namely, to enable paranoid retrieval without having to
    // immediately introduce a breaking change that may jeopardize oracle
    // queries. The point of making the mapping here is to only affect actual
    // witnessing and committing, but not the `try_data_request` function, which
    // can rather use the `InconsistentSource` error for clarity.
    // TODO: pursue a WIP that introduces `InconsistentSource` as a proper
    //  RadonError at the protocol level.
    |err| match err {
        RadError::InconsistentSource => RadError::Unknown,
        other => other,
    },
)
```

tmpolaczyk:
I just found a very funny issue, this PR breaks RNG requests because the nodes commit a "ModeTie" error instead of a random value.

rad/src/lib.rs
```diff
@@ -369,20 +386,127 @@ pub async fn run_retrieval(retrieve: &RADRetrieve, active_wips: ActiveWips) -> R
...
+/// in an inconsistent way, i.e. they return very different values when queried through different
+/// HTTP transports at once.
+pub async fn run_paranoid_retrieval(
+    retrieve: &RADRetrieve,
+    aggregate: RADAggregate,
+    settings: RadonScriptExecutionSettings,
+    active_wips: ActiveWips,
+    witnessing: WitnessingConfig,
+) -> Result<RadonReport<RadonTypes>> {
+    let futures: Result<Vec<_>> = witnessing
```
tmpolaczyk, line 408: This fixes the RNG issue

```suggestion
) -> Result<RadonReport<RadonTypes>> {
    if let RADType::Rng = &retrieve.kind {
        // RNG retrievals always return random values by design, so skip paranoid check
        return run_retrieval_report(retrieve, settings, active_wips, None).await;
    }
    let futures: Result<Vec<_>> = witnessing
```

tmpolaczyk:
When trying to use a public socks5 proxy, I get this error when trying a request with https sources (didn't test http):

```
[2022-08-11T11:36:19Z WARN  isahc::handler] request completed with error: the server certificate could not be validated
```

witnet.toml:
```
proxies = [
        "socks5://72.217.216.239:4145",
]
```

I tried other proxies and I got either the same error or a "retrive timeout" error, so not sure if I had bad luck choosing the proxies or it doesn't work as expected for some reason.

node/src/actors/chain_manager/mining.rs
tmpolaczyk: So it looks like the node already refrains from commiting errors in some cases, such as if the `RadManager` panics:

https://github.com/witnet/witnet-rust/blob/f6bcb721d11f6252daff6d72bdea6a385fb7d981/node/src/actors/chain_manager/mining.rs#L557-L561

And for some reason this pull request also enters that branch in my tests, so no commits are performed when the result is `RadError::InconsistentSource`. So not sure if we should change the behavior and commit an error in all cases, or only in case of inconsistent source, or don't commit even in case of inconsistent source.

aesedepece:
> I tried other proxies and I got either the same error or a "retrive timeout" error, so not sure if I had bad luck choosing the proxies or it doesn't work as expected for some reason.

I tried a few proxy addresses as well and found out that most of those that you can find out there are garbage, regardless of being SOCKS or HTTP.

The only one that worked like a charm was the SOCKS5 interface on port 9150 from my own Tor Browser. This added like 1-2s of latency but still all the retrievals were successful (except one API that gave me 403 because they're behind a WAF or something).

node/src/actors/chain_manager/mining.rs
aesedepece: Interesting. As nodes are already refraining from committing in such cases... am I right to assume that we can change this part and start committing `RadonError::Unknown` in this "everything went wrong" case without breaking consensus?

In that case, we would start to see reveals with `RadonError::Unknown`, which currently are rare if not non-existing.

The thing is that the lack of commits caused by the code block you are pointing out is probably not even having an impact on oracle queries as of today, because those missing commits should be allowing other valid commits to take their place, under the assumption that there are always more eligible identities than necessary.

If that's the case, and we start to default to committing an error, that would most times lead to worse tallies than not doing so, because miners can't tell errors from values while at the commitment stage.

tmpolaczyk:
> I tried a few proxy addresses as well and found out that most of those that you can find out there are garbage, regardless of being SOCKS or HTTP.

That seems to be true. I finally got it working by running a local socks server.

node/src/actors/chain_manager/mining.rs
tmpolaczyk: > I right to assume that we can change this part and start committing RadonError::Unknown in this "everything went wrong" case without breaking consensus?

Should be safe, because nodes can always commit an error. So the result would be the same as a sudden increase in http errors or timeout errors.

> If that's the case, and we start to default to committing an error, that would most times lead to worse tallies than not doing so, because miners can't tell errors from values while at the commitment stage.

True, then I guess the right approach would be to keep the old behavior for now, and investigate about the consequences of changing it in the future. Would the ability to distinguish values from errors in the commit stage be useful? Or would it introduce even more attack vectors?

rad/src/lib.rs
aesedepece: Thanks! I eventuallywent for this:

```suggestion
) -> Result<RadonReport<RadonTypes>> {
    // We can skip paranoid checks for retrieval types that don't use networking (e.g. RNG)
    if !retrieve.kind.is_http() {
        return run_retrieval_report(retrieve, settings, active_wips, None).await;
    }

    let futures: Result<Vec<_>> = witnessing
```

aesedepece: Doing it that way will save us changing it in the future every time we add a new retrieval type-

tmpolaczyk: That's better, it's future-proof for new retrieval kinds.

node/src/actors/chain_manager/mining.rs
aesedepece: > Would the ability to distinguish values from errors in the commit stage be useful? Or would it introduce even more attack vectors?

Idk, leaking that info at the commitment stage makes it to easy in theory for lazy witnesses not to perform the retrieval and simply copy the error from other commitments once they have seen `rf * consensus_requirement` of those (because at that point they will know in advance that the only value they can commit without being slashed must be an error).

aesedepece: In the other hand, that would allow to skip the reveal phase :thinking:

tmpolaczyk: I was thinking about adding a flag to the commit like `is_error: bool`, this way lazy witnesses can only guess the error. And I believe if your error is different from the consensus error you miss the reward, or something like that, no sure because that logic is complex. That would allow miners to choose non-error commits, but it will also allow miners to choose error commits to sabotage a data request.

node/src/actors/rad_manager/mod.rs
aesedepece: @tmpolaczyk which of these do you think is the right approach to performing validation of proxies upon startup?

1. Validating the addresses
2. Checking that we can connect to a socket
3. Somehow handshaking with the proxy (different for each supported proxy protocol)
4. Running `try_data_request`

Specially in this last case, I think I would give the option of disabling the validation through configuration, so a downtime of the data sources doesn't prevent Witnet nodes from (re)starting.

tmpolaczyk: Validating the addresses should be done to detect typos. Checking that we can connect or handshaking may not always work, so I prefer a more end to end test.

When running `try_data_request` from `RadManager` it is really hard to guess which proxy failed, the logs never mention the proxy url, so we need to run `try_data_request` once per proxy, and add a log that clearly says if a proxy failed (and stop the node in case of failure?). Creating a test data request to google.com should work. I guess we can simply add a config variable like `witnessing.http_test_url = "https://google.com"`, and allow `false` to disable that check.

aesedepece: Basic URL validation functionality is now in place:

```
$ RUST_LOG=witnet_rad=debug cargo run --release -- -c ~/.witnet/config/witnet.toml wallet server
Loading config from: witnet.toml
Setting log level to: DEBUG, source: Env
[2022-08-11T17:49:25Z INFO  witnet::cli] Sentry telemetry enabled
[2022-08-11T17:49:25Z DEBUG witnet_data_structures] Set environment to mainnet
[2022-08-11T17:49:25Z WARN  witnet_wallet] The local Witnet node JSON-RPC server is configured to listen at 0.0.0.0:21338 but the wallet will connect to 127.0.0.1:21338
[2022-08-11T17:49:26Z INFO  witnet_config::config] The default unproxied HTTP transport for retrieval is enabled.
[2022-08-11T17:49:26Z INFO  witnet_config::config] Configuring retrieval proxies: ["socks5://127.0.0.1", "whatever://192.111.135.17:18302", "laleche"]
[2022-08-11T17:49:26Z INFO  witnet_config::config] Paranoid retrieval percentage is set to 51%
Error: The following transport addresses are invalid:
- socks5://127.0.0.1 (the address is missing a port number)
- whatever://192.111.135.17:18302 ("whatever" is not a supported type of transport)
- laleche (relative URL without a base)
```

node/src/actors/node.rs
```diff
@@ -12,12 +12,17 @@ use crate::{
...
+use witnet_validations::witnessing::validate_witnessing_config;
 
 /// Function to run the main system
 pub fn run(config: Arc<Config>, callback: fn()) -> Result<(), failure::Error> {
     // Init system
     let system = System::new();
 
+    // Perform some initial validations on the configuration
+    let witnessing_config = config.witnessing.clone().into_config();
+    validate_witnessing_config(&witnessing_config)?;
```
tmpolaczyk, line 24: This results in the following logs when starting a node with an invalid proxy url:

```
[2022-08-12T10:14:51Z DEBUG witnet_data_structures] Set environment to development
[2022-08-12T10:14:51Z INFO  witnet_config::config] The default unproxied HTTP transport for retrieval is enabled.
[2022-08-12T10:14:51Z INFO  witnet_config::config] Configuring retrieval proxies: ["http://183.239.61.204:9091", "http://183.239.61.204:90911"]
[2022-08-12T10:14:51Z INFO  witnet_config::config] Paranoid retrieval percentage is set to 51%
Error: The following transport addresses are invalid:
- http://183.239.61.204:90911 (invalid port number)
```

The error only mentions "transport addresses", it doesn't mention that this is related to proxies or witnessing. So doing something like this would add a bit of context:

```suggestion
    validate_witnessing_config(&witnessing_config).map_err(|e| {
        failure::format_err!("Error when validating witnessing config: {}", e)
    })?;
```

Which results in this error message:

```
Error: Error when validating witnessing config: The following transport addresses are invalid:
- http://183.239.61.204:90911 (invalid port number)
```

aesedepece: I have directly changed the `impl Display for WitnessingConfigError` instead. It was totally missing a first mention to the error happening while validating witnessing configuration. This will solve the thing for both the node and the wallet.

net/src/client/http/mod.rs
```diff
@@ -0,0 +1,351 @@
...
+        let headers: Vec<(String, String)> = req
+            .header_names()
+            .map(|name| {
+                (
+                    name.to_string(),
+                    req.header(name)
+                        .map(std::string::ToString::to_string)
+                        .unwrap_or_default(),
+                )
+            })
```
tmpolaczyk, line 146: There's an `iter` method to iterate over `(name, value)`:

```suggestion
            .iter()
            .map(|(name, value)| {
                (
                    name.to_string(),
                    value.to_string(),
                )
            })
```

net/src/client/http/mod.rs
```diff
@@ -0,0 +1,351 @@
...
+impl TryFrom<&mut surf::http::Request> for WitnetHttpRequest {
+    type Error = WitnetHttpError;
+
+    fn try_from(req: &mut surf::http::Request) -> Result<Self, Self::Error> {
+        let method = isahc::http::Method::from(WitnetHttpMethod::try_from(req.method())?);
+        let version = req
+            .version()
+            .ok_or(WitnetHttpError::UnknownVersion {
+                version: String::from("None"),
+            })
```
tmpolaczyk, line 131: This error is actually ignored because of the `unwrap_or_default` below. Maybe this entire method chain can be refactored into a function

    fn(Option<surf::http::Version>) -> isahc::http::Version

to make the intention to ignore errors more clear.

net/src/client/http/mod.rs
```diff
@@ -0,0 +1,351 @@
...
+            .collect();
+        let body_reader = futures::io::BufReader::new(body);
+
+        // Create a surf response and set all the relevant parts
+        let mut res = surf::http::Response::new(status);
+        res.set_version(Some(surf::http::Version::try_from(version)?));
+        res.set_body(surf::Body::from_reader(body_reader, None));
+        for (key, value) in headers {
+            res.insert_header(key.as_str(), value);
```
tmpolaczyk, line 208: I believe this is not needed because we never check response headers, and there is no way to do that in RADON, but we can leave it for future compatibility.

tmpolaczyk: Same applies to `res.set_version`.

tmpolaczyk (commented):
Some comments for the `WitnetHttpClient`.

node/tests/actors/rad_manager.rs
```diff
@@ -22,6 +22,11 @@ fn from_config_test_success_helper(
         .cloned()
         .map(|option| option.map(|uri| uri.to_string()))
         .collect::<Vec<_>>();
+    // This is dropping RadManager explicitly so as to get rid of the panic of doing so.
+    std::panic::catch_unwind(|| {
+        std::mem::drop(manager);
+    })
+    .ok();
```
tmpolaczyk, line 29: Ideally this shouldn't be needed, I believe there is a bug in the `stop_system_if_panicking` function, which panics if the actor is being dropped in a panicking thread while no actix system is running. When testing other actors they always run inside an actix system so that's not a problem, but the RadManager is more standalone. So I believe that running inside an actix system should be optional, to avoid doing this workaround in other code that doesn't use actix. I'll see if it can be fixed.

tmpolaczyk: Done, #2276

aesedepece: Thank you!

tmpolaczyk:
Proxy URLs are validated at startup, but there is no connection attempt to ensure that they work. 

There is no easy way to manually check that proxies work:
* The witnet toolkit binary cannot be used to test proxies because it doesn't read the witnet.toml file.
* The witnet node sendRequest command reads the witnet.toml file but ignores the proxy options.
* Sheikah connected to the wallet reads the witnet.toml and uses the witnessing config, but in case of timeout there is some kind of bug and sheikah displays the loading icon instead of the error. After waiting more than one minute it finally displays the error "Source looks inconsistent when...", but that error also doesn't mention which proxy failed.

So unless I missed something there is no way to check if the configured proxies work, it would be nice to have something like a test-request CLI command.

In the node, when one proxy fails with a timeout error, this is the only log:

```
[2022-09-30T08:27:10Z ERROR witnet_node::actors::chain_manager::mining] Couldn't resolve rad request 86920ddf5d7066b1db629066ab50118855c05bf7fbf3d4ba53e1154c3ca02b43: Timeout during retrieval phase
```

It would be nice if the error message could mention the URL of the proxy that failed.

node/tests/actors/rad_manager.rs
tmpolaczyk: So this hack can be removed now?

aesedepece:
> Proxy URLs are validated at startup, but there is no connection attempt to ensure that they work.

In my mind I had implemented full validation (by querying some well-known URL). But it looks like that happened only in my mind.
Would you be OK if we go with this as it is and implement that or a CLI method (or both) in a further PR?

tmpolaczyk:
> Would you be OK if we go with this as it is and implement that or a CLI method (or both) in a further PR?

Yes, it could be a separate PR. I suggested it here in case that makes us rethink the configuration parameters or something, but the basic structure of the witnessing params looks solid enough.

Then if I haven't missed any breaking changes this is ready to merge.

tmpolaczyk (approved):

PR #2259, opened 2022-08-29
feat(wallet): automatically retry connection to alternative nodes

### LINKED ISSUE #2258: Implement a mechanism to try multiple nodes

aesedepece:
This PR implements automated multi-URL reconnections for our JSONRPC clients. That is, every time a connection fails, a new connection will be created to a newly chosen URL from a configured list. This will make Sheikah connections to nodes much more reliable.

TODO:
- [x] Make sure that no tests are broken
- [x] Verify that exponential backoff is working ~~(it is probably not)~~ it works now!
- [x] Verify that we can fail fast upon faulty addresses (reduce first connection timeout)
- [x] If multiple URLs are configured, avoid reconnecting to the same one twice in a row

fix #2258


net/src/client/tcp/actors/jsonrpc.rs
```diff
@@ -14,73 +14,111 @@ use async_jsonrpc_client::{
...
+use serde_json::value;
+
+pub use serde_json::Value;
 
 use super::Error;
 
+const DEFAULT_BACKOFF_TIME_MILLIS: u64 = 250;
+
+struct Connection {
+    backoff: Duration,
```
Tommytrg, line 28: Where are you using the backoff?

net/src/client/tcp/actors/jsonrpc.rs
```diff
@@ -14,73 +14,111 @@ use async_jsonrpc_client::{
...
+    socket: TcpSocket,
+    timestamp: Instant,
+    url: Arc<Mutex<String>>,
```
Tommytrg, line 31: Why does the `url` need to be an `Arc<Mutex<String>>` instead of a `String`? Is it to avoid trying multiple urls at the same time?

net/src/client/tcp/actors/jsonrpc.rs
aesedepece: Here: https://github.com/witnet/witnet-rust/blob/47994bc46113633244e4222c5e09ea2af470a121/net/src/client/tcp/actors/jsonrpc.rs#L89-L98

net/src/client/tcp/actors/jsonrpc.rs
aesedepece: `Arc<Mutex<T>>` is used extensively in the wallet component to safely share data across actors.

In this particular case, it helps other actors to read a reference to the URL currently being used depite the URL being changed internally by the JsonRpc actor at its own discretion.

net/src/client/tcp/actors/jsonrpc.rs
```diff
@@ -14,73 +15,110 @@ use async_jsonrpc_client::{
...
 /// Use this actor to send json-rpc requests over a websockets connection.
 pub struct JsonRpcClient {
     _handle: EventLoopHandle,
-    socket: TcpSocket,
     active_subscriptions: Arc<Mutex<HashMap<String, Subscribe>>>,
     pending_subscriptions: HashMap<String, Subscribe>,
-    url: String,
+    urls: Vec<String>,
     // Used to calculate the time since the last reconnection, and prevent multiple reconnections
     // in a short time interval
```
tmpolaczyk, line 45: This comment now belongs to `connection.timestamp`, right?

wallet/src/actors/app/params.rs
```diff
@@ -16,6 +20,25 @@ pub struct Params {
...
 pub struct NodeClient {
-    pub url: String,
     pub actor: Addr<JsonRpcClient>,
+    pub url: Arc<Mutex<String>>,
+}
+
+impl NodeClient {
+    /// Get the URL that the current client is connecting to.
+    pub fn current_url(&self) -> String {
+        self.url.lock().unwrap().to_string()
```
tmpolaczyk, line 30: This function is only used in `notify_client` to get the address of the current node. Instead of using a mutex we could ask the `actor` for the current address, but that would make this function `async`, and that would be a problem because the `notify_client` function is not `async`. However looking at the body of `notify_client`, there is a `futures::executor::block_on` which indicates that that function should actually be `async`.

So may be worth it to try to refactor this method into a `self.actor.send(GetCurrentUrl).await` or similar. This way we would avoid the mutex.

wallet/src/lib.rs
```diff
@@ -114,16 +95,18 @@ pub fn run(conf: Config) -> Result<(), Error> {
...
+                .map_err(|_| app::Error::NodeNotConnected)?;
+        let node_client = Arc::new(app::NodeClient {
             actor: node_client_actor,
-            url: node_client_url.clone(),
+            url,
         });
 
+        // Trigger connection validation. Due to how the JSON-RPC client works, it will keep
+        // retrying connection using a different URL each time.
+        node_client.valid_connection().await;
```
tmpolaczyk, line 108: What happens if none of the URLs work? For example in case the user does not have an internet connection at the moment. Does he get a "Node disconnected" error message?

aesedepece: Yes, he does, that remains unchanged.

wallet/src/actors/app/params.rs
aesedepece: Good point. I used the mutex in the first place to avoid sending messages, handling futures, etc.

So it all boils down to whether we prefer to use mutexes to save us from actor overhead, or the other way around hehehe.

net/src/client/tcp/actors/jsonrpc.rs
```diff
@@ -147,6 +190,26 @@ impl JsonRpcClient {
...
+            Duration::from_millis(MAX_BACKOFF_TIME_MILLIS),
+        );
+        self.set_backoff_time(ctx, time);
+    }
+
+    fn reset_backoff_time(&mut self, ctx: &mut <Self as Actor>::Context) {
+        self.set_backoff_time(ctx, Duration::from_millis(DEFAULT_BACKOFF_TIME_MILLIS));
+    }
+
+    fn set_backoff_time(&mut self, _ctx: &mut <Self as Actor>::Context, time: Duration) {
```
tmpolaczyk, line 206: None of these methods actually need a `ctx` parameter, right?

net/src/client/tcp/actors/jsonrpc.rs
aesedepece: Ooops I was a victim of refactoring without moving the comments. Nice catch!

wallet/src/lib.rs
tmpolaczyk: Cool, I was wondering if this await could cause some kind of infinite loop here and break everything.

wallet/src/actors/app/params.rs
tmpolaczyk: In this case the natural pattern is to just send messages I believe, because it is the `JsonRpcClient` actor who owns the `url`. If we are using a mutex then the `NodeClient` is technically allowed to change the url, but it shouldn't be, it should only be able to read it. And not sure about overhead but the actor messages are implemented using similar mutexes probably.

net/src/client/tcp/actors/jsonrpc.rs
```diff
@@ -394,3 +465,26 @@ fn subscription_topic_from_request(request: &Request) -> String {
...
+        1 => Some(input[0].clone()),
+        _ => {
+            let mut pick;
+            loop {
+                pick = input.choose(&mut rand::thread_rng())?.clone();
+
+                if pick != old {
+                    break;
+                }
+            }
```
tmpolaczyk, line 485: You can avoid the reroll like this:

* Choose 1 out of N-1 instead of 1 out of N, for example making it impossible to get the element at position 0
* If the chosen item is equal to the expected item, return the element at position 0 instead
* Otherwise, return the random item

This assumes that there is only 1 element with the same value as `old`.

```suggestion
            let pick = input[1..].choose(&mut rand::thread_rng())?.clone();

            if pick == old {
                return Some(input[0].clone());
            }
```

tmpolaczyk:
Seems to be working fine when tested in Sheikah!

tmpolaczyk:
I tried to remove the mutex, and it turned out to be mostly straightforward: https://github.com/tmpolaczyk/witnet-rust/tree/wallet-no-mutex

tmpolaczyk:
There are some conflicts, can you rebase again?

tmpolaczyk (approved):

PR #2267, opened 2022-09-06
ci: add image publishing

aeweda:


.github/workflows/bridge.yml
```diff
@@ -33,4 +33,14 @@ jobs:
...
+
+      - name: Login to Docker hub Registry
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+
+      - name: Push Image
+        run: |
+          docker push witnet/witnet-centralized-ethereum-bridge:"${{ env.IMAGE_VERSION }}
```
aesedepece, line 46: Is it really OK for this double quote to never be closed? :thinking: 
```suggestion
          docker push witnet/witnet-centralized-ethereum-bridge:"${{ env.IMAGE_VERSION }}"
```

aesedepece (changes requested):

.github/workflows/bridge.yml
aeweda: no it should most definitely be closed, great catch 😂

aesedepece (approved):

PR #2268, opened 2022-09-07
feat(wallet): add pending_tx_timeout_seconds config variable

### LINKED ISSUE #2265: feat(wallet): enable customizing UTXO timeout time through configuration
As requested by MEXC.

tmpolaczyk:
Close #2265

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -896,6 +896,9 @@ pub struct Wallet {
     pub sync_address_batch_length: u16,
     /// Allow to use outputs that have not been confirmed by a superblock in new transactions
     pub use_unconfirmed_utxos: bool,
+    /// If a transaction has not been included in a block for this amount of seconds, mark the UTXOs
+    /// as available again.
+    pub pending_transactions_timeout_seconds: u64,
 }
 
 impl Wallet {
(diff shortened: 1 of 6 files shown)


tmpolaczyk:
Setting this variable to 10 seconds in witnet.toml

```
[wallet]
pending_tx_timeout_seconds = 10
```

to try to force some double spends, and this is the result:

Captura de pantalla de 2022-09-07 11-49-58

The invalid transactions seem to stay in "Sending" state forever, and exiting from the wallet in sheikah makes the "Sending" transactions disappear. Triggering a "Resync" also makes those transactions disappear, but that's expected because a resync deletes all the local data. Creating more transactions or changing the transaction page does not make the "Sending" transactions disappear, that's also expected.

But the main use case of this feature is to increase the timeout, and in that case it works as expected, you cannot create invalid transactions while another transaction is still pending.

Captura de pantalla de 2022-09-07 12-16-58

Note that this is already the default behavior, so this pull request only changes the fact that this timeout can be increased or reduced.

Tommytrg (approved):

config/src/config.rs
```diff
@@ -896,6 +896,9 @@ pub struct Wallet {
     pub sync_address_batch_length: u16,
     /// Allow to use outputs that have not been confirmed by a superblock in new transactions
     pub use_unconfirmed_utxos: bool,
+    /// If a transaction has not been included in a block for this amount of seconds, mark the UTXOs
+    /// as available again.
+    pub pending_tx_timeout_seconds: u64,
```
aesedepece, line 901:
```suggestion
    pub pending_transactions_timeout_seconds: u64,
```

aesedepece (changes requested):

aesedepece (approved):

PR #2270, opened 2022-09-13
feat(node): more rewind params

tmpolaczyk:
Add some flags to the rewind JSON-RPC method to allow manually fixing storage with missing items, issue #2102 

Command to process all the blocks again to fix missing entries in storage:

```json
{"jsonrpc":"2.0","id":"1","method":"rewind","params":{"mode":{"write_items_to_storage": true}}}
```

The original use case of trying to recover a forked node still works with the same syntax:

```json
{"jsonrpc":"2.0","id":"1","method":"rewind","params":[10000]}
```

The main blocker is that if there is actually a missing block, the rewind process stops and the node enters the normal synchronization process which is much slower than a rewind. But eventually it fixes the issue, it is just very slow.

Marking as a draft while I check if I can implement some "resume" functionality to avoid the slow synchronization.

### DIFF
--- a/node/src/actors/chain_manager/handlers.rs
+++ b/node/src/actors/chain_manager/handlers.rs
@@ -165,7 +165,7 @@ impl Handler<EpochNotification<EveryEpochPayload>> for ChainManager {
                         }) = best_candidate
                         {
                             // Persist block and update ChainState
-                            self.consolidate_block(ctx, block, utxo_diff, false);
+                            self.consolidate_block(ctx, block, utxo_diff, None);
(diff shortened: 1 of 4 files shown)


tmpolaczyk:
So this does not work as nicely as expected, because it turns out that writing every item to the storage again does increase the total size used by rocksdb, even if the written data is exactly the same as the existing data. This resulted in my machine running out of disk space during the rewind, and that caused a fatal error leaving the storage in an invalid state, with error "OutputNotFound" when trying to synchronize. And after that the only way to recover is to do another rewind...

Issue #2282, opened 2022-10-17
Wallet serialization blocks update to secp256k1 0.24.0


tmpolaczyk:
After updating to secp256k1 0.24.0, the wallet does not allow to unlock previously created wallets with this error:

```
[2022-10-17T15:37:24Z ERROR witnet_wallet::actors::app::error] Internal Error: repository failed: database failed: bincode failed: io error: unexpected end of file
```

We need to find out why is that happening, and implement some workaround to allow us to update to secp256k1 0.24.0 or greater.

tmpolaczyk:
This seems to be because of the `ExtendedSK` type, it is used to store the master key, and the derived internal and external keys.

See the `master_key` and `account_key` functions here:
https://github.com/witnet/witnet-rust/blob/8304d479a1d78ad6b3266af30d468004384086ac/wallet/src/repository/keys.rs

PR #2284, opened 2022-10-18
test(wallet): ensure unlock wallet does not accidentally break

tmpolaczyk:
Related to #2282, does not fix the bug but it ensures we don't accidentally break that again


wallet/src/db/tests.rs
```diff
@@ -13,6 +13,30 @@ impl HashMapDb {
     pub fn new(rc: Rc<RefCell<HashMap<Bytes, Bytes>>>) -> Self {
         Self { rc }
     }
+
+    #[allow(dead_code)]
```
Tommytrg, line 17: Why do we need to specify this flag to the linter?

tmpolaczyk: Because this function is not used anywhere, and I don't want to remove it because it feels weird to have an import but no export.

aesedepece (approved):

PR #2287, opened 2022-10-20
Feat/wallet/return transaction upon creation

### LINKED ISSUE #2279: Return type transaction creation methods in Wallet API should include a BalanceMovement or WalletTransaction
This is a re-scoped follow-up to #2272, after finding out that:

> Apparently, it was a misconception that whole transactions were not being returned back when creating VT or DR transactions in through the Wallet API.
>
> If we're not being able to render inputs in Sheikah it's because we are not returning a BalanceMovement or WalletTransaction along the plain transaction data (which does not contain the actual inputs but only "output pointers"). This issue is a wontfix but I will open a different one for assessing and implementing that other feature.

aesedepece:
fix #2279


tmpolaczyk:
Can you open a separate PR with just all the `.clone()` stuff? This way this will be much easier to review.

aesedepece:
> Can you open a separate PR with just all the `.clone()` stuff? This way this will be much easier to review.

Done! (#2290)

tmpolaczyk:
Captura de pantalla de 2022-10-25 15-27-47

It works!

wallet/src/repository/wallet/tests/mod.rs
```diff
@@ -586,6 +592,14 @@ fn test_create_vtt_does_not_spend_utxos() {
         })
         .unwrap();
 
+    // the extended transaction should actually contain a value transfer transaction
+    let vtt = if let Transaction::ValueTransfer(vtt) = extended.transaction {
+        Some(vtt)
+    } else {
+        None
+    }
+    .unwrap();
```
tmpolaczyk, line 601: Using `None.unwrap()` will not show us why this test failed, prefer a panic like this:

```suggestion
    let vtt = if let Transaction::ValueTransfer(vtt) = extended.transaction {
        vtt
    } else {
        panic!("the extended transaction should actually contain a value transfer transaction, got: {:?}", extended.transaction);
    };
```

If that doesn't compile, try using a match instead of if let.

wallet/src/repository/wallet/mod.rs
```diff
@@ -1266,17 +1293,20 @@ where
             max_weight,
         )?;
 
-        let change_pkh =
-            self.calculate_change_address(dr_output.is_none(), &tx_info.inputs, state)?;
+        let change_pkh = self.calculate_change_address(state, None)?;
```
tmpolaczyk, line 1296: This is always `None`, I believe it should be `Some` when the transaction is a data request to match the old behavior.

tmpolaczyk: Yes, trying to deploy data requests from sheikah is broken, with this error:

```
[2022-10-25T14:46:31Z ERROR witnet_wallet::actors::app::error] Node Error: request failed: RPC error: Error { code: InternalError, message: "Public key hash mismatch: expected twit1v39juf5kvqapku258rdwlpch3p5zh8dt93nn0h got twit185j3rmer0kfhzm3uy3ecv2d495wge7n228cfcp", data: None }
```

aesedepece: Ooops I knew I was breaking it, but forgot to put a TODO in place so I wouldn't forget to fix it later :sweat_smile: 

It should be fixed now!

wallet/src/repository/wallet/tests/mod.rs
aesedepece: Agreed and fixed!

wallet/src/repository/wallet/tests/mod.rs
```diff
@@ -659,10 +673,18 @@ fn test_create_data_request_does_not_spend_utxos() {
...
         .create_data_req(types::DataReqParams { fee, request })
         .unwrap();
 
+    // the extended transaction should actually contain a data request transaction
+    let data_req = if let Transaction::DataRequest(drt) = extended.transaction {
+        Some(drt)
+    } else {
+        None
+    }
+    .unwrap();
```
tmpolaczyk, line 686: This also could be a panic.

And would it be possible to test that the output of the data request has the same pkh as the first input?

tmpolaczyk:
Also, I'm getting some kind of sync error when the wallet finds a tally transaction.

```
[2022-10-25T15:51:58Z WARN  witnet_wallet::repository::wallet] Found transaction to our address twit1vfuklxq4p5d7t744evcrj04r68q7j2f8yj0vqx! Amount: +3 nanowits
[2022-10-25T15:51:58Z ERROR witnet_wallet::repository::wallet] Error while indexing transaction: Err(WrongMetadataType("TallyTransaction { dr_pointer: 729293313b24b4afaff65f07e71bb8673d66fc097f443ccdfa633d30dbd0308a, tally: [216, 39, 129, 24, 82], outputs: [ValueTransferOutput { pkh: PublicKeyHash { hash: [98, 121, 111, 152, 21, 13, 27, 229, 250, 181, 203, 48, 57, 62, 163, 209, 193, 233, 41, 39] }, value: 3, time_lock: 0 }], out_of_consensus: [], error_committers: [], hash: Memoized { hash: RwLock { data: Some(e306c28390ddbe71a8bc8fe9ea264287ee4bb76936dc17b180a189efc3229ced), poisoned: false, .. } } }"))
[2022-10-25T15:51:58Z DEBUG witnet_wallet::repository::wallet] Will retrieve 0 from local, 0 from pending and 6 from DB
[2022-10-25T15:51:58Z DEBUG witnet_wallet::repository::wallet] Will retrieve 0 from local, 0 from pending and 13 from DB
[2022-10-25T15:51:58Z ERROR witnet_wallet::actors::worker::methods] Error while synchronizing (start: 2807824, end:2807824): repository failed: transaction metadata type is wrong: TallyTransaction { dr_pointer: 729293313b24b4afaff65f07e71bb8673d66fc097f443ccdfa633d30dbd0308a, tally: [216, 39, 129, 24, 82], outputs: [ValueTransferOutput { pkh: PublicKeyHash { hash: [98, 121, 111, 152, 21, 13, 27, 229, 250, 181, 203, 48, 57, 62, 163, 209, 193, 233, 41, 39] }, value: 3, time_lock: 0 }], out_of_consensus: [], error_committers: [], hash: Memoized { hash: RwLock { data: Some(e306c28390ddbe71a8bc8fe9ea264287ee4bb76936dc17b180a189efc3229ced), poisoned: false, .. } } }
```

wallet/src/repository/wallet/mod.rs
```diff
@@ -1659,9 +1707,16 @@ where
             _ => 0,
         };
 
+        // Insert resolved inputs into transaction metadata so that they can be added to the balance
+        // movement.
+        let txn = model::ExtendedTransaction {
+            metadata: Some(model::TransactionMetadata::InputValues(resolved_inputs)),
+            transaction: txn.transaction.clone(),
+        };
```
tmpolaczyk, line 1715: This looks like the most probable cause for the `WrongMetadataType` error, the metadata is always set to `InputValues` here. Since tally transactions have no inputs, this conversion shouldn't be needed if the transaction is a tally transaction.

tmpolaczyk: Alright, this seems to fix it:

```rust
// Insert resolved inputs into transaction metadata so that they can be added to the balance
// movement.
let txn = match &txn.metadata {
    Some(model::TransactionMetadata::InputValues(_input_values)) => {
        ExtendedTransaction {
            metadata: Some(model::TransactionMetadata::InputValues(resolved_inputs)),
            transaction: txn.transaction.clone(),
        }
    }
    _ => {
        txn.clone()
    }
};
```

The `txn.clone()` could be avoided but it makes the code a bit uglier.

wallet/src/repository/wallet/tests/mod.rs
aesedepece: Gotcha. Did a find-and-replace and of course this didn't match.

aesedepece: > And would it be possible to test that the output of the data request has the same pkh as the first input?

Great idea — this is tested now!

wallet/src/repository/wallet/mod.rs
aesedepece: Nice catch. I modified it so it never overwrites metadata if there's some in place (which is always the case for tally transactions).

tmpolaczyk (approved):

PR #2295, opened 2022-11-03
fix(node): check superblock parity to avoid voting desync

tmpolaczyk:


aesedepece:
Included into and superseded by #2297

PR #2296, opened 2022-11-03
chore: fix clippy warnings for Rust 1.65

tmpolaczyk:
Fix the most trivial ones, and add temporary "allow" directives for the rest.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -19,6 +19,8 @@ jobs:
         -D clippy::cast-sign-loss
         -D clippy::checked-conversions
         -A clippy::upper-case-acronyms
+        -A clippy::result-large-err
+        -A special_module_name
 
     steps:
       - uses: actions/checkout@v1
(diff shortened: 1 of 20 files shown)


.github/workflows/push.yml
```diff
@@ -19,6 +19,8 @@ jobs:
         -D clippy::cast-sign-loss
         -D clippy::checked-conversions
         -A clippy::upper-case-acronyms
+        -A clippy::result-large-err
+        -A special_module_name
```
aesedepece, line 23: Shall we create an issue not to forget to solve these?

aesedepece (changes requested):

.github/workflows/push.yml
tmpolaczyk: Oh, I already resolved those in this branch: https://github.com/tmpolaczyk/witnet-rust/tree/rust-1-65

but I just cherry-picked the first commit because it is trivial to review and it doesn't block the release. The other changes may need some comments.

aesedepece:
Included into and superseded by #2297

PR #2298, opened 2022-11-07
Fix remaining clippy lints from Rust 1.65 update

tmpolaczyk:
Follow-up to #2296


data_structures/src/error.rs
```diff
@@ -276,7 +276,7 @@ pub enum TransactionError {
     DataRequestWeightLimitExceeded {
         weight: u32,
         max_weight: u32,
-        dr_output: DataRequestOutput,
+        dr_output: Box<DataRequestOutput>,
```
aesedepece, line 279: Very interesting. Do you know why are they pushing for using `Box`es liberally like this?

toolkit/src/cli/mod.rs
```diff
@@ -7,7 +7,7 @@ mod arguments;
 pub(crate) mod commands;
 pub(crate) mod data_requests;
 
-use crate::errors::Error;
+use witnet_toolkit::errors::Error;
```
aesedepece, line 10: No more `crate::`? :cry:

tmpolaczyk: `crate::` imports from `main.rs`, `witnet_toolkit` imports from `lib.rs`

data_structures/src/error.rs
tmpolaczyk: https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err

Basically, without the box, the `TransactionError` enum is too big, and that hurts performance a bit. We could have changed all the uses of `Result<(), TransactionError>` to `Result<(), Box<TransactionError>>` but in this case I preferred to box just this one large field.

aesedepece (approved):

PR #2300, opened 2022-11-07
feat(node): allow overwriting private key through a flag when importing

### LINKED ISSUE #1693: Allow overwriting master key of node
# Problem description
Currently it is impossible to overwrite the master key of the node (see example), so if you want to reuse a synced node with another master key (e.g., check a cold wallet and quickly move some WIT), you have to start a node from scratch and synchronize it. That's not a big problem currently, but in the future you don't want to start a new node and synchronize 6 months worth of blocks if you just want to perform one action from a different identity.
```
./witnet node server --master-key-import master-1.key
...
./witnet node server --master-key-import master-2.key
...
Failed to configure master key: Tried to overwrite node master key with a different one.
    Node pkh:     wit15j0vp9wwgj3ea9z5s7fjxq3nzyx233cyj24cmh
    Imported pkh: wit1k6j82vxznn4nzysmjgr8kk2sqr7x5ye9jsxe69

    In order to import a different master key, you first need to export the current master key and delete the storage
Error: Non-zero exit code: 1
```

# Suggested solution
Allow overwriting a master key of a node by introducing an overwrite flag, which will overwrite the current master key from the database with the newly imported:
`./witnet node server --master-key-import master-2.key --overwrite`

aesedepece:
Fix #1693

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -312,6 +312,10 @@ pub struct Storage {
     #[partial_struct(skip)]
     #[partial_struct(serde(default))]
     pub master_key_import_path: Option<PathBuf>,
+    /// Whether to overwrite an existing master key when importing a new one
+    #[partial_struct(skip)]
+    #[partial_struct(serde(default))]
+    pub overwrite: bool,
     /// Keep a map of "address" to "list of UTXOs" in memory, to speed up getBalance and getUtxoInfo methods
     pub utxos_in_memory: bool,
 }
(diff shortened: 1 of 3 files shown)


aesedepece:
This is not perfect because of the concerns explained in the comments of #1693, but I'm pushing it here anyway because it can be useful in some circumstances.

tmpolaczyk (changes requested):
This will break anything related to utxos in the node, requiring at least a rewind after importing to be able to create transactions

drcpu-github:
This may give you some ideas: [overwrite-master-key-1.5.4.txt](https://github.com/witnet/witnet-rust/files/9954702/overwrite-master-key-1.5.4.txt)

drcpu-github:
I guess this can be closed given the new `snapshot-import` and `snapshot-export` capabilities?

PR #2307, opened 2022-11-15
fix(node): commit http errors in data request retrieval

### LINKED ISSUE #2306: HTTP errors are not being commited since version 1.5.3
Starting from the 1.5.3 release, when a node encounters an HTTP error or a timeout in one of the sources of a data request, it will not commit any value at all.

Run a git bisect and e33986dd4ca0c5520bc3b4e5479be89fba6bdeea is the first bad commit.

Besides, one single failing source would make the node to infer that the whole data request failed and, again, the node would refrain from committing any value or error at all. This is something that was presumably fixed before 1.5.3.

tmpolaczyk:
Fix #2306

### DIFF
--- a/node/src/actors/rad_manager/handlers.rs
+++ b/node/src/actors/rad_manager/handlers.rs
@@ -4,7 +4,7 @@ use std::time::Duration;
 
 use actix::{Handler, ResponseFuture};
 use futures::FutureExt;
-use witnet_data_structures::radon_report::{RadonReport, ReportContext};
+use witnet_data_structures::radon_report::{RadonReport, ReportContext, RetrievalMetadata, Stage};
 use witnet_rad::{
     conditions::{evaluate_tally_precondition_clause, TallyPreconditionClauseResult},
     error::RadError,
@@ -48,6 +48,8 @@ impl Handler<ResolveRA> for RadManager {
                 }
             };
             let settings = RadonScriptExecutionSettings::disable_all();
+            let retrieval_context =
+                ReportContext::from_stage(Stage::Retrieval(RetrievalMetadata::default()));
             let retrieve_responses_fut = sources
                 .iter()
                 .map(|retrieve| {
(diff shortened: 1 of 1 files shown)


aesedepece:
This change sounds like a good candidate for being shipped with 1.6 — behind a TAPI flag if necessary. What do you think?

tmpolaczyk:
I believe this can be merged without TAPI, because the old behavior is to not commit anything at all. Therefore the data requests will be unsolved until a majority of miners update, and then it will just work as expected.

It is true that we won't have a guarantee that these data requests will be solved until the next TAPI update, which ensures that miners have updated, but I believe nothing will break if this change is merged without TAPI.

aesedepece:
> It is true that we won't have a guarantee that these data requests will be solved until the next TAPI update, which ensures that miners have updated, but I believe nothing will break if this change is merged without TAPI.

I'm just fine with that approach. Thanks!

guidiaz:
In relation to what was discussed in
https://github.com/witnet/witnet-rust/discussions/2050, I'd suggest to let
a single HTTP/403 error in any of the data request's sources be interpreted
as a sufficient condition to make the witness node not to commit to such
data request. This way we'd make data requests relying on private APIs to
be systematically resolved with InsufficientCommits, unless there was a
minimal amount of witnessing nodes prepared to attend involved private API.

tmpolaczyk:
I don't think it is a good idea to special-case HTTP/403 errors, because many APIs return 404 or 401 instead.

guidiaz:
> I don't think it is a good idea to special-case HTTP/403 errors, because many APIs return 404 or 401 instead.

Yeah right, either HTTP/401 or HTTP/403 errors in one single source would be sufficient condition for not commiting to a data request. Data requests relying on a majority of private API sources returning 404 error codes would just be resolved as `HttpError { code: 404 }`.

aesedepece:
@tmpolaczyk shall we rebase and merge this onto 1.6?

tmpolaczyk:
> @tmpolaczyk shall we rebase and merge this onto 1.6?

Done

aesedepece (approved):

PR #2314, opened 2022-11-22
fix(rad): parse JSON integers larger than u64::MAX

### LINKED ISSUE #2310: Operator parseJSONMap fails on JSON objects containing long integer values
For instance: `{"data": 122637770461000000000}`

tmpolaczyk:
Fix #2310, but this will probably need TAPI to not break anything. Also the output is a bit unexpected, instead of parsing the number "122637770461000000000" as an integer or as a float, it parses it as the integer "122637770461000007680" (which differs in the last digits).

### DIFF
--- a/rad/src/script.rs
+++ b/rad/src/script.rs
@@ -555,6 +555,45 @@ mod tests {
     }
 
     #[test]
+    fn test_parse_json_map_big_integers() {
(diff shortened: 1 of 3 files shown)


tmpolaczyk:
Closing because this will need more thought.

PR #2316, opened 2022-11-24
refactor(util): remove unused local_time function

tmpolaczyk:
The `local_time` function was used to convert from timestamp to local time, and then back to timestamp. I have verified that the result is the same when using the first timestamp directly.


aesedepece (approved):
makes sense to me!

PR #2321, opened 2022-11-30
feat(bridge): add skip_first config param to allow skipping requests

### LINKED ISSUE #2320: Populate local database from certain query id set as config parameter

tmpolaczyk:
Close #2320

### DIFF
--- a/bridges/centralized-ethereum/src/actors/eth_poller.rs
+++ b/bridges/centralized-ethereum/src/actors/eth_poller.rs
@@ -22,6 +22,8 @@ pub struct EthPoller {
     pub wrb_contract: Option<Arc<Contract<web3::transports::Http>>>,
     /// Period to check for new requests in the WRB
     pub eth_new_dr_polling_rate_ms: u64,
+    /// Skip first requests up to index n when updating database
+    pub skip_first: u64,
 }
 
 impl Drop for EthPoller {
@@ -59,6 +61,7 @@ impl EthPoller {
         Self {
             wrb_contract: Some(wrb_contract),
             eth_new_dr_polling_rate_ms: config.eth_new_dr_polling_rate_ms,
+            skip_first: config.skip_first.unwrap_or(0),
         }
     }
 
@@ -66,6 +69,7 @@ impl EthPoller {
         log::debug!("Checking new DRs from Ethereum contract...");
 
         let wrb_contract = self.wrb_contract.clone().unwrap();
+        let skip_first = U256::from(self.skip_first);
         // Check requests
         let fut = async move {
             let total_requests_count: Result<U256, web3::contract::Error> = wrb_contract
@@ -89,9 +93,16 @@ impl EthPoller {
             let dr_database_addr = DrDatabase::from_registry();
             let db_request_count = dr_database_addr.send(GetLastDrId).await;
 
-            if let (Ok(total_requests_count), Ok(Ok(db_request_count))) =
+            if let (Ok(total_requests_count), Ok(Ok(mut db_request_count))) =
                 (total_requests_count, db_request_count)
             {
+                if db_request_count < skip_first {
+                    log::debug!(
+                        "Skipping first {} requests per skip_first config param",
+                        skip_first
+                    );
+                    db_request_count = skip_first;
+                }
                 if db_request_count < total_requests_count {
                     let init_index = usize::try_from(db_request_count + 1).unwrap();
                     let last_index = usize::try_from(total_requests_count).unwrap();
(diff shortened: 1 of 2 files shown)


guidiaz (changes requested):
It works perfectly. However, we need to change the version tag to something like "1.5.4.1" as to make automated docker image construction and publishing work properly.

aesedepece:
> It works perfectly. However, we need to change the version tag to something like "1.5.4.1" as to make automated docker image construction and publishing work properly.

Not strictly, I think we can just delete the existing release on DockerHub and then merge it.

tmpolaczyk:
I would prefer a separate PR for version bumps anyway, as that shouldn't block this PR.

guidiaz:
> > It works perfectly. However, we need to change the version tag to something like "1.5.4.1" as to make automated docker image construction and publishing work properly.
> 
> Not strictly, I think we can just delete the existing release on DockerHub and then merge it.

@aeweda ??

guidiaz:
> Not strictly, I think we can just delete the existing release on DockerHub and then merge it.

Perhaps in this occasion, but imagine 1.5.4 image was already been used in all testnets and mainnets and we wanted to only deploy a new binary release into (a subset of) testnets.

aesedepece:
> > Not strictly, I think we can just delete the existing release on DockerHub and then merge it.
> 
> Perhaps in this occasion, but imagine 1.5.4 image was already been used in all testnets and mainnets and we wanted to only deploy a new binary release into (a subset of) testnets.

We can always docker pull on the hosts where we need the latest one, or even refer the images by their hash.

In any case, the thing here is that CI now enforces consistency between the main witnet-rust crate and this one, so we can't do anything exotic with the versions. 

We can however make a 1.5.5 release of everything. It's all automated, so not a big deal.

aeweda:
> > It works perfectly. However, we need to change the version tag to something like "1.5.4.1" as to make automated docker image construction and publishing work properly.
> 
> Not strictly, I think we can just delete the existing release on DockerHub and then merge it.

this would be a problem as it wouldn't force a rolling update / repull unless the version actually changes


> > > Not strictly, I think we can just delete the existing release on DockerHub and then merge it.
> > 
> > 
> > Perhaps in this occasion, but imagine 1.5.4 image was already been used in all testnets and mainnets and we wanted to only deploy a new binary release into (a subset of) testnets.
> 
> We can always docker pull on the hosts where we need the latest one, or even refer the images by their hash.
> 
> In any case, the thing here is that CI now enforces consistency between the main witnet-rust crate and this one, so we can't do anything exotic with the versions.
> 
> We can however make a 1.5.5 release of everything. It's all automated, so not a big deal.

i would go with 1.5.5 version release as we can't really go into servers one by one to do a docker pull
however if we are gonna standardize it this way, let me and i can make the deployment of the bridge automated with a new tag push

tmpolaczyk:
Setting the version of `witnet-centralized-ethereum-bridge` to 1.5.5-rc1 results in this error in the script that checks version compatibility:

Version of subcrate "bridges/centralized-ethereum" (futures-util = { 0.3.8, features = ["compat"] }) does not match that of main crate (1.5.4)

guidiaz (approved):

Tommytrg:
> We can however make a 1.5.5 release of everything. It's all automated, so not a big deal.

The main problem we have with releasing a new version of `witnet-rust` every time we release a new version of the bridge is we'll tweet and announce irrelevant releases from the miner's perspective and make unnecessary noise.

A recap of all the different options we have discussed until now are:
1. Release the bridge together with the node
    
    ##### Pros
    - The easiest way to release it	
	
    ##### Cons
    - Every time we release a new witnet-rust release, we publish a tweet and a bridge release is irrelevant from the miner's perspective.


2. Release the bridge and the node independently
	
    2.1 Move the bridge to its repository
      ##### Pros
      - We don't have to change any actions in witnet-rust or modify the automatic announcements
      ##### Cons
      - To do it, we have to create a new repo and completely decouple the bridge from witnet-rust 
      - We need to keep track of compatible versions between witnet-rust and the bridge 

    
   2.2 Release the bridge and the node independently in the same repo
 
	##### Pros
    - We don't need to create a new repo and dedicate development effort
 
	##### Cons
    - We need to keep track of compatible versions between witnet-rust and the bridge
    - We need to change the action that triggers a release announcement
 
In this, I suggest releasing the bridge and the node altogether. Apparently, this is a common strategy in monorepos, for example, Vue (https://github.com/vuejs/vue/releases) or Vercel (https://github.com/vercel/vercel)

To follow this strategy, we will need to stop the Zapier trigger and update it only to publish an announcement when a change is made in the protocol.

PR #2330, opened 2022-12-27
ci: 🛂  make bridge auto-deploy on tag push

aeweda:
**There's a small caveat with this PR**

- both the bridge workflow & master workflow trigger the second a tag is released
- the bridge docker image in `docker/bridge/Dockerfile` uses `witnet-rust:latest` as a base
- naturally this will end up making the bridge use the pre-last witnet-rust image

**Courses of Action:**
- if it doesn't matter, approve PR and we can leave it as is
- if it does matter, we incorporate the workflow into the master workflow after the image deployment so that we can use that stage as a dependency
- we use/build a different image base (wouldn't recommend it!!!)

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -314,6 +314,8 @@ pub struct Storage {
     pub master_key_import_path: Option<PathBuf>,
     /// Keep a map of "address" to "list of UTXOs" in memory, to speed up getBalance and getUtxoInfo methods
     pub utxos_in_memory: bool,
+    /// RocksDB option max_open_files. -1 means unlimited.
+    pub max_open_files: i32,
 }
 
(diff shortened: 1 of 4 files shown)


aesedepece:
I'm not fully sure if I'm understanding right. Are you saying that every time we create a new tag, both the node and the bridge images will be updated, but as the bridge image is based on the node image, it will always be based on the immediately previous version of it?

aesedepece:
Isn't it possible to run one workflow after the other?

aesedepece:
> Isn't it possible to run one workflow after the other?

@aeweda what do you think? Can you figure out any way to prevent that caveat? Thanks!

aesedepece (approved):

PR #2337, opened 2023-01-11
WIP: minimum data request reward to collateral ratio

tmpolaczyk:
Continuation to #2229

* Removed required_reward_collateral_ratio from ConsensusConstants, to avoid changing the magic number. The plan is to create a separate issue to allow easily changing the ConsensusConstants, but in a future release.
* Miners that update will not include in their blocks the transactions that will be invalid after the WIP activation. This avoids an extra step of having to validate all the data requests in the TransactionsPool when the WIP status changes. And also data request creators will start to see delays in their data requests, incentivizing them to increase the reward.
* Fixed data request examples, increased the reward to 0.006 wits. This will be the minimum accepted reward with the proposed reward to collateral ratio of 1/125, and the default collateral of 1 wit.

The TAPI logic is all ready except for the voting start date, which I believe can be decided in a different PR.


aesedepece:
I have changed the base of this PR to branch `witnet-1.6`. That's where we can try the integration of all the new WIPs, and add the activation dates when decided.

drcpu-github:
Is 1 / 125 the value we will settle on?

tmpolaczyk:
> Is 1 / 125 the value we will settle on?

It is not decided yet, but I thought that was the original proposal?

drcpu-github:
> It is not decided yet, but I thought that was the original proposal?

That is one of the proposals, yes. I think it's a decent value with a sound underlying reason (but I'm biased 😄). I was just wondering if you already discussed it internally.

tmpolaczyk:
No, it has not been discussed. But it looks like a good value to me as well.

data_structures/src/data_request.rs
```diff
@@ -514,7 +518,14 @@ pub fn calculate_reward_collateral_ratio(
...
     if witness_reward > 0 {
-        dr_collateral / witness_reward
+        let ratio = dr_collateral / witness_reward;
+
+        if dr_collateral % witness_reward == 0 {
+            ratio
+        } else {
+            // If the ratio is not an integer, e.g. 123.0001, we round it to the next integer, e.g. 124.
+            ratio.saturating_add(1)
+        }
```
aesedepece, line 528: Can't we simply `dr_collateral / witness_reward).round()`?

If not, i hope we can have [`div_ceil`](https://doc.rust-lang.org/std/primitive.u64.html#method.div_ceil) soon...

aesedepece (changes requested):

aesedepece (changes requested):

witnet.toml
```diff
@@ -96,6 +96,8 @@ genesis_path = ".witnet/config/genesis_block.json"
 #mint_external_percentage = 50
 # Set a minimum fee you require before your node includes a value transfer transaction into a block
 minimum_vtt_fee_nanowits = 1
+# Set a minimum data request reward collateral ratio you expect before your node includes a data request into a block
+minimum_reward_collateral_ratio = 125
```
aesedepece, line 100: If this is exactly what this does, I can't see the point.

Imagine that you decrease this value and hence your node becomes more picky, refusing to mine under-rewarding data requests. Rest assured that someone else will leave it to a default value, letting the data request in. Your node will still participate in resolving the under-rewarding data request that you censored before, right?

aesedepece (changes requested):

witnet.toml
aesedepece: I think it would make more sense to have a setting that makes your node refrain from committing to under-rewarding data requests as a mean for witnessing risk management.

Then the fact that anyone can set that value to whatever they want would obviously mess with the replication factor and potentially end up in `InsufficientCommits`,  but idk maybe we're ok with that because the automated mechanism for adjusting the witnessing difficulty seems to work like a charm rn.

aesedepece: One way or another, let me advocate for protocol-level consensus constants over user-customizable settings. Every time we introduce a user-customizable setting, we're making things worse for the main two protocol personas:
- **Miners**: how can they assess what's a reasonable value to use for this setting? Are we explaining clearly enough that the bigger the number used here, the bigger the risk they take? Or that we can't go higher than 125 nor lower than 1? (of course that means that they can't require data requests to reward more than the collateral amount)
- **Requesters**: same story here. If we let miners reject or delay transactions based on this setting, then the request-to-collateral ratio becomes non-trivial for the requester. How can they figure out what's a reasonable ratio? Why would they not use the default value, provided that this won't affect transaction priority? What does this number mean to the requester at all, given that they're not the ones assessing risk? How will they understand that they need to decrease their ratio if a data request errors because of a legal ratio (<=125) but still not good enough for a majority of miners?

aesedepece: Summing up, you know I'm all for streamlining the user experience of Witnet, and I think that having this particular user setting works radically against that.

aesedepece: image

drcpu-github: > Rest assured that someone else will leave it to a default value, letting the data request in. Your node will still participate in resolving the under-rewarding data request that you censored before, right?

Correct on both accounts. When creating the reference implementation, I did not realize that you will still solve the request after someone else has included it, which makes that setting next to useless.

> I think it would make more sense to have a setting that makes your node refrain from committing to under-rewarding data requests as a mean for witnessing risk management.

That's the general idea I had in mind, yes. Adding a setting like that offered an easy-to-use solution for miners to 'disagree' with the protocol-set risk-rewards parameter. Given that anyone could easily modify his nodes to do this, I saw little harm in adding it out-of-the-box.

> Then the fact that anyone can set that value to whatever they want would obviously mess with the replication factor and potentially end up in InsufficientCommits, but idk maybe we're ok with that because the automated mechanism for adjusting the witnessing difficulty seems to work like a charm rn.

Given that the replication factor will result in 16 times the number of required witnesses, I doubt we'll quickly see InsufficientCommits data requests due to this additional setting. If we do, that means a very large majority of miners has decided it does not want to solve data requests at a certain ratio. And then we arrive at below point:

> how can they assess what's a reasonable value to use for this setting

I'd argue both miners and requesters can relatively easily assess this by looking at the data requests flowing network. If a majority of data requests are getting solved with a specific ratio, all participants know a high amount of miners consider that setting to be acceptable. It then seems like a relatively straightfoward conclusion that electing to set a lower ratio means requesters risk the data request not being solved and miners risk not solving them thereby losing reputation.

You can probably make the point that this may result in requesters overpaying for solving the data requests (akin to how gas pricing used to work in Ethereum before EIP-1559). However, I'd think a prediction of what a reasonable value for parameter should be, is something that could easily be included in external tooling.

I do follow your point that, overall, it may decrease the UX to the point that it is discutable whether this setting offers a significant enough advantage to warrant its existence. I have no strong preference on keeping this in or simply removing it.

tmpolaczyk: So I don't see any problem with that config parameter because it is optional, so miners aren't forced to choose a value, and also because we cannot forbid miners from implementing custom rules to decide which transactions to include into blocks.

It is true that it is a very bad user experience to have a transaction delayed or rejected for an unknown reason, but that's the nature of decentralized blockchains.

I can try to add another config parameter related to witnessing risk management, to allow node operators to choose an acceptable reward-collateral ratio before committing to a data request. Keep in mind that these config parameters can be implemented without breaking changes, so we can always easily implement them in the future.

data_structures/src/error.rs
```diff
@@ -278,6 +278,14 @@ pub enum TransactionError {
         max_weight: u32,
         dr_output: Box<DataRequestOutput>,
     },
+    #[fail(
+        display = "Data request reward to collateral percentage is {}, but must be greater than {}",
```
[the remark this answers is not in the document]
tmpolaczyk: Actually it's not a percentage, just a ratio, so that message was wrong.

witnet.toml
aesedepece: > Keep in mind that these config parameters can be implemented without breaking changes, so we can always easily implement them in the future.

Is it OK if we ditch these setting for now and then introduce it if there's such a clear feature request from the node operators community? At that point we will also have more info on how the fixed ratio plays out, so as to make a better decision.

tmpolaczyk: I don't see any problem with keeping it, but I pushed a commit that removes it.

drcpu-github: Is that a temporary commit? Right now this branch contains unused functions / variables / tests whichs makes it somewhat hard to understand?

tmpolaczyk: No, it is not temporary. Which functions are unused? The compiler should have complained about them.

drcpu-github: `set_minimum_reward_collateral_ratio` is not used anywhere anymore except in tests? I guess I'm a bit confused why we still use the `minimum_reward_collateral_ratio` variable when we could just use `required_reward_collateral_ratio`? Seems cleaner to me to reformat it?

tmpolaczyk: You are right, removed `set_minimum_reward_collateral_ratio` and its tests in the latest commit.

aesedepece (changes requested):

data_structures/src/error.rs
```diff
@@ -278,6 +278,14 @@ pub enum TransactionError {
         max_weight: u32,
         dr_output: Box<DataRequestOutput>,
     },
+    #[fail(
+        display = "The reward-to-collateral ratio for this data request is {}, but must be greater than {}",
+        reward_collateral_ratio, required_reward_collateral_ratio
+    )]
```
aesedepece, line 284: As the value printed here is the ratio, and not the percentage, then the threshold works the other way:
```suggestion
    #[fail(
        display = "The reward-to-collateral ratio for this data request is {}, but must be equal or less than {}",
        reward_collateral_ratio, required_reward_collateral_ratio
    )]
```

aesedepece (changes requested):

aesedepece (approved):

PR #2342, opened 2023-01-26
Refactor update_utxo_diff function

tmpolaczyk:
* Add tests
* Make inputs generic iterators, to avoid calls to collect
* Avoid unneeded comparison of first pkh with itself
* Stop pkh comparison and block number calculation once we know that the inputs have more than one different pkh

### DIFF
--- a/validations/src/validations.rs
+++ b/validations/src/validations.rs
@@ -1266,31 +1266,46 @@ fn increment_witnesses_counter<S: ::std::hash::BuildHasher>(
         .current += 1;
 }
 
-/// Update UTXO diff with the provided inputs and outputs
-pub fn update_utxo_diff(
+/// Update UTXO diff with the provided inputs and outputs.
+///
+/// If all the inputs have the same PKH, the outputs with that PKH will keep the maximum
+/// (most recent) block number of the inputs. Outputs with different PKH have the block number reset.
+/// And in case one of the inputs has a different PKH from another input, the block number of all the outputs is reset.
+pub fn update_utxo_diff<'a, IterInputs, IterOutputs>(
     utxo_diff: &mut UtxoDiff<'_>,
-    inputs: Vec<&Input>,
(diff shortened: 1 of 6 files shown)


validations/src/validations.rs
```diff
@@ -1266,32 +1266,47 @@ fn increment_witnesses_counter<S: ::std::hash::BuildHasher>(
...
+
     for input in inputs {
-        // Obtain the OuputPointer of each input and remove it from the utxo_diff
+        // Obtain the OutputPointer of each input and remove it from the utxo_diff
         let output_pointer = input.output_pointer();
 
-        // Returns the input PKH in case that all PKHs are the same
-        if input_pkh != utxo_diff.get(output_pointer).map(|vt| vt.pkh) {
+        if !first && input_pkh.is_none() {
+            // No need to check PKH and block number, there is more than one distinct input PKH
```
aesedepece, line 1292: :smiling_face_with_tear: 
```suggestion
        if !first && input_pkh.is_none() {
            continue;
            // No need to check PKH and block number, there is more than one distinct input PKH
```

aesedepece:
Cool way to optimize the whole thing, btw. Apparently we have been fighting with the unneeded `collect` for 3 years.

validations/src/validations.rs
tmpolaczyk: I had tried this, it breaks the code because there is a line below the if statement: `utxo_diff.remove_utxo(*output_pointer);` that must always be called. And it cannot be called before the if statement because then `utxo_diff.get` won't work.

aesedepece: > I had tried this, it breaks the code because there is a line below the if statement: `utxo_diff.remove_utxo(*output_pointer);` that must always be called. And it cannot be called before the if statement because then `utxo_diff.get` won't work.

Oh I didn't see that one on the diff. My bad!

PR #2347, opened 2023-02-02
Implement WIP0027: change value of collateral_age

tmpolaczyk:


config/src/defaults.rs
```diff
@@ -451,6 +451,10 @@ pub trait Defaults {
     }
 }
 
+/// Modified value of callateral_age after WIP0027 (1 week)
+// TODO: modify the value directly in ConsensusConstants
+pub const CONSENSUS_CONSTANTS_COLLATERAL_AGE: u32 = 13440;
```
aesedepece, line 456:
```suggestion
pub const CONSENSUS_CONSTANTS_WIP0027_COLLATERAL_AGE: u32 = 13440;
```

aesedepece (changes requested):
I'd prefer the name of the constant to clearly state that it's the post-wip0027 version, just to prevent future wrong imports.

aesedepece (approved):

PR #2349, opened 2023-02-03
feat(bridge): implement report_result_max_network_gas_price_ratio

tmpolaczyk:


bridges/centralized-ethereum/src/actors/dr_reporter.rs
```diff
@@ -32,6 +35,8 @@ pub struct DrReporter {
     pub eth_confirmation_timeout_ms: u64,
     /// Number of block confirmations needed to assume finality when sending transactions to ethereum
     pub num_confirmations: usize,
+    /// Max ratio between the gas price recommended by the provider and the gas price of the requests in the WRB
```
aesedepece, line 38:
```suggestion
    /// Max ratio between the gas price recommended by the provider and the gas price of the requests in the WRB
    /// That is, the bridge will refrain from paying more than these times the gas price originally set forth by the requesters. 
```

bridges/centralized-ethereum/src/actors/dr_reporter.rs
```diff
@@ -162,10 +178,23 @@ impl Handler<DrReporterMsg> for DrReporter {
...
+                    return;
+                }
+            };
+            let ratio = u256_div_as_f64(network_gas_price, report_gas_price);
+            if ratio <= report_result_max_network_gas_price_ratio {
+                // If not higher than the allowed ratio, set gas price
+                report_gas_price = network_gas_price;
+            } else {
+                // Higher network gas price: return error and abort
+                log::error!("Network gas price is higher than allowed ratio. Ratio: {}. Network gas price: {}. Requests gas price: {}", ratio, network_gas_price, report_gas_price);
```
aesedepece, line 197: @guidiaz Are we sure that we want to skip the reporting instead of using `report_gas_price * ratio` ?

bridges/centralized-ethereum/src/config.rs
```diff
@@ -52,12 +52,19 @@ pub struct Config {
     pub num_confirmations: usize,
     /// Miner fee for the witnet data request transactions, in nanowits
     pub dr_fee_nanowits: u64,
+    /// # Max ratio between the gas price recommended by the provider and the gas price of the requests in the WRB
```
aesedepece, line 55:
```suggestion
    /// Max ratio between the gas price recommended by the provider and the gas price of the requests in the WRB
    /// That is, the bridge will refrain from paying more than these times the gas price originally set forth by the requesters. 
```

witnet_centralized_ethereum_bridge.toml
```diff
@@ -45,6 +45,9 @@ num_confirmations = 1
 # Miner fee for the witnet data request transactions, in nanowits
 dr_fee_nanowits = 10_000
 
+# Max ratio between the gas price recommended by the provider and the gas price of the requests in the WRB
```
aesedepece, line 48:
```suggestion
# Max ratio between the gas price recommended by the provider and the gas price of the requests in the WRB
# That is, the bridge will refrain from paying more than these times the gas price originally set forth by the requesters. 
```

aesedepece (changes requested):

bridges/centralized-ethereum/src/actors/dr_reporter.rs
guidiaz: if we did so the "eth_estimateGas" would possibly fail in some providers ... we could try, but then we should still log a warning message, so we can be noticed when reading the logs what's actually happening

aesedepece: Agreed. Let's do that @tmpolaczyk

aesedepece (approved):

guidiaz (approved):

Issue #2355, opened 2023-03-21
Export / import chain state snapshots


aesedepece:
Allow exporting a snapshot of the chain state, and importing it back into a different node, with a view to near-immediate synchronization.

Tommytrg:
Closed by #2356

PR #2356, opened 2023-03-21
Basic implementation for exporting and importing chain snapshots

aesedepece:
### Already implemented
- [x] Running a JSON-RPC command that will export a chain snapshot into a local file
- [x] Loading a chain snapshot when starting a node with the ``--snapshot-import <path>` flag
- [x] Refuse to import a snapshot if our local chain is ahead of the snapshot
- [x] Allow forcing the import with `--force` flag (e.g. for faster rewinding)

~~(Be careful! it'll take some 60GB of RAM and almost the same disk space!)~~

### Missing
- [x] Trigger an export through a flag
- [x] Trigger an import through JSON-RPC
- [x] Derive own UTXOs database when importing (no need to go over every single block and transaction, just iterate over the current UTXO set trying to find PKH matches)
- [x] Exclude `own_utxos` from `ChainState` before exporting

### To improve
- [x] Speed up the importing process by leveraging `futures::future::future_all`, which enables concurrency of side-effected `async` functions (e.g. IO), and has already proven to make the exporting process much faster
- [x] Import and deserialize on the fly, with the help of `bincode::deserialize_from` consuming a file reader directly
- [x] Reduce exporting time somehow (not sure if possible, but maybe something like a )
- [x] Cut down memory footprint everywhere. Probably the best way will be to write blocks as we read them into separate export files that contain chunks of 50K blocks or so. Another approach is to encode and write different structures separately into the same file, and use markers for their bytesize and type so that the deserializer knows what to do with them.
- [x] `cargo clippy`
- [x] Simplify logging for exporting

### Left for future iterations

- My idea is to make a follow-up PR introducing some scripts for downloading the snapshots from AWS S3, torrent or whatever. The docker image will use those scripts by default, pointing to a signed snapshot for which we can verify its hash and PGP signature. Every major release should point to a new snapshot URL, hash and signature.

Solve #2355


drcpu-github:
What is the advantage of this over simply copying a storage directory and overwriting the master key (once) in storage?

aesedepece:
> What is the advantage of this over simply copying a storage directory and overwriting the master key (once) in storage?

Both approaches are rather similar, but this one is aimed to produce a cleaner output with no PII that is also more portable and compact. The end goal is to publish chain snapshots once in a while so that people can download them and then be synced almost immediately. I wouldn't share my own storage directory, even if I had stripped down my keys from it somehow.

drcpu-github:
> more portable and compact

Meaning you can further compress the RocksDB database? That would be good.

> I wouldn't share my own storage directory, even if I had stripped down my keys from it somehow.

Me neither, but syncing a node that never saves it private key and does not mine is relatively straightforward.

I'm just nitpicking a bit because this seems like a pretty involved technique with significant hardware resource requirements.

aesedepece:
> Meaning you can further compress the RocksDB database? That would be good.

The sheer chain state + superblocks + blocks backup amounts to 35.9GB before compression.

> Me neither, but syncing a node that never saves it private key and does not mine is relatively straightforward.

That's true. What really makes me lean towards this approach is just getting rid of all the database fluff, such as reverted blocks and other stuff that can get persisted at some point but doesn't really belong to the chain state.

> I'm just nitpicking a bit because this seems like a pretty involved technique with significant hardware resource requirements.

Next commits will hugely reduce the memory footprint. Putting blocks batches into different files and making the export / import processes concurrent totally did the trick.

aesedepece:
> The sheer chain state + superblocks + blocks backup amounts to 35.9GB before compression.

Compression brings it down to:
- Zip: 22.8GB
- Tar.gz: 23.0GB
- 7z: 16.8GB

aesedepece:
As of the latest commit, getting a fully synchronized node from a recent chain snapshot takes 45-60 minutes. Not bad!

I'm now trying to cut down that time by optimizing how blocks persistence is processed. I believe it can be sped up somehow. I'll keep you posted!

PR #2363, opened 2023-04-28
feat(node): discard late commits

drcpu-github:
At the moment, the witnet node will save commits and reveals which arrive over the halfway point of an epoch into a temporary vector for validation at the start of the next epoch. However, since commits are only valid in the current epoch, we can simply discard them if they arrive too late.

I have considered whether this should be handled as a warning or an error, but since there is nothing inherently wrong with the commits, I think having the node print out warnings or errors is inappropriate.

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -283,7 +283,7 @@ pub struct Witnessing {
 }
 
 /// Available storage backends
-#[derive(Debug, Default, Deserialize, Serialize, Clone, Default, PartialEq, Eq)]
+#[derive(Debug, Default, Deserialize, Serialize, Clone, PartialEq, Eq)]
 pub enum StorageBackend {
     #[serde(rename = "hashmap")]
     HashMap,
--- a/data_structures/src/chain/mod.rs
+++ b/data_structures/src/chain/mod.rs
@@ -81,7 +81,7 @@ pub struct ChainInfo {
 }
 
 /// State machine for the synchronization status of a Witnet node
-#[derive(Copy, Clone, Default, Deserialize, Debug, Default, Eq, PartialEq, Serialize)]
+#[derive(Copy, Clone, Default, Deserialize, Debug, Eq, PartialEq, Serialize)]
 pub enum StateMachine {
     /// First state, ChainManager is waiting for reaching  consensus between its peers
     #[default]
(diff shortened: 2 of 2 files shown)


config/src/config.rs
```diff
@@ -283,7 +283,7 @@ pub struct Witnessing {
 }
 
 /// Available storage backends
-#[derive(Debug, Default, Deserialize, Serialize, Clone, Default, PartialEq, Eq)]
+#[derive(Debug, Default, Deserialize, Serialize, Clone, PartialEq, Eq)]
```
aesedepece, line 286: Thanks for clean after me :clown_face:

node/src/actors/chain_manager/mod.rs
```diff
@@ -1405,7 +1405,10 @@ impl ChainManager {
                     .unwrap();
 
                 if timestamp_now > timestamp_mining {
-                    self.temp_commits_and_reveals.push(msg.transaction);
+                    // Only reveals are saved since commits which arrive too late will be invalid in the next epoch
+                    if let Transaction::Reveal(_) = &msg.transaction {
+                        self.temp_reveals.push(msg.transaction);
+                    }
```
aesedepece, line 1411: And in that case:
```suggestion
                    if let Transaction::Reveal(reveal) = &msg.transaction {
                        self.temp_reveals.push(reveal);
                    }
```

node/src/actors/chain_manager/mod.rs
```diff
@@ -229,7 +229,7 @@ pub struct ChainManager {
     /// List of superblock votes received while we are synchronizing
     temp_superblock_votes: HashSet<SuperBlockVote>,
     /// Commits and reveals to process later
-    temp_commits_and_reveals: Vec<Transaction>,
+    temp_reveals: Vec<Transaction>,
```
aesedepece, line 232: This can be made `Vec<RevealTransaction>` to enforce the behavior of only storing reveal transactions:
```suggestion
    temp_reveals: Vec<RevealTransaction>,
```

node/src/actors/chain_manager/handlers.rs
```diff
@@ -221,9 +221,9 @@ impl Handler<EpochNotification<EveryEpochPayload>> for ChainManager {
             }
         }
 
-        // After block consolidation, commits and reveals that arrive in an incorrect moment
+        // After block consolidation, reveals that arrived in an incorrect moment
         // are processed now
-        for transaction in self.temp_commits_and_reveals.drain(..) {
+        for transaction in self.temp_reveals.drain(..) {
```
aesedepece, line 226: See below on why this suggestion:
```suggestion
        for reval in self.temp_reveals.drain(..) {
            ctx.notify(AddTransaction {
                transaction: Transaction::Reveal(reveal)
```

aesedepece (changes requested):

aesedepece (approved):
Thanks gentleman.

PR #2364, opened 2023-05-04
Tapi cleanup

drcpu-github:
The first commit removes the hardcoded WIP overwrites, this can definitely be merged.

The second one removes the voting structures for past accepted WIPs, but I'm not sure whether the accepted method is to remove them now or only when new WIPs are proposed.

The `test_initialize_wip_information` test fails due to the removal of accepted WIPs, but it will fail whenever anything is updated with respect to the `BitVotesCounter` since it is specific to one set of WIPs. Not quite sure if you want to keep that test or update it every time or ...?

### DIFF
--- a/data_structures/src/chain/tapi.rs
+++ b/data_structures/src/chain/tapi.rs
@@ -272,17 +272,6 @@ impl TapiEngine {
             }
         }
 
-        // Reset the active WIPs to those statically defined in `wip_info()` and only enables those
-        // that have been marked as `old_wips` above, thus forgetting about accidentally enabled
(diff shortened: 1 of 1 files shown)


aesedepece:
> The second one removes the voting structures for past accepted WIPs, but I'm not sure whether the accepted method is to remove them now or only when new WIPs are proposed.

We usually keep it as is until a new TAPI upgrade comes.

drcpu-github:
> We usually keep it as is until a new TAPI upgrade comes.

Okay. Removed the second commit.

Issue #2367, opened 2023-05-08
Nodes that are fast-synced from snapshots will rewind to genesis if there is a rollback


aesedepece:
I know @drcpu-github faced this issue in which nodes that are fast-synced from snapshots will rewind to genesis if there is a rollback, especially that happens right after the node completes synchronization. This needs a fix.

drcpu-github:
I found another issue. After importing a snapshot and getting it to a stable chain, I sometimes see a UTXO consolidation error. I have only seen it twice or so over many testnets, so it's pretty hard to reproduce.

```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `Some((ValueTransferOutput { pkh: PublicKeyHash { hash: [254, 162, 124, 52, 222, 228, 73, 29, 60, 44, 202, 154, 232, 97, 47, 174, 124, 122, 150, 87] }, value: 1099511627776, time_lock: 0 }, 0))`,
 right: `None`: Tried to consolidate an UTXO that was already consolidated', data_structures/src/utxo_pool.rs:96:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[2023-05-08T20:10:00Z ERROR witnet_node::utils] Panic in ChainManager, shutting down system
```

This results in a node crash. Depending on how you started the node, this may result in a restart. In my case, I tend to restart them using a `while true` loop with the `import-snapshot` statement and `force` flag (because without it, the import statement seemed to not work) resulting in an overwrite of the database. Regardless of the restart issue, this seems like a bug worth root causing and solving.

PR #2369, opened 2023-05-15
Fix errors in snapshot importing functionality

aesedepece:
Previously, own UTXOs were mistakenly being tagged as unusable by inserting their value as their timeout.

Part of #2367, but not quite fixing it.


node/src/actors/chain_manager/actor.rs
```diff
@@ -875,7 +875,7 @@ impl ChainManager {
         self.chain_state.unspent_outputs_pool.visit_with_pkh(
             own_pkh,
             |(output_pointer, (vto, _))| {
```
drcpu-github, line 877: ```|(output_pointer, _)|```

aesedepece: I'd better not commit on holidays :sweat_smile:

PR #2373, opened 2023-05-22
feat(rad): decode JSON more efficiently

### LINKED ISSUE #2315: RADON parses JSON integers as floats, rounding them to a different value
So operators like StringParseJSONArray can only parse integers correctly up to MAX_SAFE_INTEGER, which is 2^53-1.

The following test shows how 2^53+1 is parsed as 2^53:

```rust
#[test]
fn test_parse_json_array_max_safe_integer_plus_2() {
    use crate::types::array::RadonArray;
    use crate::types::integer::RadonInteger;
    use crate::types::string::RadonString;

    // 2^53 + 1
    let input = RadonTypes::from(RadonString::from(r#"[9007199254740993]"#));
    let script = vec![(RadonOpCodes::StringParseJSONArray, None)];

    let output = execute_contextfree_radon_script(input, &script).unwrap();

    assert_eq!(
        output,
        RadonTypes::from(RadonArray::from(vec![RadonTypes::from(RadonInteger::from(
            9007199254740993
        ))]))
    );
}
```

Output:

```
thread 'script::tests::test_parse_json_array_biggish_integers' panicked at 'assertion failed: `(left == right)`
  left: `Array(RadonArray { value: [Integer(RadonInteger { value: 9007199254740992 })], is_homogeneous: true })`,
 right: `Array(RadonArray { value: [Integer(RadonInteger { value: 9007199254740993 })], is_homogeneous: true })`', rad/src/script.rs:607:9
```

### LINKED ISSUE #2338: RUSTSEC-2022-0081: json is unmaintained
> json is unmaintained

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Status              | unmaintained                |
| Package             | `json`                      |
| Version             | `0.12.4`                   |
| URL                 | [https://github.com/maciejhirsz/json-rust/issues/205](https://github.com/maciejhirsz/json-rust/issues/205) |
| Date                | 2022-02-01                         |

Last release was almost 3 years ago.

The maintainer is unresponsive with outstanding issues.

One of the outstanding issues include [a possible soundness issue](https://github.com/maciejhirsz/json-rust/issues/196).

## Possible Alternative(s)

The below list has not been vetted in any way and may or may not contain alternatives;

- [serde_json](https://crates.io/crates/serde_json)
- [json-deserializer](https://crates.io/crates/json-deserializer)
- [simd-json](https://crates.io/crates/simd-json)

See [advisory page](https://rustsec.org/advisories/RUSTSEC-2022-0081.html) for additional details.

aesedepece:
This should solve the stoppers for https://github.com/witnet/witnet-price-feeds/issues/210#issuecomment-1557134728.

This change may require activation through TAPI, provided that non-updated nodes will commit errors where updated nodes will report normally.

fix #2315 
fix #2338 

@Tommytrg Edit:
Also fix #2373

@aesedepece Edit:
Lol, it fixes itself?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2118,12 +2118,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "json"
-version = "0.12.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "078e285eafdfb6c4b434e0d31e8cfcb5115b651496faca5749b88fafd4f23bfd"
-
-[[package]]
 name = "jsonrpc-core"
 version = "10.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -5803,7 +5797,6 @@ dependencies = [
  "hex",
  "http",
  "if_rust_version",
- "json",
  "log 0.4.19",
  "minidom",
  "num_enum",
@@ -5811,6 +5804,7 @@ dependencies = [
  "rand 0.7.3",
  "serde",
  "serde_cbor",
+ "serde_json",
  "url",
  "witnet_config",
  "witnet_crypto",
--- a/data_structures/src/chain/tapi.rs
+++ b/data_structures/src/chain/tapi.rs
@@ -138,7 +138,7 @@ impl TapiEngine {
             }
         }
         for n in 0..self.bit_tapi_counter.len() {
-            if let Some(mut bit_counter) = self.bit_tapi_counter.get_mut(n, &epoch_to_update) {
+            if let Some(bit_counter) = self.bit_tapi_counter.get_mut(n, &epoch_to_update) {
                 if !self.wip_activation.contains_key(&bit_counter.wip)
                     && !avoid_wip_list.contains(&bit_counter.wip)
                 {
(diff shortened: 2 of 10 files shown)


aesedepece:
Nodes that implement this change will commit different values than those who do not. Namely, implementors may commit values where others may commit errors.

While this can be seen as a breaking change, strictly speaking, changes to witnessing behavior like this can't really jeopardize consensus or data integrity — provided that the affected data requests are already being tallied to an error. 

That is to say that we should be fine releasing this without a TAPI guard because it will not make data requests start to fail but rather make them end up with a different error for as long as there's no majority of implementors in the witnessing committee, to eventually make it succeed with a value when the implementors are in majority.

What do you think @witnet/devs?

drcpu-github:
I think under the assumptions that you mention, it is fine to release this without TAPI.

As long as a change does not influence validation or consensus logic and whether someone implements this change or not (to use WIP terms) has no implications to network liveness and / or correctness, it does not require TAPI. This does brings up the question whether something like [the redirect WIP](https://github.com/witnet/WIPs/blob/master/wip-0025.md) was really necessary and probably requires updating the WIP rules or a stricter compliance to them.

aesedepece:
I'm aiming at shipping this together with #2385 in an eventual 1.6.6 release.

PR #2381, opened 2023-06-09
chore: upgrade from rocksdb 0.13 to 0.21

aesedepece:
Beyond the many new APIs that we can use with this new version, this also fixes compatibility with `gcc >= 13.0`.


aesedepece:
Ok, we have an impediment here: GitHub Actions only supports Ubuntu 22.04, which locks us to `librocksdb-dev` 6.11, while crate `rocksdb = 0.21.0` needs 8.1.

aesedepece:
Until we have a newer Ubuntu on GitHub Actions, the higher we can go with the `rocksdb` crate is probably `0.15`  (uses 6.11) or if we are lucky enough, `0.18` (uses 6.28)

aesedepece:
0.18.0 didn't work — trying with 0.15.0

aesedepece:
Maybe it'll be a better solution to install a newer version of librocksdb-dev from a ppa or something. I'll give it a try on Monday.

aesedepece:
Ok, I made it work by compiling our own RocksDB 8.1.1 separately and linking dynamically. It's yet to be seen if a similar change will be needed in the cross-compilation and releasing stages.

PR #2385, opened 2023-07-11
fix(wallet): allow generating transactions without persisting change addresses

### LINKED ISSUE #2384: Wallet is using internal addresses in a non-sequential way
I recently found out that most of my wallets have enormous gaps in their internal keychains. For example, there's this one wallet whose only internal address in use are 168, 171 and 178.

The impact of this is:
- [Mild] slightly slower sync in Sheikah
- [Severe] missing balance when the same key is imported into myWitWallet

aesedepece:
This opens up the possibility of creating transactions in a wallet without persisting change addresses, which is needed in the case of fee estimations.

Fix #2384


aesedepece:
For this functionality to actually make an impact, a similar change is needed on the Sheikah side, so the `preview: true` flag is passed to `createVTT` and `createDRT`.

PR #2394, opened 2023-09-12
fix(net): prevent panicking when finding an unknown HTTP status code

aesedepece:
This prevents a potential DoS attack to where witnessing nodes could be forced into a temporal desynchronization state and slashing could be triggered on them on purpose.

### DIFF
--- a/net/src/client/http/mod.rs
+++ b/net/src/client/http/mod.rs
@@ -199,7 +199,18 @@ impl TryFrom<WitnetHttpResponse> for surf::http::Response {
         // Create a surf response and set all the relevant parts
         // Version and headers are currently not used, but they are included here for completion and
         // future readiness.
-        let mut res = surf::http::Response::new(status);
+        let code = status.status.as_u16();
+        let mut res =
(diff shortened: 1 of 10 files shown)


aesedepece:
Thanks @guidiaz for discovering this security bug! :muscle:

net/src/client/http/mod.rs
```diff
@@ -199,7 +199,14 @@ impl TryFrom<WitnetHttpResponse> for surf::http::Response {
         // Create a surf response and set all the relevant parts
         // Version and headers are currently not used, but they are included here for completion and
         // future readiness.
-        let mut res = surf::http::Response::new(status);
+        let code = status.status.as_u16();
+        let mut res =
+            std::panic::catch_unwind(|| surf::http::Response::new(status)).map_err(|_| {
```
Tommytrg, line 204: Could we add a comment explaining that we need to use catch_unwind  because the surf library panics instead of returning a `Result`?

Tommytrg (approved):

Issue #2390, opened 2023-08-30
`getBalance` node JSONRPC method ignores `simple` positional argument


aesedepece:
https://github.com/witnet/witnet-rust/blob/ad46722c2a67ebedab679751ca17939f49d30cb6/node/src/actors/json_rpc/api.rs#L1305-L1324

Thanks to @Tommytrg for reporting.

aesedepece:
`simple = params.get(1).or(false)`

Tommytrg:
Closed by #2395

PR #2395, opened 2023-09-19
gitfix(jsonrpc): read simple positional argument in getBalance

Tommytrg:
The  initial suggestion was:
```rs
  simple = params.get(1).unwrap_or(false);
```
But, we also have to call `parse` over the possible `String` so the one-liner would be:
```rs
  simple = params.get(1).and_then(|v| v.as_str().map(|s| s.parse().unwrap_or(false))).unwrap_or(false);
```

As the previous line looks a bit hard to read, I have used an `if let` which seems more readable.

### DIFF
--- a/node/src/actors/json_rpc/api.rs
+++ b/node/src/actors/json_rpc/api.rs
@@ -1316,7 +1316,7 @@ pub async fn get_balance(params: Params) -> JsonRpcResult {
             ));
         };
 
-        simple = false;
+        simple = params.get(1).and_then(Value::as_bool).unwrap_or(false);
     } else {
         let params: GetBalanceParams = params.parse()?;
         pkh = params.pkh;


aesedepece:
> ```rust
>   simple = params.get(1).and_then(|v| v.as_str().map(|s| s.parse().unwrap_or(false))).unwrap_or(false);
> ```
> 
> As the previous line looks a bit hard to read, I have used an `if let` which seems more readable.

It doesn't look that bad if you leverage the `and_then` magic and parse from `Value` instead of `str`:
```rust
simple = params.get(1).and_then(Value::as_bool).unwrap_or(false);
```

Tommytrg:
close #2390

aesedepece (approved):

Tommytrg:
close #2390

PR #2402, opened 2023-10-10
feat: implement HttpHead as new RADType

guidiaz:


rad/src/lib.rs
```diff
@@ -298,13 +301,21 @@ async fn http_response(
...
-    body.read_to_string(&mut response_string)
-        .await
-        .map_err(|x| RadError::HttpOther {
-            message: x.to_string(),
-        })?;
+
+    let (parts, mut body) = response.into_parts();
+    match retrieve.kind {
+        RADType::HttpHead => {
+            response_string = format!("{:?}", parts.headers);
```
aesedepece, line 309: Because we are relying on `impl Display for http::header::HeaderMap` here, I think we could use a test to ensure that a future version of the `http` library with a different `impl` doesn't break consensus.

aesedepece: Forget about that! You made one down below! :rocket:

rad/src/lib.rs
```diff
@@ -829,6 +841,40 @@ mod tests {
...
+                Value::Integer(RadonOpCodes::MapGetString as i128),
+                Value::Text("etag".to_string()),
+            ]),
+        ]);
+        let packed_script_r = serde_cbor::to_vec(&script_r).unwrap();
+        println!("{:?}", packed_script_r);
+
+        let retrieve = RADRetrieve {
+            kind: RADType::HttpHead,
+            url: "https://witnet.io/_nuxt/img/dragon_reading.a37f8cb.png".to_string(),
```
aesedepece, line 858: I believe this file will disappear as soon as we make any little change on the website, because Nuxt uses randomized codes in file paths upon rebuilds to deal with caching.

This may be a more future-proof URL: https://en.wikipedia.org/static/images/icons/wikipedia.png

rad/src/lib.rs
```diff
@@ -1526,6 +1572,58 @@ mod tests {
...
+        let headers = vec![("ñ", "value")];
+        let headers = headers
+            .into_iter()
+            .map(|(a, b)| (a.to_string(), b.to_string()))
+            .collect();
+        let request = RADRequest {
+            time_lock: 0,
+            retrieve: vec![RADRetrieve {
+                kind: RADType::HttpGet,
+                url: String::from("http://127.0.0.1"),
```
aesedepece, line 1589: Hmmm I don't think this test can be run by default. For any context without an actual server at `127.0.0.1:80` it might fail with a TCP connection error instead of an invalid HTTP header error.

If that's right, my suggestion is to disable this test by default with the `#[ignore]` flag below `#[test]`. Then, when you want to run it on purpose, you only need to append `--ignored` to your `cargo test` command.

rad/src/lib.rs
guidiaz: Okay. Although this test, and the others on which it's inspired, don't actually perform any HTTP request. They do build a proper a RADRetrieve struct, but then a HTTP response is hardcoded afterwards.

rad/src/lib.rs
guidiaz: All `test_try_data_request_http_*` tests, including this one, do run by default, because they're expecting some specific error to be thrown from the http lib (isahc) before an HTTP request is actually sent.

aesedepece:
This is more than approved, but we need a major release for it because it's a breaking change, so we need to wait for 2.1.0.

PR #2442, opened 2024-03-18
Upgrade witnet-rust docker image base (ubuntu)

aeweda:


docker/witnet-rust/Dockerfile
```diff
@@ -1,14 +1,14 @@
...
+RUN apt update && apt install -y --no-install-recommends \
     ca-certificates \
     curl \
-    netcat \
+    netcat-traditional \
     jq
 
 # Clean up apt packages so the docker image is as compact as possible
-RUN apt-get clean && apt-get autoremove
+RUN apt clean && apt autoremove
```
aesedepece, line 11: It is customary in the GNU world not to use `apt` in scripts. This is explained in its [man page](https://manpages.debian.org/buster/apt/apt.8#SCRIPT_USAGE_AND_DIFFERENCES_FROM_OTHER_APT_TOOLS) and the command will even complain about it when it detects that the current `tty` isn't interactive.

aesedepece (changes requested):

aesedepece (approved):

aesedepece:
Merged in as 0e1be39b3b0956656a8f364f0bba1fbeab368b50

PR #2444, opened 2024-03-22
feat(validations): raise a transaction error when too many witnesses are requested

drcpu-github:
This PR serves as a proof-of-concept of how we could validate that a data request does not request too many witnesses. The threshold is hard-coded to 1/4th of the amount of validators.

It is probably more convenient to just pass the validator count rather than the full Stakes Tracker. However, I wanted to have this PR serve as a test on how to pass the Stakes Tracker as an argument since that will be required for other validation functions.

The JSON-RPC CLI and wallet cannot really access the Stakes Tracker, so they pass it a None value. If the implementation were to be simplified to pass the validator count as a single number, they could presumably query it from the node before validating the data request.

### DIFF
--- a/data_structures/src/error.rs
+++ b/data_structures/src/error.rs
@@ -74,6 +74,14 @@ pub enum TransactionError {
     #[fail(display = "Data Request witnesses number must be greater than zero")]
     InsufficientWitnesses,
     #[fail(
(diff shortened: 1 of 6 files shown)


aesedepece:
I guess this can be closed in favor of #2446

PR #2446, opened 2024-04-13
TooManyWitnesses RAD error tally

drcpu-github:
This is an alternative implemenation to #2444 where instead of refusing to build a block with a data request requiring too many witnesses, the data request is included by a miner together with a tally. The tally has a TooManyWitnesses Radon error as result.

This is potentially more robust because of two reasons:
1) It won't keep invalid data requests in the transaction pool
2) By creating a tally the (decentralized) bridge can report an actual tally result rather than make something up


aesedepece:
Merged as 95f97ee1e5aee1ebdb7931eb884eb663c20b7da7

PR #2448, opened 2024-04-15
Track the most recently active validators and sample the superblock committee from them

drcpu-github:


aesedepece:
I'll give this a thought on how to best integrate on top of wit/2

drcpu-github:
> I'll give this a thought on how to best integrate on top of wit/2

IMHO, this is one of the many small things that is required to strengthen wit/2 against griefing attacks. I'm by no means saying the implementation is the best approach, but the concept is needed.

PR #2461, opened 2024-05-06
fix(mining): fix consolidating the genesis block for which the current epoch is still None

drcpu-github:


node/src/actors/chain_manager/mod.rs
```diff
@@ -928,8 +928,22 @@ impl ChainManager {
...
+                        return;
+                    }
+                }
+                None => {
+                    log::error!("Current epoch not loaded in ChainManager");
+                    return;
+                }
+            };
+
+            epoch
```
aesedepece, line 946: Let's abuse the magic of Rust to avoid some code repetition:
```suggestion
            // If there is no epoch set, it's because the chain is yet to be bootstrapped, or because of a data race
            match self.chain_state.chain_info.as_ref() {
                // If the chain is yet to be bootstrapped (the block we are processing is the genesis block), set the epoch to zero
                Some(chain_info) if chain_info.consensus_constants.genesis_hash == block.hash() => {
                    0
                }
                // In case of data race, shortcut the function
                _ => {
                    log::error!("Current epoch not loaded in ChainManager");
                    return;
                }
            }
```

aesedepece (changes requested):

PR #2464, opened 2024-05-16
feat(cli): validate that a validator is not already active using another withdrawer

drcpu-github:


aesedepece:
I'm just as supportive of adding this right now (to comply with the current WIP) as I'm supportive of removing it in a future 2.1 protocol revision if there's community consensus on it (as it facilitates delegation of stake).

The reason why I'm not pushing for dropping this restriction altogether is because allowing validators to operate on 3rd party stake is probably not that attractive anyway if onchain incentives are not provided. This is to say that when, at some point, we end up allowing that type of delegation, the protocol shall also allow the validators to take a fraction of the yield. If we go down that route, that percentage should be committed in the authorization message and in the stake transaction.

drcpu-github:
I am somewhat confused by this statement. Are you implying that currently it should work to stake on the same validator with different withdrawers? I tested that and that does not seems to be the case.

It could be an observation bug with the `query-stakes` method rather than one in the staking logic itself. If I stake two times 10k on the same validator with two different withdrawers that seems to work (the network allows it) and I can see that both withdrawers have 10k staked using `query-stakes`. However, if I query the validator using `query-stakes`, there is only 10k balance.

I assumed this scenario was completely unsupported and hence the necessity for this validation (which also needs to happen at the block level).

drcpu-github:
Superseded by https://github.com/witnet/witnet-rust/pull/2472/commits/059b6aa4a30b64cc9b30a573306e7d263073c00a.

Issue #2467, opened 2024-05-30
Stakes data structure does not correctly track multiple validators with the same withdrawer


drcpu-github:
Staking onto two validators with the same external withdraw address results in the `by_withdrawer` data stucture only tracking the latest one. However, this is actually a valid move and it is just the `by_withdrawer` map which cannot index that properly because it only has one SyncStake entry per withdrawer address and each of those entries is tied to a `<validator, withdrawer>` key.

It's basically the same bug as the one I already shared for `by_validator`, but more serious because the former is technically invalid if we implement all WIP validation rules.

aesedepece:
Solved a while ago.

PR #2469, opened 2024-05-31
fix(stakes): fix mining by a validator with a different withdrawer address

### LINKED ISSUE #2466: Validator with withdrawer address different from the validator address is never eligible to mine
A validator which has a stake for which the withdrawer address is not the validator address can never mine a block. This is due to the way his power is calculated. In `mining.rs`, a `StakeKey` is contructed where `own_pkh` is the pkh of the validator:
```
let key = StakeKey::from((own_pkh, own_pkh));
```
This `StakeKey` is later on used to query the power of the validator (which will determine its mining eligibility) using the `self.by_key` staking tracker. However, since that structure is indexed by a `StakeKey<validator, withdrawer>` no `StakeEntry` is found and this his power and eligibility is zero. 

Either the `query_power` should use the `by_validator` structure (not sure if this has other implications) or the validator should construct its `StakeKey` using the correct withdrawer as second argument (and I don't know how it can currently know the withdrawer address)?

drcpu-github:
This fixes #2466, #2467 and tackles the `by_validator` issue mentioned in #2464.

### DIFF
--- a/bridges/centralized-ethereum/src/actors/dr_sender/tests.rs
+++ b/bridges/centralized-ethereum/src/actors/dr_sender/tests.rs
@@ -4,14 +4,14 @@ use witnet_data_structures::chain::{RADAggregate, RADRequest, RADRetrieve, RADTa
 #[test]
 fn deserialize_empty_dr() {
     // An empty data request is invalid with error 0xE0: BridgeMalformedRequest
-    let err = deserialize_and_validate_dr_bytes(&[], 1).unwrap_err();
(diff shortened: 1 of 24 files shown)


drcpu-github:
@guidiaz I fixed all formatting and clippy issues and it finally gets to the tests, but some of the centralized bridge tests are now failing.

aesedepece:
This is merged as of eb9dc4a2976b6cabc82546d0c940c2adb49f2ba8, not without breaking a few things that I'll be fixing later!

PR #2472, opened 2024-06-14
A set of fixes and features for wit\2

drcpu-github:
The first four commits are part of https://github.com/witnet/witnet-rust/pull/2469.

The latter four commits are all new self-contained patches that introduce a fix or feature, but they build upon each other, so I kept them in the same branch.


node/src/actors/chain_manager/mod.rs
```diff
@@ -1015,6 +1016,43 @@ impl ChainManager {
...
+                            st_transaction_fee(st_tx, &utxo_diff, current_epoch, epoch_constants)
+                                .unwrap_or_default();
+                    }
+                    for ut_tx in &block.txns.unstake_txns {
+                        transaction_fees += ut_transaction_fee(ut_tx).unwrap_or_default();
+                    }
+
+                    let _ = stakes.add_reward(
+                        miner_pkh,
+                        Wit::from(50_000_000_000) + Wit::from(transaction_fees),
```
guidiaz, line 1051: Afaik, 50 WIT as block reward is not yet decided. Perhaps this proposal of reducing block time and reward should be part a dedicated PR, not mixed up with a potential fix or improvement in the mining elegibility algorithm.

validations/src/eligibility/current.rs
```diff
@@ -147,37 +151,41 @@ where
...
         //  the stakers"
-        let stakers = self.stakes_count();
-        let quantile = stakers / MINING_REPLICATION_FACTOR;
         // TODO: verify if defaulting to 0 makes sense
-        let (_, threshold) = rank.nth(quantile).unwrap_or_default();
-        if power <= threshold {
-            return Ok(IneligibilityReason::InsufficientPower.into());
+        let mut rank = self.rank(Capability::Mining, epoch);
+        let (_, max_power) = rank.next().unwrap_or_default();
+        let (_, threshold) = rank.nth(MINING_REPLICATION_FACTOR - 2).unwrap_or_default();
```
guidiaz, line 170: Shouldn't it be `rank.nth(MINING_REPLICATION_FACTOR -1)` instead?

node/src/actors/chain_manager/handlers.rs
```diff
@@ -178,6 +184,36 @@ impl Handler<EpochNotification<EveryEpochPayload>> for ChainManager {
...
+                                            stake_key.validator
+                                        );
+                                        let _ = stakes.reset_age(
+                                            stake_key.validator,
+                                            Capability::Mining,
+                                            msg.checkpoint,
+                                        );
+                                    }
+                                }
+                            }
```
guidiaz, line 216: In case of no valid block candidate to consolidate for previous epoch, the top MINING_REPLICATION_FACTOR eligible validators should be slashed, whereas with this code only those with at least 1 / MINING_REPLICATION_FACTOR of the highest power would actually get slashed.

guidiaz: Besides, I'd rather recommend to reset coin age in proportion to each one's rank position:
- First: reset coin age to zero
- Second: subtract half the current age
- Third: subtract one third of current age 
- And so on

This way we would avoid/minimize the chances of a colluding set of big stakers to end up being within the same eligible set on periodically base.

guidiaz (changes requested):

node/src/actors/chain_manager/mod.rs
drcpu-github: Of course, I fully agree, but I had to put something there as a place holder.

validations/src/eligibility/current.rs
drcpu-github: No, because `rank.next()` consumes element 0 of the iterator and you want the element up to `MINING_REPLICATION_FACTOR - 1`. If `rf` is 4, line 167 consumes element 0 and `nth(2)` returns and consumes up to the `nth` element starting from 0.

node/src/actors/chain_manager/handlers.rs
drcpu-github: > In case of no valid block candidate to consolidate for previous epoch, the top MINING_REPLICATION_FACTOR eligible validators should be slashed, whereas with this code only those with at least 1 / MINING_REPLICATION_FACTOR of the highest power would actually get slashed.

Yes, my initial implementation actually slashed all validator up to `rf` validators. The problem with that is that you can actually be slashing validators which were not eligible to propose a block which seems rather unfair.

For example, imagine validator 1 and 2 have a power of 100k and validators 3 and 4 have a power of 10k. The former two are guaranteed to propose a block, but for the latter two it's a matter of whether their VRF is smaller than a target hash. Hence the choice in above algorithm to not slash validators 3 and 4. I'm up for changing that though.

drcpu-github: > Besides, I'd rather recommend to reset coin age in proportion to each one's rank position:
> 
> * First: reset coin age to zero
> * Second: subtract half the current age
> * Third: subtract one third of current age
> * And so on

Assume validator 1 and 2 have the same power, do we rest both to zero? I'm not sure how deterministic the rank function is with respect to multiple distinct nodes since it seems to be sorted by power only.

> This way we would avoid/minimize the chances of a colluding set of big stakers to end up being within the same eligible set on periodically base.

I understand the point and actually agree that this might be a good defense, but I do wonder if it's worse to have a single epoch where multiple colluding attackers are executing an attack (and hence we only have a single epoch with a block missing) or several epochs where those attackers separately try to run a griefing attack (and we might have several epochs with missing blocks if network connectivity is not perfect or there is also censoring taking place).

guidiaz: > For example, imagine validator 1 and 2 have a power of 100k and validators 3 and 4 have a power of 10k. The former two are guaranteed to propose a block, but for the latter two it's a matter of whether their VRF is smaller than a target hash. Hence the choice in above algorithm to not slash validators 3 and 4. I'm up for changing that though.

According to the simulation source code that it's being replicated, with a replication factor of 4, validators 1 to 4 would indeed be eligible and expected to broadcast a valid block proposal. The VRF is just used for the other nodes to agree upon which to ultimately accept as best proposal, meaning that although validators 3 and 4 have less chances to win, they do have some, and further more, they are obliged to broadcast a block candidate. Therefore, in case of no blocks candidates in previous epoch, the 4 eligible validators should be staked, no matter if their absolute mining power.

guidiaz: > Assume validator 1 and 2 have the same power, do we rest both to zero? I'm not sure how deterministic the rank function is with respect to multiple distinct nodes since it seems to be sorted by power only.

Yes, I do believe that eligible validators coincidentally having same power, they should still be slashed in different proportions. The easiest and most convenient way to implement that would be to rely on the implementation of  the `rank` function, as long as it is deterministic (i.e. non dependent on underlying OS, crate version, or thinks alike), as I believe it must be.

guidiaz: > I understand the point and actually agree that this might be a good defense, but I do wonder if it's worse to have a single epoch where multiple colluding attackers are executing an attack (and hence we only have a single epoch with a block missing) or several epochs where those attackers separately try to run a griefing attack (and we might have several epochs with missing blocks if network connectivity is not perfect or there is also censoring taking place).

Considering that we're planning a replication factor as low as 4, I think it would be best to reduce the chances of multiple colluding attackers within same epoch. In fact, with this goal in mind, I'd rather suggest to consider a higher replication factor, like 8 or even 16, as long as we were certain that p2p gossiping doesn't grow to an unmanageable number of messages for the average node specs.

drcpu-github: > According to the simulation source code that I believe it's being replicated, with a replication factor of 4, validators 1 to 4 would indeed be eligible and expected to broadcast a valid block proposal. The VRF is just used for the other nodes to agree upon which to ultimately accept as best proposal, meaning that although validators 3 and 4 have less chances to win, they do still have some (even if higher ranked validators actually broadcast valid proposals), and further more, they are obliged to broadcast a block candidate whatsoever. Therefore, in case of no blocks candidates in previous epoch, the 4 eligible validators on that epoch should all be slashed, no matter if their absolute mining power is "low" compared to the others.

That is certainly not the case for the simulation and not how it is implemented in this PR. If a validator has more than `rf` less power, [this line](https://github.com/drcpu-github/proof-of-stake-tools/blob/2ed35795d4487b12bc6fe6bb0fedc52d18e9a6f3/simulate_block_mining.py#L149) combined with a node's VRF can result in them not proposing a block. It's not just that they have less chances to win, they will not even propose and broadcast a block.

> The easiest and most convenient way to implement that would be to rely on the implementation of the rank function, as long as it is deterministic (i.e. non dependent on underlying OS, crate version, or thinks alike), as I believe it must be.

I am simply not sure (the current implementation of rank](https://github.com/witnet/witnet-rust/blob/5ec6673c98f0a1f34553002d309d9945508a9033/data_structures/src/staking/stakes.rs#L232) is deterministic. I don't know if a BTreeMap in Rust maintains the order of insertion when iterated over or not and because we are only sorting on a single metric which can be the same for different validators, I imagine it could lead to non-determinism. Of course, that should be easily fixable by changing the line to something like this: `sorted_by_key(|(stake_key, power)| (stake_key.validator.clone(), *power))`

> In fact, with this goal in mind, I'd rather suggest to consider a higher replication factor, like 8 or even 16, as long as we were certain that p2p gossiping doesn't grow to an unmanageable number of messages for the average node specs.

I guess it's fine to increase the `rf` for the simple reason that the p2p traffic should reduce more or less exponentially. Any node will only forward a block if he considers it as a better block and hence most nodes will never see `rf` blocks. 8 feels fine to me, 16 feels a bit over the top.

guidiaz: > That is certainly not the case for the simulation and not how it is implemented in this PR. If a validator has more than `rf` less power, [this line](https://github.com/drcpu-github/proof-of-stake-tools/blob/2ed35795d4487b12bc6fe6bb0fedc52d18e9a6f3/simulate_block_mining.py#L149) combined with a node's VRF can result in them not proposing a block. It's not just that they have less chances to win, they will not even propose and broadcast a block. Whether a block wins or not ultimately comes down to the selector used in the simulation, but in this PR power wins, then VRF, then block hash.

Sorry, I think I got a little lost here. You were basically saying that the top `rf` stakers cannot be slashed because there may be circumstances under which some of them would never "dare" to propose a block. My point is that, despite of whatever was intended or actually implemented in the simulator, isn't the whole point of the replication factor to always force a determined number of identities (no more, no less) to propose block candidates based on their power rank, no matter the relative power among the top `rf` candidates??

So in your above mentioned example, you said:

> For example, imagine validator 1 and 2 have a power of 100k and validators 3 and 4 have a power of 10k. The former two are guaranteed to propose a block, but for the latter two it's a matter of whether their VRF is smaller than a target hash. Hence the choice in above algorithm to not slash validators 3 and 4. I'm up for changing that though.

But I do believe that validators 1 to 4 must all propose and broadcast valid block candidates. True, some of them may refrain from doing so in case they detect a valid block being broadcasted from a eligible candidate with a higher power (or same power but a higher vrf). But still, all `rf` candidates SHOULD be prone and ready to broadcast asap their block proposal whatsoever. Because if none of them happened to do so, they MUST all get slashed.

drcpu-github: Yes, we can certainly require that every node is eligible to build and broadcast a block regardless of the power distribution for the top `rf` candidates. Some of the patches here are a 100% literal translation from the simulation though. I don't really see any downside to removing the VRF eligibility check (and modify the slashing accordingly) because it will almost never matter.

@aesedepece Before I correct this, what's your opinion?

guidiaz: Moreover. Slashing ALL eligible candidates if no block gets consolidated on previous epoch, while NECESSARY, it would not yet be enough. As to minimize the chances of multiple identities colluding to perform a griefing attack (all such identities ranked within the top `rf` stakers), we should definitely consider a way to immediately slash reluctant candidates on the very same epoch on which the best block proposal gets consolidated. 

For instance, if the consolidated block proposal ultimately belongs to the `k`-est staker out of `rf` candidates, candidates ranked from `#1` to `#k-1` should be slashed immediately, whereas stakers `#k+1` to `rf` would not be slashed. This idea should be discussed and elaborated and eventually implemented on a different PR, of course.

drcpu-github:
Pertaining to above discussions, this is what is currently implemented in the branch up to commit [9ec8d09](https://github.com/witnet/witnet-rust/commit/9ec8d094952aed0d48202e709b7236866dd0b540):
1) All `rf` stakers are eligible to produce a block
2) Block winner comparison is power, vrf, block hash, in that order
3) If none of the eligible stakers produce a block, they will get slashed proportional to their rank
4) Validation code to make sure no one stakes on a validator with a different withdrawer address
5) A way to switch block times moving from V1_8 to V2_0
6) A bunch of small features and bugfixes

drcpu-github:
Things to (potentially) still implement (of the top of my head):
1) Multiplication for the `rf` factor when more than one block is missing
2) Move wit/2 constants to a central file
3) Better ratios for slashing validators

data_structures/src/staking/stakes.rs
```diff
@@ -294,17 +299,50 @@ where
...
+        stakes.iter_mut().for_each(|stake| {
+            let _ = stake.value.write().unwrap().add_stake(
+                distribute_coins,
+                current_epoch,
+                Some(0.into()),
+            );
+        });
+
+        Ok(remainder_coins)
+    }
```
aesedepece, line 344: This essentially implements a model where validators accrue rewards that are bound to become withdrawable by multiple addresses in proportion to the amounts staked in connection to those different withdrawer addresses.

I'm trying to go for a different model where stake entries are operated indepedently, i.e. a validators must compute eligibility separately for each of the stake entries in which they're designated as the validator. Then, a reward obtained by leveraging the power of one entry goes only to that one entry. Any other stake entries operated by the same validator should remain unaffected both in terms of rewards and penalties.

This does not only simplify rewards and withdrawals, but also paves the way for a future 2.X upgrade in which we open the door to allowing an operator to operate stake entries for multiple withdrawers, by means of simply removing the "1 operator = 1 withdrawer" constraint introduced here.

Why do I favor that approach? Because (with enough support from the community) we can pull off the trick of adding basic support for delegation in record time after 2.0 release, and with near-zero development effort.

drcpu-github: > This essentially implements a model where validators accrue rewards that are bound to become withdrawable by multiple addresses in proportion to the amounts staked in connection to those different withdrawer addresses.

Correct, that was the goal.

> I'm trying to go for a different model where stake entries are operated indepedently, i.e. a validators must compute eligibility separately for each of the stake entries in which they're designated as the validator. Then, a reward obtained by leveraging the power of one entry goes only to that one entry. Any other stake entries operated by the same validator should remain unaffected both in terms of rewards and penalties.

I disagree with the efficacy of this model for two reasons:
1) It complicates the power and rank functions. Rather than a simple sum of powers to calculate eligibility for a single validator, they now have to be split it up.
2) In my opinion, this defeats the main purpose of delegation: small stakers combining their power to operate as if they were a high-value validator. If each staker's power and eligibility are only proportional to their own stake, they don't really gain a whole lot by staking on a validator they do not control. Time to a block will stay as large as if they were validating with only their own stake on a self-owned validator and rather than a somewhat continuous stream of income (because their combined power significantly increases their eligibility and they split and share rewards), they need to wait a long time to get a single higher-value reward. In other words, this keeps pretty much the same variability issue we have in the current mining model.

> This does not only simplify rewards and withdrawals, but also paves the way for a future 2.X upgrade in which we open the door to allowing an operator to operate stake entries for multiple withdrawers, by means of simply removing the "1 operator = 1 withdrawer" constraint introduced here.

I don't see why this model is better for enabling a multi-staker, single-validator model. I think most of the infrastructure (such as a vector of stake entries in the `by_validator` tracker structure) is already in place.

aesedepece: > 1. It complicates the power and rank functions. Rather than a simple sum of powers to calculate eligibility for a single validator, they now have to be split it up.

It does not complicate such functions at all. Each stake entry in `by_key` has its own power and rank. Each entry has its own eligibility, and they are operated independently. There's no need for adding up stakes or anything like that because under that model there's no such thing as "a validator's eligibillity" but rather "a stake entry's eligibility". The only downside is that each validator must compute separate VRFs for each of the withdrawers.

> 2. In my opinion, this defeats the main purpose of delegation: small stakers combining their power to operate as if they were a high-value validator. If each staker's power and eligibility are only proportional to their own stake, they don't really gain a whole lot by staking on a validator they do not control. Time to a block will stay as large as if they were validating with only their own stake on a self-owned validator and rather than a somewhat continuous stream of income (because their combined power significantly increases their eligibility and they split and share rewards), they need to wait a long time to get a single higher-value reward. In other words, this keeps pretty much the same variability issue we have in the current mining model.

I'm assuming that the kind of PoS that we're implementing is fairly linear, so as you have described, the main point of pooling stake is to increase the frequency of rewards (even if long term the total amount of rewards is the same).

I push for this "independent stake entries" model because I can see a wide demographic of people in our community who will probably pass on running their own validator, but would be willing to delegate their coins into other community members' validators, knowing that they can't really steal their coins. At the same time, as I hinted before, I'm not pushing for supporting delegation just right now, but in a future update, once the wit/2 network is fully bootstrapped.

> I don't see why this model is better for enabling a multi-staker, single-validator model. I think most of the infrastructure (such as a vector of stake entries in the `by_validator` tracker structure) is already in place.

I see a clear advantage when it comes to rewards and withdrawals. Not having to split rewards and handle remainders like above is pretty convenient.

drcpu-github: > It does not complicate such functions at all. Each stake entry in by_key has its own power and rank. Each entry has its own eligibility, and they are operated independently. There's no need for adding up stakes or anything like that because under that model there's no such thing as "a validator's eligibillity" but rather "a stake entry's eligibility". The only downside is that each validator must compute separate VRFs for each of the withdrawers.

Yes, you are right. I was still reasoning from a `by_validator` iterator, but you could also iterate on `by_key` which means you do not have to flatten vectors etc.

> I'm assuming that the kind of PoS that we're implementing is fairly linear, so as you have described, the main point of pooling stake is to increase the frequency of rewards (even if long term the total amount of rewards is the same).

But if you are handling each stake entry as a different entity, the frequency of rewards does not increase? That's my whole point. Someone staking 100k WIT on a 10M WIT validator will not see a more continuous stream of rewards due to that validator splitting block rewards, but rather a single "high" block rewards after a long time.

> I push for this "independent stake entries" model because I can see a wide demographic of people in our community who will probably pass on running their own validator, but would be willing to delegate their coins into other community members' validators, knowing that they can't really steal their coins.

I agree with that point, that will likely happen.

> I see a clear advantage when it comes to rewards and withdrawals. Not having to split rewards and handle remainders like above is pretty convenient.

I see how it simplifies the reward function, I do not see how it simplifies the withdrawals. I have no idea what would have to change there because that function should not care about unified or split rewards in another step.

drcpu-github: Resolved by simplifying the `add_reward` function until a later release as discussed offline.

drcpu-github:
For the sake of an easier review, the block time patch is on https://github.com/witnet/witnet-rust/pull/2473.

PR #2473, opened 2024-06-21
feature(data_structures): switch block time when v2.0 activates

drcpu-github:
This aims to enable switching block time when v2.0 activates. I tested this on a local testnet and it seems to behave properly, but this is obviously a complicated feature.


drcpu-github:
This patch also lives on https://github.com/witnet/witnet-rust/pull/2472.

PR #2481, opened 2024-08-19
Bridge/watch dog

guidiaz:


bridges/centralized-ethereum/src/actors/watch_dog.rs
```diff
@@ -0,0 +1,364 @@
...
+                        Ok(wit_utxos_above_threshold) => wit_utxos_above_threshold,
+                        Err(err) => {
+                            if status.eq("up-and-running") {
+                                status = err;
+                            }
+                            None
+                        }
+                    }
+                }
+                None => None,
```
Tommytrg, line 165: We could move this three calls into a `fetch_wit_info` function that returns `(wit_account, wit_balance, wit_utxos_above_threshold)`

bridges/centralized-ethereum/src/actors/watch_dog.rs
```diff
@@ -0,0 +1,364 @@
...
+        let start_wit_balance = self.start_wit_balance;
+        let wit_jsonrpc_socket = self.wit_jsonrpc_socket.clone();
+        let wit_utxo_min_value_threshold = self.wit_utxo_min_value_threshold;
+        let eth_jsonrpc_url = self.eth_jsonrpc_url.clone();
+        let eth_account = self.eth_account;
+        let eth_contract_address = self.eth_contract.clone().unwrap().address();
+        let running_secs = self.start_ts.unwrap().elapsed().as_secs();
+
+        let fut = async move {
+            let mut status = "up-and-running".to_string();
```
Tommytrg, line 117: What do you think about having all the possible values of status in an enum so they are easier to handle and less prone to error using them? Something like this

```rs
enum WatchDogStatus {
    // Add all of them here
    UpAndRunning,
    WitSyncing,
    Other(String),
}

impl WatchDogStatus {
    fn to_string(&self) -> String {
        match self {
            WatchDogStatus::UpAndRunning => "up-and-running".to_string(),
            WatchDogStatus::WitSyncing => "wit-syncing".to_string(),
            WatchDogStatus::Other(message) => message.clone(),
        }
    }
}
```

bridges/centralized-ethereum/src/actors/watch_dog.rs
```diff
@@ -0,0 +1,364 @@
...
+                            }
+                            None
+                        }
+                    }
+                }
+                None => None,
+            };
+
+            let eth_balance = match check_eth_account_balance(&eth_jsonrpc_url, eth_account).await {
+                Ok(Some(eth_balance)) => {
```
Tommytrg, line 169: Can this be move inside `check_eth_account_balance`?

bridges/centralized-ethereum/src/actors/watch_dog.rs
```diff
@@ -0,0 +1,364 @@
...
+                    if status.eq("up-and-running") {
+                        status = err;
+                    }
+                    None
+                }
+            };
+
+            let dr_database = DrDatabase::from_registry();
+            let (_, drs_pending, drs_finished, _) =
+                dr_database.send(CountDrsPerState).await.unwrap().unwrap();
```
Tommytrg, line 185: We should handle this error

bridges/centralized-ethereum/src/actors/watch_dog.rs
```diff
@@ -0,0 +1,364 @@
...
+
+            let mut metrics: String = "{".to_string();
```
Tommytrg, line 187: I'd move the logic related with the population of `metrics` to its own function

Tommytrg (changes requested):

bridges/centralized-ethereum/src/actors/watch_dog.rs
guidiaz: no time before holidays, next version perhaps

bridges/centralized-ethereum/src/actors/watch_dog.rs
guidiaz: And so in other parts of the bridge where i copied that from. Consider to open a new issue so all cases can be tackled on a dedicated PR.

bridges/centralized-ethereum/src/actors/watch_dog.rs
guidiaz: i prefer not to, as tracing logs is the ultimate goal of the watch_global_status function itself. Besides, that function would only be called from single point, and it would require passing all metrics by parameter.

PR #2532, opened 2024-11-18
fix(node): fix DR eligibility for (allegedly) all protocol versions

### LINKED ISSUE #2527: Troubleshoot legacy witnessing elegibility on the testnet
Apparently, witnessing eligibility using the legacy implementation while in protocol version 1.7 / 1.8 is broken.

Namely, the eligibility proofs being produced in 1.7 / 1.8 fail to pass the verification checks that those very same protocol versions enforce.

Eligibility is covered by tests, so I'm assuming that the issue lies in actor business logic. E.g. we may have a bad comparison in one of the protocol version checks, or a wrongly versioned hash is being passed to the verification functions, etc.

aesedepece:
This adds a series of fixes and improvements, but mainly acts by swapping the branches where we construct the VRF message conditionally depending on protocol version, as they were incorrectly ordered before.

fix #2527


data_structures/src/data_request.rs
```diff
@@ -554,7 +558,7 @@ pub fn data_request_has_too_many_witnesses(
     validator_count: usize,
     epoch: Option<Epoch>,
 ) -> bool {
-    if get_protocol_version(epoch) < ProtocolVersion::V2_0 {
+    if ProtocolVersion::from_epoch(epoch) < ProtocolVersion::V2_0 {
```
drcpu-github, line 561: Why this change? Forgot to unwrap it, I guess?

Issue #2501, opened 2024-11-04
Implement superblock signing time bomb for Wit/2


aesedepece:
The superblock signing committee from #2500 will be kept for the time being, on the condition that it will be eventually disabled and that signers will be sampled from the stakes tracker. This timed switch may be relative to the eventual activation date of 2_0.

aesedepece:
We may need to discuss different strategies for sampling the committee out of the stakes tracker, which already provides functions for ranking, sampling and stuff.

drcpu-github:
https://github.com/witnet/witnet-rust/pull/2448

PR #2542, opened 2024-11-27
Update superblock committee selection

drcpu-github:


aesedepece:
This tackles #2501

drcpu-github:
> This tackles #2501

(partially)

data_structures/src/superblock.rs
```diff
@@ -450,6 +462,28 @@ impl SuperBlockState {
                     signing_committee
                 };
 
+            // Override superblock signing committee during the bootstrapping of wit/2
+            self.signing_committee = match get_environment() {
+                Environment::Mainnet => {
+                    // wit2 activates at 2922240 + 14 x 1920 + 7 (TAPI delay) x 1920 (activation delay)
+                    // disabling the activation committee at 299_712 implies we have 8 weeks time (8 x 4320)
+                    // to inplement a decentralized superblock committee selection mechanism
+                    if (292_224..299_712).contains(&superblock_index) {
```
aesedepece, line 471: Unless I got it wrong, this is assuming that:
- The start of WIP-0028's first TAPI signaling period is known.
- The TAPI signaling will pass at first try, i.e. during the first signaling period.
- V2_0 will be scheduled from the very first epoch of V1_8?

drcpu-github: But even if there is a delay anywhere, that delay is basically subtracted from the 8 weeks extra time we give ourselves (and technically the subtraction is not even fully proportional to the delay because the 8 weeks is calculated using an epoch of 20s while a delay would imply 45s epochs).

aesedepece (approved):

aesedepece:
Merged as 200cb90e76ed3d3b064fa23f2901ce6d0c76b826

PR #2550, opened 2024-12-09
feat(slashing): lazily slash reluctant block proposers

### LINKED ISSUE #2543: Revisit implementation of Validators Slashing
.

guidiaz:
Close #2543

### DIFF
--- a/data_structures/src/chain/mod.rs
+++ b/data_structures/src/chain/mod.rs
@@ -311,9 +311,14 @@ impl ConsensusConstantsWit2 {
     }
 
     /// Replication factor for data request solving
-    pub fn get_replication_factor(self, epoch: Epoch) -> u16 {
+    pub fn get_replication_factor(self, epoch: Epoch, prev_epoch: Epoch) -> u16 {
         if get_protocol_version(Some(epoch)) >= ProtocolVersion::V2_0 {
-            4
+            // linearly increasing replication factor
+            return if epoch >= prev_epoch {
+                u16::try_from(12 * (epoch - prev_epoch)).unwrap_or(u16::MAX)
+            } else {
+                0
+            }
         } else {
             0
         }
--- a/data_structures/src/chain/mod.rs
+++ b/data_structures/src/chain/mod.rs
@@ -315,7 +315,7 @@ impl ConsensusConstantsWit2 {
         if get_protocol_version(Some(epoch)) >= ProtocolVersion::V2_0 {
             // linearly increasing replication factor
             return if epoch >= prev_epoch {
-                u16::try_from(12 * (epoch - prev_epoch)).unwrap_or(u16::MAX)
+                4u16.saturating_mul(2u16.saturating_pow(epoch - prev_epoch))
             } else {
                 0
             }
--- a/data_structures/src/staking/stakes.rs
+++ b/data_structures/src/staking/stakes.rs
@@ -272,8 +272,8 @@ where
         capability: Capability,
         epoch: Epoch,
         strategy: CensusStrategy,
-    ) -> Box<dyn Iterator<Item = Address> + '_> {
-        let iterator = self.rank(capability, epoch).map(|(address, _)| address);
+    ) -> Box<dyn Iterator<Item = StakeKey<Address>> + '_> {
+        let iterator = self.by_rank(capability, epoch).map(|(address, _)| address);
 
         match strategy {
             CensusStrategy::All => Box::new(iterator),
--- a/data_structures/src/staking/stakes.rs
+++ b/data_structures/src/staking/stakes.rs
@@ -310,7 +310,8 @@ where
             .map(|stake| stake.read_value().power(capability, epoch))
             .collect::<Vec<Power>>()
             .into_iter()
-            .sum())
+            .max()
+            .unwrap_or_default())
     }
 
     /// For a given capability, obtain the full list of positive stake entries reversely ordered by their power.
--- a/data_structures/src/staking/stakes.rs
+++ b/data_structures/src/staking/stakes.rs
@@ -313,33 +313,32 @@ where
             .sum())
     }
 
-    /// For a given capability, obtain the full list of stakers ordered by their power in that
-    /// capability.
+    /// For a given capability, obtain the full list of positive stake entries reversely ordered by their power.
     /// TODO: we may memoize the rank by keeping the last one in a non-serializable field in `Self` that keeps a boxed
-    ///  iterator, so that this method doesn't have to sort multiple times if we are calling the `rank` method several
-    ///  times in the same epoch.
-    pub fn rank(
+    ///       iterator, so that this method doesn't have to sort multiple times if we are calling the `rank` method 
+    ///       several times in the same epoch.
+    pub fn by_rank(
         &self,
         capability: Capability,
         current_epoch: Epoch,
-    ) -> impl Iterator<Item = (Address, Power)> + '_ {
-        self.by_validator
+    ) -> impl Iterator<Item = (StakeKey<Address>, Power)> + '_ {
+        self.by_key
             .iter()
-            .map(move |(address, stakes)| {
-                let power = stakes
-                    .first()
-                    .unwrap()
-                    .read_value()
-                    .power(capability, current_epoch);
-
-                (address.clone(), power)
+            .filter(|(_, sync_entry)| {
+                sync_entry.read_value().epochs.get(capability) <= current_epoch
             })
-            .sorted_by(|(address_1, power_1), (address_2, power_2)| {
-                // Equal power, compare the addresses to achieve deterministic ordering
+            .map(move |(key, entry)| {
+                (key.clone(), entry.read_value().power(capability, current_epoch))
+            })
+            .sorted_by(|(key_1, power_1), (key_2, power_2)| {
                 if power_1 == power_2 {
-                    address_1.cmp(address_2)
+                    if key_1.validator == key_2.validator {
+                        key_1.withdrawer.cmp(&key_2.withdrawer)
+                    } else {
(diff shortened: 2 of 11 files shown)


data_structures/src/staking/stakes.rs
```diff
@@ -339,6 +339,50 @@ where
...
+                .take(winner_rank + 1)
+                .map(|(key, _)| key)
+                .collect();
+            // proportionally reset coin age on located entry and all those with a better mining rank:
+            let mut index: usize = 0;
+            stakers.iter().for_each(|key| {
+                let stake_entry = self.by_key.get_mut(key);
+                if let Some(stake_entry) = stake_entry {
+                    let penalty_epochs = Epoch::from((1 + winner_rank - index) as u32);
+                    log::debug!("Delaying {} as block candidate during +{} epochs", key, penalty_epochs);
```
drcpu-github, line 372: `block proposer candidate`?

drcpu-github (commented):
I guess the idea of this patch is that by increasing the replication factor, we'll eventually find a validator willing to propose a block and at that point we can enforce power slashing?

data_structures/src/staking/stakes.rs
```diff
@@ -238,33 +239,32 @@ where
...
     }
 
-    /// For a given capability, obtain the full list of stakers ordered by their power in that
-    /// capability.
+    /// For a given capability, obtain the full list of positive stake entries reversely ordered by their power.
     /// TODO: we may memoize the rank by keeping the last one in a non-serializable field in `Self` that keeps a boxed
-    ///  iterator, so that this method doesn't have to sort multiple times if we are calling the `rank` method several
-    ///  times in the same epoch.
-    pub fn rank(
+    ///       iterator, so that this method doesn't have to sort multiple times if we are calling the `rank` method
```
drcpu-github, line 244: Extraneous space at the end of the line.

data_structures/src/staking/stakes.rs
guidiaz: Perhaps even better, `"Disabling mining power to {} during +{} epochs"`?

data_structures/src/chain/mod.rs
```diff
@@ -311,9 +311,14 @@ impl ConsensusConstantsWit2 {
     }
 
     /// Replication factor for data request solving
-    pub fn get_replication_factor(self, epoch: Epoch) -> u16 {
+    pub fn get_replication_factor(self, epoch: Epoch, prev_epoch: Epoch) -> u16 {
         if get_protocol_version(Some(epoch)) >= ProtocolVersion::V2_0 {
-            4
+            // linearly increasing replication factor
+            return if epoch >= prev_epoch {
+                u16::try_from(12 * (epoch - prev_epoch)).unwrap_or(u16::MAX)
```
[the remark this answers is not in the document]
guidiaz: The higher the growing factor, the less probable that a big enough group of malicious stake entries can collude to temporarily censor the chain. For taking over up to 10 consecutive blocks, an attacker would need to place 120 stake entries on the top ranking, simultaneously. Depending on the actual census, and whether multiple stake entries per validator address is enabled or not, this number could either be considered relatively high, or most probably (in the long run, if we expect census to grow in time), relatively low. 

Nevertheless, the replication factor is proposed to be capped to some proportion of the total census. Two thirds has been initially proposed under this rationale: "if more than two thirds of census decide to censor the chain, the chain is to be censored". 

In terms of P2P network traffic, one could argue that up to 2/3rds of eligible block proposers may potentially flood the network with competing block candidates, but not really:
- Nodes should only propagate block candidates only if better than the best validated so far.
- On wit/2.1+, eligible candidates will only be able to propose one single block, no matter how many stake entries they have within the top ranking.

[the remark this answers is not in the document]
drcpu-github: I'm okay with the capping, but was thinking that it would be better to follow the same power law as with data request eligibility: `4 * 2 ** (epoch - prev_epoch - 1)`.

epoch | power | linear
--|--|--
1 | 4 | 12
2 | 8 | 24
3 | 16 | 36
4 | 32 | 48
5 | 64 | 60
6 | 128 | 72
7 | 256 | 84
8 | 512 | 96
9 | 1024 | 108
10 | 2048 | 120




In the end it's unlikely to really matter though, both for the ratio of blocks mined and actual liveness.

node/src/actors/chain_manager/mod.rs
```diff
@@ -2345,7 +2340,7 @@ impl ChainManager {
         let protocol_version = ProtocolVersion::from_epoch(block.block_header.beacon.checkpoint);
         let replication_factor = self
             .consensus_constants_wit2
-            .get_replication_factor(block.block_header.beacon.checkpoint);
+            .get_replication_factor(block.block_header.beacon.checkpoint, chain_beacon.checkpoint);
```
drcpu-github, line 2343: Need to subtract 1
```rust
// The beacon was updated with the current epoch prior to calling this function to build the block
// currently under validation, so subtract one from the beacon to get the prior epoch
let replication_factor = self
    .consensus_constants_wit2
    .get_replication_factor(block.block_header.beacon.checkpoint, chain_beacon.checkpoint - 1);
```

data_structures/src/chain/mod.rs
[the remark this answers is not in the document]
aesedepece: That's my line of thinking too. When assessing the right replication factor _RF_, we need to take into account how credibly can an attacker take over the top _RF_ positions in the rank. Under consideration of the economic criteria presented by @drcpu-github above, and based on what we've seen in the 4 years of mainnet, it is reasonable to think that 12 may be a bit too much for a start.

On the other hand, what's the downside of a high-ish initial RF value? Bloating the network with redundant messages doesn't sound specially of concern here, under the assumption that block candidates are still subject to comparison and filtering as they travel through the network.

As per the _power_ vs. _linear_ progression, I favor _power_ because it has worked pretty decently so far, but don't oppose to _linear_ if it is found to be more suitable for PoS.

node/src/actors/chain_manager/mod.rs
guidiaz: Hmm, shouldn't we rather pass `chain_beacon` instead of `beacon` when calling `future_process_validation` from https://github.com/witnet/witnet-rust/blob/master/node/src/actors/chain_manager/mining.rs#L256?

drcpu-github: I just realized above proposal does not make any sense because this implies that when validating we are always assuming blocks are consecutive which is the whole thing this patch attempts to solve.

Instead, we need something like this:
```
diff --git a/node/src/actors/chain_manager/mining.rs b/node/src/actors/chain_manager/mining.rs
index ecc0e1c9..e4ce6d7f 100644
--- a/node/src/actors/chain_manager/mining.rs
+++ b/node/src/actors/chain_manager/mining.rs
@@ -110,6 +110,7 @@ impl ChainManager {
         let checkpoint_zero_timestamp = chain_info.consensus_constants.checkpoint_zero_timestamp;
         let halving_period = chain_info.consensus_constants.halving_period;

+        let previous_block_epoch = chain_info.highest_block_checkpoint;
         let mut beacon = chain_info.highest_block_checkpoint;
         let mut vrf_input = chain_info.highest_vrf_output;
         let protocol_version = get_protocol_version(self.current_epoch);
@@ -252,6 +253,7 @@ impl ChainManager {
             .and_then(move |block, act, _ctx| {
                 act.future_process_validations(
                     block.clone(),
+                    previous_block_epoch,
                     current_epoch,
                     vrf_input,
                     beacon,
diff --git a/node/src/actors/chain_manager/mod.rs b/node/src/actors/chain_manager/mod.rs
index 7251c626..b4a7a4eb 100644
--- a/node/src/actors/chain_manager/mod.rs
+++ b/node/src/actors/chain_manager/mod.rs
@@ -1239,6 +1239,7 @@ impl ChainManager {
                     // IMPORTANT: Always perform age resets after adding rewards

                     // Reset mining power to block proposer, and higher ranked eligible candidates:
+                    log::info!("Resetting mining age for {} to {}", miner_pkh, block_epoch);
                     let _ = stakes.reset_mining_age(miner_pkh, block_epoch);

                     // Reset witnessing power
@@ -2332,6 +2333,7 @@ impl ChainManager {
     pub fn future_process_validations(
         &mut self,
         block: Block,
+        previous_block_epoch: Epoch,
         current_epoch: Epoch,
         vrf_input: CheckpointVRF,
         chain_beacon: CheckpointBeacon,
@@ -2349,7 +2351,7 @@ impl ChainManager {
         // currently under validation, so subtract one from the beacon to get the prior epoch
         let replication_factor = self
             .consensus_constants_wit2
-            .get_replication_factor(block.block_header.beacon.checkpoint, chain_beacon.checkpoint - 1);
+            .get_replication_factor(block.block_header.beacon.checkpoint, previous_block_epoch);
         let res = validate_block(
             &block,
             current_epoch,
```

aesedepece: Yes that makes much more sense.

data_structures/src/staking/stakes.rs
```diff
@@ -339,6 +340,53 @@ where
...
+                .map(|(key, _)| key)
+                .collect();
+            // proportionally reset coin age on located entry and all those with a better mining rank:
+            let mut index: usize = 0;
+            stakers.iter().for_each(|key| {
+                let stake_entry = self.by_key.get_mut(key);
+                if let Some(stake_entry) = stake_entry {
+                    let penalty_epochs = Epoch::from((1 + winner_rank - index) as u32);
+                    log::info!(
+                        "Resetting and disabling mining power of {} (ranked as #{}) during +{} epochs",
```
drcpu-github, line 374: No space needed at the end of the line.

PR #2556, opened 2024-12-13
fix(mining): always update dr_pool when processing drts w/ too many witnesses

guidiaz:
dr_state.dr_info.block_hash_dr_tx was not getting properly updated:
- if the DRT was created and included in a block by the same validator;
- if the DRT was being reinserted after a rollback.


aesedepece:
Merged as b9f4e0cd4b6cc2e0eb09f8c12269a764067b7b24

PR #2565, opened 2024-12-26
Implement proper mechanism for sampling superblock signing committees in V2_0

aesedepece:
A very simple approach to implementing decentralized selection of superblock signing committees in protocol V2_0.

IMO, sampling the committee from the top 50% validators that operate the most stake should be more than enough to prevent attackers from spamming—or potentially overtaking—the committee with sock puppet identities.

solve #2501

### DIFF
--- a/data_structures/src/chain/mod.rs
+++ b/data_structures/src/chain/mod.rs
@@ -4888,7 +4888,7 @@ mod tests {
 
     use crate::{
         proto::versioning::{ProtocolVersion, VersionedHashable},
-        superblock::{mining_build_superblock, ARSIdentities},
+        superblock::{mining_build_superblock, Census},
         transaction::{CommitTransactionBody, RevealTransactionBody, VTTransactionBody},
     };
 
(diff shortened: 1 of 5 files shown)


drcpu-github:
Does this imply we need more than 200 validators before we actually start sampling (since as long as the top 50% of stakers amounts to less than 100 validators we'd use all of them as voters)? If that is the case, I'm not sure if this strategy is sufficient because you'd need only a couple of (fairly low staking) validators to break superblock consensus?

For example, say we have 60 validators of which the top 30 is selected to vote on a superblock. Assume all validators have between 2M and 10M WIT staked with the following distribution (amounting to a total stake of exactly 300M):

- Validator 1 -> 30: 2M staked (60M total stake)
- Validator 30 -> 40: 4M staked (40M total stake)
- Validator 40 -> 60: 10M staked (200M total stake)

 In order to disrupt the superblock consensus, you'd need to control ~10 identities with a total stake of only about 40M WIT (amounting to less than 15% of the total stake). I realize I crafted an extreme example here, but if an attacker controlling less than 15% of the total stake can theoretically halt the network, I'd say the selection strategy is not as safe as I'd like.

PR #2567, opened 2025-01-02
Include validator pkh in authorization

parodyBit:


src/cli/node/json_rpc_client.rs
```diff
@@ -1039,8 +1041,23 @@ pub fn authorize_st(addr: SocketAddr, withdrawer: Option<String>) -> Result<(),
 
     let message = authorization.withdrawer.as_secp256k1_msg();
 
-    let auth_bytes = authorization.signature.to_recoverable_bytes(&message)?;
-    let auth_string = hex::encode(auth_bytes);
+    // let validator_bytes: [u8; 20] = authorization.signature.public_key.pkh().as_ref().try_into()?;
+    // let signature_bytes: [u8; 65] = authorization.signature.to_recoverable_bytes(&message).unwrap();
+    // let auth_string: String = hex::encode([&validator_bytes[..], &signature_bytes[..]].concat());
```
aesedepece, line 1046:
```suggestion
```

aesedepece (changes requested):

aesedepece (approved):

drcpu-github:
Should this patch be a bit extended to also get rid of the `--validator` flag in the `stake` method. I guess there is no point in still having that and validating it if we include the validator data in the authorization string / QR code?

guidiaz:
Either the CLI implementation for "stake" is adapted to the new format of the authorization code (hexified validator pkh + hexified validator's signature), or, the CLI implementation for "stakeAuthorization" should preserve the old schema when printing "Autherization code", and use the new one only for generating the QR code

aesedepece:
> Should this patch be a bit extended to also get rid of the `--validator` flag in the `stake` method. I guess there is no point in still having that and validating it if we include the validator data in the authorization string / QR code?

I support this for better UX.

src/cli/node/json_rpc_client.rs
```diff
@@ -978,8 +977,11 @@ pub fn send_st(
...
-    let validator_address = validator
-        .try_do_magic(|hex_str| PublicKeyHash::from_bech32(get_environment(), &hex_str))?;
+    let validator_address = {
+        let pkh_bytes =
+            hex::decode(authorization.clone().chars().take(40).collect::<String>()).unwrap();
+        PublicKeyHash::from_bytes(pkh_bytes.as_slice())?
+    };
     if validator_address != dry.validator {
         bail!(
             "The specified validator ({}) does not match the validator recovered from the authorization string ({}), please double check all arguments.",
```
drcpu-github, line 987: This message becomes a bit weird with the validator being recovered from the authorization string.

If you specify an incorrect withdrawer, the dry run will result in that withdrawer being recovered but with a validator that does not match the validator in the authorization string.

My proposal would be to update this message:
```rust
bail!(
    "The validator recovered from the authorization string ({}) does not match the validator calculated using the specified withdrawer ({}). Please double check you are supplying the same withdrawer as the one used to generate the authorization string.",
    validator_address,
    dry.validator.to_string(),
);
```

aesedepece:
Merged in as 366f2cba01e576d32a5cc6dd9853a60dda842931. Thanks gentlemen.

PR #2570, opened 2025-01-13
Json rpc/non sensitive stakes ranks utxos

guidiaz:


Tommytrg:
Could you rename the commit https://github.com/witnet/witnet-rust/pull/2570/commits/15d0e82c182fff5a8f080b2a1a14883c0c80bd07 to be less than 72 chars to avoid truncation? I propose something like "feat(json_rpc): fetch ranked stake entries by capability"

image

Tommytrg:
Regarding the JSONRPC method "getUtxoInfo", I don't see why it is a sensitive as it requires a pkh. The sensitive layer lies in the CLI where uses the node's pkh as the default value.

aesedepece:
Feel free to request a review when ready.

guidiaz:
> Regarding the JSONRPC method "getUtxoInfo", I don't see why it is a sensitive as it requires a pkh. The sensitive layer lies in the CLI where uses the node's pkh as the default value.

Exactly

node/src/actors/json_rpc/api.rs
```diff
@@ -145,6 +145,12 @@ pub fn attach_regular_methods<H>(
     server.add_actix_method(system, "queryStakes", |params: Params| {
         Box::pin(query_stakes(params.parse()))
     });
+    server.add_actix_method(system, "ranks", |params: Params| {
```
aesedepece, line 148: Just nitpicking here. What you get in return of this method is a list comprising all stake entries, ranked by power. Let me argue that's a "ranking" and not a "ranks" inasmuch it's not a list of ranks (like in military ranks, etc) but more of a ranked / sorted kind of roster.

```suggestion
    server.add_actix_method(system, "ranking", |params: Params| {
```

Alternatively, if we follow the pattern of naming the method with the action it will perform instead of what we are expecting it to return (like in methods named `getX` or `queryX`), then `rank` could make sense too:

```suggestion
    server.add_actix_method(system, "rank", |params: Params| {
```

aesedepece (changes requested):

node/src/actors/json_rpc/api.rs
guidiaz: Agreed, perhaps a better name would be "queryPowers":
- as queryStakes was not nitpicked
- and because this new method doesn't actually return a list of stake entries, but tuples containing the validator, withdrawer and current power for specified capability

aesedepece (approved):
This is 🔝

PR #2575, opened 2025-01-21
fix(session): use user_agent field to determine whether to drop protocol versions info

drcpu-github:


guidiaz:
What if we needed to release a patch release before or after V2_0 activation? Would it make sense to accept user agents containing a version either equal or greater than the one of the node processing the Version message?

drcpu-github:
The only thing this does is drop the protocol version info. It does not ice peers. I have considered this, but I'm not sure accepting protocol info from a newer node makes sense since we can assume it will contain incompatibilities with your current node?

PR #2586, opened 2025-01-29
Double check nonce before building a block

drcpu-github:


data_structures/src/staking/stakes.rs
```diff
@@ -360,19 +360,19 @@ where
...
     {
         let key = key.into();
 
-        if let Entry::Occupied(entry) = self.by_key.entry(key.clone()) {
-            let stake = entry.get().value.read()?;
+        let stake = self
+            .by_key
+            .get(&key)
+            .ok_or(StakesError::EntryNotFound { key })?
+            .read_value();
```
aesedepece, line 373: Much more elegant :ok_hand:

data_structures/src/transaction_factory.rs
```diff
@@ -680,16 +680,18 @@ pub fn transaction_inputs_sum(
...
-            && vt_time_lock > epoch_timestamp
-        {
-            return Err(TransactionError::TimeLock {
-                expected: vt_time_lock,
-                current: epoch_timestamp,
+        if vt_time_lock > epoch_timestamp {
+            if get_environment() == Environment::Mainnet
+                || ((get_environment() == Environment::Development
+                    || get_environment() == Environment::Testnet)
+                    && epoch > 22_000)
```
aesedepece, line 687: I think this is equivalent in practice, but easier to reason about:
```suggestion
            if get_environment() == Environment::Mainnet || epoch > 22_000
```

aesedepece (changes requested):

data_structures/src/transaction_factory.rs
drcpu-github: Sure, fair enough.

PR #2595, opened 2025-02-17
fix(storage): allow getting old blocks from storage

aesedepece:
Namely, it intercepts the serialized bytes and fills in some extra bytes to nullify new fields that didn't exist back when they were serialized in the first place.


aesedepece:
This was merged as of 2.0.2.

PR #2612, opened 2025-03-24
Bridge/2.1.x

guidiaz:


node/src/actors/json_rpc/api.rs
```diff
@@ -115,6 +106,9 @@ pub fn attach_regular_methods<H>(
     server.add_actix_method(system, "getBalance2", |params: Params| {
         Box::pin(get_balance_2(params.parse()))
     });
+    server.add_actix_method(system, "getBalance2", |params: Params| {
+        Box::pin(get_balance_2(params.parse()))
+    });
```
Tommytrg, line 111: This method is already implemented above
```suggestion
```

PR #2614, opened 2025-03-27
chore: avoid bootstrapping public testnet for wit/2.0

guidiaz:


guidiaz:
Changes ultimately merged by @aesedepece into master.

PR #2622, opened 2025-05-21
fix(node): solve potential DoS when passing a TX id to getBlock

aesedepece:
When passing a TX id to getBlock, a pseudomigration is run, such that it will crash the node because it can't read enough bytes.

This solution only runs the migration if there are enough bytes to operate on. In this particular case of passing a TX id, it will behave as before (produce a proper error)


aesedepece:
Thanks @guidiaz for detecting and reporting!!! :muscle:

PR #2649, opened 2025-09-08
fix(rad): processing result http-head requests as RadonMaps while all…

guidiaz:
…owing binary sources


aesedepece:
Merged as of d64f2e8183962fe6126779ec3c4829a9e3b5c573

PR #2664, opened 2025-12-09
feat(json_rpc): improve and protect getSupplyInfo* methods

guidiaz:


node/src/actors/chain_manager/mod.rs
```diff
@@ -262,6 +262,12 @@ pub struct ChainManager {
     initial_supply: u64,
     /// Populate RAD hashes index
     rad_hashes_index: bool,
+    /// Epoch when getSupplyInfo2 was last resolved.
```
drcpu-github: I suppose these variables are not saved to storage (and we do not need storage migrations)? That does imply whenever a node restarts (after a manual intervention or crash), this call will be slow again for the first time?

drcpu-github (commented):
I have not actually tested this branch, just added some questions based on a quick read through the code.

aesedepece:
Merged as of 2e99da2c607088b47ead9b75dde8bf83f5a3e39a

